PR #86, opened 2024-09-05
GH-11: Add test CI: macOS

kou:
Fix GH-11

This part:
https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L305-L360

This also imports `ci/scripts/go_{build,test}.sh` from apache/arrow.

This also introduce ShellCheck and shfmt with pre-commit.

This also adds apache/arrow-testing and apache/parquet-testing submodules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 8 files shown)


kou:
CI result: https://github.com/kou/arrow-go/actions/runs/10716893068

zeroshade:
Before we can truly claim that the CI is working, we need to update the package path of the `go.mod` file and all of the imports in the files to point to itself here instead of the old paths.

kou:
OK. How about #88?

kou:
Passed: https://github.com/kou/arrow-go/actions/runs/10745815085

I'll merge this.

PR #88, opened 2024-09-05
GH-4: Update package name

kou:
Fix GH-4

This is done by the following command line:

    find . -type f -exec sed -i'' -e 's,apache/arrow/go,apache/arrow-go,g' '{}' ';'


kou:
I want to prevent adding a new Go commit to apache/arrow after this. Because cherry-picking will be difficult after this.

kou:
@zeroshade @joellubi Can we proceed with this? This is a blocker of all CI related tasks.

zeroshade:
Let's proceed with this so we can keep moving. I'll look at the outstanding Go-related PRs and shift them here afterwards

zeroshade (approved):

kou:
Thanks!

joellubi:
This reminded me that the go package name in the `.proto` files will need to be updated as well. That also means that codegen will need to get updated so that it can be built with definitions in another repo. I've opened an issue for this: #89.

Issue #10, opened 2024-08-29
Add test CI: Windows


kou:
This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L272-L303

PR #90, opened 2024-09-07
GH-10: Add test CI: Windows

kou:
Fix GH-10

This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L272-L303


kou:
+1

Issue #2, opened 2024-08-29
Enable Dependabot


kou:
We only need this part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/dependabot.yml#L18-L33

PR #91, opened 2024-09-07
GH-2: Enable Dependabot

kou:
Fix GH-2

We only need this part:
https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/dependabot.yml#L18-L33


kou:
+1

Issue #15, opened 2024-08-29
Add lint CI


kou:
We want to use pre-commit based lint system.

Related parts:
* https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/dev.yml#L47-L62
* https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.pre-commit-config.yaml#L151-L164

PR #92, opened 2024-09-07
GH-15: Enable linter

kou:
Fix GH-15

Related parts:
* https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.pre-commit-config.yaml#L151-L164
* https://github.com/apache/arrow/blob/main/.golangci.yaml

staticcheck is also used via golangci. We need to use "nolint:staticcheck,XXX" instead of "lint:ignore XXX" for it.

See also: https://golangci-lint.run/usage/false-positives/#nolint

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,10 +53,6 @@ jobs:
       - name: Install bash via Homebrew
         run: |
           brew install bash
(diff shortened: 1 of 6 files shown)


kou:
+1

Issue #100, opened 2024-09-07
Avoid duplicated CI jobs by Dependabot


kou:
Dependabot creates a branch in apache/arrow-go and create a PR from the branch.
Our CI jobs are executed for both of push and pull request.
So Dependabot PRs run duplicated CI jobs.

For example: https://github.com/apache/arrow-go/pull/93

kou:
Done by GH-101.

PR #101, opened 2024-09-07
GH-100: Don't run CI jobs by Dependabot push

kou:
It seems that Dependabot always use `dependabot/*` branch. If we ignore `push` trigger for the branch, we can run only `pull_request` trigger for Dependabot.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,6 +19,8 @@ name: Lint
 
 on:
   push:
+    branches-ignore:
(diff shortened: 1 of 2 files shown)


kou:
Push to non`dependabot/*` branch still trigger CI: https://github.com/kou/arrow-go/actions/runs/10754458908

kou:
+1

kou:
https://github.com/apache/arrow-go/pull/104 runs only pull request CI jobs. It's expected.

Issue #102, opened 2024-09-07
Commit ID and commented tag name pair is a mismatch in actions/checkout


kou:
`actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0` is a mismatch.

https://github.com/actions/checkout/commit/1e31de5234b9f8995739874a8ce0492dc87873e2 is `v4.0.0`.

PR #103, opened 2024-09-07
GH-102: Fix commit ID for actions/checkout v4.0.0

kou:
Fix GH-102


kou:
+1

Let's see what's happen by this. I hope that Dependabot opens a setup/checkout PR with tag (like https://github.com/apache/arrow-go/pull/93 ) not commit ID (like https://github.com/apache/arrow-go/pull/94 ).

kou:
Good: https://github.com/apache/arrow-go/pull/104

PR #105, opened 2024-09-09
GH-5: Add test CI: Docker based

kou:
Fix GH-5

This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L49-L135


raulcd (approved):
LGTM

Issue #6, opened 2024-08-29
Add test CI: i386


kou:
This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L165-L190

PR #106, opened 2024-09-09
GH-6: Add test CI: i386

raulcd:
Fix https://github.com/apache/arrow-go/issues/6

This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L165-L190


kou (approved):
+1

Issue #14, opened 2024-08-29
Add test CI: TinyGo


kou:
This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L467-L482

PR #112, opened 2024-09-09
GH-14: Add test CI: TinyGo

kou:
Fix GH-14

This part:
https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L467-L482

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -169,3 +169,22 @@ jobs:
       - name: Test
         # WIP refactor, only tests in the specified dirs have been fixed
         run: GOARCH=386 go test ./parquet/file/...
+  tinygo:
+    name: TinyGo
+    runs-on: ubuntu-latest
+    env:
+      TINYGO_VERSION: 0.33.0
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -169,3 +169,22 @@ jobs:
...
         run: GOARCH=386 go test ./parquet/file/...
+  tinygo:
+    name: TinyGo
+    runs-on: ubuntu-latest
+    env:
+      TINYGO_VERSION: 0.33.0
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
zeroshade, line 180: is `v4.1.7` not a tagged verison we can use directly? Do we have to use the sha?

zeroshade (approved):

.github/workflows/test.yml
kou: We can use `v4.1.7` directly. This is for security.
See also: https://github.com/apache/arrow/pull/37676

(Honestly, I'm using tag directly in general...)

@assignUser may be able to explain security more.

Issue #89, opened 2024-09-07
Migrate Flight and FlightSql Proto Gen


joellubi:
[Flight](https://github.com/apache/arrow/blob/main/format/Flight.proto) and [FlightSql](https://github.com/apache/arrow/blob/main/format/FlightSql.proto) definitions are still in apache/arrow.

The [go_package](https://github.com/apache/arrow/blob/main/format/Flight.proto#L23) name will need to be updated to match this repo, and the actual codegen will need to be ported as well. We will no longer be able to use a direct reference to the `.proto` files, so we may need to use git submodules or a tool like [buf](https://buf.build) in order to build the needed files.

kou:
Package name change in apache/arrow: https://github.com/apache/arrow/pull/44035

kou:
How about referring apache/arrow's `format/` via an environment variable?
https://github.com/apache/arrow-go/pull/114

PR #114, opened 2024-09-10
GH-89: Add support for generating Flight related files

kou:
Fix GH-89

### DIFF
--- a/arrow/flight/gen.go
+++ b/arrow/flight/gen.go
@@ -16,5 +16,5 @@
 
 package flight
 
-//go:generate protoc --experimental_allow_proto3_optional -I../../../format --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative Flight.proto
-//go:generate protoc --experimental_allow_proto3_optional -I../../../format --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative FlightSql.proto
+//go:generate protoc --experimental_allow_proto3_optional -I${APACHE_ARROW_FORMAT_DIR} --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative Flight.proto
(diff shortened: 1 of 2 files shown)


arrow/flight/gen.go
```diff
@@ -16,5 +16,5 @@
 
 package flight
 
-//go:generate protoc --experimental_allow_proto3_optional -I../../../format --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative Flight.proto
-//go:generate protoc --experimental_allow_proto3_optional -I../../../format --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative FlightSql.proto
+//go:generate protoc --experimental_allow_proto3_optional -I${APACHE_ARROW_FORMAT_DIR} --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative Flight.proto
+//go:generate protoc --experimental_allow_proto3_optional -I${APACHE_ARROW_FORMAT_DIR} --go_out=./gen/flight --go-grpc_out=./gen/flight --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative FlightSql.proto
```
kou, line 20: Can we move these `protoc ...` command lines to `dev/sync_flight.sh` and remove this file?

zeroshade: the standard way of handling things like this would be to utilize `go generate` to run the commands rather than a separate sync script (ideally). So I'd personally prefer keeping this file and the `go:generate` lines.

zeroshade: that said, it might be better to utilize `buf` instead see the links below on the `substrait-go` repo for how I did it there to reference and generate the protobuf from the .proto files in the `substrait-io/substrait` repo

https://github.com/substrait-io/substrait-go/blob/main/doc.go#L9
https://github.com/substrait-io/substrait-go/blob/main/buf.gen.yaml

kou: OK. I withdraw this approach.

kou:
We should use buf.

PR #116, opened 2024-09-10
GH-7:  Add test CI: Debian + CGO

kou:
Fix GH-7

This part:
https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L192-L230


raulcd (approved):
LGTM, I've validated the correct tags when `ARROW_GO_TESTCGO`: `-tags assert,test,ccalloc` are in the build

Issue #12, opened 2024-08-29
Add test CI: macOS + CGO


kou:
This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L363-L402

PR #118, opened 2024-09-10
GH-12: Add test CI: macOS + CGO

raulcd:
Fix https://github.com/apache/arrow-go/issues/12

This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L363-L402

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -137,8 +137,8 @@ jobs:
       fail-fast: false
       matrix:
         go:
-          - '1.21'
           - '1.22'
+          - '1.23'
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -132,13 +132,13 @@ jobs:
...
     runs-on: macos-12
-    timeout-minutes: 15
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
         go:
-          - '1.21'
           - '1.22'
+          - '1.23'
```
kou, line 141: Thanks!

kou (approved):
+1

PR #119, opened 2024-09-10
GH-13: Add test CI: Windows + MinGW

raulcd:
Fix https://github.com/apache/arrow-go/issues/13

This part: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/.github/workflows/go.yml#L404-L465

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -276,7 +276,7 @@ jobs:
       - name: Test
         shell: bash
         run: ci/scripts/test.sh $(pwd)
-  build_test_386:
+  build-test-386:
     name: Cross-build and test for 386
     runs-on: ubuntu-latest
     timeout-minutes: 15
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -222,6 +222,60 @@ jobs:
...
+          go-version: "${{ env.GO_VERSION }}"
+          cache: true
+          cache-dependency-path: go.sum
+      - name: Build
+        shell: bash
+        run: ci/scripts/build.sh $(pwd)
+      - name: Test
+        shell: bash
+        run: ci/scripts/test.sh $(pwd)
   build_test_386:
```
kou, line 279:
```suggestion
  build-test-386:
```

kou (approved):
+1

Issue #120, opened 2024-09-10
Add initial Decimal32/Decimal64 implementation


zeroshade:
Related: apache/arrow#43956

### Rationale for this change
Widening the Decimal128/256 type to allow for bitwidths of 32 and 64 allows for more interoperability with other libraries and utilities which already support these types. This provides even more opportunities for zero-copy interactions between things such as libcudf and various databases.

### What changes are included in this PR?
This PR contains the basic Go implementations for Decimal32/Decimal64 types, arrays, builders and scalars. It also includes the minimum necessary to get everything compiling and tests passing without also extending the acero kernels and parquet handling (both of which will be handled in follow-up PRs).

### Are these changes tested?
Yes, tests were extended where applicable to add decimal32/decimal64 cases.

PR #121, opened 2024-09-10
GH-120: Add initial Decimal32/Decimal64 implementation

zeroshade:
Fix GH-120

### Rationale for this change
Widening the Decimal128/256 type to allow for bitwidths of 32 and 64 allows for more interoperability with other libraries and utilities which already support these types. This provides even more opportunities for zero-copy interactions between things such as libcudf and various databases.

This PR contains the basic Go implementations for Decimal32/Decimal64 types, arrays, builders and scalars. It also includes the minimum necessary to get everything compiling and tests passing without also extending the acero kernels and parquet handling (both of which will be handled in follow-up PRs).

### Are these changes tested?
Yes, tests were extended where applicable to add decimal32/decimal64 cases.


.gitignore
```diff
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+.vscode
```
kou, line 18:
```suggestion
/.vscode
```

kou (approved):
+1

We need integration test CI for this, right?

I'll prepare it.

.gitignore
zeroshade: I left out the forward slash because I want to ignore .vscode directories at every level, not only the top level. Does that make sense? Or should I still add the /?

zeroshade:
> We need integration test CI for this, right?

Yea, the corresponding C++ PR for decimal 32/64 has the change to datagen.py to add these cases to the integration tests, and then when we update the CI integration tests to use this repo instead of the main arrow repo, we can unskip Go.

.gitignore
kou: Oh, sorry. I thought that `.vscode` is only placed at the top directory.
If we may have `.vscode` in subdirectories, this configuration (no `/`) is preferred.

arrow/datatype_fixedwidth.go
```diff
@@ -532,19 +533,74 @@ type DecimalType interface {
 	DataType
 	GetPrecision() int32
 	GetScale() int32
+	BitWidth() int
 }
 
 func NewDecimalType(id Type, prec, scale int32) (DecimalType, error) {
 	switch id {
+	case DECIMAL32:
+		debug.Assert(prec <= int32(decimal.MaxPrecision[decimal.Decimal32]()), "invalid precision for decimal32")
```
ggodik, line 542: should this condition be an error? or is the `DecimalNType` informational

zeroshade: The assumption being made is that if a user is explicitly calling `NewDecimalType` they know what they are doing, so we only perform the precision validation in debug mode for performance.

joellubi (approved):

Issue #21, opened 2024-08-29
Add release CI


kou:
arrow-julia's implementation will help us: https://github.com/apache/arrow-julia/blob/2696105d01cfda7c55d1902951a20908a3c205e5/.github/workflows/ci.yml#L35-L66

PR #122, opened 2024-09-12
GH-21: Add release CI

kou:
Fix GH-21

This is based on
https://github.com/apache/arrow-julia/tree/main/dev/release and https://github.com/apache/arrow-adbc/tree/main/dev/release .

Workflow:

Cut a RC:

1. Update `PkgVersion` in `arrow/doc.go`
2. Run `dev/release/release_rc.sh`
   1. `dev/release/release_rc.sh` pushes `v${version}-rc${rc}` tag
   2. `.github/workflows/rc.yml` creates `apache-arrow-go-${version}.tar.gz{,.sha256,.sha512}`
   3. `.github/workflows/rc.yml` uploads `apache-arrow-go-${version}.tar.gz{,.sha256,.sha512}` to GitHub Releases
   4. `dev/release/release_rc.sh` downloads `apache-arrow-go-${version}.tar.gz` from GitHub Releases
   5. `dev/release/release_rc.sh` signs `apache-arrow-go-${version}.tar.gz` as `apache-arrow-go-${version}.tar.gz.asc`
   6. `dev/release/release_rc.sh` uploads `apache-arrow-go-${version}.tar.gz.asc` to GitHub Releases
3. Start a vote

(GitHub Actions instead of
https://dist.apache.org/repos/dist/dev/arrow/ is used like ADBC.)

Verify a RC:

1. Run `dev/release/verify_rc.sh`

Release an approved RC:

1. Run `dev/release/release.sh`
   1. `dev/release/release.sh` pushes `v${version}` tag that refers that same commit ID as `v${version}-rc${rc}`
   2. `dev/release/release.sh` downloads `apache-arrow-go-${version}.tar.gz{,.asc,.sha256,.sha512}` from GitHub Actions
   3. `dev/release/release.sh` uploads `apache-arrow-go-${version}.tar.gz{,.asc,.sha256,.sha512}` to https://dist.apache.org/repos/dist/release/arrow
   4. `dev/release/release.sh` removes old releases from https://dist.apache.org/repos/dist/release/arrow
2. Add this release to ASF's report database: https://reporter.apache.org/addrelease.html?arrow

Follow-up tasks:
* Add support for running integration test in the verification script
* Add support for releasing a release note
* Add support for creating "v${version}"'s GitHub Releases
  * We can copy "v${version}-rc${rc}"'s GitHub Releases and adjust it for the official release

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -1,0 +1,133 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: RC
+on:
+  push:
+    branches:
+      - '**'
+      - '!dependabot/**'
+    tags:
+      - '*-rc*'
+  pull_request:
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
+  archive:
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: recursive
+      - name: Prepare for tag
+        if: github.ref_type == 'tag'
+        run: |
+          version=${GITHUB_REF_NAME%-rc*}
+          version=${version#v}
+          rc=${GITHUB_REF_NAME#*-rc}
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Prepare for branch
+        if: github.ref_type == 'branch'
+        run: |
+          version=$(grep -o '^const PkgVersion = ".*"' "arrow/doc.go" |
+            sed -e 's/^const PkgVersion = "//' \
+              -e 's/"$//')
+          rc=100
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Archive
+        run: |
+          id="apache-arrow-go-${VERSION}"
+          tar_gz="${id}.tar.gz"
+          echo "TAR_GZ=${tar_gz}" >> ${GITHUB_ENV}
+          git archive HEAD --prefix "${id}/" --output "${tar_gz}"
+          sha256sum "${tar_gz}" > "${tar_gz}.sha256"
+          sha512sum "${tar_gz}" > "${tar_gz}.sha512"
+      - name: Audit
+        run: |
+          dev/release/run_rat.sh "${TAR_GZ}"
+      - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+        with:
+          name: archive
+          path: |
+            apache-arrow-go-*
+  verify:
+    name: Verify
+    needs:
+      - archive
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: recursive
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: archive
+      - name: Verify
+        run: |
+          tar_gz=$(echo apache-arrow-go-*.tar.gz)
+          version=${tar_gz#apache-arrow-go-}
+          version=${version%.tar.gz}
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            rc="${GITHUB_REF_NAME#*-rc}"
+          else
+            rc=100
+          fi
+          VERIFY_DEFAULT=0 dev/release/verify_rc.sh "${version}" "${rc}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  upload:
+    name: Upload
+    if: github.ref_type == 'tag'
+    needs:
+      - verify
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: recursive
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: archive
+      - name: Upload
+        run: |
+          # TODO: Add support for release note
+          gh release create ${GITHUB_REF_NAME} \
+            --prerelease \
+            --title "Apache Arrow Go ${GITHUB_REF_NAME}" \
+            --verify-tag \
+            apache-arrow-go-*.tar.gz \
+            apache-arrow-go-*.tar.gz.sha*
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -51,7 +51,8 @@ jobs:
         if: github.ref_type == 'branch'
         run: |
           version=$(grep -o '^const PkgVersion = ".*"' "arrow/doc.go" |
-            sed -e 's/^const PkgVersion = "//' \
+            sed \
+              -e 's/^const PkgVersion = "//' \
               -e 's/"$//')
           rc=100
           echo "VERSION=${version}" >> ${GITHUB_ENV}
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -54,7 +54,7 @@ jobs:
             sed \
               -e 's/^const PkgVersion = "//' \
               -e 's/"$//')
-          rc=100
+          rc=$(date +%Y%m%d)
           echo "VERSION=${version}" >> ${GITHUB_ENV}
           echo "RC=${rc}" >> ${GITHUB_ENV}
       - name: Archive
@@ -97,10 +97,11 @@ jobs:
           tar_gz=$(echo apache-arrow-go-*.tar.gz)
           version=${tar_gz#apache-arrow-go-}
           version=${version%.tar.gz}
+          # rc isn't used with VERIFY_DOWNLOAD=0
           if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
             rc="${GITHUB_REF_NAME#*-rc}"
           else
-            rc=100
+            rc=$(date +%Y%m%d)
           fi
           VERIFY_DEFAULT=0 dev/release/verify_rc.sh "${version}" "${rc}"
         env:
--- a/arrow/doc.go
+++ b/arrow/doc.go
@@ -34,7 +34,7 @@ To build with tinygo include the noasm build tag.
 */
 package arrow
 
-const PkgVersion = "18.0.0-SNAPSHOT"
+const PkgVersion = "18.0.0"
 
 //go:generate go run _tools/tmpl/main.go -i -data=numeric.tmpldata type_traits_numeric.gen.go.tmpl type_traits_numeric.gen_test.go.tmpl array/numeric.gen.go.tmpl array/numericbuilder.gen.go.tmpl array/bufferbuilder_numeric.gen.go.tmpl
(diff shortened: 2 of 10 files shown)


kou:
Example RC:
* https://github.com/kou/arrow-go/releases/tag/v18.0.0-rc7
* https://github.com/kou/arrow-go/actions/runs/10826402849

arrow/doc.go
```diff
@@ -34,7 +34,7 @@ To build with tinygo include the noasm build tag.
 */
 package arrow
 
-const PkgVersion = "18.0.0-SNAPSHOT"
+const PkgVersion = "18.0.0"
```
raulcd, line 37: this doesn't have to change for this PR, only when releasing, right? or do we want this to be `17.0.0` and increment on the new release?
I think we should probably continue using `SNAPSHOT` and just remove the `SNAPSHOT` bit when releasing.

.github/workflows/rc.yml
```diff
@@ -0,0 +1,133 @@
...
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Prepare for branch
+        if: github.ref_type == 'branch'
+        run: |
+          version=$(grep -o '^const PkgVersion = ".*"' "arrow/doc.go" |
+            sed -e 's/^const PkgVersion = "//' \
+              -e 's/"$//')
+          rc=100
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
```
raulcd, line 58: I am not sure I understand why this is triggered for a branch.

.github/workflows/rc.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: |
+          tar_gz=$(echo apache-arrow-go-*.tar.gz)
+          version=${tar_gz#apache-arrow-go-}
+          version=${version%.tar.gz}
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            rc="${GITHUB_REF_NAME#*-rc}"
+          else
+            rc=100
+          fi
+          VERIFY_DEFAULT=0 dev/release/verify_rc.sh "${version}" "${rc}"
```
raulcd, line 104: it took me a while to understand how were you verifying without having uploaded the release yet and then I saw the `VERIFY_DEFAULT=0` :smile:

dev/release/README.md
```diff
@@ -0,0 +1,116 @@
...
+## Overview
+
+  1. Test the revision to be released
+  2. Increment version number in `arrow/doc.go`
+  3. Prepare RC and vote (detailed later)
+  4. Publish (detailed later)
+
+### Prepare RC and vote
+
+Run `dev/release/release_rc.sh` on working copy of `git@github.com:apache/arrow-go` not your fork:
```
raulcd, line 31:
```suggestion
Run `dev/release/release_rc.sh` on a working copy of `git@github.com:apache/arrow-go` not your fork:
```

dev/release/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+```console
+$ git clone git@github.com:apache/arrow-go.git
+$ dev/release/release_rc.sh ${RC}
```
raulcd, line 35: this makes sense. As suggested below I would just go to create an apache upstream so we don't require a new clone and people can use their custom workflow without having to call the apache upstream origin, but I am fine with this

dev/release/README.md
```diff
@@ -0,0 +1,116 @@
...
+Run `dev/release/release.sh` to publish to apache.org:
+
+```console
+$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/release.sh ${VERSION} ${RC}
+```
+
+Here is an example to release 18.0.0 RC1:
+
+```console
+$ GH_TOKEN=${YOUR_GITHUB_TOKEN}dev/release/release.sh 18.0.0 1
```
raulcd, line 86:
```suggestion
$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/release.sh 18.0.0 1
```

dev/release/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+You don't need to install Go. If Go doesn't exist in system, the latest Go is automatically installed only for verification.
+
+To verify a RC, run the following command line:
+
+```console
+$ dev/release/verify_rc.sh ${VERSION} ${RC}
+```
+
+Here is an example to release 18.0.0 RC1:
```
raulcd, line 110:
```suggestion
Here is an example to verify the release 18.0.0 RC1:
```

dev/release/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+```console
+$ dev/release/verify_rc.sh 18.0.0 1
+```
+
+If the verification is succeeded, `RC looks good!` is shown.
```
raulcd, line 116:
```suggestion
If the verification is successful, the message `RC looks good!` is shown.
```

dev/release/release.sh
```diff
@@ -0,0 +1,83 @@
...
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <version> <rc>"
+  echo " e.g.: $0 18.0.0 1"
+  exit 1
+fi
+
+version=$1
+rc=$2
+
+git_origin_url="$(git remote get-url origin)"
```
raulcd, line 31: I would probably use `apache` as the remote as suggested below

raulcd (commented):
Some minor comments but looks good! Great work @kou I've tested the verification script locally against the uploaded RC to your fork and have validated successfully.

arrow/doc.go
kou: > this doesn't have to change for this PR, only when releasing, right?

No.

> or do we want this to be 17.0.0 and increment on the new release?

No.

My suggestions:
* Don't use `-SNAPSHOT` suffix for unreleased version
  * I think that it's needless because we don't create a nightly (or something) package from main
  * I think that the `-SNAPSHOT` suffix name is Maven convention: https://maven.apache.org/maven-release/maven-release-plugin/usage/prepare-release.html
    * If we want to use `-XXX` suffix for unreleased version, other name such as `-dev` is better
* We don't update this just after we release a new version
  * Because we can't know the next version (`18.0.1`? `18.1.0`? `19.0.0`?) just after we release a new version
* We update this just before we release a new version, when we add a new compatible feature (minor version up) or when we add an incompatible change (major version up)

dev/release/README.md
kou: Thanks!

dev/release/README.md
kou: The `apache` remote works but I think that it's a bit risky.
If we have the `apache` remote in our usual clone, we may push to the "apache" remote accidentally.

FYI: I use `https://github.com/apache/arrow.git` for `apache` remote in my kou/arrow clone. I change it to `git@github.com:apache/arrow.git` only when I need to push to the `apache` remote.

dev/release/README.md
kou: Good catch!

dev/release/README.md
kou: Thanks!

dev/release/README.md
kou: Thanks!

dev/release/release_rc.sh
```diff
@@ -0,0 +1,141 @@
...
+: "${RELEASE_DEFAULT:=1}"
+: "${RELEASE_PULL:=${RELEASE_DEFAULT}}"
+: "${RELEASE_PUSH_TAG:=${RELEASE_DEFAULT}}"
+: "${RELEASE_SIGN:=${RELEASE_DEFAULT}}"
+: "${RELEASE_UPLOAD:=${RELEASE_DEFAULT}}"
+
+cd "${SOURCE_TOP_DIR}"
+
+if [ "${RELEASE_PULL}" -gt 0 ] || [ "${RELEASE_PUSH_TAG}" -gt 0 ]; then
+  git_origin_url="$(git remote get-url origin)"
```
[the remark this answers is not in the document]
kou: I think that the `apache` remote approach is a bit risky as I said above.

dev/release/release_rc.sh
```diff
@@ -0,0 +1,141 @@
...
+
+if [ "${RELEASE_PULL}" -gt 0 ]; then
+  echo "Ensure using the latest commit"
+  git checkout main
+  git pull --rebase --prune
+fi
+
+version=$(grep -o '^const PkgVersion = ".*"' "arrow/doc.go" |
+  sed -e 's/^const PkgVersion = "//' \
+    -e 's/"$//')
```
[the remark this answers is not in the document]
kou: Ah, we need it mainly for major version up. We need to change version numbers in package name:
`github.com/apache/arrow-go/v18` ->
`github.com/apache/arrow-go/v19`

Let's work on it as a follow-up task.

dev/release/verify_rc.sh
```diff
@@ -0,0 +1,227 @@
...
+VERIFY_SUCCESS=no
+
+setup_tmpdir() {
+  cleanup() {
+    go clean -modcache || :
+    if [ "${VERIFY_SUCCESS}" = "yes" ]; then
+      rm -rf "${VERIFY_TMPDIR}"
+    else
+      echo "Failed to verify release candidate. See ${VERIFY_TMPDIR} for details."
+    fi
```
zeroshade, line 55: instead of calling `go clean -modcache` we could try just setting the `GOCACHE` and `GOMODCACHE` env vars to point to directories in the tmpdir. That way we still get clean caches, but we don't blow up the users cache dirs.

dev/release/verify_rc.sh
```diff
@@ -0,0 +1,227 @@
...
+    git clone \
+      https://github.com/apache/parquet-testing.git \
+      "${ARCHIVE_BASE_NAME}/parquet-testing"
+  fi
+
+  ARROW_TEST_DATA="${ARCHIVE_BASE_NAME}/arrow-testing/data"
+  export ARROW_TEST_DATA
+  PARQUET_TEST_DATA="${ARCHIVE_BASE_NAME}/parquet-testing/data"
+  export PARQUET_TEST_DATA
+}
```
zeroshade, line 133: Need to add `PARQUET_TEST_BAD_DATA` now as per https://github.com/apache/arrow-go/pull/124

zeroshade (commented):
just a few nitpicks from me

.github/workflows/rc.yml
```diff
@@ -0,0 +1,134 @@
...
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Prepare for branch
+        if: github.ref_type == 'branch'
+        run: |
+          version=$(grep -o '^const PkgVersion = ".*"' "arrow/doc.go" |
+            sed \
+              -e 's/^const PkgVersion = "//' \
+              -e 's/"$//')
+          rc=100
```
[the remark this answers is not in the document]
kou: I just want to use RC that is never used for normal RCs.
We can use `1000`, `$(date %Y%m%d)` or something instead of `100` here.

dev/release/verify_rc.sh
kou: Removing cache isn't the main purpose of it.
It's for avoiding a permission problem. See also: https://github.com/apache/arrow/pull/38222

BTW, do we need to set `GOCACHE`/`GOMODCACHE` explicitly when we set `GOPATH`? Can `GOCACHE`/`GOMODCACHE` be set automatically under `GOPATH`?

dev/release/verify_rc.sh
kou: OK. I'll add it.

arrow/doc.go
raulcd: sounds good to me to update just before releasing

raulcd (approved):
Thanks @kou

dev/release/README.md
raulcd: > FYI: I use `https://github.com/apache/arrow.git` for `apache` remote in my kou/arrow clone. I change it to `git@github.com:apache/arrow.git` only when I need to push to the `apache` remote.

This is a good idea, I've updated to do the same. Thanks!

.github/workflows/rc.yml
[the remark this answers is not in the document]
zeroshade: ah, gotcha. `$(date %Y%m%d)` is probably best i think?

dev/release/verify_rc.sh
zeroshade: Ah gotcha. 

> BTW, do we need to set `GOCACHE`/`GOMODCACHE` explicitly when we set `GOPATH`? Can `GOCACHE`/`GOMODCACHE` be set automatically under `GOPATH`?

`GOMODCACHE` should generally be `${GOPATH}/pkg/mod`, we might not need to change it explicitly since we set `GOPATH`.

`GOCACHE` can be anywhere though and typically is set to `~/.cache/go-build`, if we're not running `go clean -cache` or otherwise cleaning the build cache, and only using `go clean -modcache` then we're probably fine here without issue.

zeroshade (approved):

.github/workflows/rc.yml
```diff
@@ -0,0 +1,134 @@
...
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: recursive
+      - name: Prepare for tag
+        if: github.ref_type == 'tag'
```
zeroshade, line 43: should this be `${{ github.ref_type }}` ?

kou: We can omit `${{ ... }}`: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif

> When you use expressions in an if conditional, you can, optionally, omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. However, this exception does not apply everywhere.
>
> You must always use the ${{ }} expression syntax or escape with '', "", or () when the expression starts with !, since ! is reserved notation in YAML format.

.github/workflows/rc.yml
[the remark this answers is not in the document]
kou: OK. I'll use it.

dev/release/verify_rc.sh
kou: OK. Let's use `go clean -modcache` here.

kou:
I'll merge this.

Issue #107, opened 2024-09-09
Add test CI for arm Runners


raulcd:
We have some jobs that run on ARM self-hosted runners, we should also add those: https://github.com/apache/arrow/blob/main/.github/workflows/go.yml#L80-L91

I've created a JIRA issue asking INFRA to add this repository to the runner group so we can use the self-hosted runners that we share with the rest of repositories (arrow, arrow-nanoarrow and arrow-adbc): https://issues.apache.org/jira/browse/INFRA-26098

kou:
Thanks!

FYI: We may be able to use GitHub hosted Arm runner in this year: https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

raulcd:
Yes, as soon as they are available we should migrate the existing jobs that use self-hosted arm runners.

PR #123, opened 2024-09-13
GH-107: Add test CI for arm Runners

raulcd:
Fix https://github.com/apache/arrow-go/issues/107

We have some jobs that run on ARM self-hosted runners, we should also add those: https://github.com/apache/arrow/blob/main/.github/workflows/go.yml#L80-L91

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,6 +50,18 @@ jobs:
             "arch": "amd64",
             "go": "1.23",
             "runs-on": "ubuntu-latest"
+          },
+          {
+            "arch-label": "ARM64",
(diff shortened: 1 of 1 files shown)


raulcd:
https://github.com/apache/arrow-go/actions/runs/10846979543/job/30101080031?pr=123

kou (approved):
+1

kou:
Ah, we should have check `${GITHUB_REPOSITORY_OWNER}` like https://github.com/apache/arrow/blob/62060ab55be2dc811e5abf0c5aba9834d3056ee1/.github/workflows/go.yml#L77-L93 to prevent running these jobs on forks.

PR #124, opened 2024-09-13
[Parquet] Recover from panic in file reader

zeroshade:
see https://github.com/apache/arrow/pull/43607

original author: @don4get


kou:
Fixes apache/arrow#43605.

kou:
FYI: We can use `Co-authored-by` trailer for this case. See also: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

PR #128, opened 2024-09-16
MINOR: Bump CI test timeouts

zeroshade:
Looks like some of the test runs (particularly the ARM runs) are close to the 15 minute timeout and end up occasionally being flaky and failing. Let's bump the timeout to 20 minutes just in case so that we aren't skating the edge anymore.


kou (approved):
+1

Issue #3, opened 2024-08-29
Update license information


kou:
It seems that we can simplify https://github.com/apache/arrow/blob/main/go/LICENSE.txt because it has unrelated notes like: https://github.com/apache/arrow/blob/58415d1fac50cb829b3dcf08526033d6db8c30db/go/LICENSE.txt#L204-L240

PR #130, opened 2024-09-17
GH-3: Update license information

kou:
Fix GH-3

The current LICENSE.txt is based on apache/arrow's one. It includes unrelated notes for go/.


zeroshade (approved):

LICENSE.txt
```diff
@@ -203,203 +203,6 @@
...
-LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
-LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
-OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
-ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
---------------------------------------------------------------------------------
-
 This project includes code from the Go project, BSD 3-clause license + PATENTS
 weak patent termination clause
 (https://github.com/golang/go/blob/master/PATENTS).
```
kou, line 208: @zeroshade @joellubi Do you think that we still have code from the Go project?

This was added by https://github.com/apache/arrow/pull/2221 for the initial Go implementation: https://github.com/apache/arrow/pull/1739

zeroshade: we definitely still do for various reasons.

Some pieces of code that are adapted from internal Go things, plus when you build it it's statically built so the final binaries will include code from the Go project etc. 

This hasn't changed

kou: I see! Thanks!

Issue #131, opened 2024-09-17
Don't use self-hosted arm runner on forks


kou:
It's not available on forks.

See also: https://github.com/apache/arrow-go/pull/123#issuecomment-2350607454

PR #132, opened 2024-09-17
GH-131: Don't use self-hosted arm runner on forks

kou:
Fix GH-131

It's not available on forks.


kou:
+1

A fork doesn't have "ARM64 *":  https://github.com/kou/arrow-go/actions/runs/10895823472
apache/arrow-go has "ARM64 *": https://github.com/apache/arrow-go/actions/runs/10895825334/job/30234685059?pr=132

Issue #134, opened 2024-09-22
Add Validity compute functions


zeroshade:
We should add the validity compute methods such as IsNull, IsNotNull and IsNaN to our implemented functions, along with the corresponding mappings for substrait expressions.

zeroshade:
Closed by #135

PR #135, opened 2024-09-22
GH-134: Add Validity functions

zeroshade:
Adding `is_null`, `is_not_null`, and `is_nan` functions and tests to the Compute implementations for Go.

This also removes the `-ffast-math` option from the SIMD optimized compiler Makefile so that  our comparisons properly handle `NaN` values (`-ffast-math` assumes that no values are `NaN` and removes the checks). `is_nan` can be easily implemented for float and double values by dispatching to `A != A`.

Tests are also added for the new functions.

Ultimately the driving force behind this is as these are necessary for implementing scanning reads in iceberg-go


lidavidm (approved):

arrow/array/struct.go
```diff
@@ -43,6 +43,31 @@ func NewStructArray(cols []arrow.Array, names []string) (*Struct, error) {
...
+	for i, c := range cols {
+		if length != c.Len() {
+			return nil, fmt.Errorf("%w: mismatching child array lengths", arrow.ErrInvalid)
+		}
+		children[i] = c.Data()
+	}
+
+	data := NewData(arrow.StructOf(fields...), length, []*memory.Buffer{nil}, children, 0, 0)
+	defer data.Release()
+	return NewStructData(data), nil
```
bkietz, line 68: Does this check that the types of `cols` match those of `fields`?

arrow/compute/scalar_compare.go
```diff
@@ -134,4 +136,50 @@ func RegisterScalarComparisons(reg FunctionRegistry) {
...
+		func(ctx context.Context, opts FunctionOptions, args ...Datum) (Datum, error) {
+			switch args[0].Kind() {
+			case KindScalar:
+				arg := args[0].(*ScalarDatum)
+				switch arg.Type() {
+				case arrow.PrimitiveTypes.Float32, arrow.PrimitiveTypes.Float64:
+					// IEEE 754 says that only NAN satisfies f != f
+					return CallFunction(ctx, "not_equal", nil, arg, arg)
+				default:
+					return NewDatum(true), nil
```
bkietz, line 164: Doesn't this mean we always return `is_nan=true` for integers? This doesn't seem to match the array case below
```suggestion
return NewDatum(false), nil
```

arrow/compute/scalar_compare.go
```diff
@@ -134,4 +136,50 @@ func RegisterScalarComparisons(reg FunctionRegistry) {
...
+	if err := isNotNullFn.AddKernel(isOrNotNullKns[1]); err != nil {
+		panic(err)
+	}
+
+	reg.AddFunction(isNullFn, false)
+	reg.AddFunction(isNotNullFn, false)
+
+	reg.AddFunction(NewMetaFunction("is_nan", Unary(), EmptyFuncDoc,
+		func(ctx context.Context, opts FunctionOptions, args ...Datum) (Datum, error) {
+			switch args[0].Kind() {
```
bkietz, line 156: Nit: could you reoder the type switch to be outside the kind switch? Then we have one `return CallFunction(ctx, "not_equal", nil, arg, arg)` for all kinds of floats and we either return scalar false or broadcast it for other types

bkietz (changes requested):

arrow/compute/scalar_compare.go
```diff
@@ -134,4 +136,50 @@ func RegisterScalarComparisons(reg FunctionRegistry) {
 	reg.AddFunction(ltFn, false)
 	lteFn := makeFlippedCompare("less_equal", gteFn, EmptyFuncDoc)
 	reg.AddFunction(lteFn, false)
+
+	isOrNotNullKns := kernels.IsNullKernels()
```
kou, line 140: `isNullKns` or `isNullOrNotNullKns`?

kou (commented):
Could you check CI failures?

arrow/array/struct.go
zeroshade: good point, i've added the check

arrow/compute/scalar_compare.go
zeroshade: good catch, i screwed that up. it should be returning a scalar false

arrow/compute/scalar_compare.go
zeroshade: reordered and updated

arrow/compute/scalar_compare.go
zeroshade: renamed `kernels.IsNullKernels()` to `kernels.IsNullNotNullKernels()` to make it more obvious

zeroshade:
@kou it took a while, but i figured out the cause of the failures. it was the upgrade to the new version of `xstrings`. v1.4.0 of xstrings.ToCamelCase capitalizes the first letter of the word, while v1.5.0 of ToCamelCase *doesn't* capitalize the first letter. They really should have marked that as a breaking change :( I think i'll look for something other than xstrings to perform the camel casing there so we don't have to rely on an old version of a package

bkietz (approved):

arrow/compute/scalar_compare.go
kou: Ah, sorry. I mentioned a local variable name not a function name here.

kou:
The lint failure was caused by mixing Go 1.22 and 1.23. See also: https://github.com/golangci/golangci-lint/issues/4908

I've pushed a commit to ensure using Go 1.23 for pre-commit.

PR #162, opened 2024-10-17
GH-20: Add CONTRIBUTING.md

### LINKED ISSUE #20: Add CONTRIBUTING.md
We must follow "The Apache Way": https://www.apache.org/theapacheway/

New contributors may not know The Apache Way. We must explain it to new contributors.

apache/arrow's one: https://github.com/apache/arrow/blob/main/CONTRIBUTING.md

zeroshade:
Fixes #20

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+<!---
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# How to contribute to Apache Arrow Go
+
+## Did you find a bug?
+
+The Arrow project uses GitHub as a bug tracker.  To report a bug, sign in to
+your GitHub account, navigate to [GitHub issues](https://github.com/apache/arrow-go/issues)
+and click on **New issue** .
+
+To be assigned to an issue, add a comment "take" to that issue.
+
+Before you create a new bug entry, we recommend you first search among existing
+Arrow issues in [GitHub](https://github.com/apache/arrow-go/issues).
+
+We conventionally prefix the issue title with the component
+name in brackets, such as "[Parquet] Fix panic in record read", 
+so as to make lists more easy to navigate, and
+we'd be grateful if you did the same.
+
+## Did you write a patch that fixes a bug or brings an improvement?
+
+First create a GitHub issue as described above, selecting **Bug Report** or
+**Enhancement Request**. Then, submit your changes as a GitHub Pull Request.
+We'll ask you to prefix the pull request title with the GitHub issue number
+and the component name in brackets. (for example: "GH-100: [Parquet]
+fix panic in record read"). Respecting this convention
+makes it easier for us to process the backlog of submitted Pull Requests.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+# How to contribute to Apache Arrow Go
+
+## Did you find a bug?
+
+The Arrow project uses GitHub as a bug tracker.  To report a bug, sign in to
+your GitHub account, navigate to [GitHub issues](https://github.com/apache/arrow-go/issues)
+and click on **New issue** .
+
+To be assigned to an issue, add a comment "take" to that issue.
```
kou, line 28: We don't have this feature in apache/arrow-go.
Do we need this? If we need this, we need to port https://github.com/apache/arrow/blob/0c3206772051c52fe2e648623daaca2c4e71cb72/.github/workflows/comment_bot.yml#L178-L194 to apache/arrow-go.

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+We conventionally prefix the issue title with the component
+name in brackets, such as "[Parquet] Fix panic in record read", 
+so as to make lists more easy to navigate, and
+we'd be grateful if you did the same.
+
+## Did you write a patch that fixes a bug or brings an improvement?
+
+First create a GitHub issue as described above, selecting **Bug Report** or
+**Enhancement Request**. Then, submit your changes as a GitHub Pull Request.
```
kou, line 41: We don't have issue templates.
If we need them, we need to port https://github.com/apache/arrow/tree/main/.github/ISSUE_TEMPLATE to apache/arrow-go.

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+prefix the title of your PR with "MINOR: " if meets one of the following:
+
+*  Grammar, usage and spelling fixes that affect no more than 2 files
+*  Documentation updates affecting no more than 2 files and not more
+   than 500 words.
+
+## Do you want to propose a significant new feature or an important refactoring?
+
+We ask that all discussions about major changes in the codebase happen
+publicly on the [arrow-dev mailing-list](https://mail-archives.apache.org/mod_mbox/arrow-dev/).
```
kou, line 60: https://arrow.apache.org/community/#mailing-lists may be better because it has a bit more description.

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+we'd be grateful if you did the same.
+
+## Did you write a patch that fixes a bug or brings an improvement?
+
+First create a GitHub issue as described above, selecting **Bug Report** or
+**Enhancement Request**. Then, submit your changes as a GitHub Pull Request.
+We'll ask you to prefix the pull request title with the GitHub issue number
+and the component name in brackets. (for example: "GH-100: [Parquet]
+fix panic in record read"). Respecting this convention
+makes it easier for us to process the backlog of submitted Pull Requests.
```
kou, line 45: We can stop this convention if we want. I think that this convention is to associate Jira issue easily.
We can use `Fixes #XXX` PR description in GitHub instead.

CONTRIBUTING.md
kou: #163

CONTRIBUTING.md
zeroshade: I'd be more than willing to switch arrow-go to instead use conventional commits like ADBC does, what do you think @kou?

zeroshade: I'd be able to update the automation to have the semantic versioning automatically handled via those commit messages

CONTRIBUTING.md
zeroshade: I'd be more than willing to update to instead use conventional commits and then I can update the automation to auto generate the semantic versioning from the conventional commit messages. Thoughts?

CONTRIBUTING.md
zeroshade: I think it might be a good idea for us to port that, it's likely useful

CONTRIBUTING.md
kou: Like ADBC does, right?
It's a good idea.

CONTRIBUTING.md
kou: OK. Let's port it: GH-168

kou (approved):
+1

PR #164, opened 2024-10-17
GH-23: Add PR template

### LINKED ISSUE #23: Add pull request template
If we need it.

See also apache/arrow's one: https://github.com/apache/arrow/blob/main/.github/pull_request_template.md

zeroshade:
Fixes #23

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,60 +1,11 @@
-<!--
-Thanks for opening a pull request!
-If this is your first pull request you can find detailed information on how 
-to contribute here:
-  * [New Contributor's Guide](https://arrow.apache.org/docs/dev/developers/guide/step_by_step/pr_lifecycle.html#reviews-and-merge-of-the-pull-request)
-  * [Contributing Overview](https://arrow.apache.org/docs/dev/developers/overview.html)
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,60 @@
+<!--
```
kou, line 1: How about using raw text instead of comment text?

It seems that most developers don't remove comment in pull request template.
In apache/arrow, we remove comments in our merge script https://github.com/apache/arrow/blob/0c3206772051c52fe2e648623daaca2c4e71cb72/dev/merge_arrow_pr.py#L603-L604 but we don't use merge script in apache/arrow-go. So our commit history have many this comments...

If we use raw text instead of comment text, we can notice it before we use the merge button. We can remove needless texts before we merge.

kou (commented):
We need `Fixes #XXX` in PR description because we don't use a merge script.

zeroshade:
@kou how about a simple template with just the headings? I think we can give enough faith in developers to understand the sections without worrying about them having to remember to delete sample text?

kou (approved):
+1

Let's try the approach!

PR #165, opened 2024-10-17
GH-41: [Array] Make String and Binary consistent

### LINKED ISSUE #41: [Go] *array.Binary.ValueOffset doesn't work correctly

https://github.com/apache/arrow/blob/main/go/arrow/array/binary.go#L76 -- this place is incorrect. For example, valueOffsets contain 6 offsets for the array with 5 elements, but calling array.ValueOffset(5) will cause the panic although the element exists.

It must be 
```
func (a *Binary) ValueOffset(i int) int {
	if i < 0 || i > a.array.data.length {
		panic("arrow/array: index out of range")
	}
	return int(a.valueOffsets[a.array.data.offset+i])
}
```

similar to strings https://github.com/apache/arrow/blob/main/go/arrow/array/string.go#L66.

### Component(s)

Go

zeroshade:
Fixes #41 

Makes the behavior of `ValueOffset` consistent between String and Binary arrays.

### DIFF
--- a/arrow/array/string_test.go
+++ b/arrow/array/string_test.go
@@ -90,9 +90,6 @@ func TestStringArray(t *testing.T) {
 		if got, want := arr.ValueOffset(i), int(offsets[i]); got != want {
 			t.Fatalf("arr-offset-beg[%d]: got=%d, want=%d", i, got, want)
 		}
-		if got, want := arr.ValueOffset(i+1), int(offsets[i+1]); got != want {
-			t.Fatalf("arr-offset-end[%d]: got=%d, want=%d", i+1, got, want)
-		}
 	}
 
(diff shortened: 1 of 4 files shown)


arrow/array/string_test.go
```diff
@@ -90,9 +90,6 @@ func TestStringArray(t *testing.T) {
 		if got, want := arr.ValueOffset(i), int(offsets[i]); got != want {
 			t.Fatalf("arr-offset-beg[%d]: got=%d, want=%d", i, got, want)
 		}
-		if got, want := arr.ValueOffset(i+1), int(offsets[i+1]); got != want {
```
lidavidm, line 93: I guess we could validate ValueLen instead? Though maybe not really worth it

lidavidm (approved):

PR #170, opened 2024-10-23
fix: don't clobber saved frame pointer in arm64 assembly functions

### LINKED ISSUE #150: [Go] arm64 assembly clobbers the saved frame pointer register
This was originally https://github.com/apache/arrow/issues/44319, but the Go code has moved to this repo

The various arm64 assembly functions used by the Go library, e.g. [`_memset_neon`](https://github.com/apache/arrow/blob/3fb7777a9a656e8bcc6f59396c4a5f6024a5f78e/go/arrow/memory/memory_neon_arm64.s#L8), clobber the frame pointer saved by their caller. This causes (infrequent) crashes when running the Go runtime execution tracer, which uses frame pointer unwinding to collect call stacks.

For example, the following program reliably crashes with https://pkg.go.dev/github.com/apache/arrow/go/v17 on my arm64 macos laptop, and probably would on an arm64 linux machine as well:

```go
package main

import (
	"io"
	"runtime/trace"

	"github.com/apache/arrow/go/v17/arrow/memory"
)

func main() {
	trace.Start(io.Discard)
	defer trace.Stop()

	var b [8]byte
	for {
		memory.Set(b[:], 0xff)
	}
}
```

The crash looks like this

```
SIGSEGV: segmentation violation
PC=0x105057880 m=0 sigcode=2 addr=0x9100c3ffa97ffc05

goroutine 0 gp=0x105137740 m=0 mp=0x105137d80 [idle]:
runtime.fpTracebackPCs(...)
	/usr/local/go/src/runtime/tracestack.go:258
runtime.traceStack(0x105137d80?, 0x10504b634?, 0x1)
	/usr/local/go/src/runtime/tracestack.go:116 +0x260 fp=0x16ae02d20 sp=0x16ae028b0 pc=0x105057880
runtime.traceLocker.stack(...)
	/usr/local/go/src/runtime/traceevent.go:176
runtime.traceLocker.GoStop({0x105137d80?, 0x1?}, 0x2)
	/usr/local/go/src/runtime/traceruntime.go:478 +0x78 fp=0x16ae02db0 sp=0x16ae02d20 pc=0x105056ab8
runtime.traceLocker.GoPreempt(...)
	/usr/local/go/src/runtime/traceruntime.go:473
runtime.goschedImpl(0x140000021c0, 0x1?)
	/usr/local/go/src/runtime/proc.go:4117 +0x88 fp=0x16ae02e00 sp=0x16ae02db0 pc=0x105037a88
runtime.gopreempt_m(...)
	/usr/local/go/src/runtime/proc.go:4153
runtime.newstack()
	/usr/local/go/src/runtime/stack.go:1084 +0x2cc fp=0x16ae02f40 sp=0x16ae02e00 pc=0x1050476dc
runtime.morestack()
	/usr/local/go/src/runtime/asm_arm64.s:342 +0x70 fp=0x16ae02f40 sp=0x16ae02f40 pc=0x105066310

goroutine 1 gp=0x140000021c0 m=0 mp=0x105137d80 [running]:
github.com/apache/arrow/go/v17/arrow/memory.memory_memset_neon({0x1400009a020?, 0x8?, 0x8?}, 0xff?)
	/Users/nick.ripley/go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/arrow/memory/memory_neon_arm64.go:27 +0x64 fp=0x140000a0ef0 sp=0x140000a0ef0 pc=0x105081d44
github.com/apache/arrow/go/v17/arrow/memory.Set(...)
	/Users/nick.ripley/go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/arrow/memory/memory.go:25
main.main()
	/Users/nick.ripley/sandbox/go/badfpasm/arrow/main.go:16 +0x74 fp=0x140000a0f40 sp=0x140000a0ef0 pc=0x105081e44
runtime.main()
	/usr/local/go/src/runtime/proc.go:272 +0x288 fp=0x140000a0fd0 sp=0x140000a0f40 pc=0x10502f338
runtime.goexit({})
	/usr/local/go/src/runtime/asm_arm64.s:1223 +0x4 fp=0x140000a0fd0 sp=0x140000a0fd0 pc=0x105068574
[ ... more elided ... ]
```

The [Go internal ABI for arm64](https://go.googlesource.com/go/+/refs/tags/go1.23.1/src/cmd/compile/abi-internal.md#arm64-architecture) saves the frame pointer register one word below the stack pointer, i.e. one word below the call frame. The Go arm64 assembly functions in this repo with call frames save the frame pointer and link register at the _top_ of their call frame, and in particular the link register goes over where the Go caller would have saved its caller's frame pointer. [See this line in `_memset_neon` for example](https://github.com/apache/arrow/blob/3fb7777a9a656e8bcc6f59396c4a5f6024a5f78e/go/arrow/memory/memory_neon_arm64.s#L14). If the Go code is preempted by the runtime while execution tracing is active, after the assembly function returns, but before its caller (and maybe caller's caller?) returns, frame pointer unwinding will visit the invalid value.

Those files appear to be automatically translated from compiled C files, and presumably the C compiler isn't aware of Go's internal ABI. And https://github.com/minio/asm2plan9s is archived... Perhaps the assembly functions can be modified manually? For example, they could decrement the stack pointer to make room for the caller's saved frame pointer:

```diff
diff --git a/go/arrow/memory/memory_neon_arm64.s b/go/arrow/memory/memory_neon_arm64.s
index 18655cc7a..40c5e87b4 100755
--- a/go/arrow/memory/memory_neon_arm64.s
+++ b/go/arrow/memory/memory_neon_arm64.s
@@ -11,6 +11,11 @@ TEXT ·_memset_neon(SB), $0-24
        MOVD    len+8(FP), R1
        MOVD    c+16(FP), R2

+       // The Go ABI saves the frame pointer register one word below the
+       // caller's frame. Make room so we don't overwrite it. Needs to stay
+       // 16-byte aligned
+       SUB $16, RSP
+
        WORD $0xa9bf7bfd // stp    x29, x30, [sp, #-16]!
        WORD $0x8b010008 // add    x8, x0, x1
        WORD $0xeb00011f // cmp    x8, x0
@@ -40,4 +45,7 @@ LBB0_6:
        BNE     LBB0_6
 LBB0_7:
        WORD $0xa8c17bfd // ldp    x29, x30, [sp], #16
+
+       // Put the stack pointer back where it was
+       ADD $16, RSP
        RET
```

This kind of fix would apply to any arm64 assembly code that touches `x29`, the frame pointer register.

### Component(s)

Go

nsrip-dd:
The arm64 neon assembly functions in this repository overwrite the frame
pointer saved by their callers, leading to crashes from the Go runtime
execution tracer and profilers which use frame pointer unwinding. For
historical reasons, on arm64 Go functions save the caller's frame
pointer register (x29) one word below their stack frame. See
go.dev/s/regabi#arm64-architecture. The assembly functions here,
translated from C compiler output, save values at the top of their
frame, and overwrite the frame pointer saved by the caller. We can fix
this by decrementing the stack pointer past where that frame pointer is
saved before saving anything on the stack.

Fixed with this sed script on my macos laptop + manual cleanup to match
indentation:

```
/stp[\t ]*x29/i\
	// The Go ABI saves the frame pointer register one word below the \
	// caller's frame. Make room so we don't overwrite it. Needs to stay \
	// 16-byte aligned \
	SUB $16, RSP

/ldp[\t ]*x29/a\
	// Put the stack pointer back where it was \
	ADD $16, RSP

```

Ran the script from the root of this repository with

	find . -name '*_arm64.s' -exec sed -f fix.sed -i '' {} +

Then manually inspected the assembly for missing SUBs/ADDs at the
beginning of functions and prior to returns.

Fixes #150

### DIFF
--- a/arrow/internal/testing/tools/fpunwind_arm64.s
+++ b/arrow/internal/testing/tools/fpunwind_arm64.s
@@ -1,3 +1,19 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 TEXT ·FPUnwind(SB), $0-0
 	MOVD R29, R19
 loop:
--- a/arrow/internal/testing/tools/fpunwind_arm64.s
+++ b/arrow/internal/testing/tools/fpunwind_arm64.s
@@ -1,0 +1,9 @@
+TEXT ·FPUnwind(SB), $0-0
+	MOVD R29, R19
+loop:
+	CMP $0, R19
+	BEQ exit
+	MOVD (R19), R19
+	B loop
+exit:
+	RET
--- a/arrow/memory/memory_test.go
+++ b/arrow/memory/memory_test.go
@@ -129,6 +129,7 @@ func TestSetNoClobberFramePointer(t *testing.T) {
 	var b [8]byte
 	memory.Set(b[:], 0xff)
 	// We should be able to safely frame pointer unwind after calling this
-	// function
+	// function. The arm64 assembly implementation used to clobber the
+	// frame pointer. See #150
 	tools.FPUnwind()
 }
--- a/arrow/memory/memory_test.go
+++ b/arrow/memory/memory_test.go
@@ -130,6 +130,6 @@ func TestSetNoClobberFramePointer(t *testing.T) {
 	memory.Set(b[:], 0xff)
 	// We should be able to safely frame pointer unwind after calling this
 	// function. The arm64 assembly implementation used to clobber the
-	// frame pointer. See #150
+	// frame pointer. See GH-150
 	tools.FPUnwind()
 }
--- a/arrow/memory/memory_test.go
+++ b/arrow/memory/memory_test.go
@@ -19,6 +19,7 @@ package memory_test
 import (
 	"testing"
 
+	"github.com/apache/arrow-go/v18/arrow/internal/testing/tools"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/stretchr/testify/assert"
 )
@@ -122,4 +123,12 @@ func BenchmarkSet_8000(b *testing.B) {
 
 func BenchmarkSet_8192(b *testing.B) {
 	benchmarkSet(b, 8192)
+}
+
+func TestSetNoClobberFramePointer(t *testing.T) {
+	var b [8]byte
+	memory.Set(b[:], 0xff)
+	// We should be able to safely frame pointer unwind after calling this
+	// function
+	tools.FPUnwind()
 }
--- a/arrow/compute/internal/kernels/cast_numeric_neon_arm64.s
+++ b/arrow/compute/internal/kernels/cast_numeric_neon_arm64.s
@@ -10,6 +10,10 @@ TEXT ·_cast_type_numeric_neon(SB), $0-40
     MOVD len+32(FP), R4
 
 
+    // The Go ABI saves the frame pointer register one word below the 
+    // caller's frame. Make room so we don't overwrite it. Needs to stay 
+    // 16-byte aligned 
+    SUB $16, RSP
     WORD $0xa9bf7bfd // stp    x29, x30, [sp, #-16]!
     WORD $0x7100181f // cmp    w0, #6
     WORD $0x910003fd // mov    x29, sp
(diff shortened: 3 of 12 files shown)


zeroshade:
@nsrip-dd We do have  CI which runs on ARM machines, could we add a test appropriately for at least one or a couple of these which uses tracing as per your original reproducer to validate that this continues to work?

nsrip-dd:
> @nsrip-dd We do have CI which runs on ARM machines, could we add a test appropriately for at least one or a couple of these which uses tracing as per your original reproducer to validate that this continues to work?

Sorry for the delay. I've added a test which I confirmed fails without my fix. Right now there's only a test for `memory.Set`. In theory we could test the other functions similarly. But I found it tricky because many of the assembly functions are a few function calls deep from the public API, and it seems like most of the tests in this repo use only the public APIs for the packages they test. Any suggestion on how to proceed? Or is what I have enough for now?

arrow/memory/memory_test.go
```diff
@@ -123,3 +124,11 @@ func BenchmarkSet_8000(b *testing.B) {
 func BenchmarkSet_8192(b *testing.B) {
 	benchmarkSet(b, 8192)
 }
+
+func TestSetNoClobberFramePointer(t *testing.T) {
+	var b [8]byte
+	memory.Set(b[:], 0xff)
+	// We should be able to safely frame pointer unwind after calling this
+	// function
+	tools.FPUnwind()
```
zeroshade, line 133: let's add a comment referencing the issue so that we understand why this test exists in the future

zeroshade (changes requested):
LGTM in general, just a couple nit picks

arrow/internal/testing/tools/fpunwind_arm64.s
```diff
@@ -0,0 +1,9 @@
+TEXT ·FPUnwind(SB), $0-0
```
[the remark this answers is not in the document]
nsrip-dd: Done

arrow/memory/memory_test.go
nsrip-dd: Done 👍

zeroshade (approved):

PR #171, opened 2024-10-25
fix: compute fixes for extension types

zeroshade:
While working on github.com/apache/iceberg-go to enable reading data, I came across a few issues when dealing with UUID / extension types and some other small compute things. This fixes those issues and adds tests to account for them.

### DIFF
--- a/arrow/compute/exprs/exec.go
+++ b/arrow/compute/exprs/exec.go
@@ -571,6 +571,25 @@ func executeScalarBatch(ctx context.Context, input compute.ExecBatch, exp expr.E
 			return nil, err
 		}
 
+		var newArgs []compute.Datum
+		// cast arguments if necessary
+		for i, arg := range args {
+			if !arrow.TypeEqual(argTypes[i], arg.(compute.ArrayLikeDatum).Type()) {
+				if newArgs == nil {
+					newArgs = make([]compute.Datum, len(args))
+					copy(newArgs, args)
+				}
+				newArgs[i], err = compute.CastDatum(ctx, arg, compute.SafeCastOptions(argTypes[i]))
+				if err != nil {
+					return nil, err
+				}
+				defer newArgs[i].Release()
+			}
+		}
+		if newArgs != nil {
+			args = newArgs
+		}
+
 		kctx := &exec.KernelCtx{Ctx: ctx, Kernel: k}
 		init := k.GetInitFn()
 		kinitArgs := exec.KernelInitArgs{Kernel: k, Inputs: argTypes, Options: opts}
@@ -613,7 +632,7 @@ func executeScalarBatch(ctx context.Context, input compute.ExecBatch, exp expr.E
 			result.Release()
 		}
 
-		return result, nil
+		return result, err
 	}
 
 	return nil, arrow.ErrNotImplemented
--- a/arrow/compute/exprs/exec.go
+++ b/arrow/compute/exprs/exec.go
@@ -630,6 +630,7 @@ func executeScalarBatch(ctx context.Context, input compute.ExecBatch, exp expr.E
 
 		if ctx.Err() == context.Canceled && result != nil {
 			result.Release()
+			result = nil
 		}
 
 		return result, err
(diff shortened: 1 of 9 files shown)


arrow/compute/exprs/exec_test.go
```diff
@@ -135,8 +137,12 @@ func TestComparisons(t *testing.T) {
...
 		two  = scalar.MakeScalar(int32(2))
 
-		str = scalar.MakeScalar("hello")
-		bin = scalar.MakeScalar([]byte("hello"))
+		str           = scalar.MakeScalar("hello")
+		bin           = scalar.MakeScalar([]byte("hello"))
+		exampleUUID   = uuid.MustParse("102cb62f-e6f8-4eb0-9973-d9b012ff0967")
+		uidStorage, _ = scalar.MakeScalarParam(exampleUUID[:],
+			&arrow.FixedSizeBinaryType{ByteWidth: 16})
+		uid = scalar.NewExtensionScalar(uidStorage, extensions.NewUUIDType())
```
kou, line 145: `uid` -> `uuid`?

arrow/compute/exprs/exec.go
```diff
@@ -613,7 +632,7 @@ func executeScalarBatch(ctx context.Context, input compute.ExecBatch, exp expr.E
 			result.Release()
```
kou, line 632: Do we need `result = nil` here? (Can we return `Release()`-ed `result`?)

kou (commented):
I'm not sure what are the problems to be fixed...

It may be better that:
* We use one pull request per problem
* We add how to reproduce the problem, actual result and expected result in issue/PR description

arrow/compute/exprs/exec_test.go
zeroshade: `uuid` is used as the name of the `uuid` package, so I'd end up with a conflict / issue if I use the name for the variable here. I can rename it as something else though if we prefer.

arrow/compute/exprs/exec.go
zeroshade: hmm, probably safer to make it `nil` you're right. just in case.

zeroshade:
> I'm not sure what are the problems to be fixed...
>
> It may be better that:
> 
> * We use one pull request per problem
> * We add how to reproduce the problem, actual result and expected result in issue/PR description

I can separate the change in the compute side from the change on the parquet side. While the changes on the compute side are fixing 2 different issues, they are intertwined and would end up having duplicate code between them if i separate them into different PRs. 

I'll split the PR into those two separate ones tomorrow.  Essentially it's related to just reading UUID fields from parquet files and utilizing UUID extension types in the compute package to perform comparisons, along with a small error propagation issue.

arrow/compute/exprs/exec_test.go
kou: Ah! `uuidStorage` and `uuidScalar` may be better. (I feel that `uid` is "user id"...)

zeroshade: will do!

lidavidm (approved):

parquet/pqarrow/schema.go
```diff
@@ -514,8 +515,10 @@ func arrowFromFLBA(logical schema.LogicalType, length int) (arrow.DataType, erro
 	switch logtype := logical.(type) {
 	case schema.DecimalLogicalType:
 		return arrowDecimal(logtype), nil
-	case schema.NoLogicalType, schema.IntervalLogicalType, schema.UUIDLogicalType:
+	case schema.NoLogicalType, schema.IntervalLogicalType:
 		return &arrow.FixedSizeBinaryType{ByteWidth: int(length)}, nil
+	case schema.UUIDLogicalType:
+		return extensions.NewUUIDType(), nil
```
joellubi, line 521: This extension type is registered by default, but we currently have some tests failing when we try to explicitly unregister this type. If we want to respect the unregistration and fallback to FixedBinary then perhaps we could use `arrow.GetExtensionType()` to get the type instead of using the constructor directly.

zeroshade:
Closing in favor of #173 and #174

parquet/pqarrow/schema.go
zeroshade: I Split this out to a separate PR as per @kou's suggestion, and I've made this change in it. It's a good point!

PR #173, opened 2024-10-25
fix(parquet): Reading UUID columns

zeroshade:
Split from #171 to be a more focused PR.

Currently we will properly write arrow data with the canonical UUID extension type as a parquet UUID column via `pqarrow`. This PR enables us to read back that data using the `extensions.UUID` data type correctly even when we don't have a stored schema.

Added a test to the `ArrowExtensionTypeRoundTrip` to ensure proper round trip without a stored schema.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -990,7 +990,7 @@ func applyOriginalStorageMetadata(origin arrow.Field, inferred *SchemaField) (mo
 			return
 		}
 
-		if modified || !arrow.TypeEqual(extType, inferred.Field.Type) {
+		if modified && !arrow.TypeEqual(extType, inferred.Field.Type) {
 			if !arrow.TypeEqual(extType.StorageType(), inferred.Field.Type) {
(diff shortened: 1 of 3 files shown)


parquet/pqarrow/schema.go
```diff
@@ -984,13 +990,15 @@ func applyOriginalStorageMetadata(origin arrow.Field, inferred *SchemaField) (mo
...
+		if modified && !arrow.TypeEqual(extType, inferred.Field.Type) {
+			if !arrow.TypeEqual(extType.StorageType(), inferred.Field.Type) {
+				return modified, fmt.Errorf("%w: mismatch storage type '%s' for extension type '%s'",
+					arrow.ErrInvalid, inferred.Field.Type, extType)
+			}
 
-		inferred.Field.Type = extType
-		modified = true
+			inferred.Field.Type = extType
+			modified = true
```
kou, line 1000: It seems that this is needless.

```suggestion
```

kou (approved):
+1

joellubi (approved):

PR #174, opened 2024-10-25
fix(arrow/compute): compare kernels with UUID

zeroshade:
Split from #171 

Enable using the comparison kernels (equal, less, less_equal, greater, greater_equal) with UUID columns and extension types in general.

Tests are added to check the kernel dispatch and to ensure compute via substrait works for UUID type scalars.


kou (approved):
+1

joellubi (approved):

PR #177, opened 2024-10-25
feat(arrow/compute): make is_nan dispatchable

zeroshade:
Currently the `is_nan` compute kernel is a `MetaFunction` which cannot be dispatched via kernel dispatch making it only usable via calling it directly with `CallFunction`. By shifting it to be a proper function instead of a `MetaFunction` this improves its compatibility and makes it able to be dispatched and thus called through the substrait interface in the `exprs` package.


lidavidm:
`is_null`, or `is_nan`?

zeroshade:
oops you're right, mis-named the PR, it's `is_nan`

joellubi (approved):

PR #178, opened 2024-10-25
feat(arrow/compute): Implement kernel for "not" function

zeroshade:
Adding implementation and tests for a "not" function to the compute package.


lidavidm (approved):

arrow/compute/scalar_bool.go
```diff
@@ -130,4 +130,6 @@ func RegisterScalarBoolean(reg FunctionRegistry) {
 		andNotKleeneDoc, exec.NullComputedPrealloc)
 	makeFunction(reg, "or_kleene", 2, kernels.SimpleBinary[kernels.KleeneOrOpKernel],
 		orKleeneDoc, exec.NullComputedPrealloc)
+	makeFunction(reg, "not", 1, kernels.NotExecKernel, EmptyFuncDoc,
```
kou, line 133: Should we add a documentation instead of `EmptyFuncDoc`?

kou (approved):
+1

PR #196, opened 2024-11-22
feat(arrow/ipc): add functions to generate payloads

zeroshade:
### Rationale for this change
Brought up in https://github.com/apache/arrow/issues/39730 requesting separate functions to generate the IPC Payloads for record batches (we have one for schemas in the Flight package, but it makes sense to have one in ipc which is more efficient).

Two functions are added, `GetRecordBatchPayload` and `GetSchemaPayload` which return ipc payload objects.

### Are these changes tested?
Yes, a unit test is added.

No.

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -603,7 +603,7 @@ func (w *recordEncoder) visit(p *Payload, arr arrow.Array) error {
 			// non-zero offset: slice the buffer
 			offset := int64(data.Offset()) * typeWidth
 			// send padding if available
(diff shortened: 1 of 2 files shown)


zeroshade:
@kou i'm not familiar enough with `brew` on macOS, any ideas why it's failing on `pkgconf`?

raulcd:
> any ideas why it's failing on `pkgconf`?

`pkg-config` was moved to `pkgconf` on brew. See the same issue PR on Arrow: https://github.com/apache/arrow/pull/44798

lidavidm (approved):

PR #197, opened 2024-11-22
feat(parquet/pqarrow): Add ForceLarge option

### LINKED ISSUE #195: array.Binary and array.String should use int64 offsets

LargeBinary and LargeString use int64 offsets, however Binary and String types use int32 offsets, this makes them susceptible to slice index out of bounds errors when the column/array is larger than ~2GB ~= 2^31 bytes.

To reproduce try deserializing a parquet file that is greater than 2.2 GB. 

A workaround is to force the go library to deserialize the field/column as LargeBinary instead of Binary:
- explicitly store the arrow schema during write. see `store_schema` https://arrow.apache.org/docs/cpp/parquet.html#roundtripping-arrow-types-and-schema 
- and schema explicitly uses the [large_binary](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv412large_binaryv) or large_string type when defining the schema that is used to write the parquet files.

Error looks like:
```go
panic: runtime error: slice bounds out of range [:-2147483014]

goroutine 95 [running]:
github.com/apache/arrow/go/v17/arrow/array.(*Binary).Value(...)
	/go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/arrow/array/binary.go:59
github.com/apache/arrow/go/v17/arrow/array.(*Binary).ValueStr(0xc000178d20?, 0xc091402a00?)
	/go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/arrow/array/binary.go:67 +0xfa
extractorvalidator/data.BootstrapRecordsFromParquet({0x1de1a40, 0xcc6a9775f0}, 0x0)
	/.../data/records.go:78 +0x582
main.validationWorker({0x1dccd90, 0x2c31840}, 0x0?, {0x0?}, 0xc0000315e0, 0xc000001de0, 0xc0000fe9c0)
	/.../command.go:428 +0x125
created by main.RunValidateCmd in goroutine 1
	/.../command.go:174 +0xb90
```

 version and platform
```
Arrow Version: github.com/apache/arrow/go/v17 v17.0.0
Platform: Linux 20.04.1-Ubuntu  x86_64 x86_64 x86_64 GNU/Linux
```

### Component(s)

Parquet, Other

zeroshade:
### Rationale for this change
closes #195

For parquet files that contain more than 2GB of data in a column, we should allow a user to force using the LargeString/LargeBinary variants without requiring a stored schema.

Adds `ForceLarge` option to `pqarrow.ArrowReadProperties` and enables it to force usage of `LargeString` and `LargeBinary` data types.

### Are these changes tested?
Yes, a unit test is added.

No breaking changes, only the addition of a new option.

### DIFF
--- a/parquet/pqarrow/properties.go
+++ b/parquet/pqarrow/properties.go
@@ -165,6 +165,11 @@ type ArrowReadProperties struct {
 	Parallel bool
 	// BatchSize is the size used for calls to NextBatch when reading whole columns
 	BatchSize int64
+	// Setting ForceLarge to true will force the reader to use LargeString/LargeBinary
+	// for string and binary columns respectively, instead of the default variants. This
+	// can be necessary if you know that there are columns which contain more than 2GB of
+	// data, which would prevent use of int32 offsets.
+	ForceLarge bool
 
 	readDictIndices map[int]struct{}
 }
--- a/parquet/pqarrow/properties.go
+++ b/parquet/pqarrow/properties.go
@@ -165,13 +165,38 @@ type ArrowReadProperties struct {
 	Parallel bool
 	// BatchSize is the size used for calls to NextBatch when reading whole columns
 	BatchSize int64
-	// Setting ForceLarge to true will force the reader to use LargeString/LargeBinary
-	// for string and binary columns respectively, instead of the default variants. This
-	// can be necessary if you know that there are columns which contain more than 2GB of
-	// data, which would prevent use of int32 offsets.
-	ForceLarge bool
 
-	readDictIndices map[int]struct{}
+	readDictIndices   map[int]struct{}
+	forceLargeIndices map[int]struct{}
+}
+
+// SetForceLarge determines whether a particular column, if it is String or Binary,
+// will use the LargeString/LargeBinary variants (with int64 offsets) instead of int32
+// offsets. This is specifically useful if you know that particular columns contain more
+// than 2GB worth of byte data which would prevent use of int32 offsets.
+//
+// Passing false will use the default variants while passing true will use the large
+// variant. If the passed column index is not a string or binary column, then this will
+// have no effect.
+func (props *ArrowReadProperties) SetForceLarge(colIdx int, forceLarge bool) {
+	if props.forceLargeIndices == nil {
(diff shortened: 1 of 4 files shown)


parquet/pqarrow/properties.go
```diff
@@ -165,6 +165,11 @@ type ArrowReadProperties struct {
 	Parallel bool
 	// BatchSize is the size used for calls to NextBatch when reading whole columns
 	BatchSize int64
+	// Setting ForceLarge to true will force the reader to use LargeString/LargeBinary
+	// for string and binary columns respectively, instead of the default variants. This
+	// can be necessary if you know that there are columns which contain more than 2GB of
+	// data, which would prevent use of int32 offsets.
+	ForceLarge bool
```
wgtmac, line 172: FYI there was an attempt in parquet-cpp but the solution might not be expected: https://github.com/apache/arrow/pull/35825

lidavidm: I agree with the original poster that it's rather weird that we can generate Parquet files that we can't otherwise read. However I also agree with Antoine that it might be good to make this option per-column if possible.

lidavidm (approved):

parquet/pqarrow/properties.go
zeroshade: Well, if you don't use this option, you can still read the parquet file, it would just require manually shrinking the batch size. I can definitely change this to make it a per-column option. That's fine, albeit a larger change since we don't currently expose which column we're determining the type for to the function that does the arrow type.

Alternately, we could utilize the column metadata for the row groups and decide ahead of time to switch to utilizing the Large variant for a column if the metadata says that it is large enough to warrant it, but that would make things really complex with row groups that may or may not be large enough to require it, etc.

The other alternative would be to forcibly reduce the batchsize when reading to accomodate? 

Thoughts?

joellubi: I also agree that automatically changing the type could be confusing. Regardless of the approach used to convert types, the behavior of automatically reducing the batch size instead of exceeding the max offset of a variable width type would be very nice IMO.

lidavidm (approved):

PR #199, opened 2024-11-22
fix(internal/json): add arrow_json_stdlib build tag

toddtreece:
### Rationale for this change

Grafana and Grafana plugins both use arrow-go, but do not use arrow's `internal/json`. A decent amount of used heap is initialized by https://github.com/goccy/go-json, so it would be useful to be able to prevent the initialization.

The example below shows that 60% of this heap profile in a Grafana plugin is used by `go-json`'s encoder & decoder packages.

image

I have submitted a [PR to go-json](https://github.com/goccy/go-json/pull/490) to switch to lazy initialization (additional details in the PR), but I am having a hard time getting a response from the maintainer, so am attempting a different approach to the problem with this PR.

Adds `arrow_json_stdlib` build tag so that it's possible to switch to `encoding/json` and avoid the overhead of https://github.com/goccy/go-json.

### Are these changes tested?

I tested this locally, but I can add coverage if this seems like an acceptable approach.

### DIFF
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -14,8 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !tinygo
-// +build !tinygo
+//go:build !tinygo && !arrow_json_stdlib
+// +build !tinygo,!arrow_json_stdlib
 
 package json
 
--- a/internal/json/json_stdlib.go
+++ b/internal/json/json_stdlib.go
@@ -14,8 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build tinygo
-// +build tinygo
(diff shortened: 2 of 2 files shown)


zeroshade:
We mostly use `goccy/go-json` for performance reasons, but this seems to be a perfectly reasonable way to handle the situation. I definitely prefer fixing it upstream in `go-json` via your PR if possible. But this is a good stop-gap.

toddtreece:
@zeroshade sounds good, thanks! i will continue trying to get the `go-json` PR merged, and if i do, i'll open another PR to update the dependency here.

toddtreece:
@zeroshade it looks like maybe you pushed a fix for the failing mac OS checks, so i rebased this against main

zeroshade (approved):

PR #200, opened 2024-11-25
ci: use correct GitHub Actions permission "contents" instead of "content"

raulcd:
### Rationale for this change

The syntax for permissions should use `contents` instead of `content`, see: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions

Update to correct GitHub actions syntax.

### Are these changes tested?

No, but can be seen the usage on the Arrow repo: https://github.com/apache/arrow/blob/bbda6b1e1322c5a9b8c107a03343d18b759694c1/.github/workflows/comment_bot.yml#L28

No

### DIFF
--- a/.github/workflows/comment_bot.yml
+++ b/.github/workflows/comment_bot.yml
@@ -25,7 +25,7 @@ jobs:
   issue_assign:
     name: "Assign issue"
     permissions:
-      content: read
+      contents: read
       issues: write
     if: github.event.comment.body == 'take'
     runs-on: ubuntu-latest


assignUser (approved):
Good catch!

assignUser:
I edited the title as afaik `fix` should only be used when it's in a part of the code base that influences the semver (as it bumps the patch component). Also @zeroshade do you merge via Github UI here?

raulcd:
> I edited the title as afaik `fix` should only be used when it's in a part of the code base that influences the semver (as it bumps the patch component).

Maybe would be nice to add some examples and/or guidelines on the contributing guide like on the angular contributing guidelines:
https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

kou (approved):
+1

raulcd (commented):
:+1:

PR #216, opened 2024-12-10
feat(arrow/ipc): implement lazy loading/zero-copy for IPC files

### LINKED ISSUE #207: [Feature Request] Support lazy load/zero copy when using shared memory

**Use case**: there is an arrow table in shared memory. It was written using Python's Arrow `RecordBatchStreamWriter` (although even if it was written using `RecordBatchFileWriter`, I believe the same problem would exist due to the current state of the Go Arrow implementation). We would like to load it lazily and stream it directly to an arrow flight client. We would do so by opening up the file in process memory (lazily), then we would first aggregate the list of records (lazily) and then read the contents of the records for the first time when streaming the data to the client. 

via

```
func (s *MyServer) DoGet(
	ticket *flight.Ticket,
	server flight.FlightService_DoGetServer,
) error {
    ...
	w := flight.NewRecordWriter(server, writerOpts...)
	for _, r := range s.recordsInTable {
		if err = w.Write(r); err != nil {
			return err
		}
	}
```

Ideally, up until the records sent, no data would be actually copied/ready. However, this doesn't appear to be possible at the moment since there is a copy done when processing the records https://github.com/apache/arrow-go/blob/main/arrow/ipc/message.go#L217 instead or referencing the already allocated/used shared memory. Instead, there might need to be a new `MessageReader` implementation where this copy is avoided. 

Thank you!

### Component(s)

Other

zeroshade:
### Rationale for this change
closes #207 

Adding new method `NewMappedFileReader` to ipc package which accepts a byte slice instead of a `ReaderAtSeeker`. Updates `ipcSource` to reference the raw byte slices from the input directly instead of wrapping with `bytes.NewReader` which forces copies via `Read`, `ReadFull`, etc.

### Are these changes tested?
Unit tests added to confirm that the pointers match and that we aren't allocating unnecessarily.

Shouldn't be any user-facing changes other than a reduction in memory usage when reading non-compressed IPC data.

### DIFF
--- a/arrow/ipc/file_reader.go
+++ b/arrow/ipc/file_reader.go
@@ -35,7 +35,7 @@ import (
 
 type readerImpl interface {
 	getFooterEnd() (int64, error)
-	readFooter(*footerBlock) error
+	getBytes(offset, length int64) ([]byte, error)
 	dict(memory.Allocator, *footerBlock, int) (dataBlock, error)
 	block(memory.Allocator, *footerBlock, int) (dataBlock, error)
 }
@@ -53,52 +53,28 @@ type dataBlock interface {
 	NewMessage() (*Message, error)
 }
 
+const footerSizeLen = 4
+
+var minimumOffsetSize = int64(len(Magic)*2 + footerSizeLen)
+
 type basicReaderImpl struct {
 	r ReadAtSeeker
 }
 
-func (r *basicReaderImpl) getFooterEnd() (int64, error) {
-	return r.r.Seek(0, io.SeekEnd)
+func (r *basicReaderImpl) getBytes(offset, len int64) ([]byte, error) {
+	buf := make([]byte, len)
+	n, err := r.r.ReadAt(buf, offset)
+	if err != nil {
+		return nil, fmt.Errorf("arrow/ipc: could not read %d bytes at offset %d: %w", len, offset, err)
+	}
+	if int64(n) != len {
+		return nil, fmt.Errorf("arrow/ipc: could not read %d bytes at offset %d", len, offset)
+	}
+	return buf, nil
 }
 
-func (r *basicReaderImpl) readFooter(f *footerBlock) error {
-	var err error
-
-	if f.offset <= int64(len(Magic)*2+4) {
-		return fmt.Errorf("arrow/ipc: file too small (size=%d)", f.offset)
-	}
-
-	eof := int64(len(Magic) + 4)
-	buf := make([]byte, eof)
-	n, err := r.r.ReadAt(buf, f.offset-eof)
-	if err != nil {
-		return fmt.Errorf("arrow/ipc: could not read footer: %w", err)
-	}
-	if n != len(buf) {
-		return fmt.Errorf("arrow/ipc: could not read %d bytes from end of file", len(buf))
-	}
-
-	if !bytes.Equal(buf[4:], Magic) {
-		return errNotArrowFile
-	}
-
-	size := int64(binary.LittleEndian.Uint32(buf[:4]))
-	if size <= 0 || size+int64(len(Magic)*2+4) > f.offset {
-		return errInconsistentFileMetadata
-	}
-
-	buf = make([]byte, size)
-	n, err = r.r.ReadAt(buf, f.offset-size-eof)
-	if err != nil {
-		return fmt.Errorf("arrow/ipc: could not read footer data: %w", err)
-	}
-	if n != len(buf) {
-		return fmt.Errorf("arrow/ipc: could not read %d bytes from footer data", len(buf))
-	}
-
-	f.buffer = memory.NewBufferBytes(buf)
-	f.data = flatbuf.GetRootAsFooter(buf, 0)
-	return err
+func (r *basicReaderImpl) getFooterEnd() (int64, error) {
+	return r.r.Seek(0, io.SeekEnd)
 }
 
 func (r *basicReaderImpl) block(mem memory.Allocator, f *footerBlock, i int) (dataBlock, error) {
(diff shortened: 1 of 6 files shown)


zeroshade:
@vtk9 can you give this a try and confirm that it addresses your issue? I added unit tests which confirm that the pointers match and that we're avoiding allocations but it would be great to confirm on your end that this reduces the memory usage you were seeing.

zeroshade:
@lidavidm @kou @joellubi would one of you be able to look this over and give a review? 

I wanna merge this and then kick off an RC as requested from https://github.com/apache/arrow-go/pull/218#issuecomment-2539297076

arrow/ipc/file_reader.go
```diff
@@ -33,16 +33,166 @@ import (
...
-		buffer *memory.Buffer
-		data   *flatbuf.Footer
+func (r *basicReaderImpl) getFooterEnd() (int64, error) {
+	return r.r.Seek(0, io.SeekEnd)
+}
+
+func (r *basicReaderImpl) readFooter(f *footerBlock) error {
+	var err error
+
+	if f.offset <= int64(len(Magic)*2+4) {
```
kou, line 67: Can we add a variable for `len(Magic)*2+4` because we use this minimum size multiple times in this file?

kou, line 67: Can we give `4` a name something like `footerSizeLen` or something?

arrow/ipc/file_reader.go
```diff
@@ -33,16 +33,166 @@ import (
...
+	}
+	if n != len(buf) {
+		return fmt.Errorf("arrow/ipc: could not read %d bytes from end of file", len(buf))
+	}
+
+	if !bytes.Equal(buf[4:], Magic) {
+		return errNotArrowFile
+	}
+
+	size := int64(binary.LittleEndian.Uint32(buf[:4]))
```
kou, line 85: I think that this should be int32 not uint32: https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format

> `<FOOTER SIZE: int32>`

arrow/ipc/file_reader.go
```diff
@@ -33,16 +33,166 @@ import (
...
+		return fmt.Errorf("arrow/ipc: file too small (size=%d)", f.offset)
+	}
+
+	eof := int64((len(Magic) + 4))
+	buf := r.data[f.offset-eof:]
+	if !bytes.Equal(buf[4:], Magic) {
+		return errNotArrowFile
+	}
+
+	size := int64(binary.LittleEndian.Uint32(buf[:4]))
```
kou, line 151: ditto.

arrow/ipc/file_reader.go
```diff
@@ -768,3 +867,44 @@ func readDictionary(memo *dictutils.Memo, meta *memory.Buffer, body ReadAtSeeker
...
+	var (
+		body *memory.Buffer
+		meta *memory.Buffer
+		buf  = blk.section()
+	)
+
+	metaBytes := buf[:blk.meta]
+
+	prefix := 0
+	switch binary.LittleEndian.Uint32(metaBytes) {
```
kou, line 897: int32 not uint32?

https://github.com/apache/arrow/blob/313d11aa94c2be71142b55e3d8bb166d780c19c7/format/File.fbs#L45

> metaDataLength: int;

lidavidm:
I'll try to get to it on Monday

arrow/ipc/file_reader.go
```diff
@@ -33,16 +33,166 @@ import (
...
+	}, nil
+}
+
+type mappedReaderImpl struct {
+	data []byte
+}
+
+func (r *mappedReaderImpl) getFooterEnd() (int64, error) { return int64(len(r.data)), nil }
+
+func (r *mappedReaderImpl) readFooter(f *footerBlock) error {
```
kou, line 140: Can we unify more codes in this function with `basicReaderImple.readFooter()`?
It seems that they have many similar codes.

arrow/ipc/file_reader.go
lidavidm: I think this is because Golang only has the unsigned versions, expecting you to convert to signed integer yourself

https://pkg.go.dev/encoding/binary#ByteOrder

arrow/ipc/file_reader.go
```diff
@@ -388,34 +498,23 @@ func (src *ipcSource) buffer(i int) *memory.Buffer {
...
-		err := binary.Read(sr, binary.LittleEndian, &uncompressedSize)
-		if err != nil {
-			panic(err)
-		}
+		body := src.rawBytes.Bytes()[buf.Offset() : buf.Offset()+buf.Length()]
+		uncompressedSize := binary.LittleEndian.Uint64(body[:8])
 
-		var r io.Reader = sr
 		// check for an uncompressed buffer
 		if int64(uncompressedSize) != -1 {
```
lidavidm, line 509: nit: while this is existing code, maybe it would be safer to have `uncompressedSize := int64(...)` so you don't have to remember to convert it on use and so it's consistent with above

lidavidm (approved):

arrow/ipc/file_reader.go
zeroshade: @lidavidm is correct, only the unsigned versions are available and you are expected to convert to signed integer yourself if you need.

arrow/ipc/file_reader.go
zeroshade: It's a bit hard to unify these more since the core part of it (reading the bytes) is where the change is but I'll see what I can do

zeroshade: updated by adding a `getBytes` method to the impls and having a single implementation for `readFooter` that uses it, unifying the implementations.

kou (approved):
+1

PR #217, opened 2024-12-10
feat(parquet): Move footerOffset into FileMetaData

joechenrh:
### Rationale for this change

After looking into the code, I found that `footerOffset` is mainly used in parsing the meta data of parquet file.

I think we can store it in `FileMetaData`, so we don't need to get `footerOffset` again when we use `WithMetadata`. 

Move `Reader.footerOffset` into `FileMetadata`

### DIFF
--- a/parquet/metadata/file.go
+++ b/parquet/metadata/file.go
@@ -279,10 +279,10 @@ func NewFileMetaData(data []byte, fileDecryptor encryption.FileDecryptor) (*File
 // Size is the length of the raw serialized metadata bytes in the footer
 func (f *FileMetaData) Size() int { return f.metadataLen }
 
-// SetSourceFileSize get the total size of the source file from meta data.
+// GetSourceFileSize get the total size of the source file from meta data.
 func (f *FileMetaData) GetSourceFileSize() int64 { return f.sourceFileSize }
 
-// GetSourceFileSize set the total size of the source file in meta data.
+// SetSourceFileSize set the total size of the source file in meta data.
 func (f *FileMetaData) SetSourceFileSize(sourceFileSize int64) { f.sourceFileSize = sourceFileSize }
 
 // NumSchemaElements is the length of the flattened schema list in the thrift
@@ -398,7 +398,7 @@ func (f *FileMetaData) Subset(rowGroups []int) (*FileMetaData, error) {
 		f.FileDecryptor,
 		f.version,
 		0,
-		f.footerOffset,
+		f.sourceFileSize,
 	}
 
 	out.RowGroups = make([]*format.RowGroup, 0, len(rowGroups))
--- a/parquet/metadata/file.go
+++ b/parquet/metadata/file.go
@@ -243,6 +243,8 @@ type FileMetaData struct {
 	// size of the raw bytes of the metadata in the file which were
 	// decoded by thrift, Size() getter returns the value.
 	metadataLen int
+
+	FooterOffset int64
 }
 
 // NewFileMetaData takes in the raw bytes of the serialized metadata to deserialize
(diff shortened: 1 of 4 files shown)


parquet/metadata/metadata_test.go
```diff
@@ -114,7 +114,7 @@ func TestBuildAccess(t *testing.T) {
 	require.NoError(t, err)
 	serialized, err := faccessor.SerializeString(context.Background())
 	assert.NoError(t, err)
-	faccessorCopy, err := metadata.NewFileMetaData([]byte(serialized), nil)
+	faccessorCopy, err := metadata.NewFileMetaData([]byte(serialized), 0, nil)
```
zeroshade, line 117: Shouldn't it be invalid to use 0 for the footer offset?

parquet/metadata/file.go
```diff
@@ -275,6 +278,9 @@ func NewFileMetaData(data []byte, fileDecryptor encryption.FileDecryptor) (*File
 // Size is the length of the raw serialized metadata bytes in the footer
 func (f *FileMetaData) Size() int { return f.metadataLen }
 
+// SourceSz is the total size of the source file
+func (f *FileMetaData) SourceSz() int64 { return f.footerOffset }
```
zeroshade, line 282: Because the FileMetadata can potentially be separate from the file itself, it might make more sense to name this differently or at least use a different description for the comment.

parquet/metadata/file.go
```diff
@@ -243,11 +243,13 @@ type FileMetaData struct {
...
 	// decoded by thrift, Size() getter returns the value.
 	metadataLen int
+
+	footerOffset int64
 }
 
 // NewFileMetaData takes in the raw bytes of the serialized metadata to deserialize
 // and will attempt to decrypt the footer if a decryptor is provided.
-func NewFileMetaData(data []byte, fileDecryptor encryption.FileDecryptor) (*FileMetaData, error) {
+func NewFileMetaData(data []byte, footerOffset int64, fileDecryptor encryption.FileDecryptor) (*FileMetaData, error) {
```
zeroshade, line 252: The only thing I'm not a fan of here is that this is a breaking change. Now, I don't think there are many (if any) packages that are importing and using this function directly (as creating new file metadata would primarily be done via the Writer rather than done directly) but it's still a concern that I'd rather not create a breaking change if we can avoid it. 

Is there a way we can avoid this being a breaking change at all? If 0 is valid, maybe we create a setter?

zeroshade (commented):
I like the idea, just some nitpicks

parquet/metadata/file.go
joechenrh: Yes, I forgot the backwards compatibility. Can we name the setter/getter function like:

```go
// SetSourceFileSize get the total size of the source file from meta data.
func (f *FileMetaData) SetSourceFileSize() int64 { return f.sourceFileSize }

// GetSourceFileSize set the total size of the source file in meta data.
func (f *FileMetaData) GetSourceFileSize() int64 { return f.sourceFileSize }

// file.go
type FileMetaData struct {
    ....
    
    // sourceFileSize is not a part of FileMetaData, but it is mainly used to parse meta data.
    // Users can manually set this value and they are responsible for the validity of it.
    sourceFileSize int64
}
```

parquet/metadata/file.go
```diff
@@ -275,6 +279,12 @@ func NewFileMetaData(data []byte, fileDecryptor encryption.FileDecryptor) (*File
 // Size is the length of the raw serialized metadata bytes in the footer
 func (f *FileMetaData) Size() int { return f.metadataLen }
 
+// SetSourceFileSize get the total size of the source file from meta data.
+func (f *FileMetaData) GetSourceFileSize() int64 { return f.sourceFileSize }
+
+// GetSourceFileSize set the total size of the source file in meta data.
+func (f *FileMetaData) SetSourceFileSize(sourceFileSize int64) { f.sourceFileSize = sourceFileSize }
```
zeroshade, line 286: doc strings are reversed currently, gotta swap those :smile:

zeroshade (approved):
LGTM thanks!

PR #218, opened 2024-12-11
chore: bump github.com/goccy/go-json to v0.10.4

toddtreece:
### Rationale for this change

A follow up to https://github.com/apache/arrow-go/pull/199#issuecomment-2494887343

> i will continue trying to get the go-json PR merged, and if i do, i'll open another PR to update the dependency here.

Bumps go-json dependency to include changes from https://github.com/goccy/go-json/pull/490

### Are these changes tested?

I am assuming CI tests should cover this.

No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/andybalholm/brotli v1.1.1
 	github.com/apache/thrift v0.21.0
 	github.com/docopt/docopt-go v0.0.0-20180111231733-ee0de3bc6815
-	github.com/goccy/go-json v0.10.3
+	github.com/goccy/go-json v0.10.4
 	github.com/golang/snappy v0.0.4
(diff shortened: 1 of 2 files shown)


zeroshade:
Would it be possible to get goccy to do a release of `go-json` rather than us pointing to a specific commit?

toddtreece:
@zeroshade i will try via gophers slack, but it might also help if you could comment on the `go-json` PR asking for a release since this is a direct dependency of `arrow-go`

toddtreece:
@zeroshade `v0.10.4` was released 🎉. i updated this PR, so it should be ready for another look

zeroshade:
Once this merges, I think I would want to get a couple other PRs on here merged first and then I would be happy to start the release process

zeroshade (approved):

PR #223, opened 2024-12-19
feat(parquet): Add support for Page Indexes

### LINKED ISSUE #33: [Go][Parquet] Does this implementation support page indexes?
I would like to create page indexes (per [this doc](https://github.com/apache/parquet-format/blob/master/PageIndex.md)) for a column when writing a parquet file, and then use that index to seek to a particular row. I have a case where the file is sorted by a particular ID, and the queries often want to start with a particular ID and then read all rows thereafter. Without an index, I can find the right row group using statistics, but then have to scan through all values in the column in the row group to find the ID and determine the right row.

I've gone through all of the code and API in the parquet package and sub-packages, and all I can find are columns in Thrift-generated code for this and accessors in the column chunk metadata that return the file offset for where the index is stored. But there seems to be no API to actually read the index and use it. And there is no configuration, on the write side, for whether to create an index or not.

When will this be supported? How active is development on the Go runtime?

zeroshade:
### Rationale for this change
Adds support for writing and retrieving Page Indexes from parquet files

Added new writer properties to enable writing PageIndexes either as default or for specific columns.
Added `RowGroupPageIndexReader`, `ColumnIndexReader`, `OffsetIndexReader` to access page index information.

### Are these changes tested?
Yes, extensive unit tests have been added including round trip tests and encryption/decryption tests.

no. Just new functions available.

closes #33

### DIFF
--- a/arrow/flight/flightsql/column_metadata.go
+++ b/arrow/flight/flightsql/column_metadata.go
@@ -63,7 +63,7 @@ type ColumnMetadata struct {
 }
 
 func (c *ColumnMetadata) findStrVal(key string) (string, bool) {
-	idx := c.Data.FindKey(CatalogNameKey)
+	idx := c.Data.FindKey(key)
 	if idx == -1 {
 		return "", false
 	}
@@ -71,7 +71,7 @@ func (c *ColumnMetadata) findStrVal(key string) (string, bool) {
 }
 
 func (c *ColumnMetadata) findBoolVal(key string) (bool, bool) {
-	idx := c.Data.FindKey(CatalogNameKey)
+	idx := c.Data.FindKey(key)
 	if idx == -1 {
 		return false, false
 	}
@@ -79,7 +79,7 @@ func (c *ColumnMetadata) findBoolVal(key string) (bool, bool) {
 }
 
 func (c *ColumnMetadata) findInt32Val(key string) (int32, bool) {
-	idx := c.Data.FindKey(CatalogNameKey)
+	idx := c.Data.FindKey(key)
 	if idx == -1 {
 		return 0, false
 	}
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -85,6 +85,17 @@ func (p *page) Data() []byte              { return p.buf.Bytes() }
 func (p *page) NumValues() int32          { return p.nvals }
 func (p *page) Encoding() format.Encoding { return p.encoding }
 
+// DataPageConfig is a struct for passing configuration params to data page creation
+// which can be expanded in the future without causing any breaking changes.
+type DataPageConfig struct {
+	Num              int32
+	Encoding         parquet.Encoding
+	UncompressedSize int32
+	Stats            metadata.EncodedStatistics
+	FirstRowIndex    int64
+	SizeStats        SizeStatistics
+}
+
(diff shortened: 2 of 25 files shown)


arrow/flight/flightsql/column_metadata.go
```diff
@@ -63,23 +63,23 @@ type ColumnMetadata struct {
 }
 
 func (c *ColumnMetadata) findStrVal(key string) (string, bool) {
-	idx := c.Data.FindKey(CatalogNameKey)
+	idx := c.Data.FindKey(key)
```
kou, line 66: This is unrelated fix, right?

kou (commented):
(Sorry, I couldn't review this carefully... I took a look at all changes but I don't have enough knowledge of the Parquet specification and Go. And this change is too large for me... I hope that others can review this.)

arrow/flight/flightsql/column_metadata.go
zeroshade: yea, it was a metadata issue I came across that turned out to be small enough that it made sense to just incorporate

zeroshade:
Thanks for taking a look at all @kou!

@wgtmac and @pitrou I specifically tagged you both for review here given your Parquet experience hoping that one or both of you would be able to take a look at this from that perspective. Much of this is based on the way that Parquet C++ handles Page Indexes

zeroshade:
I'd like to get this in before I kick off the next release in the next week or so. Would anyone be free to do a review soon?

parquet/file/page_reader.go
```diff
@@ -144,6 +161,13 @@ func NewDataPageV1WithStats(buffer *memory.Buffer, num int32, encoding, defEncod
 	return ret
 }
 
+// NewDataPageV1WithExtras is like WithStats but adds the first row index and size statistics
+func NewDataPageV1WithExtras(buffer *memory.Buffer, num int32, encoding, defEncoding, repEncoding parquet.Encoding, uncompressedSize int32, stats metadata.EncodedStatistics, firstRowIdx int64, szStats SizeStatistics) *DataPageV1 {
```
lidavidm, line 165: Is it worth putting some of these parameters in a struct so they can be extended in the future? Or is this very likely to be the "last" extension? (Since now we have NewV1, NewV1WithStats, and NewV1WithExtras)

parquet/metadata/page_index.go
```diff
@@ -0,0 +1,889 @@
...
+
+type BoundaryOrder = format.BoundaryOrder
+
+const (
+	Unordered  BoundaryOrder = format.BoundaryOrder_UNORDERED
+	Ascending  BoundaryOrder = format.BoundaryOrder_ASCENDING
+	Descending BoundaryOrder = format.BoundaryOrder_DESCENDING
+)
+
+type ColumnIndex interface {
```
lidavidm, line 44: It might be useful to have docstrings for public types at least?

lidavidm (approved):

parquet/metadata/page_index.go
zeroshade: bah! I thought I had gone through and added the docstrings, I guess I missed a few. I'll fix that!

zeroshade:
Thanks for the review @lidavidm!!

Issue #202, opened 2024-11-26
Prepare https://arrow.apache.org/go/


kou:
### Describe the enhancement requested

My idea:

* We don't generate the documentation for https://arrow.apache.org/go/
* We just have a simple HTML that have links to https://pkg.go.dev/github.com/apache/arrow-go/arrow and so on in the `asf-site` branch (that is published to https://arrow.apache.org/go/ )

See also: https://github.com/apache/arrow-go/issues/139#issuecomment-2390638600

### Component(s)

Documentation

zeroshade:
Hey @kou, is this considered done given the merged commits? Can we close it?

kou:
Yes. I close this.

(I don't know why "Fixes GH-202" didn't close this automatically...)

PR #231, opened 2025-01-01
chore(website): redirect to https://pkg.go.dev/github.com/apache/arrow-go/

kou:
### Rationale for this change

Fixes GH-202

We need our canonical URL.

Redirect to https://pkg.go.dev/github.com/apache/arrow-go/ .

### Are these changes tested?

No.

Yes.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,1 @@
+Redirect permanent /go/ https://pkg.go.dev/github.com/apache/arrow-go/


zeroshade (approved):
Perfect. The canonical URL here should always automatically end up at the current version :) Thanks!!

kou:
I'm not sure whether we can use `.htaccess` or not on https://arrow.apache.org/ .
Let's try this.

kou:
`.htaccess` should work because apache/arrow-site uses it: https://github.com/apache/arrow-site/blob/asf-site/.htaccess
But https://arrow.apache.org/go/ doesn't work.

We need to add `.asf.yaml` to the `asf-site` branch too.

PR #249, opened 2025-01-13
fix(release): fix svn add command and add script to generate release notes

zeroshade:
### Rationale for this change
Ease of utility for release process

Fixing `svn add` command to explicitly refer to the directory so it doesn't present an error anymore.
Add script to generate the markdown page for adding release notes to the arrow-site repo.

### DIFF
--- a/dev/release/post-website.sh
+++ b/dev/release/post-website.sh
@@ -69,7 +69,6 @@ fi
 export TZ=UTC
 release_date_iso8601=$(LC_TIME=C date "+%Y-%m-%d")
 
-git_tag=v${version}
 git_range=v${previous_version}..v${version}
 
 contributors_command_line="git shortlog -sn ${git_range}"
--- a/dev/release/post-website.sh
+++ b/dev/release/post-website.sh
@@ -22,7 +22,7 @@ set -u
 
 SOURCE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ARROW_DIR="${SOURCE_DIR}/../../"
-: ${ARROW_SITE_DIR:="${ARROW_DIR}/../arrow-site"}
+: "${ARROW_SITE_DIR:="${ARROW_DIR}/../arrow-site"}"
 
 if [ "$#" -ne 2 ]; then
   echo "Usage: $0 <previous-version> <version>"
@@ -32,7 +32,7 @@ fi
 previous_version=$1
 version=$2
 
-branch_name=release-note-arrow-go-${version}
+branch_name="release-note-arrow-go-${version}"
 release_dir="${ARROW_SITE_DIR}/_posts"
 announce_file="${release_dir}/$(date +%Y-%m-%d)-arrow-go-${version}.md"
 
@@ -52,10 +52,10 @@ popd
 
 pushd "${ARROW_DIR}"
 
-previous_major_version="$(echo ${previous_version} | cut -d. -f1)"
-previous_minor_version="$(echo ${previous_version} | cut -d. -f2)"
-major_version="$(echo ${version} | cut -d. -f1)"
-minor_version="$(echo ${version} | cut -d. -f2)"
+previous_major_version="$(echo "${previous_version}" | cut -d. -f1)"
+previous_minor_version="$(echo "${previous_version}" | cut -d. -f2)"
+major_version="$(echo "${version}" | cut -d. -f1)"
+minor_version="$(echo "${version}" | cut -d. -f2)"
 if [ "${previous_major_version}" -eq "${major_version}" ]; then
   if [ "${previous_minor_version}" -eq "${minor_version}" ]; then
     release_type="patch"
@@ -74,7 +74,7 @@ git_range=v${previous_version}..v${version}
 contributors_command_line="git shortlog -sn ${git_range}"
 contributors=$(${contributors_command_line} | grep -v dependabot)
 
-n_commits=$(git log --pretty=oneline ${git_range} | grep -c -i -v "chore: Bump")
+n_commits=$(git log --pretty=oneline "${git_range}" | grep -c -i -v "chore: Bump")
 n_contributors=$(${contributors_command_line} | grep -c -v dependabot)
 
 git_changelog="$(gh release view --json body --jq .body | grep -v '@dependabot' | sed -e 's/^#/##/g')"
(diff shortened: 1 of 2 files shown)


dev/release/release.sh
```diff
@@ -57,7 +57,7 @@ gh release download "${rc_tag}" \
 
 echo "Uploading to release/"
 pushd "${dist_base_dir}"
-svn add .
+svn add "${release_id}"
```
kou, line 60: Oh. I thought that "." also works.
Do you still have an actual error message?

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+branch_name=release-note-arrow-go-${version}
+release_dir="${ARROW_SITE_DIR}/_posts"
+announce_file="${release_dir}/$(date +%Y-%m-%d)-arrow-go-${version}.md"
+
+pushd "${ARROW_SITE_DIR}"
+DEFAULT_BRANCH="$(git rev-parse --abbrev-ref origin/HEAD | sed s@origin/@@)"
+git fetch --all --prune --tags --force -j$(nproc)
+git checkout ${DEFAULT_BRANCH}
+git branch -D ${branch_name} || :
+git checkout -b ${branch_name}
```
kou, line 44:
```suggestion
git switch -c ${branch_name}
```

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+version=$2
+
+branch_name=release-note-arrow-go-${version}
+release_dir="${ARROW_SITE_DIR}/_posts"
+announce_file="${release_dir}/$(date +%Y-%m-%d)-arrow-go-${version}.md"
+
+pushd "${ARROW_SITE_DIR}"
+DEFAULT_BRANCH="$(git rev-parse --abbrev-ref origin/HEAD | sed s@origin/@@)"
+git fetch --all --prune --tags --force -j$(nproc)
+git checkout ${DEFAULT_BRANCH}
```
kou, line 42: `switch` is better with recent Git:

```suggestion
git switch ${DEFAULT_BRANCH}
```

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+git branch -D ${branch_name} || :
+git checkout -b ${branch_name}
+popd
+
+pushd "${ARROW_DIR}"
+
+previous_major_version="$(echo v${previous_version} | grep -o '^[0-9]*')"
+major_version="$(echo v${version} | grep -o '^[0-9]*')"
```
kou, line 50:
```suggestion
previous_major_version="$(echo ${previous_version} | grep -o '^[0-9]*')"
major_version="$(echo ${version} | grep -o '^[0-9]*')"
```

BTW, `cut` may be better:

```shell
previous_major_version="$(echo ${previous_version} | cut -d. -f1)"
major_version="$(echo ${version} | cut -d. -f1)"
```

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+if [ ${previous_major_version} -eq ${major_version} ]; then
+  release_type=patch
+else
+  release_type=major
+fi
```
kou, line 55: Can we add "minor" release detection?

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+{% endcomment %}
+-->
+
+The Apache Arrow team is pleased to announce the v${version} release ofApache Arrow Go.
```
kou, line 108:
```suggestion
The Apache Arrow team is pleased to announce the v${version} release of Apache Arrow Go. 
```

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+git_range=v${previous_version}..v${version}
+
+contributors_command_line="git shortlog -sn ${git_range}"
+contributors=$(${contributors_command_line} | grep -v dependabot)
+
+n_commits=$(git log --pretty=oneline ${git_range} | grep -i -v "chore: Bump" | wc -l)
+n_contributors=$(${contributors_command_line} | grep -v dependabot | wc -l)
+
+git_tag_hash=$(git log -n 1 --pretty=%H ${git_tag})
+git_changelog="$(gh release view --json body --jq .body | grep -v '@dependabot')"
```
kou, line 76: Could you replace `##` section markup to `###` because we use `## Changelog` section?

```suggestion
git_changelog="$(gh release view --json body --jq .body | grep -v '@dependabot' | sed -e 's/^#/##/g')"
```

dev/release/post-website.sh
```diff
@@ -0,0 +1,138 @@
...
+previous_version=$1
+version=$2
+
+branch_name=release-note-arrow-go-${version}
+release_dir="${ARROW_SITE_DIR}/_posts"
+announce_file="${release_dir}/$(date +%Y-%m-%d)-arrow-go-${version}.md"
+
+pushd "${ARROW_SITE_DIR}"
+DEFAULT_BRANCH="$(git rev-parse --abbrev-ref origin/HEAD | sed s@origin/@@)"
+git fetch --all --prune --tags --force -j$(nproc)
```
kou, line 41: Could you add a check whether this working copy is a fork not apache/arrow-site?

dev/release/release.sh
zeroshade: yea, I get

```console
svn: warning: W150002: '/tmp/arrow-go/dev/release/dist' is already under version control
svn: E200009: Could not add all targets because some targets are already versioned
svn: E200009: Illegal target for the requested operation
```

when I use ".", which is fixed if I use the directory name

dev/release/post-website.sh
zeroshade: I simply pulled the logic from the version of this script in the original Arrow repo, switching to cut is definitely a better idea. Good catch

dev/release/release.sh
kou: Thanks!

I'll fix related codes in apache/arrow-java.

kou: https://github.com/apache/arrow-java/pull/513

dev/release/post-website.sh
zeroshade: added

dev/release/post-website.sh
zeroshade: added

kou (approved):
+1

Issue #58, opened 2023-04-12
[Go][CI]: Run `gofmt` check in CI


yevgenypats:
### Describe the enhancement requested

This would help save contributors and reviewers time so they can focus on code logic and bugs and offload formatting issues to the CI. 

### Component(s)

Go

singh1203:
Hello @yevgenypats Is this issue still available to work on? 
If yes I would love to work on it.

zeroshade:
I believe this is still available to work on. Go for it!

kou:
Done by #92.

PR #263, opened 2025-01-29
CI: Add gofmt check in lint

singh1203:
Fixes: #58 

Added the `gofmt` check in the pre-commit-config.yml, which identifies files that need `gofmt` linting and make proper linting.

### Are these changes tested?
Yes, I tested it on my forked repository example [here](https://github.com/singh1203/arrow-go/actions/runs/13046944649/job/36399021645)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -49,6 +49,13 @@ jobs:
           go-version: '1.23'
           cache: true
           cache-dependency-path: go.sum
+      - name: Run gofmt
+        run: |
+          FILES=$(gofmt -l .)
+          if [ -n "$FILES" ]; then
+            echo "$FILES"
+            exit 1
+          fi
       - name: Install pre-commit
         run: |
           python -m pip install pre-commit
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -49,13 +49,6 @@ jobs:
           go-version: '1.23'
           cache: true
           cache-dependency-path: go.sum
-      - name: Run gofmt
-        run: |
-          FILES=$(gofmt -l .)
-          if [ -n "$FILES" ]; then
-            echo "$FILES"
-            exit 1
-          fi
       - name: Install pre-commit
         run: |
           python -m pip install pre-commit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,3 +59,10 @@ repos:
     rev: v0.13.0
     hooks:
       - id: yamlfmt
+  - repo: local
+    hooks:
+      - id: go-fmt
+        name: go fmt
+        language: system
+        entry: bash -c "go get golang.org/x/tools/cmd/goimports@latest && go fmt ./..."
+        pass_filenames: false
--- a/arrow/internal/flatbuf/Decimal.go
+++ b/arrow/internal/flatbuf/Decimal.go
@@ -22,10 +22,10 @@ import (
 	flatbuffers "github.com/google/flatbuffers/go"
 )
 
-/// Exact decimal value represented as an integer value in two's
-/// complement. Currently 32-bit (4-byte), 64-bit (8-byte), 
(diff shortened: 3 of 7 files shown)


singh1203:
> Would it make more sense to include this in the golangci configuration that we use with `pre-commit` instead? Particularly if we could have it so that pre-commit runs gofmt for the user too

Yes, you are correct. I raised the same question in yesterday's Arrow community meeting. I will incorporate the changes as per the review.

singh1203:
> Would it make more sense to include this in the golangci configuration that we use with `pre-commit` instead? Particularly if we could have it so that pre-commit runs gofmt for the user too

@zeroshade I made changes by adding a hook, which you can see [here](https://github.com/singh1203/arrow-go/actions/runs/13046944649/job/36399021645). This raises the question: do I need to commit the changed files after running pre-commit cmd locally?

zeroshade:
> @zeroshade I made changes by adding a hook, which you can see [here](https://github.com/singh1203/arrow-go/actions/runs/13046944649/job/36399021645). This raises the question: do I need to commit the changed files after running pre-commit cmd locally?

That's the way it would work, yes. I'm not sure if we'd want it to automatically commit the changed files.

singh1203:
> > @zeroshade I made changes by adding a hook, which you can see [here](https://github.com/singh1203/arrow-go/actions/runs/13046944649/job/36399021645). This raises the question: do I need to commit the changed files after running pre-commit cmd locally?
> 
> That's the way it would work, yes. I'm not sure if we'd want it to automatically commit the changed files.

Yes you were right I need to commit, and also, instead of hook, I chose to go ahead with `go fmt` cmd since that pre-commit hook is more being maintained. Aside from that, I think we can run the pre-commit run before pushing rather than autocommitting.
@kou Any thoughts for this?

kou:
Can we use https://github.com/golangci/golangci-lint ?
It seems that it includes `gofmt` support: https://golangci-lint.run/usage/configuration/

We don't need to commit changes by pre-commit automatically. Developers should do it after they checked.

singh1203:
> Can we use https://github.com/golangci/golangci-lint ? It seems that it includes `gofmt` support: https://golangci-lint.run/usage/configuration/
> 
> We don't need to commit changes by pre-commit automatically. Developers should do it after they checked.

Yes, we have integrated https://github.com/golangci/golangci-lint in our `pre-commit-config.yaml` and have explicitly configured the [gofmt linter](https://github.com/apache/arrow-go/blob/main/.golangci.yaml#L25). However, linting does not appear to be taking effect. Running `gofmt` manually produces the expected formatting results. Additionally, if `gofumpt` is used as the linter, the output differs from `gofmt`, indicating that it enforces formatting rules that deviate from the standard `gofmt` behavior.

singh1203:
> Can we use https://github.com/golangci/golangci-lint ? It seems that it includes `gofmt` support: https://golangci-lint.run/usage/configuration/
> 
> We don't need to commit changes by pre-commit automatically. Developers should do it after they checked.

Also, I just learned that `gofmt` is integrated as a linter, not as a formatter, so its scope is limited to the current module in https://github.com/golangci/golangci-lint

kou:
Oh, I forgot that we already have golangci/golangci-lint configuration!

It seems that it works well. So I think that we don't need to do this.

The changed files in this PR have "DO NOT EDIT" comment because they are auto generated files. So we should not target them.

It seems that golangci/golangci-lint excepts them automatically. And it's an expected behavior.

singh1203:
> Oh, I forgot that we already have golangci/golangci-lint configuration!
> 
> It seems that it works well. So I think that we don't need to do this.
> 
> The changed files in this PR have "DO NOT EDIT" comment because they are auto generated files. So we should not target them.
> 
> It seems that golangci/golangci-lint excepts them automatically. And it's an expected behavior.

Okay :+1:  Then I should proceed with closing this PR.
cc- @zeroshade

kou:
Sorry. I should have closed #58 when I did #92...

Issue #267, opened 2025-02-04
Use `ubuntu-24.04-arm` instead of self-hosted arm runner in CI


kou:
### Describe the enhancement requested

We're using self-hosted arm runner at: https://github.com/apache/arrow-go/blob/b709be7d5def1c3160f4168943070e14ae949da3/.github/workflows/test.yml#L58-L72

### Component(s)

Continuous Integration

singh1203:
Hey @kou I would be happy to assist here; kindly assign

kou:
Thanks!

PR #268, opened 2025-02-05
chore: Update self-hosted arm runner to ubuntu-24.04-arm

singh1203:
Fixes: #267 

Replaced the running platform from `self-hosted arm` runner to `ubuntu-24.04-arm` 

### Are these changes tested?
Not locally, but can be tested by triggering the runner

No

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,59 +30,30 @@ concurrency:
 permissions:
   contents: read
 jobs:
-  docker-targets:
-    name: Docker targets
-    runs-on: ubuntu-latest
-    outputs:
-      targets: ${{ steps.detect-targets.outputs.targets }}
-    steps:
-      - name: Detect targets
-        id: detect-targets
-        run: |
-          echo "targets<<JSON" >> "$GITHUB_OUTPUT"
-          echo "[" >> "$GITHUB_OUTPUT"
-          cat <<JSON >> "$GITHUB_OUTPUT"
-          {
-            "arch-label": "AMD64",
-            "arch": "amd64",
-            "go": "1.22",
-            "runs-on": "ubuntu-latest"
-          },
-          {
-            "arch-label": "AMD64",
-            "arch": "amd64",
-            "go": "1.23",
-            "runs-on": "ubuntu-latest"
-          }
-          JSON
-          if [ "$GITHUB_REPOSITORY_OWNER" = "apache" ]; then
-            echo "," >> "$GITHUB_OUTPUT"
-            cat <<JSON >> "$GITHUB_OUTPUT"
-          {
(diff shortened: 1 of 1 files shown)


raulcd (commented):
Thanks for the PR!
I would remove the if check for the GitHub repository owner and run on arm as we do for amd64. I mean removing:
```
          if [ "$GITHUB_REPOSITORY_OWNER" = "apache" ]; then
            echo "," >> "$GITHUB_OUTPUT"
            cat <<JSON >> "$GITHUB_OUTPUT"
```
and reorganize the code to a single JSON write.

raulcd (approved):
LGTM

kou:
Could you remove the `docker-targets` job? Could you define `matrix` in the `docker` job statically? We don't need to define `matrix` dynamically now.

raulcd:
If you want an example from what @kou suggests you can see this diff on the main apache/arrow repo where the `docker-targets` was removed for a `matrix`:
https://github.com/apache/arrow/pull/45308/files

singh1203:
Sure @kou. I'd love to help here, and Thank you @raulcd for pointing me in right direction

singh1203:
> Could you remove the `docker-targets` job? Could you define `matrix` in the `docker` job statically? We don't need to define `matrix` dynamically now.

Done. Let's run the runner for testing.

.github/workflows/test.yml
```diff
@@ -30,59 +30,30 @@ concurrency:
...
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
-        include: ${{ fromJson(needs.docker-targets.outputs.targets) }}
+        include:
+          - arch: amd64
+            go: 1.22
+            runs-on: ubuntu-latest
+            arch-label: AMD64
```
kou, line 44: Could you keep these keys in alphabetical order?

```suggestion
          - arch-label: AMD64
            arch: amd64
            go: 1.22
            runs-on: ubuntu-latest
```

singh1203: Done

kou (approved):
+1

Thanks!

zeroshade:
@singh1203 @kou @raulcd  It looks like the ARM jobs are still failing unable to find a runner :( any ideas?

kou:
Which jobs did you see?
https://github.com/apache/arrow-go/pull/264 ?
https://github.com/apache/arrow-go/pull/266 ?

They are opened before this PR is merged. I'll rebase them.

singh1203:
> Which jobs did you see? #264 ? #266 ?
> 
> They are opened before this PR is merged. I'll rebase them.

After rebase, #266 is good to go, but in #264 couple of runners are failing.
@zeroshade I guess you are referring to this?

PR #274, opened 2025-02-11
Drop executable file bit of source files

jas4711:
Hello.  I'm packaging this project for Debian.

Our linter noticed that a bunch of source code files are marked as executable in the source repository.  I don't think this is a good idea, and it is done inconsistency, so I suspect it is leftover somehow.  How about the attached patch?  Generated like this:

```
chmod -x $(find . -type f -executable|grep -v -e ^./.git -e ^./ci -e ^./dev)
```

/Simon


zeroshade:
Thanks for this @jas4711! This makes sense, there's no need for the .go/.c/.s/.h files to be executable.

Could you link to the packaging of this that you're doing for debian? It would be fantastic to add a link on the README indicating how it can be installed on debian/ubuntu etc.

zeroshade:
@jas4711 sounds good. Once it is (hopefully) formally accepted, I hope you'll file a PR to add a link to the README :smile:

zeroshade (approved):

jas4711:
Thank you!  I expect approval to take at least a month, the formal process is fairly slow, so don't hold your breath :)

Issue #32, opened 2024-07-10
[Go] Improve 32-bit support (`GOARCH=386`)


joellubi:
### Describe the enhancement requested

There are several parts of the code that currently fail on 32-bit systems like 386.

1. One problem is that the `refCount int64` we currently use for reference-counted resources requires manual alignment on 32-bit systems when used with `atomic.AddInt64`. This was raised in apache/arrow#43186.
2. Another problem specific to 386 arch is that certain functions with archictecture-specific implementations do not have variants defined for 386, causing them to panic when compiled for that system.

apache/arrow#43206 starts to address these issues for a narrow subset of the codebase. This issue is intended to track the remaining work needed to add full support across the codebase.

### Component(s)

Go

jas4711:
Debian supports three 32-bit platforms: i386, armhf, and armel.  So we would like to see arrow-go work on 32-bit platforms.  On i386, our stripped down build works fine, but there are self-tests failures that only occur on i386 and not amd64:

https://salsa.debian.org/jas/golang-github-apache-arrow-go/-/jobs/7074351

Build log with >1000 FAIL's:

https://salsa.debian.org/jas/golang-github-apache-arrow-go/-/jobs/7074351/artifacts/raw/debian/output/golang-github-apache-arrow-go_18.1.0-1+salsaci+20250211+5_i386.build

For example the first failure:


```
=== NAME  TestBitmapOps/TestBitmapAnd/aligned/left_offset_3/right_offset_3/out_offset_3
    panic.go:262: test panicked: runtime error: invalid memory address or nil pointer dereference
        goroutine 192 [running]:
        runtime/debug.Stack()
        	/usr/lib/go-1.23/src/runtime/debug/stack.go:26 +0x83
        github.com/stretchr/testify/suite.failOnPanic(0xa8ff408, {0x843d980, 0x8769920})
        	/builds/jas/golang-github-apache-arrow-go/debian/output/source_dir/_build/src/github.com/stretchr/testify/suite/suite.go:89 +0x38
        github.com/stretchr/testify/suite.recoverAndFailOnPanic(0xa8ff408)
        	/builds/jas/golang-github-apache-arrow-go/debian/output/source_dir/_build/src/github.com/stretchr/testify/suite/suite.go:83 +0x3b
        panic({0x843d980, 0x8769920})
        	/usr/lib/go-1.23/src/runtime/panic.go:785 +0x103
[...]

PR #277, opened 2025-02-11
fix(arrow/bitutil): fix bitmap ops on 32-bit platforms

zeroshade:
### Rationale for this change
Fixes the broken 32-bit arch tests identified by #32 

Defining the `opAligned` pure go fallback methods for bitmap operations.

### Are these changes tested?
Yes, it makes the tests succeed properly on `GOARCH=386`

No.

### DIFF
--- a/arrow/bitutil/bitmaps.go
+++ b/arrow/bitutil/bitmaps.go
@@ -465,20 +465,24 @@ type bitOp struct {
 
 var (
 	bitAndOp = bitOp{
-		opWord: func(l, r uint64) uint64 { return l & r },
-		opByte: func(l, r byte) byte { return l & r },
+		opWord:    func(l, r uint64) uint64 { return l & r },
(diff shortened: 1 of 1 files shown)


zeroshade:
@jas4711 Can you test this on your end?

@kou Do you know if there's an available `i386` image on GHA that we could use and add a CI run for?

lidavidm (approved):
We could maybe test with qemu?

lidavidm:
https://github.com/docker/setup-qemu-action

kou:
I think that we can use x86_64 ubuntu-latest runner for it.

In general, x86_64 Linux kernel can execute i386 binary. Go can build i386 binary on x86_64 by `GOARCH=386`, right?

zeroshade:
True, I could create a docker config that just runs the scripts with `GOARCH=386` set in the environment. I'll open a separate PR for that though rather than hold this off.

PR #281, opened 2025-02-15
feat(parquet/cmd/parquet_reader): Add command to dump the column and offset indices

zeroshade:
### Rationale for this change
Makes it easier to visualize the Column and Offset indices that may exist in a parquet file.

A new command `column-index` added to the `parquet-reader` CLI which dumps the column and offset indices to the terminal. 

The format of the output is based on the parquet-java CLI


parquet/schema/schema.go
```diff
@@ -161,6 +163,11 @@ func (s *Schema) Column(i int) *Column {
 	return s.leaves[i]
 }
 
+// Columns returns an iterator over the leave columns of the schema
```
lidavidm, line 166:
```suggestion
// Columns returns an iterator over the leaf columns of the schema
```

lidavidm (approved):

PR #289, opened 2025-02-18
fix(arrow/memory): Align allocations always

### LINKED ISSUE #282: mallocator: align allocations to 64 bytes

Some types want 16 byte (or more) alignment, such as decimal256, and it's easiest just to overallocate + align to avoid issues (even if it's not strictly _required_ by the spec, many things in practice assume it). We have to overallocate since the allocator API is not given an alignment. 

Context: https://github.com/apache/arrow-adbc/issues/2526

### Component(s)

Other

lidavidm:
### Rationale for this change

We pay a bit of up-front memory to make sure allocations are always aligned, making in-process FFI easier.

Always align in Buffer.

### Are these changes tested?

Yes

~Maybe: the default Go allocator already aligned allocations. Other allocators may now have a bit more overhead. We will now effectively never use `realloc` with the mallocator, instead always preferring a fresh `malloc` + `memcpy` + `free`.

Closes #282.


lidavidm:
Hmm, something about the memory accounting is no longer right...

arrow/memory/allocator.go
```diff
@@ -25,3 +25,58 @@ type Allocator interface {
 	Reallocate(size int, b []byte) []byte
 	Free(b []byte)
 }
+
+// AlignedAllocator is an Allocator that returns buffers aligned to 64-byte boundaries.
+type AlignedAllocator interface {
+	Allocator
+	// AllocateAligned returns a buffer of the requested size aligned to 64-byte boundaries. buf is the aligned buffer, and alloc is the actual allocation that should be passed to Free.  If alloc is nil, then it is the same as buf.
```
zeroshade, line 32: can we make this a multi-line comment just to make things cleaner?

arrow/memory/allocator.go
```diff
@@ -25,3 +25,58 @@ type Allocator interface {
...
+	AllocateAligned(size int) (buf []byte, alloc []byte)
+	ReallocateAligned(size int, buf []byte, alloc []byte) (newBuf []byte, newAlloc []byte)
+}
+
+// MakeAlignedAllocator makes an AlignedAllocator for the given Allocator.  If the Allocator already implements AlignedAllocator, it is returned as-is.  There is usually no need to call this directly as Buffer will call this for you.
```
zeroshade, line 37: same here, can we make this a multi-line comment instead?

arrow/memory/allocator.go
```diff
@@ -25,3 +25,58 @@ type Allocator interface {
 	Reallocate(size int, b []byte) []byte
 	Free(b []byte)
 }
+
+// AlignedAllocator is an Allocator that returns buffers aligned to 64-byte boundaries.
```
zeroshade, line 29: any particular reason to pick 64-byte boundary as opposed to letting it be configurable?

arrow/memory/allocator.go
```diff
@@ -25,3 +25,58 @@ type Allocator interface {
...
+	return a.mem.Allocate(size)
+}
+
+func (a *alignedAllocator) Reallocate(size int, b []byte) []byte {
+	return a.mem.Reallocate(size, b)
+}
+
+func (a *alignedAllocator) Free(b []byte) {
+	a.mem.Free(b)
+}
```
zeroshade, line 59: probably simpler to just embed the allocator instead of having to recreate these:

```go
type alignedAllocator struct {
         Allocator
}
```

and creating it would be:

```go
return &alignedAllocator{Allocator: mem}
```

that way it just picks up the embedded methods instead of having to write them ourselves. If we ever change anything about the Allocator interface we won't have to also make a change here.

arrow/memory/allocator.go
```diff
@@ -25,3 +25,58 @@ type Allocator interface {
...
+
+func (a *alignedAllocator) AllocateAligned(size int) ([]byte, []byte) {
+	buf := a.mem.Allocate(size + alignment)
+	addr := int(addressOf(buf))
+	next := roundUpToMultipleOf64(addr)
+	if addr != next {
+		shift := next - addr
+		return buf[shift : size+shift : size+shift], buf
```
zeroshade, line 67: shouldn't the capacity be just `size`?

zeroshade (commented):
Any particular reason to do it this way as opposed to just making the `Mallocator` always aligned (like the Go allocator does) and not having to add that logic to the buffer?

At a worst case you could essentially have the mallocator keep a `map[uintptr]uintptr` to map the slice data pointer to the true allocation pointer or something.

lidavidm:
> At a worst case you could essentially have the mallocator keep a `map[uintptr]uintptr` to map the slice data pointer to the true allocation pointer or something.

Ah....That _is_ an option, but yikes :sweat_smile:

lidavidm:
But yeah, unless we do that we have to somehow keep the "real" allocation around since the allocator interface works in term of `[]byte` only

lidavidm:
I reworked this using a `sync.Map`. It's...not intellectually pleasing, but works much simpler!

lidavidm:
At some point we should consider your idea of vendoring jemalloc/mimalloc and just building a simple bump allocator on top in go xD

arrow/memory/mallocator/mallocator.go
```diff
@@ -58,28 +70,44 @@ func (alloc *Mallocator) Allocate(size int) []byte {
 	if size < 0 {
 		panic("mallocator: negative size")
 	}
-	ptr, err := C.calloc(C.size_t(size), 1)
+	paddedSize := C.size_t(size) + C.size_t(alloc.alignment)
```
zeroshade, line 73: `C.size_t(size + alloc.alignment)` instead?

arrow/memory/mallocator/mallocator.go
```diff
@@ -58,28 +70,44 @@ func (alloc *Mallocator) Allocate(size int) []byte {
...
 		// where for some reason calloc return ENOMEM even though there is definitely memory
 		// available for use. So we attempt to fallback to simply doing malloc + memset in
 		// this case. If malloc returns a nil pointer, then we know we're out of memory
 		// and will surface the error.
-		if ptr = C.malloc(C.size_t(size)); ptr == nil {
+		if ptr = C.malloc(paddedSize); ptr == nil {
 			panic(err)
 		}
-		C.memset(ptr, 0, C.size_t(size))
+		C.memset(ptr, 0, C.size_t(paddedSize))
```
zeroshade, line 84: `paddedSize` is already a `C.size_t` now, so you don't need the cast

arrow/memory/mallocator/mallocator.go
```diff
@@ -58,28 +70,44 @@ func (alloc *Mallocator) Allocate(size int) []byte {
...
 
+	buf := unsafe.Slice((*byte)(ptr), paddedSize)
+	aligned := roundToPowerOf2(uintptr(ptr), uintptr(alloc.alignment))
+	alloc.realAllocations.Store(aligned, uintptr(ptr))
 	atomic.AddUint64(&alloc.allocatedBytes, uint64(size))
-	return unsafe.Slice((*byte)(ptr), size)
+
+	if uintptr(ptr) != aligned {
+		shift := aligned - uintptr(ptr)
+		return buf[shift : uintptr(size)+shift : uintptr(size)+shift]
```
zeroshade, line 96: should the capacity still just be `size`?

lidavidm: The third slice parameter is an index and not a capacity, so this sets the capacity to `size`.

lidavidm: > The primary expression
> 
> a[low : high : max]
> 
> constructs a slice of the same type, and with the same length and elements as the simple slice expression a[low : high]. Additionally, it controls the resulting slice's capacity by setting it to max - low.

zeroshade: Ah weird. I always thought it was an absolute capacity for some reason...

arrow/memory/mallocator/mallocator.go
lidavidm: Fixed.

arrow/memory/mallocator/mallocator.go
lidavidm: Fixed.

zeroshade (approved):

PR #290, opened 2025-02-18
docs(license): update LICENSE.txt

### LINKED ISSUE #275: Please clarify copyright holder

Hi.  Who is the copyright holder of the files in this project?  There are no copyright statements in the code, so this is a bit unclear.  I can guess that it is `the Apache Software Foundation (ASF)` but guessing when it comes to licensing issues is rarely a good idea.   Debian has strict policies on licensing, so if you could add a short paragraph to `README` to explain things, that would be best.

Also the `LICENSE` files includes below the typical Apache 2.0 license, the following blurb:

```
This project includes code from the Go project, BSD 3-clause license + PATENTS
weak patent termination clause
(https://github.com/golang/go/blob/master/PATENTS).

Copyright (c) 2009 The Go Authors. All rights reserved.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are
met:

   * Redistributions of source code must retain the above copyright
notice, this list of conditions and the following disclaimer.
   * Redistributions in binary form must reproduce the above
copyright notice, this list of conditions and the following disclaimer
in the documentation and/or other materials provided with the
distribution.
   * Neither the name of Google Inc. nor the names of its
contributors may be used to endorse or promote products derived from
this software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

Which files contain code derived from this license?  I wonder if this is really the case any more.

Thanks,
Simon

### Component(s)

Documentation

zeroshade:
### Rationale for this change
Fixes #275 

Removing "safe-math" from being vendored as it's no longer being used. Updating the LICENSE.txt with all appropriate information including files which contain code that has been copied from other projects.


jas4711:
Thank you!  Looks good to me, even though it does not add any copyright notices for ASF.  I was expecting something like:

```
Copyright (C) Apache Software Foundation
```

Is there any reason you do not want this to be in the git repo?  I don't care strongly about legal reasons here, but there is a lot of license scanners that search for 'Copyright' as a string to find licenses and authors.  It would have saved me some uncertainty when trying to figure out Debian's debian/copyright file for this package if there were more clear copyright notices pointing at the ASF.  Sometimes the exact legal name is significant too.

zeroshade:
The `NOTICE.txt` file contains the line `Copyright 2016-2025 The Apache Software Foundation` and the heading of every source file has the standard Apache License which includes `See the NOTICE file distributed with this work for additional information regarding copyright ownership.` 

Is that sufficient?

kou (approved):
+1

In general, ASF projects write "Copyright YYYY The Apache Software Foundation" in `NOTICE` not `LICENSE`.

See also: https://infra.apache.org/licensing-howto.html#step-by-step

jbonofre:
Thanks ! As you removed safe-math, the `LICENSE` doesn't have to be updated.

About the code copied from LLVM, it's OK in the `LICENSE` and no need to update `NOTICE` as LLVM doesn't provide a `NOTICE`.

The dependency copyright doesn't have to be in `NOTICE` (`NOTICE` is to host relevant information from dependency `NOTICE`).
The copyright should be in `LICENSE` with the code copied (similar to https://github.com/apache/arrow-java/blob/main/LICENSE.txt#L211).

On the other hand, the project copyright should be in the `NOTICE`, and we are good here:

```
Apache Arrow Go
Copyright 2016-2025 The Apache Software Foundation

This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).
```

So, @zeroshade you are good here ! Thanks !

assignUser (approved):

jbonofre (approved):

Issue #56, opened 2023-04-16
ApproxEqual for strings


yevgenypats:
### Describe the enhancement requested

In lot's of databases null characters (`\x00`) are not accepted in strings (For example [postgres](https://stackoverflow.com/questions/28813409/are-null-bytes-allowed-in-unicode-strings-in-postgresql-via-python) but not only). I think given it's such a common gotch in dbs this would be a great use-case for an `arrayApproxEqualString` method just like we have `arrayApproxEqualFloat16`.

### Component(s)

Go

singh1203:
@yevgenypats I would be delighted to help out here.

cc - @zeroshade @kou

zeroshade:
Please do!

PR #291, opened 2025-02-19
feat: Add arrayApproxEqualString to handle null characters in string.

singh1203:
Fixes: #56 

### Rationale for this change
Many databases, including PostgreSQL, do not accept null characters (\x00) in strings. This PR introduces `arrayApproxEqualString` & `arrayApproxEqualLargeString`, functions that enables approximate equality comparison for string arrays while ignoring trailing null characters. This helps improve consistency with database behavior and aligns with existing arrayApproxEqualFloat16.

- Added `arrayApproxEqualString`:
   - Compares two string arrays for approximate equality.
   - Ignores trailing null characters (\x00).
- Added unit tests in `compare_test.go` to verify correctness.

### Are these changes tested?
Yes, unit tests have been added to validate:

- Identical strings return true.
- Strings differing only in trailing null characters return true.
- Completely different strings return false.

No

### DIFF
--- a/arrow/array/compare.go
+++ b/arrow/array/compare.go
@@ -487,19 +487,19 @@ func arrayApproxEqual(left, right arrow.Array, opt equalOption) bool {
 		return arrayEqualBinary(l, r)
 	case *String:
 		r := right.(*String)
-		return arrayEqualString(l, r)
+		return arrayApproxEqualString(l, r)
 	case *LargeBinary:
 		r := right.(*LargeBinary)
 		return arrayEqualLargeBinary(l, r)
 	case *LargeString:
 		r := right.(*LargeString)
-		return arrayEqualLargeString(l, r)
+		return arrayApproxEqualLargeString(l, r)
 	case *BinaryView:
 		r := right.(*BinaryView)
 		return arrayEqualBinaryView(l, r)
 	case *StringView:
 		r := right.(*StringView)
-		return arrayEqualStringView(l, r)
+		return arrayApproxEqualStringView(l, r)
 	case *Int8:
 		r := right.(*Int8)
 		return arrayEqualInt8(l, r)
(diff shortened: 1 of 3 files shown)


arrow/array/compare_test.go
```diff
@@ -725,4 +774,4 @@ func TestTableEqual(t *testing.T) {
 			}
 		})
 	}
-}
+}
```
zeroshade, line 777: needs the trailing newline

zeroshade (changes requested):
should also add the same for StringView too please

arrow/array/compare_test.go
singh1203: Done

singh1203:
> should also add the same for StringView too please

Added!

arrow/array/compare.go
```diff
@@ -644,6 +645,46 @@ func validityBitmapEqual(left, right arrow.Array) bool {
...
+			return false
+		}
+	}
+	return true
+}
+
+func stripNulls(s string) string {
+	return strings.TrimRight(s, "\x00")
+}
+
```
singh1203, line 687: Maybe this could be moved to util.go ?

arrow/array/compare.go
```diff
@@ -644,6 +645,46 @@ func validityBitmapEqual(left, right arrow.Array) bool {
...
+	for i := 0; i < left.Len(); i++ {
+		if left.IsNull(i) {
+			continue
+		}
+		if stripNulls(left.Value(i)) != stripNulls(right.Value(i)) {
+			return false
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
zeroshade: It's fine for them to live either here or in `string.go`, so they can live here just fine.

arrow/array/compare.go
zeroshade: sure, i'm not opposed to that

singh1203:
@zeroshade I think we are GTG here.
Thank you.

zeroshade (approved):

PR #302, opened 2025-03-04
feat(parquet): add byte buffer when disable buffered stream

joechenrh:
### Rationale for this change

When we create a reader with `BufferedStreamEnabled = false`, we have to first create a `data` slice. 

https://github.com/apache/arrow-go/blob/efecae3596e6734ce8244a5ed1befdcff09884df/parquet/reader_properties.go#L76-L91

And we will create another slice with same size in the `NewBufferedReader` again, which is redundant.

https://github.com/apache/arrow-go/blob/efecae3596e6734ce8244a5ed1befdcff09884df/internal/utils/buf_reader.go#L45-L51

- Add a new struct `byteReader`, which is a wrapper of `bytes.NewReader` and implement `BufferedReader` interface.
- Remove `BufferSize()` method from `BufferedReader` interface

### Are these changes tested?

`BufferedReader` created from `bytes.NewReader` in tests are all replaced by new function `NewByteReader`.

No.

### DIFF
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -533,7 +533,7 @@ func (p *serializedPageReader) GetDictionaryPage() (*DictionaryPage, error) {
 		hdr := format.NewPageHeader()
 		rd := utils.NewBufferedReader(
 			io.NewSectionReader(p.r.Outer(), p.dictOffset-p.baseOffset, p.dataOffset-p.baseOffset),
-			p.r.BufferSize())
+			int(parquet.DefaultBufSize))
 		if err := p.readPageHeader(rd, hdr); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


parquet/file/page_reader.go
```diff
@@ -533,7 +533,7 @@ func (p *serializedPageReader) GetDictionaryPage() (*DictionaryPage, error) {
 		hdr := format.NewPageHeader()
 		rd := utils.NewBufferedReader(
 			io.NewSectionReader(p.r.Outer(), p.dictOffset-p.baseOffset, p.dataOffset-p.baseOffset),
-			p.r.BufferSize())
+			int(parquet.DefaultBufSize))
```
zeroshade, line 536: This would remove our respecting of the BufferSize property in the reader properties, we shouldn't be removing the `BufferSize()` method

joechenrh: So how about change it to:


```go
readBufSize := min(p.dataOffset-p.baseOffset, p.r.BufferSize())
rd := utils.NewBufferedReader(
	io.NewSectionReader(p.r.Outer(), p.dictOffset-p.baseOffset, p.dataOffset-p.baseOffset),
	readBufSize)
```

joechenrh:
Can I manually rerun the CI?

kou:
Done!

zeroshade (approved):
Thanks!

PR #311, opened 2025-03-12
fix(pqarrow): respect list element nullability during conversion

zeroshade:
### Rationale for this change
While working on iceberg-go I discovered a bug in how we converted Arrow <--> Parquet schemas when dealing with the nullability of the element for a list field.

Fixing the proper handling for nullability instead of blanket setting it to nullable for elements of lists. Thus correctly propagating the nullability.

### Are these changes tested?
Yes, unit tests are added.

Parquet schemas that are being created from Arrow Schemas will now properly respect the required/optional for elements of lists.


lidavidm (approved):

kou (approved):
+1

PR #315, opened 2025-03-16
refactor(arrow/array): replace some codegen with generics

zeroshade:
### Rationale for this change
Simplifying code and removing old code-gen

Replacing the old `numeric.gen.go` and `numeric.gen.go.tmpl` with a new version built using go generics.

### Are these changes tested?
All the same unit tests will test the new versions of the objects

There shouldn't be any user-facing changes. Everything was designed to avoid any public breaking changes.


arrow/array.go
```diff
@@ -127,3 +127,15 @@ type Array interface {
 	// When the reference count goes to zero, the memory is freed.
 	Release()
 }
+
+// ValueType is a generic constraint for valid Arrow primitive types
+type ValueType interface {
+	bool | FixedWidthType | string | []byte
```
lidavidm, line 133: FixedWidthType sticks out here but I guess it's because there's no equivalent Golang type; however does it make sense to also allow StringType etc. here?

lidavidm (approved):

arrow/array.go
zeroshade: `FixedWidthType` is not `FixedWidthDataType`, `FixedWidthType` is a generic constraint while `FixedWidthDataType` is the actual interface. See `type_traits.go`

in this case, `FixedWidthType` is just short-hand for all of the signed and unsigned int types, float types, decimal types, and temporal types.

Notice we have `| string | []byte` which would include all of the string and binary types too.

lidavidm: ah I see

PR #319, opened 2025-03-19
feat(arrow/compute): implement "is_in" function

zeroshade:
### Rationale for this change
Since we use arrow-go for iceberg-go and utilize the compute libraries for filtering, we need to ensure we add support for the minimal number of functions that iceberg requires. 

Implementing the `is_in` function for the function registry, registering it by default, and ensuring we also allow using `is_in` from substrait. 

### Are these changes tested?
Yes, unit tests are included.

There shouldn't be any user-facing changes.

### DIFF
--- a/arrow/array/builder.go
+++ b/arrow/array/builder.go
@@ -176,12 +176,12 @@ func (b *builder) resize(newBits int, init func(int)) {
 }
 
 func (b *builder) reserve(elements int, resize func(int)) {
-	if b.nullBitmap == nil {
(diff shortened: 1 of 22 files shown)


lidavidm (approved):
Is it just me or are there a lot of unrelated changes mixed in here? :sweat_smile:

zeroshade:
@lidavidm while i was doing this i went a fixed a bunch of linting thingies that i came across and didn't feel like it was important enough to shift them to a separate PR haha.

Issue #71, opened 2022-07-20
[Go][Parquet] PlainFixedLenByteArrayEncoder behaves differently from DictFixedLenByteArrayEncoder with null values where schema has Nullable: false


asfimport:
I have created a small repro to illustrate this bug: https://gist.github.com/phillipleblanc/5e3e2d0e6914d276cf9fd79e019581de

When writing a Decimal128 array to a Parquet file the pqarrow package will prefer to use DictFixedLenByteArrayEncoder. If the size of the array goes over some threshold, it will switch to using PlainFixedLenByteArrayEncoder.

The DictFixedLenByteArrayEncoder tolerates null values in a Decimal128 array with the arrow schema set to Nullable: false, however the PlainFixedLenByteArrayEncoder will not tolerate null values and will panic.

Having null values in an array marked as non-nullable is an issue in the user code - however, it was surprising that my buggy code was working some times and not working other times. I would expect the PlainFixedLen encoder to handle nulls the same way as the DictFixedLen encoder or for the DictFixedLen encoder to panic.

An observation is that most other array types handle nulls with the schema marked as non-nullable when writing to Parquet; this was the first instance I found in the pqarrow package where having the Arrow schema marked as Nullable was necessary for Parquet writing arrays with null values. Again, debatable if this is desirable or not.

**Reporter**: [Phillip LeBlanc](https://issues.apache.org/jira/browse/ARROW-17133)

<sub>**Note**: *This issue was originally created as [ARROW-17133](https://issues.apache.org/jira/browse/ARROW-17133). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

singh1203:
@zeroshade To be straightforward, I lack context, but if it's still available, I'd love to give it a try.

zeroshade:
I thought this was addressed and fixed, go for it and give it a try. If you find this is already fixed (not reproducible) then let me know so I can close this.

singh1203:
> I thought this was addressed and fixed, go for it and give it a try. If you find this is already fixed (not reproducible) then let me know so I can close this.

Yes, the issue still persists [here,](https://github.com/apache/arrow-go/blob/main/parquet/internal/encoding/fixed_len_byte_array_encoder.go#L46) and IMO both (DictFixedLenByteArrayEncoder & PlainFixedLenByteArrayEncoder) should enforce the schema strictly, i.e., give panic.
Also, I'll add it to today's community meeting. 👍

PR #320, opened 2025-03-20
Fix: Handle null values in PlainFixedLenByteArrayEncoder gracefully

singh1203:
Fixes: #71 

### Rationale for this change
Ensures `PlainFixedLenByteArrayEncoder` handles null values gracefully in non-nullable fields by writing zero-filled bytes instead of panicking, maintaining consistency with `DictFixedLenByteArrayEncoder`.

- Replaces panics on null values with writing zero-filled bytes to maintain byte alignment.
- Ensures encoding consistency between dictionary and plain encoding.
- Prevents unexpected failures when handling schema inconsistencies.

### Are these changes tested?
Yes:

- Passed existing test suite.
- Verified with local Arrow-Go modifications.

Not Sure (Need possible guidance if required)

### DIFF
--- a/arrow/ipc/reader.go
+++ b/arrow/ipc/reader.go
@@ -39,7 +39,7 @@ type Reader struct {
 	r      MessageReader
 	schema *arrow.Schema
 
-	refCount int64
+	refCount atomic.Int64
 	rec      arrow.Record
 	err      error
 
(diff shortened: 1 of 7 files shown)


zeroshade (changes requested):
Can we add a new test that would fail without this change?

singh1203:
> Can we add a new test that would fail without this change?

Sure @zeroshade I'm on it.

zeroshade (approved):

singh1203:
Thank you. @zeroshade

Issue #278, opened 2025-02-12
How to best optimize reading from S3?


stevbear:
### Describe the usage question you have. Please include as many useful details as possible.

Hi!
I have a use case of reading certain row groups from S3.
I see that there is an option BufferedStreamEnabled. 
When I set BufferedStreamEnabled to false, it seems to try to read all of the data of a column for a row group at once, which will, unfortunately, result in OOM for us.
When I set BufferedStreamEnabled to true, the library seems to be reading the row group page by page, which is not optimal for cloud usage.
How can I improve this? I imagine that the best way to improve this would be to read multiple pages in one read() sys call?

### Component(s)

Parquet

zeroshade:
Hi @stevbear 

Optimizing the reading from S3 has been on my list for a while and I just hadn't gotten around to it. It didn't get prioritized because no one had filed any issues concerning it, so thank you for filing this!

> When I set BufferedStreamEnabled to true, the library seems to be reading the row group page by page, which is not optimal for cloud usage.

This is likely because the default buffer size is 16KB, have you tried increasing the `BufferSize` member of the `ReaderProperties` so that it buffers more pages at a time? Perhaps try a buffer of a few MB?

I have a few ideas to further optimize via pre-buffering (which all have different trade-offs) so can you give me a bit more context to make sure that your use case would be helped and to identify which idea would work best for you?

Specifically: If reading an entire column for a single row group gives you OOM, you either have a significantly large row group, or I'm guessing it's string data with a lot of large strings? That leads to the question of what you're doing with the column data after you read it from the row group. If you can't hold the entire column in memory from a single row group, are you streaming the data somewhere? Are you reading only a single column at a time or multiple columns from the row group?    Can you give me more of an idea of the sizes of the columns / row group of the file and the memory limitations of your system?

Is the issue the copy that happens when decoding/decompressing the column data? etc.

The more information the better so we can figure out a good solution here, gives me the opportunity to improve the memory usage of the parquet package like I've been wanting to! :smile:

stevbear:
Hi @zeroshade, thanks for getting back to me.
The OOM is not my concern at the moment, as I was able to get away with using BufferSize and specifying a reasonable batch size. My concern is more on efficiency. The BufferSize inside of ReaderProperties doesn't matter that much If I'm reading the source code correctly, as it creates a new buffer when reading the page data (https://github.com/apache/arrow-go/blob/main/parquet/file/page_reader.go#L543).
[...]

PR #321, opened 2025-03-20
feat(parquet/pqarrow): Add SeekToRow for RecordReader

zeroshade:
### Rationale for this change
As suggested by https://github.com/apache/arrow-go/issues/278#issuecomment-2741136394 allowing the RecordReader from the `pqarrow` package to also leverage the `SeekToRow` functionality to skip records from a parquet file while respected skipping particular row groups.

Implementing a `SeekToRow` method for `pqarrow.RecordReader` to seek the record reader to a specific row where the next read will start from.

### Are these changes tested?
Yes, unit tests are added for this.

Just the new functions.

### DIFF
--- a/parquet/file/record_reader.go
+++ b/parquet/file/record_reader.go
@@ -63,6 +63,7 @@ type RecordReader interface {
 	// ReleaseValues transfers the buffer of data with the values to the caller,
 	// a new buffer will be allocated on subsequent calls.
 	ReleaseValues() *memory.Buffer
+	ResetValues()
 	// NullCount returns the number of nulls decoded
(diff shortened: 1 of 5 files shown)


parquet/file/record_reader.go
```diff
@@ -78,6 +79,8 @@ type RecordReader interface {
 	// Release decrements the ref count by one, releasing the internal buffers when
 	// the ref count is 0.
 	Release()
+
+	SeekToRow(int64) error
```
lidavidm, line 83: Should we add docstrings for new methods?

lidavidm (approved):

Issue #269, opened 2025-02-07
Learning Curve of the API


maxpoletaev:
Don't get me wrong, this is an incredible piece of software, and the platform-specific optimizations and complexity it handles are very impressive - mad respect to everyone involved in building it.

I do, however, find myself struggling with the API's learning curve to the point that it makes me feel very dumb and frustrated every time I want to read a few parquet files.

Currently, I rely exclusively on `pqarrow/*_test.go` files as my primary reference to understand usage patterns, which doesn't seem like the proper place to learn. The godoc, while being comprehensive, covers so many types and subpackages that it's hard to piece together a clear picture of how everything fits together.

I suspect other people might be hitting the same wall. The gap between understanding the API documentation and actually implementing something feels unusually wide here compared to other packages in the Go ecosystem. 

I thought that maybe adding some foundational examples in the README or the main godoc page could make a huge difference in helping devs get over that initial hump and make the project appear more approachable.

zeroshade:
Thanks @maxpoletaev, you're definitely right here. 

We're aware of the need for improving the docs and examples here, and the issue is honestly two-fold:

1. Resources and time to dedicate to creating and writing examples
2. Myself and others are too close to the material. I actually find it difficult myself to figure out which aspects of the API specifically need more example coverage. 

If you could contribute any PRs with examples, that would be amazing. Otherwise, I invite you to please add comments here describing any specific scenarios you ran into that were difficult to figure out. It would significantly help with dedicating resources to creating examples for known situations that devs have trouble with. 

Either way, thanks for bringing this up! I'll see if I can find some time to work on this, and will happily review any PRs that get filed.

zeroshade:
Just had another idea, I've been wanting to simplify the API a bit by adding generic interfaces. This would hopefully make it much easier to interact with this lib. 

Again, prioritizing where to start this effort will come down to hopefully getting examples from users as to what the rough edges and difficulties in the current API. 

Looking forward to any responses here!

loicalleyne:
@maxpoletaev

I also ran into this difficulty, I've found it difficult to find working examples of some corners of the API (especially some like `compute`). 

I've started making helper libraries that both solve the problem at hand I'm trying to solve and that can also be useful as a reference. Perhaps these might be helpful to you.

**Getting generic data into Arrow format/generating Arrow schema from data:**
[bodkin](https://github.com/loicalleyne/bodkin) convert data to Arrow (includes json2parquet utility that does uses both Arrow and Parquet)

[...]

PR #322, opened 2025-03-20
feat(arrow/memory): experimenting with addcleanup

zeroshade:
### Rationale for this change
Experimenting with using AddCleanup for automatic cleanup of the Buffers without requiring `Retain` and `Release` everywhere even for custom Allocators.

Changes to the underlying Buffer object to leverage AddCleanup when using go1.24 or higher.

### Are these changes tested?
Yes, though more tests need to be added to ensure more complex scenarios don't fail.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -316,7 +316,7 @@ jobs:
     name: TinyGo
     runs-on: ubuntu-latest
     env:
-      TINYGO_VERSION: 0.38.0
+      TINYGO_VERSION: 0.39.0
     timeout-minutes: 20
     steps:
       - name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -316,7 +316,7 @@ jobs:
     name: TinyGo
     runs-on: ubuntu-latest
     env:
-      TINYGO_VERSION: 0.39.0
+      TINYGO_VERSION: 0.38.0
     timeout-minutes: 20
     steps:
       - name: Checkout
--- a/arrow/array.go
+++ b/arrow/array.go
@@ -83,6 +83,8 @@ type ArrayData interface {
 	Dictionary() ArrayData
 	// SizeInBytes returns the size of the ArrayData buffers and any children and/or dictionary in bytes.
 	SizeInBytes() uint64
+
+	Equal(ArrayData) bool
 }
 
 // Array represents an immutable sequence of values using the Arrow in-memory format.
--- a/arrow/array/array_test.go
+++ b/arrow/array/array_test.go
@@ -17,6 +17,7 @@
 package array_test
 
 import (
+	"runtime"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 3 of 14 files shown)


lidavidm (commented):
So the idea is that via [AddCleanup](https://pkg.go.dev/runtime#AddCleanup) we can avoid explicit retain/release calls? I suppose if it makes basic usage of the library simpler, then it makes sense. I think we should caution that for "library" code, it should not be used and retain/release should be used instead.

Is there a runtime cost to attaching the cleanup callback? An application that does intend to use retain/release may not want to pay the price if there is one. 

FWIW, Java may be of interest here. Java had community libraries like Netty that effectively relied on explicit retain/release callbacks to avoid GC overhead and access native memory. The modern FFM API offers an [API that is instead based around explicit "arenas"](https://docs.oracle.com/en/java/javase/22/core/memory-segments-and-arenas.html), and one of the arenas is managed by the garbage collector (= no manual memory management), while others require you to close the arena (but not individual allocations). (Incidentally, arrow-java has "the worst of both worlds": both allocators and individual allocations must be closed.) That may be too restrictive of a paradigm here, though.

zeroshade:
> I suppose if it makes basic usage of the library simpler, then it makes sense. I think we should caution that for "library" code, it should not be used and retain/release should be used instead.

Would make sense to put this into the documentation.

Alternatively, we could use a build tag to control whether it uses addcleanup or not? (turn whichever isn't in use into no-ops)

> Is there a runtime cost to attaching the cleanup callback? An application that does intend to use retain/release may not want to pay the price if there is one.

Yes, there's a runtime cost to attaching the cleanup callback (albeit, AFAIK it's very small). In theory, it would be overall cheaper for the AddCleanup solution vs the repeated calls for Retain and Release.

One important caveat to keep in mind (that technically still exists even with the current retain/release paradigm) is that users should never maintain a reference to the `[]byte` from the `Bytes()` method of `memory.Buffer` longer than the scope of the buffer (or call to Release). In the case of a non-go allocator, those bytes could get wiped out from under them when the releasing occurs. It may make sense to utilize the new weak.Ptr in some way for how we expose the byte slices from the buffer.

pixelherodev:
> Is there a runtime cost to attaching the cleanup callback? An application that does intend to use retain/release may not want to pay the price if there is one.

> Can we make Retain and Release noops in go 1.25+ and mark them as deprecated?

For what it's worth, Retain and Release are slower in my profiling than the native GC. We have a database using Arrow for part of the API, and patching out the refcounting from arrow-go cut CPU usage significantly.

zeroshade:
@lidavidm @pixelherodev what would you both think about a build flag that would make the retain/release calls no-ops and solely use `AddCleanup` so that a consumer could essentially "opt-in" if they aren't concerned about the non-deterministic nature of the GC?

lidavidm:
> @lidavidm @pixelherodev what would you both think about a build flag that would make the retain/release calls no-ops and solely use `AddCleanup` so that a consumer could essentially "opt-in" if they aren't concerned about the non-deterministic nature of the GC?

I think that's reasonable, I guess we'd need to duplicate some testing but that's OK. I assume the Go compiler will inline/dead-code-eliminate calls properly so that the overhead doesn't increase.

pixelherodev:
> One other thought, though: if we're already using memory.DefaultAllocator, the Go allocator, Free is a no-op. We shouldn't even need the cleanups, _or_ the release/retain. IMO it almost makes more sense to move the AddCleanup logic _into memory.Allocator_, and not invoke it at _all_ for the Go allocator.
> 
> That, combined with Retain/Release being disabled, would allow users of the _default allocator_ to not be paying overhead for special cases - does that design make sense?

Bonus: Cleanups are 16 bytes; this avoids an additional 16 bytes for each Buffer for the Go allocator, which will _never use them._

pixelherodev:
Copying from #269:

> The two problems with totally dropping retain/release:

> 1. AddCleanup is non-deterministic on when it runs, so for situations where the user is using a custom allocator (instead of the default Go allocator) we'd still want to make sure that the Retain/Release paradigm allows users to control when memory is freed
> 2. It's also not guaranteed to run before program exit, which may not be a problem given that typically it's just calling Free on some allocations so maybe this isn't an issue.

2 is a non-issue IMO. The non-determinism of 1 is also, I think, a bit overstated? AddCleanup is allowed to batch cleanups but the documentation says this is only likely to happen with objects that are less than sixteen bytes _and_ that do not contain pointers, neither of which applies to most of the types currently using Retain/Release.

_Practically_ speaking, the cleanups will most likely run within 1 or 2 GC cycles. Are there cases where this is insufficient? How important is it that memory be freed as quickly as possible?

pixelherodev:
> Ah, the other issue is that the memory is allocated via things like mallocator, or Go is given handles to external allocations - so the GC isn't necessarily aware of the (full extent of the) memory used. (And with external handles it's nice to be able to say, release _this handle right now_ instead of hoping the GC gets it.)

That's mostly just about timing, yeah? Having the memory be freed faster?

If the goal is to bound memory usage of the Go components, whcih aren't as controlled as the others, would it not be sufficient to set a soft memory limit and use the Go allocator instead?

pixelherodev:
> Not permissible with cgo.

Ahhhh, gotcha. That's tricky then... is this for Go calling into C, or vice versa? Is there a defined operational timespan? Trying to think if something like an arena allocator would be sufficient, to at least reuse the same memory...

pixelherodev:
If it's helpful, I can probably build on top of this PR with those changes?

pixelherodev:
> that would be super helpful, thanks! Sorry for not finding the time yet to focus on this

No worries, completely understandable!

pixelherodev:
> It sounds like the best option that doesn't break anything for anyone, while still getting at least most of the benefits, is to add a build flag to switch from retain/release to AddCleanup, and then to move the cleanup logic into the Allocators, so that anyone using the GoAllocator has no additional runtime overhead for memory management?

How to best disable retain/release is another interesting question. We could move every single copy of the Retain/Release implementation into their own files with the build tag, but that's kinda gross IMO, even if we just have it as one per package.

Might be best to have a single implementation of retain/release and refcounting, and then just embed that as a substructure wherever needed - which, actually, is probably a good change to make _on its own anyways_, so I'll start by sending in a PR with that I think

Edit: ah, except the problem is that each one needs to be able to free all of its refcounted subfields, too, not just to hand it back to the allocator. I'll think this over and se if I can find a way to clean this up...

pixelherodev:
Opened a proof-of-concept to add a build flag to disable refcounting: https://github.com/apache/arrow-go/pull/578

I'm wondering: for those of using just the Go allocator, is AddCleanup even necessary?

If retain/release is intended to be kept as is for other users, should we _just_ have RR by default, and a norc flag for Go users to turn it off?

zeroshade:
> I'm wondering: for those of using just the Go allocator, is AddCleanup even necessary?

In that scenario, it wouldn't be. But the problem is that there's no good way to know whether or not we need the addcleanup without pushing it as a responsibility onto the allocator which would defeat the purpose.

> If retain/release is intended to be kept as is for other users, should we just have RR by default, and a norc flag for Go users to turn it off?

At least for our first pass at this, I think that's the right way to go, and we can start introducing using the AddCleanup (behind a flag) behind a potential flag. If it becomes worthwhile, we might look at making the breaking change and a new major version release

PR #323, opened 2025-03-21
fix: go's atomic operations require 64bit alignment in structs on ARM

sahib:
### Rationale for this change

I've tried to run a pqarrow based writer/reader on a RaspberryPi. This lead to an almost immediate crash however, due to an unaligned struct member in `ipc.Reader`.

I did not check yet if there are other structs that might have similar issues. It would probably be nice to 

Swapping the struct field order is enough to fix the issue. See also the link in the diff.

### Are these changes tested?

Yes, no additional unit tests due to hardware constraints, but my program runs through now.

Less crashing?

### Other note

I noticed that this will also crash if I do not pass the `noasm` build tag, which is probably also a little unfortunate. This seems to be because the `funclist` in `parquet/internal/bmi/bmi.go` is empty. If you want I can add a default, so the funclist is always initialized with the fallback go implementation, unless overwritten.

### DIFF
--- a/arrow/ipc/reader.go
+++ b/arrow/ipc/reader.go
@@ -38,7 +38,7 @@ import (
 type Reader struct {
 	// refCount needs to be 64-bit aligned
 	// https://pkg.go.dev/sync/atomic#pkg-note-BUG
-	refCount int64
+	refCount atomic.Int64
 
 	r      MessageReader
 	schema *arrow.Schema
@@ -73,13 +73,14 @@ func NewReaderFromMessageReader(r MessageReader, opts ...Option) (reader *Reader
 
 	rr := &Reader{
 		r:        r,
-		refCount: 1,
+		refCount: atomic.Int64{},
 		// types:    make(dictTypeMap),
 		memo:               dictutils.NewMemo(),
 		mem:                cfg.alloc,
 		ensureNativeEndian: cfg.ensureNativeEndian,
 		expectedSchema:     cfg.schema,
 	}
+	rr.refCount.Add(1)
 
 	if !cfg.noAutoSchema {
 		if err := rr.readSchema(cfg.schema); err != nil {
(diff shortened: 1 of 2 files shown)


zeroshade:
> I noticed that this will also crash if I do not pass the noasm build tag, which is probably also a little unfortunate. This seems to be because the funclist in parquet/internal/bmi/bmi.go is empty. If you want I can add a default, so the funclist is always initialized with the fallback go implementation, unless overwritten.

`bmi_arm64.go` should be used if `noasm` is not passed which will populate the funclist accordingly. We also have CI that runs on ARM64 both with and without the `noasm` build tag, so I'm confused why it would crash in your case.

arrow/ipc/reader.go
```diff
@@ -36,12 +36,15 @@ import (
 // Reader expects a schema (plus any dictionaries) as the first messages
 // in the stream, followed by records.
 type Reader struct {
+	// refCount needs to be 64-bit aligned
+	// https://pkg.go.dev/sync/atomic#pkg-note-BUG
+	refCount int64
```
zeroshade, line 41: does this actually ensure alignment? Or should we use atomic.Int64 ?

sahib:
> `bmi_arm64.go` should be used if `noasm` is not passed which will populate the funclist accordingly. We also have CI that runs on ARM64 both with and without the `noasm` build tag, so I'm confused why it would crash in your case.

Well, that's because it's not arm64, but "regular" 32-bit arm in our case.

arrow/ipc/reader.go
sahib: It does because it's now at the start of the struct. The issue was that `schema *arrow.Schema` is a pointer, which is 32-bit on `arm` (without it being `arm64`). 

Using `atomic.Int64` would be indeed a safer option since it does [alignment automatically](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/sync/atomic/type.go;l=101) . Good point! :+1:

zeroshade:
> Well, that's because it's not arm64, but "regular" 32-bit arm in our case.

Ah! *sigh* yea, i guess it would be best to ensure we default to the pure go impls then

arrow/ipc/reader.go
sahib: I swapped with `atomic.Int64` now. I think this would need to be done on the complete code base though. If I have some time next week I can offer to do this. Along with that I could also double check that we don't have more alignment issues I did not stumble into yet.

zeroshade (approved):
Looks good! Thanks! will merge once CI passes

PR #325, opened 2025-03-21
fix(parquet/pqarrow): fix definition levels with non-nullable lists

zeroshade:
### Rationale for this change
Related to https://github.com/apache/arrow/issues/38503, it was found in https://github.com/apache/iceberg-go/issues/357 that the Parquet file being written by pqarrow was incompatible with pyarrow in some testing. After some digging I determined the cause. So we should fix this to finally put #38503 to bed

Fixing the computation of definition levels when writing from Arrow data with lists of non-nullable elements. Previously we were basing the `nullableInParent` on whether it was a list or a map, assuming that the element was *always* nullable in a list. This, of course, is incorrect and we need to actually *check* the nullability of the list element to compute the correct definition level. This way we don't produce incongruous levels that are larger than what the max definition level should be.

### Are these changes tested?
Yes, I've updated the corresponding test, which wasn't *actually* testing a non-nullable element until now.

Only fixing writing of files.

### DIFF
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -20,12 +20,14 @@ import (
 	"bytes"
 	"encoding/binary"
 	"io"
+	"strconv"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/arrow/memory"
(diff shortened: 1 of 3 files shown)


lidavidm (approved):
Is this the kind of thing where we should try to push out a new release ASAP to avoid "bad" files being written?

Is there a way to "fix" a bad file?

zeroshade:
There isn't really a way to fix a "bad" file other than re-writing it unfortunately. It might be worth a patch release to fix this

PR #326, opened 2025-03-22
fix: move from atomic.(Add|Load|Store) to atomic.Int64{}

sahib:
### Rationale for this change

This is a follow-up to https://github.com/apache/arrow-go/pull/323.

ARM 32-bit requires atomic operations to be 32-bit aligned. This has not been the case in a rather large amount of cases in the code base. Failing to align causes crashes during runtime.

I replaced all uses of `atomic.LoadInt64`, `atomic.StoreInt64` and `atomic.AddInt64` with `atomic.Int64{}` and its methods. This type has built-in alignment, so it does not matter in which order struct fields appear, which makes it generally harder to screw up alignment during changes.

### Are these changes tested?

Briefly on the 32-bit architecture with the same program mentioned in #323. The best way to move forward here would be to also add a 32bit CI environment to make sure `arrow` builds and tests there. But I guess this is out of scope for this PR.

I've noticed that `go vet ./...` produces many warnings over the project, most of them already there before this PR. The new ones are all of the kind:

```
arrow/array/dictionary.go:614:42: literal copies lock value from bldr: github.com/apache/arrow-go/v18/arrow/array.dictionaryBuilder contains github.com/apache/arrow-go/v18/arrow/array.builder contains sync/atomic.Int64 contains sync/atomic.noCopy
```

I.e. false positives due to initialization in another struct and then copying it in a constructor function.

In the hope that everything was done right: no.

### DIFF
--- a/arrow/array/array.go
+++ b/arrow/array/array.go
@@ -35,7 +35,7 @@ const (
 )
 
 type array struct {
-	refCount        int64
+	refCount        atomic.Int64
 	data            *Data
 	nullBitmapBytes []byte
 }
@@ -43,16 +43,16 @@ type array struct {
 // Retain increases the reference count by 1.
 // Retain may be called simultaneously from multiple goroutines.
 func (a *array) Retain() {
-	atomic.AddInt64(&a.refCount, 1)
+	a.refCount.Add(1)
 }
 
 // Release decreases the reference count by 1.
 // Release may be called simultaneously from multiple goroutines.
 // When the reference count goes to zero, the memory is freed.
 func (a *array) Release() {
-	debug.Assert(atomic.LoadInt64(&a.refCount) > 0, "too many releases")
+	debug.Assert(a.refCount.Load() > 0, "too many releases")
 
-	if atomic.AddInt64(&a.refCount, -1) == 0 {
+	if a.refCount.Add(-1) == 0 {
 		a.data.Release()
 		a.data, a.nullBitmapBytes = nil, nil
 	}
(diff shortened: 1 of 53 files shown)


zeroshade:
@sahib can you address the linting and other CI issues here?

sahib:
> @sahib can you address the linting and other CI issues here?

I tried to fix the linting (I think you have to trigger the pipeline explicitly?).

Interestingly, those  were not broken due to my changes, but the template was wrong. I still have (at least one) [one failing test](https://github.com/sahib/arrow-go/blob/0dc5d50a8cf07caea8bb615d354b3020490e9c6a/arrow/array/numericbuilder.gen_test.go#L240) locally. I think this [fails rightfully so]( https://github.com/sahib/arrow-go/blob/77af38fa6cbbcf097117d180fe88406c3d20dd93/arrow/array/numericbuilder.gen.go#L221), since `strcvonv.ParseInt()` will not parse stuff like `Nan` or `+Inf`. Should it be `strconv.ParseFloat()` or should this just not work for Ints?

sahib:
@zeroshade: I've gone with the more likely option of fixing the test cases.

OT question: How hard would you think would it be to enable the asm optimizations also for arm(32)? I noticed they *do* help a lot on arm64 and would be great to have them there as well.

zeroshade:
@sahib I haven't tested generating the asm for ARM32 and I don't have an ARM32 system that I could test it on if I were to generate it. 

In the ideal scenario, it should be really easy to generate the assembly via adding the entries to the Makefiles, but I'm sure it won't be that simple.

zeroshade:
Yup, I'll approve and merge this.

> Most of the asm functions are selected over function variables during runtime (i.e. during init() different implementation are set based on cpu functions). Overall a nice approach, but it comes with a small performance penalty, as this prohibits inlining (and maybe some more things?).

If I remember correctly, the Go compiler isn't able to inline assembly functions like this in the first place so we wouldn't be able to inline anyways.

> I did push a dummy branch here: https://github.com/sahib/arrow-go/tree/bench/build-tag - seems like the whole platform selection gets less convoluted using this approach as well. All in all more minor stuff, but I wanted your opinion on this first.

I like the idea, as long as the benchmarks in your branch support the idea and don't cause any performance regressions

zeroshade (approved):

PR #335, opened 2025-03-25
add UnifySchemas()

loicalleyne:
### Rationale for this change
Implement schema unification function similar to [pyarrow.unify_schemas](https://arrow.apache.org/docs/python/generated/pyarrow.unify_schemas.html)

func UnifySchemas(promotePermissive bool, schemas ...*arrow.Schema) (*arrow.Schema, error)

### Are these changes tested?
yes

util.UnifySchemas()

### DIFF
--- a/arrow/util/schemas/unify.go
+++ b/arrow/util/schemas/unify.go
@@ -1,0 +1,490 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemas
+
+import (
+	"errors"
+	"fmt"
+	"slices"
+
+	"maps"
+
+	"github.com/apache/arrow-go/v18/arrow"
+)
+
+var (
+	ErrPathNotFound     = errors.New("path not found")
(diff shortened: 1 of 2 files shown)


singh1203 (commented):
I’m diving deep to really grasp how this implementation works!

zeroshade (changes requested):
I think converting the entire schema to a tree of nodes might actually be overkill and unnecessarily complex.

It might make more sense to do the following:

1. Create a SchemaBuilder object which contains a `[]arrow.Field` and a map of name -> list of indexes (we can error on duplicate field names)
2. Add a `MergeWith` method to `arrow.Field` which contains the logic for how to merge one field with another
3. Finally, create a `mergeTypes` function that would be called from the `MergeWith` method of `arrow.Field`, which can then handle just how do you merge the current type with another (possibly promoting, etc.). This is where the logic for the individual types would be handled.

This way you don't need to create an entire tree from each schema. You only need to create a SchemaBuilder, then loop over the schemas calling `AddField` for each top level field of the schema. AddField would simply lookup whether or not the field exists, calling `Field.MergeWith` if it does or appending if it doesn't.

The implementation of merging a struct type would be to just create a schema builder and do the same thing between the struct types (since a struct type is essentially the same as a schema in a lot of ways).

Full disclosure: I got this idea by looking at the current implementation of UnifySchemas in arrow C++ (https://github.com/apache/arrow/blob/7c18001f0d7bd97471237719702c33165858bba7/cpp/src/arrow/type.cc#L426). But I do believe the result would be cleaner, simpler, and more efficient (mostly because we wouldn't need to create an entire tree of nodes for each schema).

Thoughts?

arrow/util/schemas/unify.go
```diff
@@ -0,0 +1,490 @@
...
+}
+
+func (f *treeNode) schema() (*arrow.Schema, error) {
+	var s *arrow.Schema
+	defer func(s *arrow.Schema) (*arrow.Schema, error) {
+		if pErr := recover(); pErr != nil {
+			return nil, fmt.Errorf("schema problem: %v", pErr)
+		}
+		return s, nil
+	}(s)
```
singh1203, line 257: I have a doubt: Since `s` is passed as a parameter to the `defer` function, isn't it captured as `nil` at that moment? Also, is this the correct way to pass `s` to `defer`, or should we rely on closure instead?

PR #340, opened 2025-04-01
feat(nested): export StructType metadata

AnatolyRugalev:
### Rationale for this change

`StructType` contains private `meta Metadata` field, which is not exported and only used in tests. Analogous to `Schema` and `Field` structs, I want to keep metadata of `StructType`, and it seems that it being not export is an oversight.

* Add `StructType.Metadata` method
* Add `StructOfWithMetadata` method
* Rewire `StructOf` to call `StructOfWithMetadata` with empty metadata struct

### Are these changes tested?

Yes

Yes

### DIFF
--- a/arrow/datatype_nested.go
+++ b/arrow/datatype_nested.go
@@ -404,6 +404,13 @@ type StructType struct {
 //
 // StructOf panics if there is a field with an invalid DataType.
 func StructOf(fs ...Field) *StructType {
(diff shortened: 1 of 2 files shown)


AnatolyRugalev:
I just realized that `struct` type is flatbuffers doesn't have metadata, so I will close this PR and try to figure out why that's the case.

PR #344, opened 2025-04-05
feat(parquet): add variant encoder/decoder

sfc-gh-mbojanczyk:
### Rationale for this change
This adds a basic Variant encoder/decoder to start the process of supporting the new [Variant encoding spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) in the Apache Go Parquet library. Variants are useful for efficiently storing and accessing data, especially in things like Iceberg tables.

This adds logic to encode and decode Variants, but does not yet plumb that logic through to either Arrow or Parquet. The PR's getting beefy as is, and this seems to be a good standalone unit to get feedback on.

Still to implement are the handling of decimal primitives.

For ease of implementation, the Metadata keys are only stored in unsorted order. This makes the creation of an encoded Variant simpler as one can serialize data as its being added. For sorted Metadata keys to work, you'd need to buffer data and only create objects at the very end so that the appropriate width of indicies can be chosen.

### Are these changes tested?
There are unit tests throughout to test that marshaling produces the expected binary output as per the spec, and to ensure that unmarshaling can spit out the expected values. There are many levels of unit tests, from testing individual marshaling bits to testing the marshaling and unmarshaling of entire Variants.

With this PR, no. This is simply a library to create Variants, but does not plumb the output into Parquet or Arrow.


parquet/variants/util.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	if maxPos := offset + size; maxPos > len(raw) {
+		return 0, fmt.Errorf("out of bounds: trying to access position %d, max position is %d", maxPos, len(raw))
+	}
+	var ret uint64
+	for i := range size {
+		ret |= uint64(raw[i+offset]) << (8 * i)
+	}
+	return ret, nil
+}
```
zeroshade, line 40: `binary.LittlEndian.Uint64(raw[offset:size])` ?

parquet/variants/util.go
```diff
@@ -0,0 +1,154 @@
...
+	} else if maxSize < 0xFFFFFF {
+		return 3
+	}
+	return 4
+}
+
+// Checks that a given range is in the provided raw buffer.
+func checkBounds(raw []byte, low, high int) error {
+	maxPos := len(raw)
+	if low >= maxPos {
```
zeroshade, line 65: we should be also validating that both values are >= 0 and that high >= low

parquet/variants/util.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Encodes a number of a specified width in little-endian format and writes to a writer.
+func encodeNumber(val int64, size int, w io.Writer) {
+	buf := make([]byte, size)
+	for i := range size {
+		buf[i] = byte(val)
+		val >>= 8
+	}
+	w.Write(buf)
```
zeroshade, line 81: why not use the `encoding/binary` functions to do this instead? such as `binary.LittleEndian.Put*` or `binary.LittleEndian.Append*` etc.

parquet/variants/util.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func isLarge(numItems int) bool {
+	return numItems > 0xFF
+}
+
+// Returns the basic type the passed in value should be encoded as, or undefined if it cannot be handled.
+func kindFromValue(val any) BasicType {
```
zeroshade, line 89: we should cache the conversions somewhere if possible, reflection is slow so let's avoid duplicating calls to it if we don't need to

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	case bool:
+		return marshalBoolean(val, w), nil
+	case int:
+		return marshalInt(int64(val), w), nil
+	case int8:
+		return marshalInt(int64(val), w), nil
+	case int16:
+		return marshalInt(int64(val), w), nil
+	case int32:
+		return marshalInt(int64(val), w), nil
```
zeroshade, line 155: why marshal  everything as `int64`? Why not marshal them as their actual type via generics?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+		return marshalInt(val, w), nil
+	case float32:
+		return marshalFloat(val, w), nil
+	case float64:
+		return marshalDouble(val, w), nil
+	case string:
+		if allOpts&MarshalAsUUID != 0 {
+			return marshalUUID([]byte(val), w), nil
+		}
+		return marshalString(val, w), nil
```
zeroshade, line 169: can we accept `uuid.UUID` instead?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+// local timezone set.
+func marshalPrimitive(v any, w io.Writer, opts ...MarshalOpts) (int, error) {
+	var allOpts MarshalOpts
+	for _, o := range opts {
+		allOpts |= o
+	}
+	switch val := v.(type) {
+	case bool:
+		return marshalBoolean(val, w), nil
+	case int:
```
zeroshade, line 148: what about uints?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	prim, err := primitiveFromHeader(raw[offset])
+	if err != nil {
+		return err
+	}
+
+	switch prim {
+	case primitiveNull:
+		dest.Set(reflect.Zero(dest.Type()))
+	case primitiveTrue, primitiveFalse:
+		if kind != reflect.Bool && kind != reflect.Interface {
```
zeroshade, line 212: why `kind != reflect.Interface`?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+			return fmt.Errorf("cannot decode Variant value of %s into dest %s", prim, kind)
+		}
+		dest.Set(reflect.ValueOf(prim == primitiveTrue))
+	case primitiveInt8, primitiveInt16, primitiveInt32, primitiveInt64:
+		iv, err := decodeIntPhysical(raw, offset)
+		if err != nil {
+			return err
+		}
+		if kind == reflect.Interface {
+			dest.Set(reflect.ValueOf(iv))
```
zeroshade, line 222: I don't understand why you're checking and using `reflect.Interface` here. Can you explain? For reference, `reflect.Interface` just means the type is *an* interface, it doesn't mean it is the empty interface (i.e. `interface{}` or `any`)

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+		} else if dest.CanInt() {
+			if dest.OverflowInt(iv) {
+				return fmt.Errorf("int value of %d will overflow dest", iv)
+			}
+			switch kind {
+			case reflect.Int:
```
zeroshade, line 228: what about the uint types?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+		if err != nil {
+			return err
+		}
+		if kind == reflect.Slice && dest.Type().Elem().Kind() == reflect.Uint8 || kind == reflect.Interface {
+			dest.Set(reflect.ValueOf(bytes))
+		} else if kind == reflect.String {
+			dest.Set(reflect.ValueOf(string(bytes)))
+		} else {
+			return fmt.Errorf("cannot decode Variant value of %s into dest %s", prim, kind)
+		}
```
zeroshade, line 351: this could be simplified with a generic most likely

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+		if err != nil {
+			return err
+		}
+		if kind == reflect.Slice && dest.Type().Elem().Kind() == reflect.Uint8 || kind == reflect.Interface {
+			dest.Set(reflect.ValueOf(bytes))
+		} else if kind == reflect.String {
+			dest.Set(reflect.ValueOf(string(bytes)))
+		} else {
+			return fmt.Errorf("cannot decode Variant UUID into dest %s", kind)
+		}
```
zeroshade, line 363: what if they used uuid.UUID? or `[16]byte`?

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+
+	return nil
+}
+
+func marshalNull(w io.Writer) int {
+	hdr, _ := primitiveHeader(primitiveNull)
+	w.Write([]byte{hdr})
+	return 1
+}
```
zeroshade, line 375: the return for all of these should probably be (int, error) and just do `return w.Write(...)`

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+func marshalUUID(uuid []byte, w io.Writer) int {
+	hdr, _ := primitiveHeader(primitiveUUID)
+	w.Write([]byte{hdr})
+
+	// A UUID is 16 bytes. Either pad or truncate to this length.
+	if len(uuid) > 16 {
+		uuid = uuid[:16]
+	} else if pad := 16 - len(uuid); pad > 0 {
+		uuid = append(uuid, make([]byte, pad)...)
+	}
```
zeroshade, line 530: a uuid must be exactly 16 bytes, we should error if it is not exactly 16 bytes

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	w.Write(uuid)
+	return 17
+}
+
+func unmarshalUUID(raw []byte, offset int) ([]byte, error) {
+	if err := checkBounds(raw, offset, offset+17); err != nil {
+		return nil, err
+	}
+	return raw[offset+1 : offset+17], nil
+}
```
zeroshade, line 540: use `[16]byte` or `uuid.UUID` please

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+			return "", fmt.Errorf("end index is out of bounds: trying to access position %d, max position is %d", endIdx, maxPos)
+		}
+		return string(raw[offset+1 : endIdx]), nil
+	}
+
+	b, err := getBytes(raw, offset+1)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
```
zeroshade, line 563: use `unsafe.String` to ensure we don't perform an unnecessary copy here.

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+	bt := BasicTypeFromHeader(raw[offset])
+
+	if bt == BasicShortString {
+		l := int(raw[offset] >> 2)
+		endIdx := 1 + l + offset
+		if endIdx > maxPos {
+			return "", fmt.Errorf("end index is out of bounds: trying to access position %d, max position is %d", endIdx, maxPos)
+		}
+		return string(raw[offset+1 : endIdx]), nil
```
zeroshade, line 556: use `unsafe.String` to avoid an unnecessary copy

parquet/variants/primitive.go
```diff
@@ -0,0 +1,678 @@
...
+	return 9
+}
+
+// func decodeTimestamp(raw []byte, offset int) (int64, error) {
+// 	ts, err := readUint(raw, offset+1, 8)
+// 	if err != nil {
+// 		return -1, err
+// 	}
+// 	return int64(ts), nil
+// }
```
zeroshade, line 640: remove commented out code?

zeroshade (commented):
Made an initial pass on this and left a bunch of comments. I'll try to take another look later on

sfc-gh-mbojanczyk:
> Made an initial pass on this and left a bunch of comments. I'll try to take another look later on

Thanks so much! I'll get to addressing the comments here shortly (was out for a few days)

parquet/variants/util.go
sfc-gh-mbojanczyk: Similar to the comment in `readUint()`, the `binary` package only provides Put/Append functionality for `uint{16,32,64}`, but we need it for the whole range from 1-8 bytes.

parquet/variants/util.go
sfc-gh-mbojanczyk: Good call. Done, plus I realized that I forgot tests for this function, so I added that as well.

parquet/variants/util.go
sfc-gh-mbojanczyk: Hmm, this is an interesting point, I need to chew on this a bit.

My gut instinct is that this isn't terribly cacheable, the only place where we _could_ potentially cache is for elements in a slice/array (assuming we don't have nested slices) or for elements in a map (ditto). NB, this also only gets hit in the `builder.Marshal()` codepath- one can also totally build things with the provided builders (`Primitive(), Object() and Array()`), and avoid reflection entirely.

sfc-gh-mbojanczyk (commented):
Mostly addressed your comments here, added some context for the others. The biggest diff IMO is using `uuid.UUID` now, which definitely simplifies some stuff around there.

I've also moved this review out of draft. Figure it's in pretty decent reviewable shape as-is.

Thanks for the first pass- looking forward to polishing this up!

parquet/variants/util.go
zeroshade: Then you could probably just do something like:

```go
switch size {
case 1:
    var val uint8
    _, err := binary.Decode(raw[offset:size], binary.LittleEndian, &val)
    return val, err
case 2:
    var val uint16
    _, err := binary.Decode(raw[offset:size], binary.LittleEndian, &val)
    return val, err
case 4:
    var val uint32
    _, err := binary.Decode(raw[offset:size], binary.LittleEndian, &val)
    return val, err
case 8:
    var val uint64
    _, err := binary.Decode(raw[offset:size], binary.LittleEndian, &val)
    return val, err
}
```

right?

parquet/variants/util.go
zeroshade: you can use `binary.Encode(buf, binary.LittleEndian, val)` which will handle the whole range for you

parquet/variants/util.go
sfc-gh-mbojanczyk: The subtlety here: you need to be able to decode things of widths 1-8 _inclusive_. For example, the [field offset size](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#value-header-for-object-basic_type2) of an object can be 1, 2, 3, or 4 bytes. `binary.Decode` can handle the 1, 2, and 4 case, but cannot handle the 3-byte width case (see [this example](https://go.dev/play/p/5QKZ1ove5Ab) in the Go playground).

For the odd-width cases I suppose we could pad the buffer with additional zeros, but that's getting gnarly just to make `binary.Decode` work IMO

parquet/variants/util.go
sfc-gh-mbojanczyk: Similar vein to `Decode()`- this will work for the powers-of-two widths, but if we've got a width that's not (ie. 3) we'd encode with additional padding since we've got to encode to the next power of two width.

This isn't the end of the world (encoding over the minimal necessary width _is_ within spec), so I have fewer objections here. I'd argue with `Decode()` rolling our own is more of a necessity due to not being in control of what comes in (eg. the Java library _can_ encode 3-byte wide numbers, so it's gotta be handled). This keeps the encode and decode logic fairly similar. Also, FWIW, I feel like the intention of the spec authors is to minimize the number of bytes used to encode things, see the existence of `Short String` even though a primitive `String` type exists.

zeroshade: The problem with rolling our own here is that we're going to need to manage the big/little endian logic ourselves then so that this runs properly on big-endian systems.

parquet/variants/util.go
sfc-gh-mbojanczyk: Ah, I see where your confusion is coming from (the spec doesn't make it easy to find all these in one place).

There are _so many places_ where widths are set, `is_large` is solely one place. In this case `is_large` indicates whether `num_elements` in the [Object Data](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#value-data-for-object-basic_type2) is 1 or 4 bytes. I'm talking about `field_id` and `field_offset` sizes. From the header documentation:

> `field_offset_size_minus_one` and `field_id_size_minus_one` are 2-bit values that represent the number of bytes used to encode the field offsets and field ids.

In this case, `field_id` and `field_offset` can be 1-4 bytes.

Perhaps a better example here is in the [header](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#value-header-for-short-string-basic_type1) for `Short String`:

> When `basic_type` is 1, value_header is a 6-bit `short_string_header` ... The `short_string_header` value is the length of the string.

So a short string's length can be represented by 1-6 bytes inclusive.

sfc-gh-mbojanczyk: (Sorry, misread the Short String header, it's six bits. Removed that example)

parquet/variants/primitive.go
```diff
@@ -0,0 +1,674 @@
...
+	case int:
+		return marshalInt(int64(val), w), nil
+	case int8:
+		return marshalInt(int64(val), w), nil
+	case int16:
+		return marshalInt(int64(val), w), nil
+	case int32:
+		return marshalInt(int64(val), w), nil
+	case int64:
+		if allOpts&MarshalAsTime != 0 {
```
zeroshade, line 160: instead of needing this option, why not just use `arrow.Time32` or `arrow.Time64`? if they want a Time they should pass that.

parquet/variants/primitive.go
zeroshade: do you mind if I push some changes to the branch with a possible solution?

sfc-gh-mbojanczyk: Please do! I'm not territorial about this :)

zeroshade: I didn't delete any of your code, but rather commented it out where I replaced it with some generic stuff.

My personal opinion is that we should encode based on the type they give us rather than trying to downcast for them. If they want it to be smaller and use fewer bytes, they can downcast it themselves :smile:

Anyway, let me know what you think of the changes I made

sfc-gh-mbojanczyk:
> @sfc-gh-mbojanczyk Hey, sorry for the long delay here. I've gone through multiple iterations and discussions with people, plus pulling inspiration from the C++, Spark, and parquet-java implementations to figure out a good design. Please let me know what you think of the updated version! Thanks!

No worries- I got caught in a whirlwind over here myself and was about to dust this off too :) Lemme take a peek here after the long weekend.

lidavidm (approved):
I like this, this is very clean and straightforward

parquet/variant/variant.go
```diff
@@ -0,0 +1,725 @@
...
+const (
+	BasicUndefined   BasicType = iota - 1 // Unknown
+	BasicPrimitive                        // Primitive
+	BasicShortString                      // ShortString
+	BasicObject                           // Object
+	BasicArray                            // Array
+)
+
+func basicTypeFromHeader(hdr byte) BasicType {
+	return BasicType(hdr & basicTypeMask)
```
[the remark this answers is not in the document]
zeroshade: turns out there's no need for a `debug.Assert` here, because we're doing `& basicTypeMask` we're guaranteed the result is going to be in the correct range for the `BasicType` enum: `[0,3]`

parquet/variant/variant_test.go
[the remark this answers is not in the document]
lidavidm: Thanks. I wonder if fuzzing the variants specifically would be a more manageable case to start with.

lidavidm (approved):

parquet/variant/variant.go
```diff
@@ -0,0 +1,722 @@
...
+	// to be outside of the range of BasicType. Therefore, we don't
+	// need to perform any checks. The value will always be [0,3]
+	return BasicType(hdr & basicTypeMask)
+}
+
+// PrimitiveType represents specific primitive data types within the variant format.
+type PrimitiveType int
+
+const (
+	PrimitiveInvalid            PrimitiveType = iota - 1 // Unknown
```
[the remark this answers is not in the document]
zeroshade: These constants have specific values defined by the spec, so the zero value needs to be Null as the constant for "Null" type is 0. I don't really have wiggle room to change that. Though, technically if all variant building is done through the builder, we could probably get away without exporting this enum and these constants.

parquet/variant/builder.go
```diff
@@ -0,0 +1,852 @@
...
+//
+// Importantly, the value data is the returned variant value is not copied here. This will
+// return the raw buffer data owned by the builder's buffer. If you wish to reuse a builder,
+// then the [Value.Clone] method must be called on the returned value to copy the data before
+// calling [Builder.Reset]. This enables trivial cases that don't reuse the builder to avoid
+// performing this copy.
+func (b *Builder) Build() (Value, error) {
+	nkeys := len(b.dictKeys)
+	totalDictSize := 0
+	for _, k := range b.dictKeys {
```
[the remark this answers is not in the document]
zeroshade: Mostly just out of convenience and saving the couple bytes in the size of the builder itself (which is negligible honestly). There's no reason I can't keep a running tally in AddKey, I'll make that change.

parquet/variant/basic_type_string.go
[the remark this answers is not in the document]
zeroshade: sure, i'll update the go-generate lines and have it rename the files.

parquet/variant/doc.go
```diff
@@ -0,0 +1,142 @@
...
+// Values can also be converted to JSON:
+//
+//	jsonBytes, err := json.Marshal(v)
+//
+// # Low-level Construction
+//
+// For direct construction of complex nested structures, you can use the low-level
+// methods:
+//
+//	var b variant.Builder
```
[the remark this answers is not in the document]
zeroshade: >  Things also get tricky with sorting of metadata keys and building (as you can't know the index of the dictionary key until build time, which then changes an Object's index, which can then change the size of offsets, etc...).

Well that's an entirely different ball of worms. Even if you use the nested object approach, things are still tricky if you want to sort the metadata keys as you end up having to patch, modify, or entirely rewrite any already written objects when you change the field index by sorting the metadata keys. (See the [rust impl](https://github.com/apache/arrow-rs/pull/7452/files#diff-83c4572d83951bc1234fd12d3cc6537b1e025188f06c39b378b0f7b9c32aa4f3)). So using the nested approach with multiple buffers doesn't help us make sorting metadata keys more efficient.

> I don't think it can be done

yea, that's why the API looks the way it does right now, so that we can build everything using a single buffer instead of needing multiple buffers and copying a bunch of stuff around. 

I've managed to sketch something out that enables using parent/child builders to manage the nested building while still requiring that the object be built in order. Kind of a compromise between the approaches, I'll try to clean it up a bit and see if it's worth pushing it as an option.

> Once again, up to you. At least with child builders the parent builder can keep track of all nested things, and either error out if its children haven't called Finish(), or the parent can call each child's Finish()

Yea, there's definitely benefits to having the parent builders to keep track of the nested children in terms of usability. Just trying to figure out if the code complexity is worth it.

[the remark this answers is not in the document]
zeroshade: Okay, what do you think of something like this:

```go
	var b variant.Builder

	obj, err := b.NewObject()
        // handle err

	obj.AppendInt("id", 1)

	observe, err := obj.AppendObject("observation")
        // handle err

	observe.AppendString("location", "In the Volcano")
	observe.AppendString("time", "12:34:56")

	value, err := observe.AppendObject("value")
	// handle err

	value.AppendInt("humidity", 456)
	value.AppendInt("temperature", 123)
	value.Finish()
	observe.Finish()

	species, err := obj.AppendObject("species")
	// handle err
	species.AppendString("name", "lava monster")
	species.AppendInt("population", 6789)
	species.Finish()

	obj.Finish()
        v, err := b.Build()
	// handle err
	out, err := json.Marshal(v)	
	/* result is this `{
		"id": 1,
		"observation": {
			"location": "In the Volcano",
			"time": "12:34:56",
			"value": {
				"humidity": 456,
				"temperature": 123
			}
		},
		"species": {
			"name": "lava monster",
			"population": 6789
		}
	}` */
```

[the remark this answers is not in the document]
sfc-gh-mbojanczyk: I like it. I assume you'll do some handling if the parent has `Finish()` called without the children's `Finish()` (either panic, or just call their `Finish()` methods, right?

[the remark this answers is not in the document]
lidavidm: (Also, presumably we can add the higher level builders separately/on top of this API? Does it have to be built into the builder for this first pass?)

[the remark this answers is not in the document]
zeroshade: > I like it. I assume you'll do some handling if the parent has Finish() called without the children's Finish() (either panic, or just call their Finish() methods, right?

Yea, `Finish()` errors if the children haven't had `Finish()` called first.

> You can also still just directly append a map or slice instead of keeping track manually - would that solve your use case? (Also, presumably we can add the higher level builders separately/on top of this API? Does it have to be built into the builder for this first pass?)

I think @lidavidm has a good point here. The "compromise" example I put together is just built on top of the existing Builder API I've put together by creating the "higher-level" API on top of the "lower-level" one that I've already built. Any objections to pushing through what we've already got here and then following up with the higher-level builder that handles the offsets and fields for you as a separate PR afterwards @sfc-gh-mbojanczyk?

sfc-gh-mbojanczyk:
Can't actually toss in my approval since I'm the original author, but this PR looks good to me.

parquet/variant/doc.go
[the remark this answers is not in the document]
zeroshade: > At the end of the day, this is your codebase and I'm just tossing in my 2c.

Yup, but I still value the feedback. Especially as I am often *way* too deep in the weeds and will miss things that can affect usability by consumers. So the feedback is important.

>  `Append()` does the overwhelming majority of the heavy lifting for users and is probably going to be the most used entry point, so this PR is fully featured enough (ie. it's a good hole for folk to drop into).

Cool. I'll push this as is for now and then we can look at creating the higher-level "nicer" API for simplifying object/array construction later on. Another reason is that I want to get the basic variant support out and my "compromise" version has a lot of duplicated code in there that I wasn't able to get to my liking yet.

PR #352, opened 2025-04-11
feat(parquet/schema): initial variant logical type

### LINKED ISSUE #348: Support Variant Logical Type

Initial implementation of the Logical type definition and an arrow extension type for Variant as per following the pattern used in https://github.com/apache/arrow/pull/45375

### Component(s)

Parquet

zeroshade:
### Rationale for this change
Closes #348 

This is heavily based on https://github.com/apache/arrow/pull/45375 for reference.

Initial implementation of the Variant Logical Type in the Parquet schema package and an arrow extension type to represent it.

This *does not* implement actually encoding or decoding the variant data, nor does it implement actually reading or writing data. To this end, the extension type *is not exposed* or otherwise exported from this package currently so it can only be constructed internally for now until there is a standardized canonical extension representation for it.

### Are these changes tested?
Yes, unit tests are added for the changes.

Only the new types and ability to manipulate schemas with Variant annocated fields.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -18,6 +18,7 @@ package pqarrow
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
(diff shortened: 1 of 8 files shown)


zeroshade:
CC sfc-gh-mbojanczyk

parquet/pqarrow/schema.go
```diff
@@ -306,6 +323,11 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
 			}, fieldIDFromMeta(field.Metadata))
 		}
 		return schema.MapOf(field.Name, keyNode, valueNode, repFromNullable(field.Nullable), fieldIDFromMeta(field.Metadata))
+	case arrow.EXTENSION:
+		extType := field.Type.(arrow.ExtensionType)
+		if extType.ExtensionName() == "parquet.variant" {
```
paleolimbot, line 328: This is the name that's currently in C++ too, but should this just be "arrow.variant" to avoid having both canonical Parquet extension types and canonical Arrow extension types?

paleolimbot (commented):
Awesome!

It seems as though the complexity of variant is in the shredding, but this seems to me like the scope of the C++ PR.

parquet/pqarrow/schema.go
zeroshade: For now this isn't exported so it can't be created and isn't being registered so it won't be utilized except in this spot. When we create a canonical arrow extension type, I figure all the implementations would just update and change the name?

I'm mostly just using this name to maintain compatibility with the C++ impl for now.

lidavidm (approved):

PR #357, opened 2025-04-17
Compress: add MarshalText and UnmarshalText

jdemeyer:
### Rationale for this change

Right now, there is no public API to convert a string like `"ZSTD"` to a compression codec. There is only https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.2.0/parquet/internal/gen-go/parquet#CompressionCodecFromString but that can't be used since it's in an `internal` package.

Adding `UnmarshalText` can be used for applications taking such a string as configuration option. Adding `MarshalText` isn't strictly needed since it overlaps with `String()` but makes sense for symmetry.

Add `UnmarshalText` and `MarshalText` as thin wrappers around the same methods from the internal https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.2.0/parquet/internal/gen-go/parquet#CompressionCodec type.

### Are these changes tested?

Yes

Minor new API for `Compression` type

### DIFF
--- a/parquet/compress/compress_test.go
+++ b/parquet/compress/compress_test.go
@@ -139,18 +139,39 @@ func TestCompressReaderWriter(t *testing.T) {
 	}
 }
 
+var marshalTests = []struct {
+	text  string
+	codec compress.Compression
+}{
+	{"UNCOMPRESSED", compress.Codecs.Uncompressed},
+	{"SNAPPY", compress.Codecs.Snappy},
+	{"GZIP", compress.Codecs.Gzip},
+	{"LZO", compress.Codecs.Lzo},
+	{"BROTLI", compress.Codecs.Brotli},
+	{"LZ4", compress.Codecs.Lz4},
+	{"ZSTD", compress.Codecs.Zstd},
+	{"LZ4_RAW", compress.Codecs.Lz4Raw},
+}
+
 func TestMarshalText(t *testing.T) {
-	compression := compress.Codecs.Zstd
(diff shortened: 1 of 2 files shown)


parquet/compress/compress_test.go
```diff
@@ -138,3 +138,23 @@ func TestCompressReaderWriter(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, "ZSTD", string(data))
+}
+
+func TestUnmarshalText(t *testing.T) {
+	var compression compress.Compression
+	err := compression.UnmarshalText([]byte("ZSTD"))
+	assert.NoError(t, err)
+	assert.Equal(t, compress.Codecs.Zstd, compression)
+}
```
zeroshade, line 154: I figured we should probably use a table test and test all the codes. i.e.

```go
func TestMarshalText(t *testing.T) {
    tests := []struct{
            codec ....
            text string
    }{
         {compress.Codecs.Zstd, "ZSTD"},
         {compress.Codecs.Snappy, "SNAPPY"},
          ...
      }

      for _, tt := range tests {
          t.Run(tt.text, func(t *testing.T) {
               data, err := tt.codec.MarshalText()
               assert.NoError(t, err)
               assert.Equal(t, tt.text, string(data))
               ...
          })
      }
}
```

and so on

zeroshade (approved):
Thanks for this! This is helpful and I'll probably end up updating iceberg-go to utilize it

PR #359, opened 2025-04-21
fix(arrow/array): optional struct array with required field

zeroshade:
### Rationale for this change
https://github.com/apache/iceberg-go/issues/398 discovered that the current `NewStructArrayWithFields` fails if any child is marked as non-nullable but has nulls (as would be the case in an optional struct array full of nulls with a required field). So we need to allow constructing such an array.

A new function is created, `NewStructArrayWithFieldsAndNulls` which takes in the top level struct null bitmap, the number of nulls, offset columns and list of fields.

### Are these changes tested?
Yes, a unit test was created for it.

The above case that would error, now will no longer error.


lidavidm (approved):
I feel like Arrow should either sit down and specify the semantics of a nullable field, or else it should explicitly declare that validation is up to the application

PR #362, opened 2025-04-28
chore(arrow): remove most lock copies

zeroshade:
### Rationale for this change
#326 moved to using `atomic.Int64{]` for the refcount, as a result we found many places that we were copying the value. For linting reasons we should remove those.

Utilizing generics to clean up the dictionary array code significantly, along with removing most of the cases where we were copying the refcount which contains a lock.

### Are these changes tested?
Yes, the existing unit tests cover it all.

Nope.


arrow/array/dictionary.go
```diff
@@ -1402,143 +1107,178 @@ func (b *BinaryDictionaryBuilder) ValueStr(i int) string {
...
-	data := arrow.DayTimeIntervalTraits.CastToBytes(arr.values)
-	for len(data) > 0 {
-		if err = b.insertDictValue(data[:arrow.DayTimeIntervalSizeBytes]); err != nil {
-			break
-		}
-		data = data[arrow.DayTimeIntervalSizeBytes:]
-	}
-	return
-}
+// type Decimal32DictionaryBuilder struct {
```
lidavidm, line 1169: Did you mean to leave this code commented out?

lidavidm (approved):

arrow/array/dictionary.go
zeroshade: Nope, oops. Thanks!

lidavidm (approved):

PR #369, opened 2025-05-07
fix: TestDeltaByteArray implementation and fix

MetalBlueberry:
This exact test seems to be passing on v17 version. More investigation is needed

### Rationale for this change

After updating from v17 to v18 I noticed a test case failing that was passing before.

This PR contains a test case to reproduce the issue. A reference PR for version v17 can be found here
- https://github.com/apache/arrow/pull/46353

### Are these changes tested?

The PR itself just contains the test

### DIFF
--- a/parquet/file/file_reader_test.go
+++ b/parquet/file/file_reader_test.go
@@ -860,11 +860,9 @@ func TestDeltaByteArray(t *testing.T) {
 
 	for rr.Next() {
 		rec := rr.Record()
-		defer rec.Release()
-
-		for i := 0; i < int(rec.NumCols()); i++ {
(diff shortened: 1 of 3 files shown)


zeroshade:
Can you provide the csv file that you're testing with so that I can try to reproduce?

MetalBlueberry:
@zeroshade see last commit for the fix.

I'm just pattern matching, I do not really know what those values mean :sweat_smile: .

zeroshade:
I figured out what the underlying issue was and updated this to be a better fix that won't cause the other tests to fail :smile:

zeroshade (approved):

PR #375, opened 2025-05-12
feat: Extend arrow csv writter

MetalBlueberry:
### Rationale for this change

I've been looking for a way to convert parquet files to CSV. The arrow library seems to be doing a good job, until I found out that.

- Not every type was supported
- It was impossible to change how types were formatted

My goal is to generate CSV that can be used with `COPY TO` statement in postgresql, For example, I needed to make sure binary fields are hex encoded instead of base64.

After thinking a bit about it, I considered adding a custom type converted option that will allow any user of the csv.Writer to change the behavior for specific types. This is just a function that will be called before the standard mapping to allow any custom logic to handle the type. 

Tests `TestCustomTypeConversion` in this PR show how it can be used.

- Option for csv.Writer to set a `CustomTypeConversion`
- Remove csv.Writer schema type validation. (It will fail on write if type is not handled)
- Add tests for csv.Writer based on apache/parquet_testing `alltypes_plain.parquet` and `delta_byte_array.parquet`
- Add test for `CustomTypeConversion` 

### Are these changes tested?

Yes

- Invalid schemas won't fail on csv.Writer creation but on first write. I believe this is easier as it sets a single source of truth for type validation.
- if CustomTypeConverstion option is not set, there are no changes in the behavior.


lidavidm:
> My goal is to generate CSV that can be used with `COPY TO` statement in postgresql, For example, I needed to make sure binary fields are hex encoded instead of base64.

BTW, you can directly COPY Arrow data to Postgres with the ADBC driver, depending on what you're doing this might save you a step. These examples happen to be in Python, but you can do the same in Go (loading the PostgreSQL driver with the Go driver manager): https://arrow.apache.org/adbc/current/python/recipe/postgresql.html#create-append-to-a-table-from-an-arrow-dataset

MetalBlueberry:
Thanks for you advice @lidavidm
I found https://github.com/apache/arrow-adbc/tree/main/go/adbc but it is not clear at the first glance how to use it. I tried to quickly set it up but failed. It complains about `Must SetSqlQuery before ExecuteQuery` but I'm not even calling ExecuteQuery :boom: I'm using ExecuteUpdate.

Also I do not have any strong reasons to improve performance, for now I prefer the simplicity and compatibility of CSV files. So this change is still valid.

I'll be happy to discuss the adbc approach further. But provably this thread is not the right place to do so. Send me an email if you are willing to discuss it as well :smile: vperez@timescale.com is provably the best place to find me

zeroshade:
@MetalBlueberry The current scripts only initialize and sync the parquet-testing submodules right before running the tests for the `parquet` package. The tests for the csv package are in the `testdata` directory, can we either avoid relying on the parquet-testing files for the csv package? Or we have to update the scripts.

MetalBlueberry:
I think relying on parquet-testing data gives us a common foundation. Otherwise I'll end up writing the go code to generate the exact same parquet file :sweat_smile: 

I've updated the test script in the last commit. I think I got it right :crossed_fingers:

arrow/csv/common.go
```diff
@@ -237,15 +237,26 @@ func WithStringsReplacer(replacer *strings.Replacer) Option {
 	}
 }
 
-func validate(schema *arrow.Schema) {
+func WithCustomTypeConverter(converter func(typ arrow.DataType, col arrow.Array) (result []string, handled bool)) Option {
```
zeroshade, line 240: can we add a docstring comment to this to explain how it works etc. ?

zeroshade (changes requested):
just one change, otherwise all looks great to me :smile:

MetalBlueberry:
Thanks for pointing out the doc string :rocket:

zeroshade (approved):
Thanks!

PR #380, opened 2025-05-19
feat(parquet/pqarrow): parallelize SeekToRow

### LINKED ISSUE #379: Follow up for https://github.com/apache/arrow-go/issues/278

I had a try with `RecordReader`.`SeekToRow` as mentioned in https://github.com/apache/arrow-go/issues/278. I find the performance for seeking is slower than actually read the same amount of data (or even more).
I see parallel is used during read but not during seek, do you think it is possible to also use parallel to improve the performance of seek operation? Thanks a lot!

### Component(s)

Parquet

zeroshade:
### Rationale for this change
Closes #379

Update the `SeekToRow` method of the record reader to parallelize the calls to `SeekToRow` for the columns if the Parallel option is set to true.

### Are these changes tested?
Yes, the RecordReader SeekToRow is already tested via unit tests with multiple columns.

There should only be the performance benefit

### DIFF
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -697,13 +697,22 @@ func (r *recordReader) SeekToRow(row int64) error {
 		return fmt.Errorf("invalid row index %d, file only has %d rows", row, r.numRows)
 	}
 
-	for _, fr := range r.fieldReaders {
-		if err := fr.SeekToRow(row); err != nil {
-			return err
-		}
+	var (
+		np = 1
+		g  errgroup.Group
+	)
+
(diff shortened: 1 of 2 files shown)


zeroshade (commented):
@Sunyue can you try this branch out and let me know if it is sufficient for your issue?

lidavidm (approved):

Sunyue:
> @Sunyue can you try this branch out and let me know if it is sufficient for your issue?

Looks good to me. I tested with a 400+ column parquet file from ADLV2, seek used took 4mins+ and now it decreased to 15s.

PR #395, opened 2025-05-30
feat(arrow/extensions): Add Variant extension type, array, and builder

### LINKED ISSUE #351: Variant Shredding

Shred variants in accordance with the [Variant Shredding Spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) referencing https://github.com/apache/arrow/issues/45948

### Component(s)

Parquet

zeroshade:
### Rationale for this change
closes #351

Moves the definition of the Variant extension type to `arrow/extensions` and automatically registers it as a canonical extension type.

In addition to creating `VariantType`, and `VariantArray` types, it also adds a `VariantBuilder` for building variant extension arrays, including performing shredding if necessary based on the schema being used.

### Are these changes tested?
Yes

Just the new types and functionality.


parquet/variant/builder.go
```diff
@@ -777,6 +791,18 @@ func (b *Builder) FinishObject(start int, fields []FieldEntry) error {
 	return nil
 }
 
+// AppendEncoded is a special case where we directly append a pre-encoded variant
+// value. Its keys must already be in the dictionary and v must already be
+// a properly encoded variant value. No checking is performed here currently, so
+// be careful as this can easily lead to an invalid variant result.
+func (b *Builder) AppendEncoded(v []byte) error {
```
lidavidm: Should this be named UnsafeApppendEncoded perhaps? I guess it still returns an error, but only in the case of Write itself failing...

parquet/variant/builder.go
```diff
@@ -848,3 +882,27 @@ func (b *Builder) Build() (Value, error) {
...
+	constraints.Integer | constraints.Float | string | []byte |
+		arrow.Date32 | arrow.Time64 | arrow.Timestamp | bool |
+		uuid.UUID | DecimalValue[decimal.Decimal32] |
+		DecimalValue[decimal.Decimal64] | DecimalValue[decimal.Decimal128]
+}
+
+// Encode is a convenience function that produces the encoded bytes for a primitive
+// variant value. At the moment this is just delegating to the [Builder.Append] method,
+// but in the future it will be optimized to avoid the extra overhead and reduce allocations.
+func Encode[T variantPrimitiveType](v T, opt ...AppendOpt) ([]byte, error) {
```
lidavidm, line 896: Should variantPrimitiveType be public if it's used in a public signature?

arrow/array/dictionary.go
```diff
@@ -1652,6 +1652,32 @@ func UnifyTableDicts(alloc memory.Allocator, table arrow.Table) (arrow.Table, er
...
 }
 
+type dictWrapper[T arrow.ValueType] struct {
+	*Dictionary
+
+	typedDict arrow.TypedArray[T]
+}
+
+// NewDictWrapper creates a simple wrapper around a Dictionary array provides the
+// a Value method which will use the underlying dictionary to return the value
```
lidavidm:
```suggestion
// NewDictWrapper creates a simple wrapper around a Dictionary array that provides
// a Value method which will use the underlying dictionary to return the value
```

arrow/extensions/variant.go
```diff
@@ -0,0 +1,1536 @@
...
+// The rules for a variant storage type are:
+//  1. MUST be a struct
+//  2. MUST have required field named "metadata" that is binary/largebinary/binary_view
+//  3. Must satisfy exactly one of the following:
+//     a. MUST have required field named "value" that is binary/largebinary/binary_view
+//     b. MUST have an optional field named "value" that is binary/largebinary/binary_view
+//     and another optional field named "typed_value" that is either a primitive type or
+//     a list/large_list/list_view or struct which also satisfies the following requirements:
+//     i. The elements must be REQUIRED
+//     ii. There must either be a single REQUIRED field named "value" which is
```
lidavidm: I think Arrow tends to use "non-nullable" instead of "required"

arrow/extensions/variant.go
```diff
@@ -0,0 +1,1536 @@
...
+//     binary/largebinary/binary_view or havve an optional "value" field and an optional
```
lidavidm:
```suggestion
//     binary/largebinary/binary_view or have an optional "value" field and an optional
```

arrow/extensions/variant.go
```diff
@@ -0,0 +1,1536 @@
...
+}
+
+func (v *VariantType) NewBuilder(mem memory.Allocator) array.Builder {
+	return NewVariantBuilder(mem, v)
+}
+
+func isBinary(dt arrow.DataType) bool {
+	return dt.ID() == arrow.BINARY || dt.ID() == arrow.LARGE_BINARY ||
+		dt.ID() == arrow.BINARY_VIEW
+}
```
lidavidm, line 197: There's IsBinaryLike already, but that doesn't cover BINARY_VIEW (should it?)

arrow/extensions/variant.go
```diff
@@ -0,0 +1,1536 @@
...
+	}
+}
+
+func (b *binaryDictBuilderAdapter) UnsafeAppend(v []byte) {
+	b.Append(v)
+}
+
+func (b *binaryDictBuilderAdapter) Append(v []byte) {
+	if err := b.BinaryDictionaryBuilder.Append(v); err != nil {
+		panic(fmt.Sprintf("error appending value %s to binary dictionary builder: %v", string(v), err))
```
lidavidm, line 926: Why does Append panic? Shouldn't it return an error and let UnsafeAppend panic?

parquet/variant/builder.go
zeroshade: I didn't want it to be public as I don't intend for others to use this defined type set externally, it's just for defining the constraints on this function (or elsewhere we use it) rather than intending it to be used externally to this lib. This is sufficient to enable users to call this, while restraining what types can be set, without allowing it to get referenced directly externally.

arrow/extensions/variant.go
zeroshade: The primary reason why the `BinaryDictionaryBuilder` doesn't meet the usual builder interface is that the standard builders *don't* return an error for their `Append` methods, the dictionary builder is unique in that way. Technically the only way you could really end up with an error coming from the dictionary builder is if something goes *really* wrong with the internal map used for the dictionary building. In practice, unless you run out of memory (leading to further problems) it's unlikely that you'd hit this at all. But in order to meet the Append interface without returning an error I didn't want to throw away the error if a user somehow hit this condition.

lidavidm: Ah, ok, sounds good.

arrow/extensions/variant.go
zeroshade: I was thinking about that, I hadn't decided yet but I think you're right that we should update it to cover BINARY_VIEW

zeroshade: hmm, `IsBinaryLike` returns true for both String and Binary, whereas we want this to only be binary so I don't think that updating `IsBinaryLike` would work here.

parquet/variant/builder.go
zeroshade: that's fair, I'll rename it.

arrow/extensions/variant.go
lidavidm: Ah, fair enough.

lidavidm (approved):

sfc-gh-mbojanczyk (approved):
The arrow stuff is definitely outside of my wheelhouse, but from a Go standpoint looks good to me. Variant stuff also looks good!

PR #397, opened 2025-06-05
fix(parquet/pqarrow): Fix propagation of field-ids for Lists

zeroshade:
### Rationale for this change
An issue was found in apache/iceberg-go when attempting to retrieve data from a table containing a List Column that had a struct as the element. It was failing to propagate the element-id for the fields when fetching. I tracked it down to the schema handling here.

Changes the `getNestedFactory` method in pqarrow/schema.go to use `ListOfField` instead of `ListOf` so that it preserves the metadata, i.e. the field id.

### Are these changes tested?
Yes, a test has been added to cover this scenario.

Previously this situation would result in a field-id of -1, now users will see the field-id get propagated correctly.


zeroshade (commented):
CC @shubham-tomar this fixes the issue you brought up to me on the iceberg-go slack. After this is merged, I'll update the dependency in iceberg-go.

PR #409, opened 2025-06-17
ci(benchmark): Fix benchmark runs

### LINKED ISSUE #408: Benchmark result upload fails

https://github.com/apache/arrow-go/actions/runs/15616399947/job/43990198178#step:8:19
```
python3 -m pip install benchadapt@git+https://github.com/conbench/conbench.git@main#subdirectory=benchadapt/python
  python3 ci/scripts/bench_adapt.py
python3: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /__t/Python/3.9.23/x64/lib/libpython3.9.so.1.0)
python3: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /__t/Python/3.9.23/x64/lib/libpython3.9.so.1.0)
```
The issue is caused by the debian image, I am not sure why we are even running this in a container instead on the runner directly? Missed that in review. @singh1203 @zeroshade 

### Component(s)

Continuous Integration

zeroshade:
### Rationale for this change
Fixes #408 

Run the benchmark directly on the runner instead of using a container

Using a branch on the repo itself instead of a fork so that it has access to the secrets to actually run the benchmark and confirm this fixes the issue.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -29,8 +29,7 @@ permissions:
   contents: read
 jobs:
   benchmark:
-    runs-on: ubuntu-latest
-    container: debian:12
+    runs-on: ubuntu-latest    
     strategy:
       matrix:
         go: ['1.22.7']


kou (approved):
+1

PR #412, opened 2025-06-18
fix(arrow/flight/flightsql): fix sqlite dependency for windows

zeroshade:
### Rationale for this change
modernc.org/sqlite doesn't build on windows, but there's a shim that will switch to a different sqlite lib for unsupported platforms.

Using github.com/uptrace/bun/sqliteshim to dynamically pick the proper driver to use based on the platform.

### Are these changes tested?
Yes

No


zeroshade:
ASAN issue with modernc.org/sqlite i think we're gonna have to find a better library...

PR #414, opened 2025-06-18
ci: make additional checks to prevent flaky EOF

zeroshade:
### Rationale for this change
Additional checks for `io.EOF` to avoid flaky tests

### DIFF
--- a/arrow/flight/flightsql/example/sql_batch_reader.go
+++ b/arrow/flight/flightsql/example/sql_batch_reader.go
@@ -21,7 +21,6 @@ package example
 
 import (
 	"database/sql"
-	"errors"
 	"io"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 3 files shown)


arrow/flight/flightsql/example/sql_batch_reader.go
```diff
@@ -264,7 +265,7 @@ func (r *SqlBatchReader) Next() bool {
 
 	rows := 0
 	for rows < maxBatchSize && r.rows.Next() {
-		if err := r.rows.Scan(r.rowdest...); err != nil {
+		if err := r.rows.Scan(r.rowdest...); err != nil && err.Error() != io.EOF.Error() {
```
lidavidm, line 268: Huh...it's quite weird that this can return EOF

lidavidm (approved):

arrow/flight/flightsql/example/sql_batch_reader.go
zeroshade: From what I can tell, it's something weird with the sqlite handler implementation. it shouldn't be returning EOF, which looks like it isn't even returning `io.EOF` and instead just returns the string "EOF" it's really weird.

Issue #415, opened 2025-06-19
Support hamba decoded time.Time into arrow timestamp


Willem-J-an:
### Describe the enhancement requested

I've run into the following problem:
- We have field with "logicalType" : "timestamp-millis"
- This is internally decoded by hambaavro to time.Time
- time.Time is [not supported in the creation of arrow record](https://github.com/apache/arrow-go/blob/2e570031ca8c8ffcb04ef98b6f6618e0905c0574/arrow/avro/reader_types.go#L861-L875)
- OCFReader panics because it expected values for the timestamp field but no values were added

I'm wondering if adding support for time.Time in the appendTimestampData function is the way to go, or if there are any other suggestions to resolve this issue?
I've created this PR as a suggested way to resolve:
https://github.com/apache/arrow-go/pull/416

### Component(s)

Integration

zeroshade:
resolved by #416

PR #416, opened 2025-06-19
feat: support avro decoded time struct

Willem-J-an:
### Rationale for this change
#415

Suggestion to add support for supporting time.Time decoded value on timestamps for arrow record creation

### Are these changes tested?
I have some tests in my own project where with this change:
```									
"type" : [ "null", {"type" : "long","logicalType" : "timestamp-millis"}]
is converted to:
arrow.TimestampType{Unit: arrow.Millisecond, TimeZone: "UTC"}
```

No API changes, just broader compatability.

### DIFF
--- a/arrow/avro/reader_types.go
+++ b/arrow/avro/reader_types.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
@@ -871,5 +872,11 @@ func appendTimestampData(b *array.TimestampBuilder, data interface{}) {
 		case int64:
 			b.Append(arrow.Timestamp(v))
 		}
+	case time.Time:
+		v, err := arrow.TimestampFromTime(dt, b.Type().(*arrow.TimestampType).Unit)
+		if err != nil {
+			panic(err)
+		}
+		b.Append(v)
 	}
 }


zeroshade:
Thanks for this, can you add a test for it?

Willem-J-an:
> Thanks for this, can you add a test for it?

Will do when I get around to it!

Willem-J-an:
> Thanks for this, can you add a test for it?

@zeroshade I've added a test with some sample data based on the pre-existing avro schema that was in the repo. It encodes the data to avro & to json, and validates when the avro is decoded that it matches the json. For the json encoder I aligned it to what I saw the arrow record as json produced.

I'm wondering what happened here, as quite a few types were not working. Maybe hamba had some breaking changes and due to no tests they were not detected?

What do you think about the reader_type switch cases that do not seem to be applicable anymore; are there valid cases where these are still needed or does it make sense to delete them?

zeroshade:
Which cases do you see are no longer applicable and not working? the tests that are being removed seem to be types that should be supported

Willem-J-an:
Tests that are being removed? I think no tests are being removed?

As an example, the cases that I'm referring to is this:
```
func appendFixedSizeBinaryData(b *array.FixedSizeBinaryBuilder, data interface{}) {
	switch dt := data.(type) {
	case nil: # <- is fine of course 
		b.AppendNull()
	case []byte: # <- this case does not apply at least in my testing because hamba decoder returns array, not a slice
		b.Append(dt)
	case map[string]any: # <- this case also didn't apply in my testing
		switch v := dt["bytes"].(type) {
		case nil:
			b.AppendNull()
		case []byte:
			b.Append(v)
		}
```
Not sure if these are legacy or still relevant in cases that I cannot oversee!

zeroshade:
You can add the file to the `dev/release/rat_exclude_files.txt` that should exclude it from the license check

zeroshade:
A `fixed` should probably map to a `fixed_len_byte_array[N]` if we can easily detect it, if we can't then I think it's okay to remove it for now if it's not currently working.

Willem-J-an:
I added working fixes for all the types already, including for fixed!

Willem-J-an:
Added the schema file to the rat exclude, that should fix the windows tests. Any other feedback I'd need to look at? Thanks!

PR #418, opened 2025-06-20
feat(arrrow/compute/expr): support substrait timestamp and decimal properly

### LINKED ISSUE #404: Support arrow.TIME64 arrow.TIMESTAMP .. in exprs.ToSubstraitType()

I am trying to do filter on arrow table/record with `compute` and `exprs` package, and found filtering for arrow.TIME64, arrow.TIMESTAMP, arrow.DECIMAL128, arrow.DECIMAL256 is not working as expected.

Could you please help to support arrow.TIME64 arrow.TIMESTAMP ... in exprs.ToSubstraitType()
https://pkg.go.dev/github.com/apache/arrow-go/v18/arrow/compute/exprs#ToSubstraitType

Looking at the code arrow.DECIMAL128, arrow.DECIMAL256 seems are supported, but not work well, maybe a bug?

Expected: SetInputSchema(), exprs.ExecuteScalarExpression() can work well with arrow.TIME64 arrow.TIMESTAMP, arrow.DECIMAL128, arrow.DECIMAL256

Note: I find it relatively simple to support arrow.TIME64 arrow.TIMESTAMP by adding cases to `exprs.ToSubstraitType()`, but it seems only `arrow.Microsecond` is supported as it is hard coded in `FromSubstraitType`, any possibility to derive the unit from the input schema?

### LINKED ISSUE #417: substrait-go@v4
### Describe the usage question you have. Please include as many useful details as possible.

I noticed that there was a major release from `substrait-go` a few months ago. Is there a plan to upgrade `arrow-go` to this new version?

### Component(s)

Other

zeroshade:
### Rationale for this change
Fixes #404
Fixes #417 

Upgrades substrait-go to v4 and adds handling and support for PrecisionTime and PrecisionTimestamp, fixes substrait Decimal128Type handling.

### Are these changes tested?
Yes, unit test is added.

only the new features being usable.

Relies on https://github.com/substrait-io/substrait-go/pull/139 getting merged before this can get merged


arrow/compute/utils.go
```diff
@@ -353,19 +354,45 @@ func commonTemporal(vals ...arrow.DataType) arrow.DataType {
...
 
-	switch {
-	case zone != nil:
-		// at least one timestamp seen
-		return &arrow.TimestampType{Unit: finestUnit, TimeZone: *zone}
-	case sawDate64:
-		return arrow.FixedWidthTypes.Date64
-	case sawDate32:
-		return arrow.FixedWidthTypes.Date32
+	sawTimestampOrDate := zone != nil || sawDate32 || sawDate64 || sawDuration
```
lidavidm: nit: this also includes duration

arrow/compute/utils.go
```diff
@@ -353,19 +354,45 @@ func commonTemporal(vals ...arrow.DataType) arrow.DataType {
...
+
+	if sawTime && sawTimestampOrDate {
+		// no common type possible
+		return nil
+	}
+
```
lidavidm: This blocks (time AND any other type) but I would assume (duration AND any other type) is also out of the question?

lidavidm: Maybe all these checks could be combined into the three cases (saw timestamp or date and not time or duration) and (saw time and not anything else) and (saw duration and not anything else)?

arrow/compute/utils.go
lidavidm, line 1: Personally, I think commonTemporal is complicated enough to potentially warrant specific unit tests for it

arrow/compute/utils.go
zeroshade: Was following the existing [`CommonTemporal`](https://github.com/apache/arrow/blob/dacec3080f8a7c245dc884d278ea6280942086f0/cpp/src/arrow/compute/kernels/codegen_internal.cc#L314) in arrow C++. Should we also change that function too?

arrow/compute/utils.go
zeroshade: updated

arrow/compute/utils.go
lidavidm: Ah, probably, though I guess it's small enough to not really warrant its own PR...

lidavidm (approved):

PR #428, opened 2025-06-27
refactor(parquet/internal/encoding): Refactor parquet logic to use generics

zeroshade:
### Rationale for this change
Refactoring the code to utilize generics allows for simplification and removal of the code generation in many locations. This will also allow for pushing the generics further up the stack to make more of the public APIs generic-friendly for easier future development.

The code generation for the internal/hashing/MemoTable and it's subsequent objects along with the creation and population of encoders/decoders in parquet/internal/encoding are also replaced by utilizing generics. To maintain the existing interface code, aliases are created for encoders/decoders/MemoTables which alias to the generic versions so that we can slowly push the generics through rather than having to propagate it as part of this change.

### Are these changes tested?
There should be no changes to functionality, as a result the existing unit tests will confirm that we don't have any regressions.

There shouldn't be.


internal/hashing/xxh3_memo_table.go
```diff
@@ -97,14 +106,14 @@ const (
...
-}
+// func max(a, b uint64) uint64 {
+// 	if a > b {
+// 		return a
+// 	}
+// 	return b
+// }
 
-var isNan32Cmp = func(v float32) bool { return math.IsNaN(float64(v)) }
+// var isNan32Cmp = func(v float32) bool { return math.IsNaN(float64(v)) }
```
kou: Can we remove them instead of commenting out them?

kou (approved):
+1

lidavidm (approved):

internal/hashing/xxh3_memo_table.go
zeroshade: bah, i knew i missed something. i'll remove those

PR #429, opened 2025-06-28
feat(parquet/examples): enhance library examples

milden6:
### Rationale for this change
Enhancing the library examples

2 new examples on how to use the `arrow-go` library:
- Writing parquet file
- Reading parquet file

### Are these changes tested?
Not applicable

No, just enhancing examples


zeroshade:
Instead of standalone examples like this, can we use Go testable examples that will show up in the docs and be tested to ensure they don't get out of date?

milden6:
> Instead of standalone examples like this, can we use Go testable examples that will show up in the docs and be tested to ensure they don't get out of date?

I did it like here, what the difference?
https://github.com/apache/arrow-go/tree/main/arrow/examples/table_creation

zeroshade:
My preference would be to follow [this pattern] (https://github.com/apache/arrow-go/blob/main/arrow/array/example_type_conversion_test.go)

Here's a good explanation of it: https://go.dev/blog/examples

zeroshade:
That path works great

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Create parquet writer props with snappy compression
+	writerProps := parquet.NewWriterProperties(
+		parquet.WithCompression(compress.Codecs.Snappy),
+	)
+
+	// Create arrow writer props to store the schema in the parquet file
+	arrowWriterProps := pqarrow.NewArrowWriterProperties(
+		pqarrow.WithStoreSchema(),
+	)
```
zeroshade: Should we elaborate in the comment about the benefit to storing the Arrow schema in the metadata?

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+	recordBuilder := array.NewRecordBuilder(memory.DefaultAllocator, schema)
+
+	// Create a builder for each field
+	intFieldIdx := schema.FieldIndices("intField")[0]
+	stringFieldIdx := schema.FieldIndices("stringField")[0]
+	listFieldIdx := schema.FieldIndices("listField")[0]
+
+	intFieldBuilder := recordBuilder.Field(intFieldIdx).(*array.Int32Builder)
+	stringFieldBuilder := recordBuilder.Field(stringFieldIdx).(*array.StringBuilder)
+	listFieldBuilder := recordBuilder.Field(listFieldIdx).(*array.ListBuilder)
```
zeroshade: do we need to do the `FieldIndices` calls when we know what the schema is and the indices are since we created the schema? There might be benefits to keeping this as simple as possible

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+		log.Fatalf("Failed to write record: %v", err)
+
+	}
+
+	record.Release()
+	recordBuilder.Release()
+
+	// IMPORTANT: Close the writer to finalize the file
+	if err := writer.Close(); err != nil {
+		log.Printf("Failed to close parquet writer: %v", err)
```
zeroshade: `log.Fatalf`?

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+		log.Fatalf("Failed to get schema: %v", err)
+	}
+
+	// colIndices can be nil to read all columns. Here, we specify which columns to read
+	colIndices := make([]int, len(colNames))
+	for idx := range colNames {
+		colIndices[idx] = schema.FieldIndices(colNames[idx])[0]
+	}
+
+	// Get the current record from the reader
```
zeroshade:
```suggestion
	// Get a record reader from the file to iterate over
```

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+	recordReader, err := arrowReader.GetRecordReader(context.TODO(), colIndices, nil)
+	if err != nil {
+		log.Fatalf("Failed to get record reader: %v", err)
+	}
+	defer recordReader.Release()
+
+	for recordReader.Next() {
+		// Create a record
+		record := recordReader.Record()
+		record.Retain()
```
zeroshade: We only need to call `retain` if we need to utilize the record beyond and outside of this loop. Since we only interact with the record in the loop and aren't storing it for use beyond the loop we don't need to call `Retain` and `Release`. Might be better to simplify this

parquet/example_write_read_pq_test.go
```diff
@@ -0,0 +1,183 @@
...
+		listCol := record.Column(colIndices[2]).(*array.List)
+		listValueCol := listCol.ListValues().(*array.Float32)
+
+		// Iterate over the rows within the current record
+		for idx := range int(record.NumRows()) {
+			// For the list column, get the start and end offsets for the current row
+			start, end := listCol.ValueOffsets(idx)
+
+			fmt.Printf("%d  %s  %v\n", intCol.Value(idx), stringCol.Value(idx), listValueCol.Float32Values()[start:end])
+		}
```
zeroshade: Do we need to grab each column separately like this instead of just calling `fmt.Println(record)` ?

parquet/example_write_read_pq_test.go
milden6: Yes, you are right, what about this comment?

```go
// WithStoreSchema embeds the original Arrow schema into the Parquet file metadata,
// allowing it to be accurately restored when reading. This ensures correct handling
// of advanced types like dictionaries and preserves full type fidelity across write/read
```

parquet/example_write_read_pq_test.go
milden6: Yep, sure, I'll specify the indices explicitly for simplicity:

```go
	intFieldBuilder := recordBuilder.Field(0).(*array.Int32Builder)
	stringFieldBuilder := recordBuilder.Field(1).(*array.StringBuilder)
	listFieldBuilder := recordBuilder.Field(2).(*array.ListBuilder)
```

parquet/example_write_read_pq_test.go
milden6: Thanks, I will fix it

parquet/example_write_read_pq_test.go
milden6: Thanks

parquet/example_write_read_pq_test.go
milden6: Okay, thanks for clarification

parquet/example_write_read_pq_test.go
milden6: Are you sure? If I write it like this:

```go
for recordReader.Next() {
	// Create a record
	record := recordReader.Record()
	fmt.Println(record)
}
```

The output will be like this:
```
record:
  schema:
  fields: 3
    - intField: type=int32
          metadata: ["PARQUET:field_id": "-1"]
    - stringField: type=utf8
             metadata: ["PARQUET:field_id": "-1"]
    - listField: type=list<element: float32, nullable>
           metadata: ["PARQUET:field_id": "-1"]
  rows: 2
  col[0][intField]: [38 13]
  col[1][stringField]: ["val1" "val2"]
  col[2][listField]: [[1 2 4 8] [1 2 4 8]]

record:
  schema:
  fields: 3
    - intField: type=int32
          metadata: ["PARQUET:field_id": "-1"]
    - stringField: type=utf8
             metadata: ["PARQUET:field_id": "-1"]
    - listField: type=list<element: float32, nullable>
           metadata: ["PARQUET:field_id": "-1"]
  rows: 2
  col[0][intField]: [53 93]
  col[1][stringField]: ["val3" "val4"]
  col[2][listField]: [[1 2 4 8] [1 2 4 8]]

record:
  schema:
  fields: 3
    - intField: type=int32
          metadata: ["PARQUET:field_id": "-1"]
    - stringField: type=utf8
             metadata: ["PARQUET:field_id": "-1"]
    - listField: type=list<element: float32, nullable>
           metadata: ["PARQUET:field_id": "-1"]
  rows: 1
  col[0][intField]: [66]
  col[1][stringField]: ["val5"]
  col[2][listField]: [[1 2 4 8]]
```

Instead of:
```
38  val1  [1 2 4 8]
13  val2  [1 2 4 8]
53  val3  [1 2 4 8]
93  val4  [1 2 4 8]
66  val5  [1 2 4 8]
```

Or did I misunderstand you?

parquet/example_write_read_pq_test.go
zeroshade: that looks good! Perhaps also mention that some of the other Parquet libraries also support this so it helps ensure cross-language type consistency too during reading.

parquet/example_write_read_pq_test.go
zeroshade: Nope you didn't misunderstand, perhaps change the batch size to be 3 instead of 2 so that it only outputs two records?

milden6: I think it’s best to leave it as is, since it demonstrates how to access cell values for each row. Working with records isn’t always obvious for beginners.

Like here: https://github.com/apache/arrow-go/blob/084371ed663ee877561f64749e5f4f7d8cf454d3/parquet/file/file_reader_test.go#L687-L696

What do you think?

zeroshade: fair enough, we can leave it as is.

zeroshade (approved):
Thanks so much for this. It will significantly help people with interacting with this library!

PR #430, opened 2025-06-30
feat(arrow/compute): support some float16 casts

zeroshade:
### Rationale for this change
Support casting float16 arrays to/from int and float32/float64

Implementation of new casting kernels for cast_float, cast_half_float, cast_int32

### Are these changes tested?
Unit tests are added to account for this.

The only user-facing change is that float16.Num is no longer a struct with bits member but instead a type definition.

As a result, using `float16.Num{}` must be replaced with `float16.Num(0)` if it is used.


zeroshade (commented):
CC @Sunyue

zeroshade:
Yea, I'm trying to figure out if there's a way I can do this easily without making the breaking change part.

PR #433, opened 2025-07-02
feat(parquet/pqarrow): Correctly handle Variant types in schema

zeroshade:
### Rationale for this change
Updating the `pqarrow` package to handle the variant extension type when converting between arrow and parquet schemas.

Replacing the TODOs with implementations to handle shredded variant structures in schema conversion.

### Are these changes tested?
A unit test is added for shredded variant handling.

Only that this is now supported instead of erroring.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -865,8 +865,6 @@ func mapToSchemaField(n *schema.GroupNode, currentLevels file.LevelInfo, ctx *sc
 }
 
 func variantToSchemaField(n *schema.GroupNode, currentLevels file.LevelInfo, ctx *schemaTree, _, out *SchemaField) error {
-	// this is for unshredded variants. shredded variants may have more fields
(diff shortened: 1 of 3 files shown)


parquet/pqarrow/schema.go
```diff
@@ -859,8 +867,10 @@ func mapToSchemaField(n *schema.GroupNode, currentLevels file.LevelInfo, ctx *sc
 func variantToSchemaField(n *schema.GroupNode, currentLevels file.LevelInfo, ctx *schemaTree, _, out *SchemaField) error {
 	// this is for unshredded variants. shredded variants may have more fields
 	// TODO: implement support for shredded variants
```
lidavidm: Is this comment still up to date?

zeroshade: bah, missed one! I can remove that now :) thanks!

lidavidm (approved):

PR #437, opened 2025-07-14
fix(arrow/flight/flightsql): drain channel in flightSqlServer.DoGet

### LINKED ISSUE #435: flightsql.BaseServer does not exhaust channel in case of an error writing StreamChunks in DoGet(...)

Hey everyone,

As this is my first time discussing on this repository, kindly forgive me in case of a mistake.

While exploring Arrow Flight SQL servers with DuckDB, I came across an issue.

Let's assume we have a Flight SQL client that sends SQL queries to a FlightSQL server, while the data is being streamed [here (sqlite example)](https://github.com/apache/arrow-go/blob/b196d3b316d09f63786f021d4f1baa1fdd7620d2/arrow/flight/flightsql/example/sqlite_server.go#L488), if the client receives a non handled SIGINT, the communication between server and client stops (without a cancel request), and the server catches the error [here](https://github.com/apache/arrow-go/blob/b196d3b316d09f63786f021d4f1baa1fdd7620d2/arrow/flight/flightsql/server.go#L928).
However, I still have an arrow array somewhere buffered in memory (server side) and the record reader that has not been released and therefore I still have arrow arrays in the wild waiting to be `free`d.

When modifying the source code [here](https://github.com/apache/arrow-go/blob/b196d3b316d09f63786f021d4f1baa1fdd7620d2/arrow/flight/flightsql/server.go#L928), while checking for an error adding a for loop that exhaust the channel to release the StreamChunk resolves the issue.

```go
func exhaustChannelAndReleaseStreamChunks(cc <-chan flight.StreamChunk) {
	for chunk := range cc {
		chunk.Data.Release()
	}
}
```

```go
if err = wr.WriteWithAppMetadata(chunk.Data, chunk.AppMetadata); err != nil {
    exhaustChannelAndReleaseStreamChunks(cc)
    return err
}
```

Did I misunderstand something?
I can take care of the PR if need be.
I've been testing that manually. Do you think I can test the behaviour rigorously?

Thank you for your time.

### Component(s)

Other

arnoldwakim:
### Rationale for this change
This PR fixes #435. 
If writing to the RecordWriter fails, the current implementation does not drain the channel when an error occurs while writing to the record writer.

We add a deferred function that will drain the channel and call `.Release()` on the underlying `StreamChunks` in order to be able to release the RecordReader.

### Are these changes tested?
@zeroshade as you suggested, I'll gladly accept a recommendation for the test setup. 

No.

### DIFF
--- a/arrow/flight/flightsql/server.go
+++ b/arrow/flight/flightsql/server.go
@@ -916,6 +916,14 @@ func (f *flightSqlServer) DoGet(request *flight.Ticket, stream flight.FlightServ
 		return err
 	}
 
+	defer func() {
+		for chunk := range cc {
+			if chunk.Data != nil {
+				chunk.Data.Release()
+			}
+		}
+	}()
+
 	wr := flight.NewRecordWriter(stream, ipc.WithSchema(sc))
 	defer wr.Close()
 


zeroshade:
Curious, somehow this change is causing a deadlock in the tests. I'll see if I can dig a bit and figure out the cause.

awakim:
> Curious, somehow this change is causing a deadlock in the tests. I'll see if I can dig a bit and figure out the cause.

I see what the issue was, and that's my bad. For some reason I did not push the code I intended in the first place.
On my local machine, I ran the test when draining the channel after [this error check](https://github.com/apache/arrow-go/blob/b196d3b316d09f63786f021d4f1baa1fdd7620d2/arrow/flight/flightsql/server.go#L915), which was fine. 
The version I pushed was deferring before the error check causing the deadlocks since it was the equivalent of this: 

```go
var cc <-chan flight.StreamChunk

for chunk := range cc {
    ...
}
``` 

I have amended the commit and pushed the intended version. Sorry for the inconvenience.

zeroshade (approved):
Since I plan on cutting a release, I think I'll get this in now and work out a test for it as a separate PR just to ensure this fix goes out with the release.

arnoldwakim:
Thanks for taking care of this 🫡

PR #440, opened 2025-07-14
fix(arrow/compute/exprs): Handle large types in expr handling

zeroshade:
### Rationale for this change
An issue was discovered in https://github.com/apache/iceberg-go when working with compacted files and parquet files written by pyiceberg. Essentially it boiled down to the fact that if an expression contained a Field Reference to a LargeString field, evaluation would fail as it expected `String` and not `LargeString` when converting from the Substrait types. This difference shouldn't cause an issue so we need to ensure that LargeString is allowed when String is expected, same for LargeBinary and Binary

Simply adding a case to the error condition when the types aren't equal to allow String + LargeString or Binary + LargeBInary to be allowed.

### Are these changes tested?
A unit test has been added to test it.

Just the new case being allowed when it was previously disallowed.

### DIFF
--- a/arrow/compute/exprs/exec.go
+++ b/arrow/compute/exprs/exec.go
@@ -395,6 +395,11 @@ func literalToDatum(mem memory.Allocator, lit expr.Literal, ext ExtensionIDSet) 
 // You can provide an allocator to use through the context via compute.WithAllocator.
 //
 // You can provide the ExtensionIDSet to use through the context via WithExtensionIDSet.
+//
+// Note: Substrait expressions do not have an equivalent to LargeString/LargeBinary, so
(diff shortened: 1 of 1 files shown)


lidavidm (commented):
Hmm. Isn't there a binding step where this could be resolved and the "real" type propagated? Would silently allowing large string in place of string (or vice versa) potentially trip up some kernels inside an expression tree?

lidavidm:
(And would we eventually have to do this with StringView too?)

zeroshade:
Sure, I'll update the docs. Though you wouldn't end up getting anything different than what the Arrow schema has unless you're field referencing a mix of string and large string in the same function call, resulting in one getting casted for the kernel. Otherwise you're just gonna get the same type as the input.

PR #447, opened 2025-07-22
fix(arrow/compute/exprs): fix literalToDatum for precision types

### LINKED ISSUE #446: Follow up on issue https://github.com/apache/arrow-go/issues/404

I had a try with enhancement from https://github.com/apache/arrow-go/issues/404 using v18.4.0, when executing exprs.ExecuteScalarExpression it still gives me an unimplemented error. The error is raised from arrow/compute/exprs/exec.go literalToDatum(), maybe those new PrecisionTimeType or PrecisionTimestampType should also be supported in this method? Thanks.

### Component(s)

Other

zeroshade:
### Rationale for this change
While #404 fixed support for substrait precision types on inputs, it did not address support for literals that use the Precision types. 

Fixes #446

Fix `literalToDatum` so that it properly supports the precision types for literals.

### Are these changes tested?
Yes, a unit test is added.

Just the increased support


Sunyue:
> @Sunyue please take a look at this and confirm this fixes your issue. Thanks!

Looks good! Any chance we can get a patch release? Thanks!

PR #450, opened 2025-07-25
fix(arrow/array): update timestamp json format

### LINKED ISSUE #445: timestamps are marshaled as non-RFC3339 format despite Go convention of using that format

[`(*array.Timestamp).MarshalJSON`](https://github.com/apache/arrow-go/blob/513dc09a09384ae8ec8dad9ae4ea1ecc88680209/arrow/array/timestamp.go#L106-L113) calls [`(*array.Timestamp).ValueStr`](https://github.com/apache/arrow-go/blob/513dc09a09384ae8ec8dad9ae4ea1ecc88680209/arrow/array/timestamp.go#L90-L97) via [`(*array.Timestamp).GetOneForMarshal`](https://github.com/apache/arrow-go/blob/513dc09a09384ae8ec8dad9ae4ea1ecc88680209/arrow/array/timestamp.go#L99-L104) where the time format layout is "2006-01-02 15:04:05.999999999Z0700" which does not agree with the Go convention of using RFC3339 for JSON-encoded timestamp values (shown [here](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/time/format_rfc3339.go;l=62-80) called via [`time.Time.MarshalJSON`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/time/time.go;l=1584-1593)).

Doing this breaks downstream users of these JSON documents since most will use encoding/json, which expects valid RFC3339 timestamps ([here](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/time/format_rfc3339.go;l=155-188) via [here](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/time/time.go;l=1595-1609)).

### Component(s)

Parquet

zeroshade:
### Rationale for this change
resolves #445

Use `time.RFC3339Nano` for the format to output the time value

### Are these changes tested?
Yes

Changes the format of timestamp strings for `ValueStr` and marshalling to JSON to match JSON and Go conventions.

### DIFF
--- a/arrow/array/timestamp.go
+++ b/arrow/array/timestamp.go
@@ -93,7 +93,7 @@ func (a *Timestamp) ValueStr(i int) string {
 	}
 
 	toTime, _ := a.DataType().(*arrow.TimestampType).GetToTimeFunc()
-	return toTime(a.values[i]).Format("2006-01-02 15:04:05.999999999Z0700")
+	return toTime(a.values[i]).Format(time.RFC3339Nano)
 }
 
 func (a *Timestamp) GetOneForMarshal(i int) interface{} {
--- a/arrow/array/timestamp_test.go
+++ b/arrow/array/timestamp_test.go
@@ -23,7 +23,9 @@ import (
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/internal/json"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


zeroshade (commented):
@efd6 does this solve your issue?

amoeba (approved):

efd6:
Yeah, this change is essentially what I was hoping for. Thanks.

Thanks @amoeba for the test suggestion; this will prevent future skew away from stdlib JSON behaviour.

zeroshade:
I'd be open to adding a configuration to control the format, but given the current release the default should be the current format used in the most recent release.

PR #453, opened 2025-07-26
refactor: switch golang.org/x/exp to standard library packages

ufUNnxagpM:
### Rationale for this change

This change migrates several of usages of `golang.org/x/exp` packages to use the standard library equivalents that have been introduced into newer go versions.

### Are these changes tested?

Tested with `go test ./...`. I get the exact same behavior (including some failures) as on master.

n/a

### DIFF
--- a/arrow/compute/exec/utils.go
+++ b/arrow/compute/exec/utils.go
@@ -19,8 +19,10 @@
 package exec
 
 import (
+	"cmp"
 	"fmt"
 	"math"
+	"slices"
 	"sync/atomic"
 	"unsafe"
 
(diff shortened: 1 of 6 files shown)


arrow/compute/exec/utils.go
```diff
@@ -51,14 +51,14 @@ func GetSpanOffsets[T int32 | int64](span *ArraySpan, i int) []T {
...
 		return a
 	}
 	return b
 }
 
-func Max[T constraints.Ordered](a, b T) T {
+func Max[T cmp.Ordered](a, b T) T {
 	if a > b {
 		return a
 	}
```
zeroshade, line 64: I need to get around to removing these entirely and just using the updated built-ins.... but that doesn't have to happen for this PR

PR #461, opened 2025-08-05
feat(parquet): only allocate index buffer once for `RleDecoder`

joechenrh:
### Rationale for this change

I think the `Decode` function of each reader should be call sequentially? Then we don't need to allocations of these slices each time. If I understand wrong, I will close this. Thanks!

Move `indexbuffer` into `RleDecoder`


joechenrh:
Thanks for the advise, I just wrote a simple test code to read parquer and ran pprof, then found this. Let me further check it again.

Btw, is the CI only triggered after any review?

PR #466, opened 2025-08-11
refactor(arrow): rename Record to RecordBatch and add deprecated alias

Mandukhai-Alimaa:
### Rationale for this change

Rename the Record interface to RecordBatch for clarity, since Record commonly means a single row but this type represents a batch of rows.

  - Renamed the Record interface to RecordBatch in /arrow/record.go
  - Updated interface method signatures to use RecordBatch in return types (SetColumn and NewSlice)
  - Added a deprecated type alias type Record = RecordBatch for backward compatibility to avoid breaking existing
  code and ease migration
  - Added proper Go deprecation annotation (// Deprecated:) to ensure tooling support

### Are these changes tested?

All existing tests in ./arrow and ./arrow/array packages are passing.

Users can now use the semantically correct RecordBatch type name.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,6 +31,10 @@ linters:
       - third_party$
       - builtin$
       - examples$
+    rules:
+      - linters:
+          - staticcheck
+        text: "SA1019"
 issues:
   fix: true
 formatters:


zeroshade (commented):
At a minimum we should update the examples to also use `RecordBatch` instead of `Record`. We can then incrementally update the rest of the library as we go

arrow/record.go
```diff
@@ -37,13 +37,18 @@ type Record interface {
...
 	//
 	// NewSlice panics if the slice is outside the valid range of the record array.
 	// NewSlice panics if j < i.
-	NewSlice(i, j int64) Record
+	NewSlice(i, j int64) RecordBatch
 }
+
+// Deprecated: Record is deprecated, use RecordBatch instead.
+// Record represents a single row, but this type actually represents a batch of rows.
+// This alias is provided for backwards compatibility.
```
zeroshade: the convention for Go would be for the `deprecated` notice to be at the bottom. i.e.

```suggestion
// Record as a term typically refers to a single row, but this type represents a batch of rows, known in Arrow parlance
// as a RecordBatch. This alias is provided for backwards compatibility.
//
// Deprecated: This is deprecated to avoid the confusion of the terminology where Record refers to a single row,
// use [RecordBatch] instead.
```

PR #476, opened 2025-08-16
Generate variant tests in GO language

aihuaxu:
### Rationale for this change
This is to generate the test files by reading the files from https://github.com/apache/parquet-testing/pull/91 and writing out the same files through GO implementation. The files are generated in https://github.com/apache/parquet-testing/pull/93. A few invalid cases are not generated from GO.

Then I tested out using Parquet-Java against those test files. 

- Overall the implementation is compatible.
- Issues:
1. Variant logical type should write as `VARIANT(1)` instead of `VARIANT(0)` since the variant spec version should be 1.
2. Type for time should `TIME(MICROS,false)` per spec, not `TIME(MICROS,true)`. 

```
aixu@K7YJWY4PK6 go_variant % parquet2 meta case-032.parquet

File path:  case-032.parquet
Created by: parquet-go version 18.4.0
Properties: (none)
Schema:
message schema {
  required int32 id (INTEGER(32,true)) = 1;
  optional group var (VARIANT(0)) = 2 {
    required binary metadata;
    optional binary value;
    optional int64 typed_value (TIME(MICROS,true));
  }
}
```

### DIFF
--- a/arrow/extensions/variant.go
+++ b/arrow/extensions/variant.go
@@ -18,6 +18,7 @@ package extensions
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math"
 	"reflect"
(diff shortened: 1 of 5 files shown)


zeroshade:
I had something for this locally that I just hadn't gotten to uploading yet. I'll finish it up and get it up by Monday

PR #477, opened 2025-08-18
Implement RLE dictionary decoder using generics

daniel-adam-tfs:
### Rationale for this change

Performance improvement of reading data from RLE-dictionary encoded parquet columns.

The DictionaryConverter interface was changed to DictionaryConverter[T] interface, so the methods could take []T as a parameter instead of interface{}. This required changed to several types implementing the interface.

The template code generation for RleDecoder was removed and the methods for the parquet types are now handled by the generic TypedRleDecoder type.

Additionally, some utils.Min/utils.Max calls were replace with calls to min/max intrinsics

### Are these changes tested?

Yes, tests verifying the RLE dictionary already existed and I've added BenchmarkReadInt32Column to measure the performance improvement.

Some exported types functions were changed:
DictionaryConverter -> DictionaryConverter[T]
RleDecoder -> split into RleDecoder and TypeRleDecoder[T]

func (b *BitReader) GetAligned(nbytes int, v interface{}) bool -> split into getAlignedUint8, getAlignedUint16, getAlignedUint32, getAlignedUint64

### DIFF
--- a/parquet/internal/utils/rle.go
+++ b/parquet/internal/utils/rle.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/internal/bitutils"
-	"github.com/apache/arrow-go/v18/internal/utils"
 	"golang.org/x/xerrors"
 )
 
@@ -37,7 +36,7 @@ const (
 func MinRLEBufferSize(bitWidth int) int {
 	maxLiteralRunSize := 1 + bitutil.BytesForBits(int64(MaxValuesPerLiteralRun*bitWidth))
 	maxRepeatedRunSize := binary.MaxVarintLen32 + bitutil.BytesForBits(int64(bitWidth))
-	return int(utils.Max(maxLiteralRunSize, maxRepeatedRunSize))
+	return int(max(maxLiteralRunSize, maxRepeatedRunSize))
 }
 
 func MaxRLEBufferSize(width, numValues int) int {
@@ -48,7 +47,7 @@ func MaxRLEBufferSize(width, numValues int) int {
 	minRepeatedRunSize := 1 + int(bitutil.BytesForBits(int64(width)))
 	repeatedMaxSize := int(bitutil.BytesForBits(int64(numValues))) * minRepeatedRunSize
 
-	return utils.Max(literalMaxSize, repeatedMaxSize)
+	return max(literalMaxSize, repeatedMaxSize)
 }
 
 // Utility classes to do run length encoding (RLE) for fixed bit width values.  If runs
@@ -181,11 +180,11 @@ func (r *RleDecoder) Discard(n int) int {
 		remain := n - read
 
 		if r.repCount > 0 {
-			repbatch := int(math.Min(float64(remain), float64(r.repCount)))
+			repbatch := min(remain, int(r.repCount))
 			r.repCount -= int32(repbatch)
 			read += repbatch
 		} else if r.litCount > 0 {
-			litbatch := int(math.Min(float64(remain), float64(r.litCount)))
+			litbatch := min(remain, int(r.litCount))
 			n, _ := r.r.Discard(uint(r.bitWidth), litbatch)
 			if n != litbatch {
 				return read
@@ -211,7 +210,7 @@ func (r *RleDecoder) GetBatch(values []uint64) int {
 		remain := size - read
 
 		if r.repCount > 0 {
-			repbatch := int(math.Min(float64(remain), float64(r.repCount)))
+			repbatch := min(remain, int(r.repCount))
 			for i := 0; i < repbatch; i++ {
 				out[i] = r.curVal
 			}
@@ -220,7 +219,7 @@ func (r *RleDecoder) GetBatch(values []uint64) int {
 			read += repbatch
 			out = out[repbatch:]
 		} else if r.litCount > 0 {
-			litbatch := int(math.Min(float64(remain), float64(r.litCount)))
+			litbatch := min(remain, int(r.litCount))
 			n, _ := r.r.GetBatch(uint(r.bitWidth), out[:litbatch])
 			if n != litbatch {
 				return read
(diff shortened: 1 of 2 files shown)


parquet/file/file_reader_test.go
```diff
@@ -927,3 +928,61 @@ func TestListColumns(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkReadInt32Column(b *testing.B) {
+	b.Skip("rle-dict-int32-snappy.parquet not available")
```
daniel-adam-tfs: @zeroshade I've generated this file with a single int32 column, 1 000 000 rows overall, 10 row groups and 100 unique values, v2 format and snappy compression. It's 113KB in size, if there's no other file you'd prefer to use for testing then we can add it to the data repo and remove the `b.Skip` call.

parquet/internal/utils/bit_reader.go
```diff
@@ -152,24 +169,67 @@ func (b *BitReader) GetAligned(nbytes int, v interface{}) bool {
 	if n != nbytes {
 		return false
 	}
+
 	// zero pad the bytes
-	memory.Set(b.raw[n:typBytes], 0)
-
-	switch v := v.(type) {
```
daniel-adam-tfs, line 158: We could make this generic:

func GetAligned[T ...](b *BitReader, v *T)

I didn't because I wanted to avoid casting to interface{} so I can do a type switch, but in my cpu profiles that switch was pretty fast. The slowest thing in the original code was the call to reflect package, but in the generic version unsafe.Sizeof(*v) can be called, which is much quicker.

parquet/file/file_reader_test.go
zeroshade: Would it be difficult to generate the file in `t.TempDir()` during the test, do the benchmark, then clean it up? That would be my preference, otherwise we can add a `testdata` folder and add it there.

parquet/file/file_reader_test.go
```diff
@@ -927,3 +928,61 @@ func TestListColumns(t *testing.T) {
...
+	defer reader.Close()
+
+	int32ColIdx := reader.MetaData().Schema.Root().FieldIndexByName("int32")
+	if int32ColIdx < 0 {
+		b.Fatalf("Expected to find int32 column in schema, got index %d", int32ColIdx)
+	}
+
+	numValues := reader.NumRows()
+	values := make([]int32, numValues)
+	b.StopTimer()
```
zeroshade: you don't need `StopTimer` before you call `ResetTimer`

parquet/file/file_reader_test.go
```diff
@@ -927,3 +928,61 @@ func TestListColumns(t *testing.T) {
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		startIndex := 0
+		for rg := 0; rg < reader.NumRowGroups(); rg++ {
+			b.StopTimer()
```
zeroshade: Swap this with the line above it

parquet/internal/utils/bit_reader.go
zeroshade: i think making it generic is the good call, using `unsafe.Sizeof` is the right way to do this

parquet/file/file_reader_test.go
daniel-adam-tfs: I definitely shouldn't have 3 calls to `StopTimer` in this benchmark, I think I forgot to remove the one on 961. 

The timer is in `On` state as this benchmark runs and `ResetTimer` doesn't stop it. I want to measure just the `ReadBatch` call, so I stop it to avoid including execution of lines 958-971 into the benchmark timing. Or do you think we should include it?

parquet/file/file_reader_test.go
daniel-adam-tfs: Hmm I've used a python to generate it, it's simple so it shouldn't be to hard to rewrite the code to Go and generate a file with the same parameters. 

I haven't used the writers available in this lib, is there a test or something that writes a RLE-encoded column so I check how to do it?

zeroshade: The default logic currently will enable dictionary encoding, but if you want to be explicit you can use the writer property `parquet.WithDictionaryFor("path.to.column", true)` or `parquet.WithDictionaryDefault(true)` to explicitly tell the writer to use dictionary encoding when writing the column.

parquet/file/file_reader_test.go
zeroshade: you would want to swap 956 and 957 (call `StopTimer` after `ResetTimer`). Then you can remove the one at 961 so you just start, do readbatch, and then stop.

daniel-adam-tfs: `ResetTimer` doesn't restart the timer, so what's the reason for the swap? The only difference I can think of is that the clock counters will be updated with data from the interval. Ie.

The current way:
1) timer is stopped
2) timer counters are zero-ed

The swapped way:
1) timer counters are zero-ed (but the timer keeps running)
2) timer is stopped (and the counters are updated with the data from interval between 1) and 2) )

Am I missing something?

parquet/internal/utils/bit_reader.go
daniel-adam-tfs: Annoying, I benchmarked it now and the non-generic version is about 2% faster according to the benchmark. In this case I'm alright with a little bit of code duplication for 2% better performance, but what's your call?

zeroshade: Agreed, a little duplication for the better performance is the way to go here

parquet/file/file_reader_test.go
zeroshade: Ah, that's my mistake. The current version looks good!

zeroshade (approved):

zeroshade:
LGTM are we all set here @daniel-adam-tfs ?

parquet/internal/utils/bit_reader.go
daniel-adam-tfs: Reverted back to the non-generic code.

PR #479, opened 2025-08-20
fix(parquet/internal/encoding): Fix typed dictionary encoding

MasslessParticle:
### Rationale for this change
The typed dict encoder tries to cast `memory.Allocator` to `TypedMemoTable`. This results in a panic. 

This PR changes `mem` to `memo` in the encoder so the `MemoTable` is the variable cast to `TypedMemoTable`.

### Are these changes tested?
I've tested this locally and panics no longer happen 🎉

### DIFF
--- a/parquet/internal/encoding/typed_encoder_test.go
+++ b/parquet/internal/encoding/typed_encoder_test.go
@@ -1,3 +1,19 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


zeroshade:
@MasslessParticle any luck on adding a test for this?

MasslessParticle:
@zeroshade Sorry, I got distracted last week. I've added a very basic test. I've only just started learning about the arrow ecosystem so it doesn't capture any other context than "we can call this without a panic or error." 

Let me know what you think.

parquet/internal/encoding/typed_encoder_test.go
```diff
@@ -0,0 +1,29 @@
+package encoding
```
zeroshade, line 17: this needs the Apache license header, then we should be good

MasslessParticle: done!

PR #485, opened 2025-08-27
feat(parquet): utilize memory allocator in `serializedPageReader`

joechenrh:
### Rationale for this change

Previously, only data page v2 in the reader will use the memory allocated from allocator. And seems there exists some misuage of the allocator.

Use memory allocator to allocate buffer for `serializedPageReader`, and introduce `Close` function to make sure all the memory will be freed after usage.

### Are these changes tested?

Track the memory usage in existing test to ensure all the allocated memory are freed.

For user with no explicitly set allocators, they don't need to change anything. Otherwise, they should remember to call `ColumnChunkReader.Close` after reading after chunk.

### DIFF
--- a/parquet/file/column_writer_test.go
+++ b/parquet/file/column_writer_test.go
@@ -298,22 +298,25 @@ func (p *PrimitiveWriterTestSuite) readColumnFully(compression compress.Compress
 
 // checkReadColumnByPage is used to check the data read from two interface is same
 func (p *PrimitiveWriterTestSuite) checkReadColumnByPage(compression compress.Compression) {
+	if p.descr.PhysicalType() != parquet.Types.FixedLenByteArray && p.descr.PhysicalType() != parquet.Types.ByteArray {
+		return
+	}
 	totalValues := int64(len(p.DefLevelsOut))
 	pagereader, _ := file.NewPageReader(arrutils.NewByteReader(p.readbuffer.Bytes()), totalValues, compression, mem, nil)
 	reader := file.NewColumnReader(p.descr, pagereader, mem, &p.bufferPool)
 
 	valuesRead := int64(0)
 	for valuesRead < totalValues {
-		read := p.ReadBatchInPage(reader, totalValues-valuesRead, valuesRead, p.DefLevelsOut[valuesRead:], p.RepLevelsOut[valuesRead:])
+		read := p.ReadBatchInPage(reader, totalValues-valuesRead, p.DefLevelsOut, p.RepLevelsOut)
 		switch reader.(type) {
 		case *file.ByteArrayColumnChunkReader:
 			batchOut := p.ValuesOut.([]parquet.ByteArray)
 			pageOut := p.ValuesOutPage.([]parquet.ByteArray)
-			p.Equal(batchOut[valuesRead:valuesRead+read], pageOut[valuesRead:valuesRead+read])
+			p.Equal(batchOut[valuesRead:valuesRead+read], pageOut[:read])
 		case *file.FixedLenByteArrayColumnChunkReader:
 			batchOut := p.ValuesOut.([]parquet.FixedLenByteArray)
 			pageOut := p.ValuesOutPage.([]parquet.FixedLenByteArray)
-			p.Equal(batchOut[valuesRead:valuesRead+read], pageOut[valuesRead:valuesRead+read])
+			p.Equal(batchOut[valuesRead:valuesRead+read], pageOut[:read])
 		}
 		valuesRead += read
 	}
--- a/parquet/internal/testutils/primitive_typed.go
+++ b/parquet/internal/testutils/primitive_typed.go
@@ -254,31 +254,31 @@ func (p *PrimitiveTypedTest) ReadBatch(reader file.ColumnChunkReader, batch, val
 	}
 }
 
-func (p *PrimitiveTypedTest) ReadBatchInPage(reader file.ColumnChunkReader, batch, valuesRead int64, defLevels, repLevels []int16) int64 {
+func (p *PrimitiveTypedTest) ReadBatchInPage(reader file.ColumnChunkReader, batch int64, defLevels, repLevels []int16) int64 {
 	switch r := reader.(type) {
 	case *file.Int32ColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]int32)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]int32), defLevels, repLevels)
 		return int64(read)
 	case *file.Int64ColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]int64)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]int64), defLevels, repLevels)
 		return int64(read)
 	case *file.Float32ColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]float32)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]float32), defLevels, repLevels)
 		return int64(read)
 	case *file.Float64ColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]float64)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]float64), defLevels, repLevels)
 		return int64(read)
 	case *file.Int96ColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.Int96)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.Int96), defLevels, repLevels)
 		return int64(read)
 	case *file.ByteArrayColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.ByteArray)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.ByteArray), defLevels, repLevels)
 		return int64(read)
 	case *file.BooleanColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]bool)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]bool), defLevels, repLevels)
 		return int64(read)
 	case *file.FixedLenByteArrayColumnChunkReader:
-		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.FixedLenByteArray)[valuesRead:], defLevels, repLevels)
+		_, read, _ := r.ReadBatchInPage(batch, p.ValuesOutPage.([]parquet.FixedLenByteArray), defLevels, repLevels)
 		return int64(read)
 	default:
 		panic("unimplemented")


parquet/file/column_reader.go
```diff
@@ -113,6 +113,8 @@ type ColumnChunkReader interface {
 	// automatically read the first page of the page reader passed in until
 	// HasNext which will read in the next page.
 	setPageReader(PageReader)
+	// Close releases the resources held by the column reader.
+	Close()
```
zeroshade: if we're adding a `Close()` method to the reader we should follow the convention and the signature should be `Close() error`

parquet/file/page_reader.go
```diff
@@ -53,6 +53,8 @@ type PageReader interface {
 	// Get the dictionary page for this column chunk
 	GetDictionaryPage() (*DictionaryPage, error)
 	SeekToPageWithRow(rowIdx int64) error
+	// Close releases the resources held by the reader.
+	Close()
```
zeroshade: same comment as above

parquet/file/page_reader.go
```diff
@@ -441,7 +457,6 @@ func (p *serializedPageReader) Reset(r parquet.BufferedReader, nrows int64, comp
 	if p.err != nil {
 		return
 	}
-	p.decompressBuffer.Reset()
```
zeroshade, line 444: why remove this?

parquet/file/page_reader.go
```diff
@@ -574,7 +588,7 @@ func (p *serializedPageReader) GetDictionaryPage() (*DictionaryPage, error) {
 
 		return &DictionaryPage{
 			page: page{
-				buf:      memory.NewBufferBytes(data),
+				buf:      buf,
```
zeroshade, line 592: this can become a problem due to race conditions. If a new DictionaryPage is read in by the reader while this page is still in use, you'll end up overwriting data that is still being used. That's why we were creating a new buffer here, to avoid the potential race condition.

parquet/file/page_reader.go
```diff
@@ -728,7 +740,7 @@ func (p *serializedPageReader) Next() bool {
 			// make dictionary page
 			p.curPage = &DictionaryPage{
 				page: page{
-					buf:      memory.NewBufferBytes(data),
+					buf:      buf,
```
zeroshade, line 744: same comment as above. we should utilize a buffer pool instead most likely

parquet/file/page_reader.go
```diff
@@ -759,7 +774,7 @@ func (p *serializedPageReader) Next() bool {
 			// make datapagev1
 			p.curPage = &DataPageV1{
 				page: page{
-					buf:      memory.NewBufferBytes(data),
+					buf:      buf,
```
zeroshade, line 778: same issue as above

parquet/file/page_reader.go
joechenrh: Because this PR change it to `memory.Buffer`, it's redundant to call `Reset`

parquet/file/page_reader.go
joechenrh: I understand your meaning, but I think there should be no concurrency, or, data race here.
- First, dictionary page is only read once for each column chunk reader, either reading it at first, or reading it after we skip some rows and find it's not initialized. 
- Second, the page read by `serializedPageReader` is stored in the `columnChunkReader`, so there should be only one page in use at any time. And user shouldn't access one chunk reader in parallel, which I think is an incorrect usage.

https://github.com/apache/arrow-go/blob/321159468711a8fac9a6c52a956424a390274da2/parquet/file/column_reader.go#L326-L345

parquet/file/column_reader_types.gen.go.tmpl
```diff
@@ -45,8 +46,13 @@ func (cr *{{.Name}}ColumnChunkReader) Skip(nvalues int64) (int64, error) {
...
-    return cr.curDecoder.(encoding.{{.Name}}Decoder).Decode(values[start:start+len])
-  })
+  return cr.readBatch(batchSize, defLvls, repLvls,func(start, len int64) (int, error) {
+    n, err := cr.curDecoder.(encoding.{{.Name}}Decoder).Decode(values[start:start+len])
+    if err == nil && cr.postProcess != nil {
+      cr.postProcess(values[start:start+len])
+    }
+    return n, err
+  }, 
+  )
```
joechenrh, line 56: I found another problem with the current `serializedPageReader`, either using memory allocator or buffer pool.

For example, suppose each data page contains 4 `ByteArray`, and we want to read 16 data. We first read the data in 1st and 2nd page, store them in `values[0:8]`. When reading the 3rd page, if the pool/allocator give us a previously used buffer, the value in `values[0:8]` might be overwritten (depends on the decoder we use).

So I add the logic here to make a copy of the buffer. But I don't think it's the best way to solve this. Could you have some suggestions? @zeroshade

parquet/file/column_reader.go
```diff
@@ -437,16 +446,27 @@ func (c *columnChunkReader) initDataDecoder(page Page, lvlByteLen int64) error {
...
+	}
+
+	switch c.descr.PhysicalType() {
+	case parquet.Types.FixedLenByteArray:
+		c.curDecoder = &encoding.FixedLenByteArrayDecoderWrapper{
+			FixedLenByteArrayDecoder: c.curDecoder.(encoding.FixedLenByteArrayDecoder),
+		}
+	case parquet.Types.ByteArray:
+		c.curDecoder = &encoding.ByteArrayDecoderWrapper{
+			ByteArrayDecoder: c.curDecoder.(encoding.ByteArrayDecoder),
```
joechenrh: I found another problem with the current `serializedPageReader`, either using memory allocator or buffer pool.

For example, suppose each data page contains 4 `ByteArray`, and we want to read 12 data. We first read the data in 1st and 2nd page, store them in `values`. And these values may point to the internal buffer in the reader (depends on the decoder we use). If the pool/allocator give us a previously used buffer when reading the 3rd page, old values might be overwritten.

So I add the wrapper here to make a copy of the output. Do you think it's acceptable? @zeroshade

zeroshade: The pool/allocator shouldn't be able to return a previously used buffer until it has been put back into the pool, which shouldn't happen until we release it. Did you find a situation where this was happening?

joechenrh: I didn't meet this situation, but I think it could happen. Just like the above example:

- Get buffer1 -> Read data from page 1 -> Release buffer1
- Get buffer2 -> Read data from page 2 -> Release buffer2
- Get buffer 1 again

Since the `values` might directly point to these buffers, they could be modified when reading subsequent pages

https://github.com/apache/arrow-go/blob/c6ce2ef4e55009a786cf04b3845eba5170c98066/parquet/file/column_reader_types.gen.go#L216-L220

zeroshade: I think that's also part of why we were creating new buffers and copying the data rather than being able to reuse the buffers in the page reader (the thing you're changing in this PR)

zeroshade: I think it would be good if you could craft a test case that validates this issue so that we can prove to ourselves that the updates you're making here prevent it from happening.

parquet/internal/encoding/typed_encoder.go
```diff
@@ -556,6 +572,36 @@ func (enc *DictByteArrayEncoder) Type() parquet.Type {
...
+	ByteArrayDecoder
+}
+
+func (d *ByteArrayDecoderWrapper) Decode(out []parquet.ByteArray) (int, error) {
+	n, err := d.ByteArrayDecoder.Decode(out)
+	if err != nil {
+		return n, err
+	}
+	cloneByteArray(out[:n])
+	return n, nil
```
zeroshade: it would be better for the cloning to be done by the ByteArrayDecoder itself inside of the `Decode` method rather than needing this wrapper.

parquet/file/column_reader.go
joechenrh: > I think that's also part of why we were creating new buffers and copying the data rather than being able to reuse the buffers in the page reader (the thing you're changing in this PR)

But `memory.NewBufferBytes` doesn't copy the data, just ref it. That's the reason I open this PR.
https://github.com/apache/arrow-go/blob/c6ce2ef4e55009a786cf04b3845eba5170c98066/arrow/memory/buffer.go#L50-L53

zeroshade: Yes, but remember that the code you're changing called `	buf := memory.NewResizableBuffer(p.mem)` which does allocate a new buffer that the data is copied into before the changes you're making. The changes in this PR are introducing buffer re-use that wasn't there before. 

I'm not saying it's a bad thing, just pointing out part of the reason for the previous behavior that allocated new buffers was to prevent the problem you've identified. Now that we're reusing these buffers in this change, we have to do the copying you're proposing.

Thats all

joechenrh:
```diff
From eec53776b4aba9af1011a5a419472768c7c4dd0e Mon Sep 17 00:00:00 2001
From: Ruihao Chen <joechenrh@gmail.com>
Date: Fri, 29 Aug 2025 12:41:17 -0400
Subject: [PATCH] Add simple test case

Signed-off-by: Ruihao Chen <joechenrh@gmail.com>
---
 parquet/file/column_writer_test.go | 36 ++++++++++++++++++++++++++++++
 1 file changed, 36 insertions(+)

diff --git a/parquet/file/column_writer_test.go b/parquet/file/column_writer_test.go
index 90b239e4..3670de55 100644
--- a/parquet/file/column_writer_test.go
+++ b/parquet/file/column_writer_test.go
@@ -56,6 +56,41 @@ const (
 	DictionaryPageSize = 1024 * 1024
 )
 
+type simpleAllocator struct {
+	q chan []byte
+}
+
+func (a *simpleAllocator) Allocate(n int) []byte {
+	l := len(a.q)
+	for range l {
+		b := <-a.q
+		if cap(b) >= n {
+			return b[:n]
+		}
+		a.q <- b
+	}
+	return make([]byte, n)
+}
+
+func (a *simpleAllocator) Free(b []byte) {
+	if b == nil {
+		return
+	}
+	select {
+	case a.q <- b:
+	default: // discard if full
+	}
+}
+
+func (a *simpleAllocator) Reallocate(n int, old []byte) []byte {
+	a.Free(old)
+	return a.Allocate(n)
+}
+
+func newSimpleAllocator() *simpleAllocator {
+	return &simpleAllocator{q: make(chan []byte, 64)}
+}
+
 type mockpagewriter struct {
 	mock.Mock
 }
@@ -258,6 +293,7 @@ func (p *PrimitiveWriterTestSuite) TearDownTest() {
 }
 
 func (p *PrimitiveWriterTestSuite) buildReader(nrows int64, compression compress.Compression) file.ColumnChunkReader {
+	mem := newSimpleAllocator()
 	p.readbuffer = p.sink.Finish()
 	pagereader, _ := file.NewPageReader(arrutils.NewByteReader(p.readbuffer.Bytes()), nrows, compression, mem, nil)
 	return file.NewColumnReader(p.descr, pagereader, mem, &p.bufferPool)
-- 
2.47.3


```

I've crafted a customized allocator, which could make the UT in column_writer_test on main branch fail.

parquet/internal/encoding/typed_encoder.go
joechenrh: Because there are several different encoder for `ByteArray` type, comparing to modify every implementation, I prefer add a wrapper for this work. Besides, by only applying the wrapper to page reader (which the target of this PR), we can ensure that other part of the code won't be affected by this additional overhead.

joechenrh: After inspect the code again, I guess the problem arises from `ColumnChunkReader.ReadBatch` function itself, which is used in test to read data.

For other place that read data from column chunk, like `recordReaderImpl`, seems it has special logic for `ByteArray` and `FixedLenByteArray`. That is, copy the value from the decoder to a new, separate buffer. Like the below code, `bldr` itself contains a buffer allocated from allocator.

https://github.com/apache/arrow-go/blob/c6ce2ef4e55009a786cf04b3845eba5170c98066/parquet/file/record_reader.go#L842-L846

Since it's used in test, let me revert the change and try to rewrite this function separately.

parquet/file/column_reader.go
```diff
@@ -38,6 +38,22 @@ const (
 	defaultPageHeaderSize = 16 * 1024
 )
 
+func CloneByteArray[T parquet.ByteArray | parquet.FixedLenByteArray](src []T) {
```
zeroshade: We shouldn't export this.

Also the constraint should probably be `~[]byte`, yea?

zeroshade (changes requested):
Just one nitpick change, but otherwise looks great to me. thanks for all this!

zeroshade:
An additional test case would be fantastic! Thanks!

zeroshade:
@joechenrh do you want me to hold this open for you to add a test? Or would you want to add the test as a separate PR given this is already a fairly large one?

PR #489, opened 2025-09-03
fix(parquet/pqarrow): Fix null_count column stats

MasslessParticle:
### Rationale for this change
When dictionary encoding is enabled and and repetitions are set to `required`, the `null_count` statistic is negative because `defLevels` is always 0.

This PR sets the null count to 0 if `defLevels - nonNullCount < 0`

### Are these changes tested?
Yes. I've added a new test that exposes another bug. I started with the `fullTypeList` variable like [this test](https://github.com/apache/arrow-go/blob/c6ce2ef4e55009a786cf04b3845eba5170c98066/parquet/pqarrow/encode_dictionary_test.go#L43) and discovered that not all types are supported. 

If an unsupported type is encoded as a dictionary, it results in a nasty panic in the typed dictionary encoder because the types don't line up.

Yes, the stats written to parquet files are currently wrong for these sorts of columns. This PR should fix that!

### DIFF
--- a/parquet/pqarrow/encode_arrow_test.go
+++ b/parquet/pqarrow/encode_arrow_test.go
@@ -1284,6 +1284,30 @@ func (ps *ParquetIOTestSuite) writeColumn(mem memory.Allocator, sc *schema.Group
 	return buf.Bytes()
 }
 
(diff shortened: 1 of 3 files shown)


MasslessParticle:
I've added a new issue for the encoding panic I mentioned [here](https://github.com/apache/arrow-go/issues/490)

PR #492, opened 2025-09-05
docs: fix redirect link

### LINKED ISSUE #491: The link on the main repository page is broken

On https://github.com/apache/arrow-go, the main link is https://arrow.apache.org/go/ and redirects to https://pkg.go.dev/github.com/apache/arrow-go

Image

It looks broken:

Image

### Component(s)

Documentation

zeroshade:
fixes #491

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-Redirect permanent /go/ https://pkg.go.dev/github.com/apache/arrow-go/
+Redirect permanent /go/ https://pkg.go.dev/github.com/apache/arrow-go/v18/


amoeba (approved):

ianmcook:
> -1 on fixing it this way. Other packages on pkg.go.dev don't need the version number in the URL to find the package page. Can you find out why the Arrow package needs it?

Upon further investigation: It seems like pkg.go.dev doesn't provide any better way to do this :(

kou (approved):
+1

How about updating the `v18` part automatically in release process?

apache/arrow-dotnet will do similar thing https://github.com/apache/arrow-dotnet/pull/33 . So we can do it in apache/arrow-go too.

zeroshade:
I'll look into automating the update for that. For now we'll merge this so that we can get the redirect fixed

kou:
If you need, I can help the automation. :-)

PR #497, opened 2025-09-08
chore: Automate GitHub Releases creation and site redirect

### LINKED ISSUE #496: Automate GitHub Releases creation and https://arrow.apache.org/go/ redirect

Currently, we create a GitHub Release manually after we run `dev/release/release.sh` (right?) .

We also need to update https://github.com/apache/arrow-go/blob/asf-site/.htaccess when we release a new major version.

We should automate them to reduce maintenance cost.

### Component(s)

Release

kou:
### Rationale for this change

Fixes #496

We should automate release process as much as possible to reduce maintenance cost.

* Add a workflow that is used on release tag such as `v18.4.1`
* It creates a GitHub Release for the release automatically
  * Asserts are copied from the GitHub Released of the passed RC
  * It also creates a discussion for the release
* It updates the asf-site branch

### Are these changes tested?

No...

Yes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,8 +19,8 @@ name: Release
 on:
   push:
     tags:
-      - "*"
-      - "!*-rc*"
+      - "v*"
+      - "!v*-rc*"
 permissions:
   contents: write
 env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,101 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Release
+on:
+  push:
+    tags:
+      - "*"
```
zeroshade: should probably be `v*`

kou: I think that we don't use tags for no release use cases but let's use strict pattern.

Issue #493, opened 2025-09-06
equal kernel giving incorrect results


ostafen-m2d:
### Describe the bug, including details regarding any error messages, version, and platform.

Hi, I'm experimenting a weird behavior for the equal kernel, when using it to compare boolean values.

```golang
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/apache/arrow-go/v18/arrow/array"
	"github.com/apache/arrow-go/v18/arrow/compute"
	"github.com/apache/arrow-go/v18/arrow/memory"
)

func main() {
	ctx := context.Background()
	pool := memory.NewGoAllocator()
	n := 5 // number of elements in the array

	// --- Build first Boolean array filled with false ---
	bldr1 := array.NewBooleanBuilder(pool)
	defer bldr1.Release()
	for i := 0; i < n; i++ {
		bldr1.Append(false)
	}
	arr1 := bldr1.NewBooleanArray()
	defer arr1.Release()

	// --- Build second Boolean array filled with false ---
	bldr2 := array.NewBooleanBuilder(pool)
	defer bldr2.Release()
	for i := 0; i < n; i++ {
		bldr2.Append(false)
	}
	arr2 := bldr2.NewBooleanArray()
	defer arr2.Release()

	// Wrap arrays as compute.Datum
	datum1 := compute.NewDatum(arr1)
	datum2 := compute.NewDatum(arr2)

	// --- Call equal kernel ---
	resDatum, err := compute.CallFunction(ctx, "equal", nil, datum1, datum2)
	if err != nil {
		log.Fatal(err)
	}
	defer resDatum.Release()

	// Convert result Datum to BooleanArray
	resArr := resDatum.(*compute.ArrayDatum).MakeArray().(*array.Boolean)
	defer resArr.Release()

	// Print element-wise comparison
	for i := 0; i < int(resArr.Len()); i++ {
		fmt.Printf("arr1[%d] == arr2[%d] -> %v\n", i, i, resArr.Value(i))
	}
}
```

I would expect this snippet to print true for all result values, but it print false.

Is this expected behavior? Am I missing something?

### Component(s)

Other

PR #505, opened 2025-09-14
use xnor for boolean equals function

Dhruvit96:
### Rationale for this change
fixes https://github.com/apache/arrow-go/issues/493

Add bitxnor operation, utilize bitxnor for boolean equality.

### Are these changes tested?
Yes

Boolean equality operations now return correct results. No changes to the public API.

### DIFF
--- a/arrow/compute/internal/kernels/scalar_comparisons.go
+++ b/arrow/compute/internal/kernels/scalar_comparisons.go
@@ -36,9 +36,11 @@ import (
 
 type binaryKernel func(left, right, out []byte, offset int)
 
-type cmpFn[LeftT, RightT arrow.FixedWidthType] func([]LeftT, []RightT, []uint32)
-type cmpScalarLeft[LeftT, RightT arrow.FixedWidthType] func(LeftT, []RightT, []uint32)
-type cmpScalarRight[LeftT, RightT arrow.FixedWidthType] func([]LeftT, RightT, []uint32)
+type (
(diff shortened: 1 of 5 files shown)


arrow/compute/internal/kernels/scalar_comparisons.go
```diff
@@ -589,7 +591,7 @@ func compareTimestampKernel(ty exec.InputType, op CompareOperator) exec.ScalarKe
 var (
 	boolEQ = binaryBoolOps{
 		arrArr: func(_ *exec.KernelCtx, lhs, rhs, out bitutil.Bitmap) error {
-			bitutil.BitmapAnd(lhs.Data, rhs.Data, lhs.Offset, rhs.Offset, out.Data, out.Offset, out.Len)
+			bitutil.BitmapXnor(lhs.Data, rhs.Data, lhs.Offset, rhs.Offset, out.Data, out.Offset, out.Len)
```
zeroshade, line 594: Can you add a test to `scalar_comparisons_test.go` which tests Boolean scalar equal? I'm assuming we must not have a test in there for this, otherwise it would have been caught sooner.

zeroshade (changes requested):
Looks good to me, just one request!

arrow/compute/internal/kernels/scalar_comparisons.go
Dhruvit96: Added unit test for checking boolean equals and not equals functions.

zeroshade (approved):

PR #510, opened 2025-09-16
feat: Add support for specifying `Timestamp` `valueStr` output format

erezrokah:
### Rationale for this change

See discussion in https://github.com/apache/arrow-go/pull/450

Allow setting the formatting of timestamps when using `ValueStr` (or JSON marshalling)

### Are these changes tested?

Yes - re-added the original test that was modified in https://github.com/apache/arrow-go/pull/450

Yes - users will now be able to revert back to the old format before https://github.com/apache/arrow-go/pull/450, to avoid unintentional breakage

### DIFF
--- a/arrow/array/timestamp.go
+++ b/arrow/array/timestamp.go
@@ -34,11 +34,16 @@ import (
 type Timestamp struct {
 	array
 	values []arrow.Timestamp
+	format string
 }
 
 // NewTimestampData creates a new Timestamp from Data.
 func NewTimestampData(data arrow.ArrayData) *Timestamp {
-	a := &Timestamp{}
+	return NewTimestampDataWithFormat(data, time.RFC3339Nano)
+}
+
+func NewTimestampDataWithFormat(data arrow.ArrayData, format string) *Timestamp {
+	a := &Timestamp{format: format}
 	a.refCount.Add(1)
 	a.setData(data.(*Data))
 	return a
@@ -93,7 +98,11 @@ func (a *Timestamp) ValueStr(i int) string {
 	}
 
 	toTime, _ := a.DataType().(*arrow.TimestampType).GetToTimeFunc()
-	return toTime(a.values[i]).Format(time.RFC3339Nano)
+	format := a.format
+	if format == "" {
+		format = time.RFC3339Nano
+	}
+	return toTime(a.values[i]).Format(format)
 }
 
 func (a *Timestamp) GetOneForMarshal(i int) interface{} {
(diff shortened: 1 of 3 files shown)


arrow/avro/testdata/testdata.go
```diff
@@ -51,7 +51,6 @@ type TimestampMicros int64
 
 func (t TimestampMicros) MarshalJSON() ([]byte, error) {
 	ts := time.Unix(0, int64(t)*int64(time.Microsecond)).UTC().Format(time.RFC3339Nano)
-	// arrow record marshaller trims trailing zero digits from timestamp so we do the same
```
erezrokah, line 54: This comment seems no longer relevant

arrow/array/timestamp.go
```diff
@@ -34,11 +34,16 @@ import (
...
+	format string
 }
 
 // NewTimestampData creates a new Timestamp from Data.
 func NewTimestampData(data arrow.ArrayData) *Timestamp {
-	a := &Timestamp{}
+	return NewTimestampDataWithFormat(data, time.RFC3339Nano)
+}
+
+func NewTimestampDataWithFormat(data arrow.ArrayData, format string) *Timestamp {
```
efd6:
```suggestion
func NewTimestampDataWithLayout(data arrow.ArrayData, layout string) *Timestamp {
```
reflecting the terminology used in stdlib/time (`s/format/layout/g` throughout)

Also probably want godoc for this.

efd6 (commented):
this seems reasonable

arrow/array/timestamp.go
```diff
@@ -34,11 +34,18 @@ import (
...
 
 // NewTimestampData creates a new Timestamp from Data.
 func NewTimestampData(data arrow.ArrayData) *Timestamp {
-	a := &Timestamp{}
+	return NewTimestampDataWithLayout(data, time.RFC3339Nano)
+}
+
+// NewTimestampDataWithLayout creates a new Timestamp from Data with a custom string value layout.
+// This is useful for cases where consumers expect a non standard layout
+func NewTimestampDataWithLayout(data arrow.ArrayData, layout string) *Timestamp {
```
zeroshade: I'd rather slightly different naming to indicate we're talking about the layout *when converting to a string*

Also, we should indicate in the comment that the layout refers to the layout from `[time.Time.Format]`

arrow/array/timestamp.go
```diff
@@ -130,10 +141,15 @@ type TimestampBuilder struct {
...
 	rawData []arrow.Timestamp
+	layout  string
 }
 
 func NewTimestampBuilder(mem memory.Allocator, dtype *arrow.TimestampType) *TimestampBuilder {
-	tb := &TimestampBuilder{builder: builder{mem: mem}, dtype: dtype}
+	return NewTimestampBuilderWithLayout(mem, dtype, time.RFC3339Nano)
+}
+
+func NewTimestampBuilderWithLayout(mem memory.Allocator, dtype *arrow.TimestampType, layout string) *TimestampBuilder {
```
zeroshade: same as above, not a big fan of this naming and we should indicate in a comment that the layout is referring to the `[time.Time.Format]` layout

arrow/array/timestamp.go
erezrokah: How about `NewTimestampDataWithValueStrLayout`? Happy to change it if this is not what you have in mind

zeroshade (changes requested):
Looks good to me, just one nitpick: can we add a JSON test which verifies the JSON marshalling output respects the layout

erezrokah:
Thank you @zeroshade and @efd6 for the swift review process, appreciate it

Issue #28, opened 2024-08-07
[Go] Fatal Error in pqarrow.writeDenseArrow : invalid pointer


Niivii:
Got a runtine error for this code in Python with polars and adbc engine to Snowflake:
```python
with adbc_driver_snowflake.dbapi.connect(db_kwargs={
    "username": user,
    "adbc.snowflake.sql.account": host,
    "adbc.snowflake.sql.auth_type": "auth_ext_browser",
    "adbc.snowflake.sql.warehouse": warehouse,
    "adbc.snowflake.sql.role": role,
    "adbc.snowflake.sql.db": database,
    "adbc.snowflake.sql.schema": db_schema,
}) as connection:
    self.write_database(
        table_name=table,
        connection=connection,
        if_table_exists="replace",
        engine="adbc",
    )
```

Stacktrace:
```
runtime: bad pointer in frame github.com/apache/arrow/go/v17/parquet/pqarrow.writeDenseArrow at 0xc00100d5e0: 0x1
fatal error: invalid pointer found on stack

runtime stack:
runtime.throw({0x7fa6b183582f?, 0x7fa6b2daf908?})
	/usr/local/go/src/runtime/panic.go:1077 +0x5e fp=0x7fa6af8ea590 sp=0x7fa6af8ea560 pc=0x7fa6afdba69e
runtime.adjustpointers(0x7fa6af8ea808?, 0x7fa6af8ea650, 0x7fa6afdddd65?, {0x7fa6af8ea808?, 0x0?})
	/usr/local/go/src/runtime/stack.go:627 +0x1ad fp=0x7fa6af8ea5f0 sp=0x7fa6af8ea590 pc=0x7fa6afdd306d
runtime.adjustframe(0x7fa6af8ea808, 0x7fa6af8ea6e8)
	/usr/local/go/src/runtime/stack.go:684 +0xdb fp=0x7fa6af8ea680 sp=0x7fa6af8ea5f0 pc=0x7fa6afdd319b
runtime.copystack(0xc0005da680, 0x7fa6af8eab20?)
	/usr/local/go/src/runtime/stack.go:935 +0x2c5 fp=0x7fa6af8ea978 sp=0x7fa6af8ea680 pc=0x7fa6afdd3945
runtime.shrinkstack(0xc0005da680)
	/usr/local/go/src/runtime/stack.go:1218 +0x113 fp=0x7fa6af8ea998 sp=0x7fa6af8ea978 pc=0x7fa6afdd4833
[...]

PR #513, opened 2025-09-22
fix(arrow/cdata): Avoid calling unsafe.Slice on zero-length pointers

orlp:
### Rationale for this change

Slices from FFI may have an arbitrary pointer when the length is zero, but this is not allowed in Go, where the pointer must always be valid. I believe this fixes https://github.com/apache/arrow-go/issues/28.

I changed the instances of `unsafe.Slice` in `arrow/cdata` I could find to be robust when used with length zero.

### Are these changes tested?

No, I don't have a Go setup at all.

No.

### DIFF
--- a/arrow/cdata/cdata_fulltest.c
+++ b/arrow/cdata/cdata_fulltest.c
@@ -384,8 +384,8 @@ void export_str_array_with_nulls(int64_t nitems, struct ArrowArray* out) {
     int64_t bitmap_nbytes = (nitems + 7) / 8;
     out->buffers[0] = malloc(bitmap_nbytes);
     memset((void*)out->buffers[0], 0, bitmap_nbytes);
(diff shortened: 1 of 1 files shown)


zeroshade:
thanks for putting this together @orlp. You beat me to putting up the change :smile: The test was able to confirm that the `0x1` got through before the changes, and afterwards it is a proper Go empty slice.

zeroshade:
That is technically correct, though the checks don't look in C/C++ memory. Essentially, as long as we don't attempt to store `ArrowArray.buffers[2]` into a Go variable when it's not a valid pointer (which I believe we successfully avoid now) we should be safe because you only find the bad pointer by first dereferencing the void** in `ArrowArray.buffers`. 

If we do manage to hit this still, then we can look into adding a check in the trampoline function we have

PR #517, opened 2025-09-23
fix(arrow/array): fix panic in dictionary builders

### LINKED ISSUE #516: dictionaryBuilder: panic on mismatch between arrow.Duration and int64

Go 1.24.2 on linux/amd64

github.com/apache/arrow-go/v18 v18.4.1

The following program panics on a mismatch in `dictionaryBuilder` between `arrow.Duration` and `int64`:

```
package main

import (
        "fmt"

)

func main() {
        dictType := &arrow.DictionaryType{IndexType: &arrow.Int8Type{}, ValueType: arrow.FixedWidthTypes.Duration_ns}
        bldr := array.NewDictionaryBuilder(memory.DefaultAllocator, dictType)
	defer bldr.Release()

	builder := bldr.(*array.DurationDictionaryBuilder)
	if err := builder.Append(arrow.Duration(1)); err == nil {
		fmt.Println("success")
	} else {
                fmt.Println(err)
	}
}
```

Output:

```
panic: interface conversion: interface {} is arrow.Duration, not int64

goroutine 1 [running]:
github.com/apache/arrow-go/v18/internal/hashing.(*Table[...]).GetOrInsert(0x469039?, {0x6622a0?, 0x73dd08?})
	/home/.../go/pkg/mod/github.com/apache/arrow-go/v18@v18.4.1/internal/hashing/xxh3_memo_table_types.go:255 +0x49
github.com/apache/arrow-go/v18/arrow/array.(*dictionaryBuilder).appendValue(0xc0000b40e0, {0x6622a0?, 0x73dd08?})
	/home/.../go/pkg/mod/github.com/apache/arrow-go/v18@v18.4.1/arrow/array/dictionary.go:785 +0x2b
github.com/apache/arrow-go/v18/arrow/array.(*dictBuilder[...]).Append(...)
	/home/.../go/pkg/mod/github.com/apache/arrow-go/v18@v18.4.1/arrow/array/dictionary.go:994
main.main()
	/home/.../go/arrow-bug/main.go:17 +0x118
exit status 2
```

I am not particularly familiar with arrow-go but I believe the calling code is valid, and that the construction of the internal MemoTable may be losing the distinction between arrow.Duration and int64. The same problem may also exist for a handful of other types but I didn't investigate closely.

(Please let me know if the calling code _isn't_ valid. The program above is extracted from a complicated open-telemetry deployment that is panicking when trying to use arrow on the wire; if something about the calling pattern above is incorrect then I will have to go looking elsewhere for the problem.)

### Component(s)

Other

zeroshade:
### Rationale for this change
The underlying hash table is in terms of primitive types (int64/int32/int16, etc.) not the top-level types (arrow.Duration, arrow.Timestamp, etc.) which results in a panic mismatch when appending values. 

fixes #516 

Checking and validating the type when appendValue is called to ensure correct type coercion.

### Are these changes tested?
A unit test is added in this change.

A case that previously caused a panic will now be successful.

### DIFF
--- a/arrow/array/dictionary.go
+++ b/arrow/array/dictionary.go
@@ -991,6 +991,22 @@ type dictBuilder[T arrow.ValueType] struct {
 }
 
 func (b *dictBuilder[T]) Append(v T) error {
+	switch val := any(v).(type) {
+	case arrow.Duration:
+		return b.appendValue(int64(val))
+	case arrow.Timestamp:
+		return b.appendValue(int64(val))
+	case arrow.Time32:
+		return b.appendValue(int32(val))
+	case arrow.Time64:
+		return b.appendValue(int64(val))
+	case arrow.Date32:
+		return b.appendValue(int32(val))
+	case arrow.Date64:
+		return b.appendValue(int64(val))
+	case arrow.MonthInterval:
+		return b.appendValue(int32(val))
+	}
 	return b.appendValue(v)
 }
 
--- a/arrow/array/dictionary_test.go
+++ b/arrow/array/dictionary_test.go
@@ -1899,6 +1899,20 @@ func TestBinaryDictionaryPanic(t *testing.T) {
 	assert.True(t, allocator.paniced)
 }
 
+func TestDictionaryBuilderDuration(t *testing.T) {
+	// verify fix for https://github.com/apache/arrow-go/issues/516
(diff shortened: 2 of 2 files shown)


amoeba (approved):
+1

It looks like the switch handles all the types it needs as far as I can tell. CI failure looks unrelated.

PR #518, opened 2025-09-23
fix(arrow/compute): fix scalar comparison panic

### LINKED ISSUE #503: Followup on issue https://github.com/apache/arrow-go/issues/464

Hi @zeroshade , I just notice here https://github.com/apache/arrow-go/blob/d6e0850d1ea239cbe4e678f600ce2f8ba1823f57/arrow/compute/internal/kernels/scalar_comparisons.go#L111 can end in the same panic as issue https://github.com/apache/arrow-go/issues/464. Could you help to come up with another fix? Thanks!

What I don't understand is when arrow uses `comparePrimitiveScalarArray` or `comparePrimitiveArrayScalar` internally?
I didn't manage to create a test case that can reproduce this issue permenantly. It looks like the original test case sometimes can reproduce it and sometimes not.

### Component(s)

Other

zeroshade:
### Rationale for this change
fixes #503 

Properly limit the slice when calling `op` in `comparePrimitiveArrayScalar` like we do for `comparePrimitiveScalarArray` as per #464 

### Are these changes tested?
Yes, the unit test is updated to account for this situation

scenario will no longer panic

### DIFF
--- a/arrow/compute/exprs/exec_test.go
+++ b/arrow/compute/exprs/exec_test.go
@@ -743,7 +743,7 @@ func TestDecimalFilterLarge(t *testing.T) {
 				db.Append(d)
 			}
 
-			rec := array.NewRecord(schema, []arrow.Array{db.NewArray()}, int64(tc.n))
+			rec := array.NewRecordBatch(schema, []arrow.Array{db.NewArray()}, int64(tc.n))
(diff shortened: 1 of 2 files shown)


kou (approved):
+1

PR #520, opened 2025-09-24
fix(parquet/pqarrow): unsupported dictionary types in pqarrow

### LINKED ISSUE #490: [Go][Parquet] unsupported dictionary encoding types result in panic

When a column is dictionary encoded but the type doesn't exist as a typed dictionary encoder, the columns writer panics.

To reproduce: run [this test](https://github.com/apache/arrow-go/blob/3c01efba60abe1421a49ac21c95858dbfccd78d6/parquet/pqarrow/encode_dictionary_test.go#L70-L111) with `arrow.PrimitiveTypes.Uint8` and observe a panic resulting from [this](https://github.com/apache/arrow-go/blob/3c01efba60abe1421a49ac21c95858dbfccd78d6/parquet/internal/encoding/typed_encoder.go#L165) type assertion. The underlying cause, in this specific case, is because the underlying type is `uint8` but the encoder is `int32`

### Component(s)

Parquet

zeroshade:
### Rationale for this change
fixes #490 

When using `PutDictionary` to write an arrow dictionary array to Parquet directly, we will now normalize the data type to a parquet physical storage type to avoid type problems. e.g. A dictionary with `int8` values will get cast to `int32` values before we process it so that all of the remaining encoding code can continue to be based on the physical storage type. This alleviates panics when using `pqarrow` to write Dictionary arrays.

### Are these changes tested?
Yes, the unit tests are updated to include tests for int8/uint8/etc. as dictionary value types.

No, just fixing panics.

### DIFF
--- a/parquet/internal/encoding/types.go
+++ b/parquet/internal/encoding/types.go
@@ -117,9 +117,7 @@ type DictEncoder interface {
 	// native type. e.g. a dictionary of type int8 will be cast to an int32
 	// dictionary for parquet storage.
 	//
-	// The returned array must be released by the caller if different from
-	// the input array. If no normalization is needed, the input array
-	// may be returned as-is.
+	// The returned array must always be released by the caller.
 	NormalizeDict(arrow.Array) (arrow.Array, error)
 }
 
--- a/parquet/internal/encoding/types.go
+++ b/parquet/internal/encoding/types.go
@@ -113,6 +113,14 @@ type DictEncoder interface {
 	// of [0,dictSize) and is not validated. Returns an error if a non-integral
 	// array is passed.
 	PutIndices(arrow.Array) error
+	// NormalizeDict takes an arrow array and normalizes it to a parquet
+	// native type. e.g. a dictionary of type int8 will be cast to an int32
+	// dictionary for parquet storage.
+	//
+	// The returned array must be released by the caller if different from
+	// the input array. If no normalization is needed, the input array
+	// may be returned as-is.
+	NormalizeDict(arrow.Array) (arrow.Array, error)
 }
 
 var bufferPool = sync.Pool{
--- a/parquet/pqarrow/encode_arrow_test.go
+++ b/parquet/pqarrow/encode_arrow_test.go
@@ -1353,6 +1353,9 @@ var fullTypeList = []arrow.DataType{
 
 var dictEncodingSupportedTypeList = []arrow.DataType{
 	arrow.PrimitiveTypes.Int8,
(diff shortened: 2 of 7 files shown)


parquet/internal/encoding/types.go
```diff
@@ -113,6 +113,14 @@ type DictEncoder interface {
 	// of [0,dictSize) and is not validated. Returns an error if a non-integral
 	// array is passed.
 	PutIndices(arrow.Array) error
+	// NormalizeDict takes an arrow array and normalizes it to a parquet
+	// native type. e.g. a dictionary of type int8 will be cast to an int32
+	// dictionary for parquet storage.
+	//
+	// The returned array must be released by the caller if different from
+	// the input array. If no normalization is needed, the input array
+	// may be returned as-is.
```
lidavidm: That's kind of annoying, right? You have to branch. Maybe it should always require releasing (normalize can call retain)?

parquet/pqarrow/encode_arrow_test.go
```diff
@@ -1352,6 +1352,7 @@ var fullTypeList = []arrow.DataType{
 }
 
 var dictEncodingSupportedTypeList = []arrow.DataType{
+	arrow.PrimitiveTypes.Int8,
```
amoeba, line 1355: I'm not that familiar with the Go package yet, should this list include more of the possible types?

amoeba (approved):
Two tangential notes:

- Would it be good to title PRs like this fix(pqarrow): ... so users can skim changelogs for parquet changes?
- Does pqarrow smuggle Arrow schemas inside Parquet metadata so users can round-trip? This PR doesn't introduce any breakage of course but it strikes me that arrow -> parquet -> arrow won't round-trip well if the arrow input has to be cast like this.

parquet/pqarrow/encode_arrow_test.go
zeroshade: testing at least one is generally sufficient, but I can add the other types easily to confirm.

parquet/internal/encoding/types.go
zeroshade: fair enough, i'll update to that.

parquet/pqarrow/encode_arrow_test.go
amoeba: Thanks. I like having more tests when they're cheap.

amoeba:
Thanks for the answers to my questions.

lidavidm (approved):

PR #522, opened 2025-09-30
refactor(arrow): last increment of the Record -> RecordBatch migration

### LINKED ISSUE #457: Alias `Record` to `RecordBatch` and soft-deprecate `Record`

The name `Record` is a semantic crime. We should try to make it gradually fade away.

### Component(s)

Documentation, Other

Mandukhai-Alimaa:
### Rationale for this change

Rename the Record interface to RecordBatch for clarity, since Record commonly means a single row but this type represents a batch of rows.

The rest of the code is now changed to use the RecordBatch method and updated interface.

### Are these changes tested?

Yes. All affected packages build successfully and core tests pass.

No user facing API changes. 

Closes #457


.golangci.yaml
```diff
@@ -31,10 +31,6 @@ linters:
       - third_party$
       - builtin$
       - examples$
-    rules:
-      - linters:
-          - staticcheck
-        text: "SA1019"
```
zeroshade, line 37: Why removing this?

zeroshade (approved):
Looks good generally, just one question

.golangci.yaml
Mandukhai-Alimaa: I originally added this rule at the start of the migration to suppress deprecated warnings during pre-commit checks. Now that the code has been updated to use RecordBatch instead of the deprecated Record, the linter should pass without needing to ignore those warnings, so this rule is no longer necessary.

zeroshade: Ah! Nice

PR #545, opened 2025-10-23
feat(arrow/extensions): add support for geoarrow.point

Mandukhai-Alimaa:
### Rationale for this change

There is no support for geoarrow in go

Add support for geoarrow.point and register it as extension type and add tests for the geoarrow.point

### Are these changes tested?

Yes, the newly added test and all the other tests are passing.

Users will be able to use geoarrow.point as an extension type.

### DIFF
--- a/arrow/extensions/geoarrow.go
+++ b/arrow/extensions/geoarrow.go
@@ -1,0 +1,186 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/apache/arrow-go/v18/arrow"
+)
+
+// CoordinateDimension represents the dimensionality of coordinates
+type CoordinateDimension int
+
+const (
+	DimensionXY CoordinateDimension = iota
+	DimensionXYZ
+	DimensionXYM
+	DimensionXYZM
+)
+
+// String returns the string representation of the coordinate dimension
+func (d CoordinateDimension) String() string {
+	switch d {
+	case DimensionXY:
+		return "xy"
+	case DimensionXYZ:
+		return "xyz"
+	case DimensionXYM:
+		return "xym"
+	case DimensionXYZM:
+		return "xyzm"
+	default:
+		return "unknown"
+	}
+}
+
+// Size returns the number of coordinate values per point
+func (d CoordinateDimension) Size() int {
+	switch d {
+	case DimensionXY:
+		return 2
+	case DimensionXYZ, DimensionXYM:
+		return 3
+	case DimensionXYZM:
+		return 4
+	default:
+		return 2
+	}
+}
+
+// GeometryEncoding represents the encoding method for geometry data
+type GeometryEncoding int
+
+const (
+	EncodingGeoArrow GeometryEncoding = iota
+)
+
+// String returns the string representation of the encoding
+func (e GeometryEncoding) String() string {
+	switch e {
+	case EncodingGeoArrow:
+		return "geoarrow"
+	default:
+		return "unknown"
+	}
+}
+
+// EdgeType represents the edge interpretation for geometry data
+type EdgeType string
+
+const (
+	EdgePlanar    EdgeType = "planar"
+	EdgeSpherical EdgeType = "spherical"
+)
+
+// String returns the string representation of the edge type
+func (e EdgeType) String() string {
+	return string(e)
+}
+
+// CoordType represents the coordinate layout type
+type CoordType string
+
+const (
+	CoordSeparate    CoordType = "separate"
+	CoordInterleaved CoordType = "interleaved"
+)
+
+// String returns the string representation of the coordinate type
+func (c CoordType) String() string {
+	return string(c)
+}
+
+// GeometryMetadata contains metadata for GeoArrow geometry types
+type GeometryMetadata struct {
+	// Encoding specifies the geometry encoding format
+	Encoding GeometryEncoding `json:"encoding,omitempty"`
+
+	// CRS contains PROJJSON coordinate reference system information
+	CRS json.RawMessage `json:"crs,omitempty"`
+
+	// Edges specifies the edge interpretation for the geometry
+	Edges EdgeType `json:"edges,omitempty"`
+
+	// CoordType specifies the coordinate layout (separate vs interleaved)
+	CoordType CoordType `json:"coord_type,omitempty"`
+}
+
+// NewGeometryMetadata creates a new GeometryMetadata with default values
+func NewGeometryMetadata() *GeometryMetadata {
+	return &GeometryMetadata{
+		Encoding:  EncodingGeoArrow,
+		Edges:     EdgePlanar,
+		CoordType: CoordSeparate,
+	}
+}
+
+// Serialize serializes the metadata to a JSON string
+func (gm *GeometryMetadata) Serialize() (string, error) {
+	if gm == nil {
+		return "", nil
+	}
+	data, err := json.Marshal(gm)
+	if err != nil {
+		return "", fmt.Errorf("failed to serialize geometry metadata: %w", err)
+	}
+	return string(data), nil
+}
+
+// DeserializeGeometryMetadata deserializes geometry metadata from a JSON string
+func DeserializeGeometryMetadata(data string) (*GeometryMetadata, error) {
+	if data == "" {
+		return NewGeometryMetadata(), nil
+	}
+
+	var gm GeometryMetadata
+	if err := json.Unmarshal([]byte(data), &gm); err != nil {
+		return nil, fmt.Errorf("failed to deserialize geometry metadata: %w", err)
+	}
+
+	return &gm, nil
+}
+
+// createCoordinateType creates an Arrow data type for coordinates based on dimension
+func createCoordinateType(dim CoordinateDimension) arrow.DataType {
+	var fieldNames []string
+
+	switch dim {
+	case DimensionXY:
+		fieldNames = []string{"x", "y"}
+	case DimensionXYZ:
+		fieldNames = []string{"x", "y", "z"}
+	case DimensionXYM:
+		fieldNames = []string{"x", "y", "m"}
+	case DimensionXYZM:
+		fieldNames = []string{"x", "y", "z", "m"}
+	default:
+		fieldNames = []string{"x", "y"}
+	}
+
+	fields := make([]arrow.Field, len(fieldNames))
+	for i, name := range fieldNames {
+		fields[i] = arrow.Field{
+			Name:     name,
+			Type:     arrow.PrimitiveTypes.Float64,
+			Nullable: false,
+		}
+	}
+
+	return arrow.StructOf(fields...)
+}
--- a/arrow/extensions/geoarrow_point.go
+++ b/arrow/extensions/geoarrow_point.go
@@ -1,0 +1,407 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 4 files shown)


arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+// NewEmptyPoint creates a new empty point
+func NewEmptyPoint() Point {
+	return Point{X: math.NaN(), Y: math.NaN(), Z: math.NaN(), M: math.NaN(), Dimension: DimensionXY}
+}
+
+// IsEmpty returns true if this is an empty point (all coordinates are NaN or zero with no dimension)
+func (p Point) IsEmpty() bool {
+	return math.IsNaN(p.X) && math.IsNaN(p.Y) && math.IsNaN(p.Z) && math.IsNaN(p.M)
+}
```
zeroshade, line 65: the comment says "NaN or zero" but you're only checking for NaN here

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+		return 2
+	}
+}
+
+// GeometryEncoding represents the encoding method for geometry data
+type GeometryEncoding int
+
+const (
+	EncodingGeoArrow GeometryEncoding = iota
+)
```
zeroshade, line 71: what would be the alternate to this that would be the other constants?

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+
+// String returns the string representation of the coordinate type
+func (c CoordType) String() string {
+	return string(c)
+}
+
+// GeometryMetadata contains metadata for GeoArrow geometry types
+type GeometryMetadata struct {
+	// Encoding specifies the geometry encoding format
+	Encoding GeometryEncoding `json:"encoding,omitempty"`
```
zeroshade, line 112: I don't see this in the spec, where is this member coming from?

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// CRS contains PROJJSON coordinate reference system information
+	CRS json.RawMessage `json:"crs,omitempty"`
```
zeroshade, line 115: While it should be PROJJSON it could also be other formats

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+	CoordInterleaved CoordType = "interleaved"
+)
+
+// String returns the string representation of the coordinate type
+func (c CoordType) String() string {
+	return string(c)
+}
+
+// GeometryMetadata contains metadata for GeoArrow geometry types
+type GeometryMetadata struct {
```
zeroshade, line 110: you're missing the optional `crs_type` member which should be one of

* "projjson"
* "wkt2:2019"
* "authority_code"
* "srid"

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+	default:
+		return "unknown"
+	}
+}
+
+// EdgeType represents the edge interpretation for geometry data
+type EdgeType string
+
+const (
+	EdgePlanar    EdgeType = "planar"
```
zeroshade, line 87: according to the docs, this isn't a valid value for the "edges" key, Planar is the type if the edges key is omitted entirely

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+	EdgeSpherical EdgeType = "spherical"
+)
```
zeroshade, line 89: missing the other edge types: "vincenty", "thomas", "andoyer" and "karney"

arrow/extensions/geoarrow.go
```diff
@@ -0,0 +1,186 @@
...
+	var gm GeometryMetadata
+	if err := json.Unmarshal([]byte(data), &gm); err != nil {
+		return nil, fmt.Errorf("failed to deserialize geometry metadata: %w", err)
+	}
+
+	return &gm, nil
+}
+
+// createCoordinateType creates an Arrow data type for coordinates based on dimension
+func createCoordinateType(dim CoordinateDimension) arrow.DataType {
```
zeroshade, line 160: you need to have a switch on the coordinate type. While "separated" uses a struct with fields, "interleaved" is instead a `FixedSizeList<float64>[n_dim]`

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+	return Point{X: x, Y: y, M: m, Dimension: DimensionXYM}
+}
+
+// NewPointZM creates a new 3D point with Z and M coordinates
+func NewPointZM(x, y, z, m float64) Point {
+	return Point{X: x, Y: y, Z: z, M: m, Dimension: DimensionXYZM}
+}
+
+// NewEmptyPoint creates a new empty point
+func NewEmptyPoint() Point {
```
zeroshade, line 58: The comment for `IsEmpty` says that an empty point would have no dimension, so we shouldn't initialize it with `DimensionXY`

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+// String returns a string representation of the point
+func (p Point) String() string {
+	if p.IsEmpty() {
+		return "POINT EMPTY"
+	}
+
+	switch p.Dimension {
+	case DimensionXY:
+		return fmt.Sprintf("POINT(%.6f %.6f)", p.X, p.Y)
```
zeroshade, line 75: move this to the bottom and have it be a `fallthrough` instead of repeating it

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+// NewPointTypeWithDimension creates a new Point extension type with specified dimension
+func NewPointTypeWithDimension(dim CoordinateDimension) *PointType {
+	metadata := NewGeometryMetadata()
+	coordType := createCoordinateType(dim)
+
+	return &PointType{
+		ExtensionBase: arrow.ExtensionBase{Storage: coordType},
+		metadata:      metadata,
+		dimension:     dim,
+	}
```
zeroshade, line 108:
```suggestion
	return NewPointTypeWithMetadata(dim, nil)
```

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+// ExtensionEquals checks if two extension types are equal
+func (p *PointType) ExtensionEquals(other arrow.ExtensionType) bool {
+	if p.ExtensionName() != other.ExtensionName() {
+		return false
+	}
+	if otherPoint, ok := other.(*PointType); ok {
+		return p.dimension == otherPoint.dimension
+	}
+	return arrow.TypeEqual(p.Storage, other.StorageType())
```
zeroshade, line 148: The logic here is incorrect. The storage type should always be compared and you're forgetting to compare the metadata

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// Serialize serializes the extension type metadata
+func (p *PointType) Serialize() string {
+	if p.metadata == nil {
+		return ""
+	}
+	serialized, _ := p.metadata.Serialize()
+	return serialized
```
zeroshade, line 157: personally I'd have the json.Marshal and such in here rather than having an explicit `Serialize` method on the metadata object.

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// Deserialize deserializes the extension type metadata
+func (*PointType) Deserialize(storageType arrow.DataType, data string) (arrow.ExtensionType, error) {
+	metadata, err := DeserializeGeometryMetadata(data)
+	if err != nil {
+		return nil, err
+	}
```
zeroshade, line 165: I would do the json.Unmarshal in here instead of having the exported `DeserializeGeometryMetadata`

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+	// Determine dimension from storage type
+	dim := DimensionXY
+	if structType, ok := storageType.(*arrow.StructType); ok {
```
zeroshade, line 169: need to check for the interleaved version too

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+// PointArray represents an array of Point geometries
+type PointArray struct {
+	array.ExtensionArrayBase
+}
+
+// String returns a string representation of the array
+func (p *PointArray) String() string {
+	o := new(strings.Builder)
+	o.WriteString("PointArray[")
```
zeroshade, line 218: don't need `PointArray` in front, we'll automatically add the type when printing

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+		}
+	}
+	o.WriteString("]")
+	return o.String()
+}
+
+// Value returns the Point at the given index
+func (p *PointArray) Value(i int) Point {
+	pointType := p.ExtensionType().(*PointType)
+	structArray := p.Storage().(*array.Struct)
```
zeroshade, line 237: need to determine whether it is interleaved or separate before you assume it's a struct

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+
+	point := Point{Dimension: pointType.dimension}
+
+	if p.IsNull(i) {
+		return point
```
zeroshade, line 242: why not use `NewEmptyPoint`?

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+	case DimensionXYZ:
+		return []float64{point.X, point.Y, point.Z}
+	case DimensionXYM:
+		return []float64{point.X, point.Y, point.M}
+	case DimensionXYZM:
+		return []float64{point.X, point.Y, point.Z, point.M}
+	default:
+		// Should never happen but defensive programming
+		panic(fmt.Sprintf("unknown coordinate dimension: %v", point.Dimension))
+	}
```
zeroshade, line 307: implement `MarshalJSON` on the Point type and then just use that

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+	return &PointBuilder{
+		ExtensionBuilder: array.NewExtensionBuilder(mem, dtype),
+	}
+}
+
+// Append appends a Point to the array
+func (b *PointBuilder) Append(point Point) {
+	pointType := b.Type().(*PointType)
+	structBuilder := b.Builder.(*array.StructBuilder)
+	structBuilder.Append(true)
```
zeroshade, line 335: same comment as above needs to handle the interleaved OR separate cases

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// AppendXYZM appends a 3D point with M coordinate to the array
+func (b *PointBuilder) AppendXYZM(x, y, z, m float64) {
+	b.Append(NewPointZM(x, y, z, m))
+}
+
+// AppendNull appends a null value to the array
+func (b *PointBuilder) AppendNull() {
+	b.ExtensionBuilder.Builder.(*array.StructBuilder).AppendNull()
```
zeroshade, line 384: same comment as above, need to handle the FixedSizeList case

paleolimbot (commented):
Very cool!

I will take a closer look over the weekend, but I did want to quickly point out up front that `geoarrow.wkb` is the only one required for GeoParquet/Parquet Geometry/Geography support (because in Parquet land everything is WKB binary). Functionally that just means there's a tiny bit of CRS translation that has to happen on read/write (and a bit of WKB parsing if you would like to support writing statistics).

Implementing the other extension types is also very cool but fairly difficult (the matrix expansion of point/linestring/polygon/multipoint/multilinestring/multipolygon by xy/xyz/xym/xyzm by interleaved/separated is sort of insane to deal with and is not widely used).

zeroshade:
@dwilson1988 feel free to ping me when you update this so I know to take another pass/look at this!

arrow/extensions/geoarrow_point.go
```diff
@@ -0,0 +1,407 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+)
+
+// Point represents a point geometry with coordinates
+type Point struct {
```
dwilson1988, line 32: @paleolimbot - I meant for these types. For all the concrete geoarrow native encodings, I wanted to use go-geom types for each instead of redefining new ones.

dwilson1988: https://github.com/twpayne/go-geom/blob/407ce645ab08a9a093e599ca07522d8fa483806d/point.go#L15

dwilson1988:
Thanks, I'll see what I can can come up with.

PR #547, opened 2025-10-24
Utilize memory allocator in ReadProperties.GetStream

### LINKED ISSUE #540: [Go][Parquet] Use custom allocator in ReaderProperties.GetStream

We're testing reading parquet files from S3 and I'm looking at the memory profile. We're using a 2GB file for testing and in the profile I see that the `make` call in https://github.com/apache/arrow-go/blob/c8d24781add849d22169a27a8bd6d00a86fbdde1/parquet/reader_properties.go#L80 allocates about 2GB of data, which is about a third of the total allocations that our test that reads the whole file does.

I see that the ReaderProperties has an allocator field, but the allocation there is made using the standard Go call. Is there a reason why the custom allocator is not used there? I wanted to test using a custom allocator, so maybe we can switch the `make` call to `r.alloc.Allocate` ?

### Component(s)

Parquet

daniel-adam-tfs:
### Rationale for this change
Optimization of memory usage, enables the use of custom allocators when reading column data with both buffered and unbuffered readers.

- Optimization of memory allocation and buffer management. 
- Enable the use of custom allocators when reading column data with both buffered and unbuffered readers.

Page Reader Optimizations:

- Unified V1 and V2 page reading with getPageBytesV1() and getPageBytesV2()
- Added readOrStealData() to optimize buffer stealing vs direct reads
- Strategic buffer reuse based on compression/encryption state

Custom Allocator Support:

- Modified utils.NewBufferedReader to accept memory.Allocator parameter
- Added Free() and Buffered() methods to BufferedReader interface
- Added NewBytesBufferReader for allocator-backed byte buffers
- ReaderProperties.GetStream() now uses the configured allocator for all buffer allocations

### Are these changes tested?

- All existing unit tests pass with refactored implementation
- New TestDecryptColumns validates 8 encryption/compression configurations
- Comprehensive benchmarks test V1/V2 pages, compression, and encryption
- Memory leak tests in TestReaderPropsGetStreamWithAllocator and TestReaderPropsGetStreamBufferedWithAllocator verify proper cleanup

New Features:

- Users can now provide custom allocators via ReaderProperties to control memory allocation during page reading

Breaking Changes:

- utils.NewBufferedReader now requires a memory.Allocator parameter (pass nil for default)
- BufferedReader interface adds Buffered() int and Free() methods

Closes #540

### DIFF
--- a/internal/utils/buf_reader.go
+++ b/internal/utils/buf_reader.go
@@ -22,6 +22,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+
+	"github.com/apache/arrow-go/v18/arrow/memory"
 )
 
 type Reader interface {
@@ -38,6 +40,7 @@ type byteReader struct {
 
 // NewByteReader creates a new ByteReader instance from the given byte slice.
 // It wraps the bytes.NewReader function to implement BufferedReader interface.
+// It is considered not to own the underlying byte slice, so the Free method is a no-op.
 func NewByteReader(buf []byte) *byteReader {
 	r := bytes.NewReader(buf)
 	return &byteReader{
@@ -108,10 +111,43 @@ func (r *byteReader) Reset(Reader) {}
 
 func (r *byteReader) BufferSize() int { return len(r.buf) }
 
+func (r *byteReader) Free() {}
+
+// bytesBufferReader is a byte slice with a bytes reader wrapped around it.
+// It uses an allocator to allocate and free the underlying byte slice.
+type bytesBufferReader struct {
+	alloc memory.Allocator
+	byteReader
+}
+
+// NewBytesBufferReader creates a new bytesBufferReader with the given size and allocator.
+func NewBytesBufferReader(size int, alloc memory.Allocator) *bytesBufferReader {
+	buf := alloc.Allocate(size)
+	return &bytesBufferReader{
+		alloc: alloc,
+		byteReader: byteReader{
+			bytes.NewReader(buf),
+			buf,
+			0,
+		},
+	}
+}
+
+// Outer returns the underlying byte slice.
+func (r *bytesBufferReader) Buffer() []byte {
+	return r.buf
+}
+
+// Free releases the underlying byte slice back to the allocator.
+func (r *bytesBufferReader) Free() {
+	r.alloc.Free(r.buf)
+}
+
 // bufferedReader is similar to bufio.Reader except
 // it will expand the buffer if necessary when asked to Peek
 // more bytes than are in the buffer
 type bufferedReader struct {
+	alloc    memory.Allocator // allocator used to allocate the buffer
 	bufferSz int
 	buf      []byte
 	r, w     int
@@ -122,9 +158,10 @@ type bufferedReader struct {
 // NewBufferedReader returns a buffered reader with similar semantics to bufio.Reader
 // except Peek will expand the internal buffer if needed rather than return
 // an error.
-func NewBufferedReader(rd Reader, sz int) *bufferedReader {
+func NewBufferedReader(rd Reader, sz int, alloc memory.Allocator) *bufferedReader {
 	r := &bufferedReader{
-		rd: rd,
+		alloc: alloc,
+		rd:    rd,
 	}
 	r.resizeBuffer(sz)
 	return r
@@ -140,11 +177,9 @@ func (b *bufferedReader) Reset(rd Reader) {
 
 func (b *bufferedReader) resetBuffer() {
 	if b.buf == nil {
-		b.buf = make([]byte, b.bufferSz)
+		b.buf = b.alloc.Allocate(b.bufferSz)
 	} else if b.bufferSz > cap(b.buf) {
-		buf := b.buf
-		b.buf = make([]byte, b.bufferSz)
-		copy(b.buf, buf)
+		b.buf = b.alloc.Reallocate(b.bufferSz, b.buf)
 	} else {
 		b.buf = b.buf[:b.bufferSz]
 	}
@@ -297,4 +332,8 @@ func (b *bufferedReader) Read(p []byte) (n int, err error) {
 	n = copy(p, b.buf[b.r:b.w])
 	b.r += n
 	return n, nil
+}
+
+func (b *bufferedReader) Free() {
+	b.alloc.Free(b.buf)
 }
--- a/internal/utils/buf_reader.go
+++ b/internal/utils/buf_reader.go
@@ -40,7 +40,7 @@ type byteReader struct {
 
 // NewByteReader creates a new ByteReader instance from the given byte slice.
 // It wraps the bytes.NewReader function to implement BufferedReader interface.
-// It is considered not to own the underlying byte slice, so the Free method is a no-op.
+// It is considered not to own the underlying byte slice.
 func NewByteReader(buf []byte) *byteReader {
 	r := bytes.NewReader(buf)
 	return &byteReader{
@@ -113,7 +113,11 @@ func (r *byteReader) BufferSize() int { return len(r.buf) }
 
 func (r *byteReader) Buffered() int { return len(r.buf) - r.pos }
 
-func (r *byteReader) Free() {}
+func (r *byteReader) Free() {
+	r.r = nil
+	r.buf = nil
+	r.pos = 0
+}
 
 // bytesBufferReader is a byte slice with a bytes reader wrapped around it.
 // It uses an allocator to allocate and free the underlying byte slice.
--- a/internal/utils/buf_reader.go
+++ b/internal/utils/buf_reader.go
@@ -111,6 +111,8 @@ func (r *byteReader) Reset(Reader) {}
 
 func (r *byteReader) BufferSize() int { return len(r.buf) }
 
+func (r *byteReader) Buffered() int { return len(r.buf) - r.pos }
+
 func (r *byteReader) Free() {}
 
 // bytesBufferReader is a byte slice with a bytes reader wrapped around it.
--- a/internal/utils/buf_reader.go
+++ b/internal/utils/buf_reader.go
@@ -122,6 +122,9 @@ type bytesBufferReader struct {
 
 // NewBytesBufferReader creates a new bytesBufferReader with the given size and allocator.
 func NewBytesBufferReader(size int, alloc memory.Allocator) *bytesBufferReader {
+	if alloc == nil {
+		alloc = memory.DefaultAllocator
+	}
 	buf := alloc.Allocate(size)
 	return &bytesBufferReader{
 		alloc: alloc,
@@ -159,6 +162,9 @@ type bufferedReader struct {
 // except Peek will expand the internal buffer if needed rather than return
 // an error.
 func NewBufferedReader(rd Reader, sz int, alloc memory.Allocator) *bufferedReader {
+	if alloc == nil {
+		alloc = memory.DefaultAllocator
+	}
 	r := &bufferedReader{
 		alloc: alloc,
 		rd:    rd,
--- a/parquet/file/column_reader_test.go
+++ b/parquet/file/column_reader_test.go
@@ -33,13 +33,26 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/internal/utils"
 	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/compress"
 	"github.com/apache/arrow-go/v18/parquet/file"
+	"github.com/apache/arrow-go/v18/parquet/internal/encryption"
 	"github.com/apache/arrow-go/v18/parquet/internal/testutils"
 	"github.com/apache/arrow-go/v18/parquet/pqarrow"
 	"github.com/apache/arrow-go/v18/parquet/schema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+)
+
+const (
+	FooterEncryptionKey    = "0123456789012345"
+	ColumnEncryptionKey1   = "1234567890123450"
+	ColumnEncryptionKey2   = "1234567890123451"
+	ColumnEncryptionKey3   = "1234567890123452"
+	FooterEncryptionKeyID  = "kf"
+	ColumnEncryptionKey1ID = "kc1"
+	ColumnEncryptionKey2ID = "kc2"
+	ColumnEncryptionKey3ID = "kc3"
 )
 
 func initValues(values reflect.Value) {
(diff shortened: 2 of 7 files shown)


parquet/file/page_reader.go
```diff
@@ -501,7 +504,16 @@ func (p *serializedPageReader) Page() Page {
...
 
 func (p *serializedPageReader) decompress(rd io.Reader, lenCompressed int, buf []byte) ([]byte, error) {
-	p.decompressBuffer.ResizeNoShrink(lenCompressed)
+	// As of go1.25.3: There is an issue when bytes.Buffer and io.CopyN are used together. io.CopyN
+	// uses io.LimitReader, which does an additional read on the underlying reader to determine EOF.
+	// However, bytes.Buffer always attempts to read at least bytes.MinRead (which is 512 bytes) from the
+	// underlying reader, even if there is less data available than that. So even if there are no more bytes,
+	// the buffer must have at least bytes.MinRead capacity remaining to avoid a relocation.
+	allocSize := lenCompressed
+	if p.decompressBuffer.Cap() < lenCompressed+bytes.MinRead {
```
daniel-adam-tfs: This is dependent on the combined behavior of io.LimitReader and bytes.Buffer. Which seems fragile to me, but I don't have any other ideas how to deal with it. I'll at least add unit tests that the reallocation happens when I don't add bytes.MinRead to the allocation size and doesn't happen when I do.

zeroshade: I agree that this seems really fragile. Maybe `io.ReadFull` directly into `p.decompressBuffer.Bytes()[:lenCompressed]` instead of using the intermediate `bytes.Buffer`?

daniel-adam-tfs: Yes, lets go with ReadFull and we can skip ``bytes.Buffer` altogether.

parquet/file/page_reader.go
```diff
@@ -500,14 +508,31 @@ func (p *serializedPageReader) Page() Page {
...
 
+func (p *serializedPageReader) readUncompressed(rd io.Reader, lenUncompressed int, buf []byte) ([]byte, error) {
+	n, err := io.ReadFull(rd, buf[:lenUncompressed])
+	if err != nil {
+		return nil, err
+	}
+	if n != lenUncompressed {
+		return nil, fmt.Errorf("parquet: expected to read %d bytes but only read %d", lenUncompressed, n)
+	}
+	if p.cryptoCtx.DataDecryptor != nil {
```
daniel-adam-tfs: I'm not sure if this is needed or not. But for data page v2, the data is just read by ReadFull and Decrypt is not called:

https://github.com/apache/arrow-go/blob/95b3f76a5c82529784402ff992c511fe9998ec6f/parquet/file/page_reader.go#L815-L824

So maybe the Decrypt call is not needed for data age v1 or dictionary page either?

zeroshade: The reverse actually. Looks like this is a bug we just never came across, I'm guessing no one was using DataPageV2 with uncompressed data but still encrypted that was using this library.

daniel-adam-tfs: Alright, I'll fix it for DataPageV2 then. I'll add a unit tests without compression and with encryption that should fail with the current main, if I have the time.

I reran the profiler with the current commit in this PR, with a 2.8GB parquet file stored in S3, uncompressed. And cpu profiler is showing that more time is spent in runtime.memmove (copying memory) than in syscall.Syscall6 (read). Which is annoying me. :-D

I think it should be still possible to eliminate at least one copy for the uncompressed case.

So this is my scenario:
1) `ReaderProperties.GetStream` reads column chunk from a TLS and stores it in a buffer (or just allocates the buffer if `BufferedStreamEnabled`, but lets go with the unbuffered case for now)
2) `serializedPageReader` is created with the buffer returned from ReaderProperties.GetStream
3) `serializedPageReader.Next` get the page header calls `serializedPageReader.readUncompressed` / `serializedPageReader.decompress` which reads data from the GetStream buffer into dictPageBuffer/dataPageBuffer
  3a) for the uncompressed case the this is just a copy
4) `page` struct is created from the bytes written to the dictPageBuffer/dataPageBuffer

I think I could avoid the copy in 3a and create page directly from the bytes in the buffer returned by `ReaderProperties.GetStream` by using combination of calls Peek (to get the bytes)+Discard (to move the internal position inside the buffer). This should hold when `BufferedStreamEnabled` is false, I have to check what happens when it is true.

zeroshade: Awesome! Thanks for diving into this!

daniel-adam-tfs: Alright, so I "steal" the buffer by using `Peek`/`Discard` if the data has been read previously and it is available of the `BufferedReader`. So in the uncompressed and unencrypted case -> data is read and stored into a buffer in `ReaderProperties.GetStream` and copied to the user provided buffer to `Float32ColumnChunkReader.ReadBatch`. 
Now, if we have a plainEncoder and no compression, it should be possible to write the data directly to the user provided buffer, so that would eliminate even that copy, but one is more complicated and I need to be start doing other stuff. :D

Also, the decryption types allocate buffers for the decrypted data. We could send it an already allocated buffer to use, or maybe do an in place decryption (if possible), or give it the custom allocator if it is set.

Anyway, I'll fix the decryption for DataPageV2 next and I'll consider this one done.

parquet/file/column_reader_test.go
```diff
@@ -813,6 +826,145 @@ func TestFullSeekRow(t *testing.T) {
...
+	rdr, err := file.NewParquetReader(bytes.NewReader(buf.Bytes()), file.WithReadProps(readProps))
+	require.NoError(t, err)
+	defer rdr.Close()
+
+	rgr := rdr.RowGroup(0)
+	col0, err := rgr.Column(0)
+	require.NoError(t, err)
+
+	icr := col0.(*file.Int64ColumnChunkReader)
+	// require.NoError(t, icr.SeekToRow(3)) // TODO: this causes a panic currently
```
daniel-adam-tfs, line 867: If I uncomment this, then this causes a panic

parquet/file/column_reader_test.go
```diff
@@ -813,6 +826,145 @@ func TestFullSeekRow(t *testing.T) {
...
+		{"c0": 6,    "c1": "f",  "c2": null}
+	]`})
+	require.NoError(t, err)
+	defer tbl.Release()
+
+	schema := tbl.Schema()
+	arrWriterProps := pqarrow.NewArrowWriterProperties()
+
+	var buf bytes.Buffer
+	wr, err := pqarrow.NewFileWriter(schema, &buf, writerProps, arrWriterProps)
```
daniel-adam-tfs, line 852: @zeroshade I think there is a bug in pqarrow writer. For the data page v2 buffer, it first writes the levels (definition and repetition) and then the values. Only values are compressed. However, this whole buffer is then encrypted. And unless ChatGPT is hallucinating on me then only the compressed values should also be encrypted, levels should stay unencrypted and uncompressed.
I'll check with some encrypted parquet create in a different way on Monday and see what happens.

daniel-adam-tfs: @zeroshade OK, I should have some time this week to finish this. In fact, I think the memory allocation is done, but the decryption needs fixing before the tests pass.

I've been trying to use PyArrow to encrypt/decrypt files, but there seems to be some discrepancy in implementations. I cannot get files encrypted by PyArrow to decrypt using go-arrow and vice versa. I'll open an issue for the encryption/decryption.

zeroshade: Please link the issue here when you file it, as PyArrow and arrow-go should be agreeing on encrypting/decrypting and vice versa since pyarrow binds to Arrow C++ and the tests for parquet C++ and arrow-go should be the same tests that are passing. I'd be interested to reproduce the failure and debug it

parquet/file/column_reader_test.go
zeroshade: That shouldn't  be panicing as the SeekToRow works correctly based on my last tests.... I'll see if i can debug this

daniel-adam-tfs:
Alright, decryption should be correct now. Let me rebase this (...and try remember what was happening here 😄 )

parquet/file/column_reader_test.go
daniel-adam-tfs: This works correctly now.

parquet/file/column_reader_test.go
daniel-adam-tfs: This still panics:

```
--- FAIL: TestDecryptColumns (0.00s)
    --- FAIL: TestDecryptColumns/DataPageV2_BufferedRead (0.00s)
panic: cipher: message authentication failed [recovered, repanicked]

goroutine 8 [running]:
testing.tRunner.func1.2({0x2899940, 0xc000054100})
	/usr/local/Cellar/go/1.25.6/libexec/src/testing/testing.go:1872 +0x237
testing.tRunner.func1()
	/usr/local/Cellar/go/1.25.6/libexec/src/testing/testing.go:1875 +0x35b
panic({0x2899940?, 0xc000054100?})
	/usr/local/Cellar/go/1.25.6/libexec/src/runtime/panic.go:783 +0x132
github.com/apache/arrow-go/v18/parquet/internal/encryption.(*aesDecryptor).Decrypt(0xc0002c58f0, {0xc000026800, 0x4000, 0x4000}, {0xc0002c5b10?, 0x600c00003ff20?, 0x44d3560?}, {0xc0002c5e90, 0xd, 0x10})
	/daniel-adam-tfs/arrow-go/parquet/internal/encryption/aes.go:262 +0x26d
github.com/apache/arrow-go/v18/parquet/internal/encryption.(*decryptor).Decrypt(0xc0003326c0?, {0xc000026800?, 0xc00003fd40?, 0x1397b05?})
	/daniel-adam-tfs/arrow-go/parquet/internal/encryption/decryptor.go:268 +0x45
github.com/apache/arrow-go/v18/parquet/file.(*serializedPageReader).readPageHeader(0xc0003326c0, {0x2aceb20, 0xc00003ff20}, 0xc0002ff740)
	/daniel-adam-tfs/arrow-go/parquet/file/page_reader.go:704 +0x139
github.com/apache/arrow-go/v18/parquet/file.(*serializedPageReader).Next(0xc0003326c0)
	/daniel-adam-tfs/arrow-go/parquet/file/page_reader.go:798 +0xdd
github.com/apache/arrow-go/v18/parquet/file.(*serializedPageReader).SeekToPageWithRow(0xc0003326c0, 0x3)
	/daniel-adam-tfs/arrow-go/parquet/file/page_reader.go:749 +0x186
github.com/apache/arrow-go/v18/parquet/file.(*columnChunkReader).SeekToRow(0xc0002fa780, 0x3)
	/daniel-adam-tfs/arrow-go/parquet/file/column_reader.go:584 +0x2a
github.com/apache/arrow-go/v18/parquet/file_test.checkDecryptedValues(0xc000326380, 0xc000210870, 0xc0003283f0)
	/daniel-adam-tfs/arrow-go/parquet/file/column_reader_test.go:867 +0x591
github.com/apache/arrow-go/v18/parquet/file_test.TestDecryptColumns.func1(0xc000326380)
	/daniel-adam-tfs/arrow-go/parquet/file/column_reader_test.go:964 +0x18a
testing.tRunner(0xc000326380, 0xc0002fe680)
	/usr/local/Cellar/go/1.25.6/libexec/src/testing/testing.go:1934 +0xea
created by testing.(*T).Run in goroutine 7
	/usr/local/Cellar/go/1.25.6/libexec/src/testing/testing.go:1997 +0x465
FAIL	github.com/apache/arrow-go/v18/parquet/file	1.922s
```

For each of the cases defined, doesn't matter whether it is a V1 or V2 page, or if is it compressed or not, or if BufferedStream is used or not. SeekToRow works in the unencrypted case, so this is again encryption related.

daniel-adam-tfs: It seems to be related to this line:

https://github.com/apache/arrow-go/blob/38dc64bb93890aa0c4b292f58ac3344ced9d12da/parquet/file/page_reader.go#L699-L702

I think this is intended to skip past the dictionary page to the data page. But for some reason after this Seek the 

https://github.com/apache/arrow-go/blob/38dc64bb93890aa0c4b292f58ac3344ced9d12da/parquet/file/page_reader.go#L663

call fails. The offset to the data page is correct, so I'm thinking that some internal state of the decryptor or the page reader is affected by skip past the parsing of the dictionary page. 

But I'll remove the SeekToRow and park this for now in https://github.com/apache/arrow-go/issues/566. Because we are benchmarking various formats + readers of unencrypted data, so I wanna make some optimization (including this PR, which according to my benchmarks speeds things up a little).

internal/utils/buf_reader.go
```diff
@@ -108,10 +111,48 @@ func (r *byteReader) Reset(Reader) {}
 
 func (r *byteReader) BufferSize() int { return len(r.buf) }
 
+func (r *byteReader) Buffered() int { return len(r.buf) - r.pos }
+
+func (r *byteReader) Free() {}
```
zeroshade: should `Free` do `r.r = nil` and `r.buf = nil` ?

parquet/file/row_group_reader.go
```diff
@@ -134,11 +134,13 @@ func (r *RowGroupReader) GetColumnPageReader(i int) (PageReader, error) {
 	}
 
 	if r.fileDecryptor == nil {
+		stream.Free()
```
zeroshade, line 137: should we just put `defer stream.Free()` at line 121 instead?

parquet/reader_properties.go
```diff
@@ -50,8 +50,11 @@ type BufferedReader interface {
 	Peek(int) ([]byte, error)
 	Discard(int) (int, error)
 	Outer() utils.Reader
+	// Buffered returns the number of bytes already read and stored in the buffer
+	Buffered() int
```
zeroshade: this will potentially break other users, we instead need to create a new interface the embeds this one so that we don't break other consumers. Something like adding to `internal/utils/buf_reader.go`

```go
type BufferedReaderV2 interface {
    BufferedReader

    Buffered() int
    Free()
}
```

zeroshade (changes requested):

parquet/file/row_group_reader.go
daniel-adam-tfs: Couldn't think of an simple way of doing it. The most robust way is something this: 

```
shouldFreeStream := true
takeStream := func(s BufferedReaderV2) BufferedReaderV2 {
  shouldFreeStream = false
  return s
}
```

And use `takeStream` when stream is assigned and returned and check `shouldFreeStream` in the deferred function.

parquet/file/page_reader.go
```diff
@@ -451,12 +477,17 @@ func NewPageReader(r parquet.BufferedReader, nrows int64, compressType compress.
 func (p *serializedPageReader) Reset(r parquet.BufferedReader, nrows int64, compressType compress.Compression, ctx *CryptoContext) {
```
daniel-adam-tfs, line 477: If I could change the signature of Reset of the `PageReader` interface to use `parquet.BufferedReaderV2` I could skip the whole `bufferedReaderV2Adapter` shenanigans, but `NewPageReader` and `RecordReader.SetPageReader` are public API. If they get ever removed from public API then we can get rid of `bufferedReaderV2Adapter` as well.

daniel-adam-tfs:
I've rebased it, so it has the fix and test from yesterday. But let me add some additional tests, since going by the coverage report not all code branches in the new functions is covered.

zeroshade:
@daniel-adam-tfs A few things need addressing before review — see our [Pull Request quality criteria](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md).

- :x: **Merge conflicts** — this branch conflicts with `main`. See [docs](https://arrow.apache.org/docs/developers/guide/index.html).

> **Note:** Your branch is **364 commits behind `main`**. Please rebase and push again to get up-to-date CI results.

No rush.

---

_Note: This comment was drafted by an AI-assisted triage tool run by a maintainer, and may contain mistakes. Once you have addressed the points above, an Apache Arrow Go maintainer — a real person — will take the next look at your PR. If anything here looks wrong, say so on the PR and a maintainer will follow up. See [CONTRIBUTING.md](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md) for the project's contribution conventions._

Danielius1922:
Our next 3 week sprint will start next Wednesday, so let me try to squeeze this in.

Issue #511, opened 2025-09-17
[Go][Parquet] pqarrow to support RowGroupTotalBytesWritten() for writer using WriteBuffered()


DuanWeiFan:
### Describe the enhancement requested

The method **RowGroupTotalBytesWritten()** is available but only make sense for the ones that use [fileWriter.Write()](https://github.com/apache/arrow-go/blob/main/parquet/pqarrow/file_writer_test.go#L54C21-L54C41)

For the ones that use [fileWriter.WriteBuffered()](https://github.com/apache/arrow-go/blob/main/parquet/pqarrow/file_writer_test.go#L131) , it won't populate the bytes written until the writer is closed.

We are wondering if there is a way of populating or give an estimation of the size before closing the rowGroupWriter. This would be helpful to determine how large of a parquet file we want to build.

### Component(s)

Parquet

DuanWeiFan:
After some investigation, I realize `RowGroupTotalCompressedBytes()` will get populated as soon as one of the DataPage got flushed. 
Wondering if that makes sense to add a `totalCompressedBytes` & `totalBytesWritten` in FileWriter such that it can track the total number of bytes for every row group instead of just the current row group.

The two new methods: `TotalBytesWritten()` & `TotalCompressedBytes()` will then be able to report the total bytes written by the FileWriter.
```
type FileWriter struct {
  ...
  totalCompressedBytes int64
  totalBytesWritten    int64
}
// NewRowGroup does what it says on the tin, creates a new row group in the underlying file.
// Equivalent to `AppendRowGroup` on a file.Writer
func (fw *FileWriter) NewRowGroup() {
  if fw.rgw != nil {
    fw.totalCompressedBytes += fw.rgw.TotalCompressedBytes()
    fw.totalBytesWritten += fw.rgw.TotalBytesWritten()
    fw.rgw.Close()
  }
  fw.rgw = fw.wr.AppendRowGroup()
  fw.colIdx = 0
}

// NewBufferedRowGroup starts a new memory Buffered Row Group to allow writing columns / records
// without immediately flushing them to disk. This allows using WriteBuffered to write records
// and decide where to break your row group based on the TotalBytesWritten rather than on the max
// row group len. If using Records, this should be paired with WriteBuffered, while
// Write will always write a new record as a row group in and of itself.
func (fw *FileWriter) NewBufferedRowGroup() {
  if fw.rgw != nil {
    fw.totalCompressedBytes += fw.rgw.TotalCompressedBytes()
[...]

PR #548, opened 2025-10-25
[Go] [Parquet] pqarrow file-writer & row-group-writer tracking total & total compressed bytes

DuanWeiFan:
### Rationale for this change
As discussed in [issue](https://github.com/apache/arrow-go/issues/511), pqarrow parquet writer when writing more than 1 row group, it cannot track the bytesWritten & compressedbytesWritten for closed row group.
The method `RowGroupTotalCompressedBytes` & `RowGroupTotalBytesWritten` only returns the **current** row group.
The ideal is to introduce `TotalCompressedBytes()` & `TotalBytesWritten()` as an enhancement for user to track all row group written bytes & compressed bytes.

As part of making this change, I notice the totalCompressedBytes is not being populated correctly in `row_group_writer.go`.
It should follow the same approach as `bytesWritten` where it tracks all the closed column_writer.

### Are these changes tested?
Yes. I added test cases to ensure those metrics are populated correctly

Yes. There are two new methods introduced:
1. TotalCompressedBytes()
2. TotalBytesWritten()

It won't impact on existing users.

### DIFF
--- a/parquet/pqarrow/file_writer.go
+++ b/parquet/pqarrow/file_writer.go
@@ -49,16 +49,14 @@ func WriteTable(tbl arrow.Table, w io.Writer, chunkSize int64, props *parquet.Wr
 
 // FileWriter is an object for writing Arrow directly to a parquet file.
 type FileWriter struct {
-	wr                   *file.Writer
-	schema               *arrow.Schema
-	manifest             *SchemaManifest
-	rgw                  file.RowGroupWriter
-	arrowProps           ArrowWriterProperties
-	ctx                  context.Context
-	colIdx               int
-	closed               bool
-	totalCompressedBytes int64
-	totalBytesWritten    int64
+	wr         *file.Writer
+	schema     *arrow.Schema
+	manifest   *SchemaManifest
+	rgw        file.RowGroupWriter
+	arrowProps ArrowWriterProperties
+	ctx        context.Context
+	colIdx     int
+	closed     bool
 }
 
 // NewFileWriter returns a writer for writing Arrow directly to a parquetfile, rather than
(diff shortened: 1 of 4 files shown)


parquet/pqarrow/file_writer.go
```diff
@@ -111,11 +115,27 @@ func (fw *FileWriter) NewRowGroup() {
 func (fw *FileWriter) NewBufferedRowGroup() {
 	if fw.rgw != nil {
 		fw.rgw.Close()
+		fw.totalCompressedBytes += fw.rgw.TotalCompressedBytes()
+		fw.totalBytesWritten += fw.rgw.TotalBytesWritten()
```
zeroshade: In `FileWriter.Close` when we write the metadata, we should probably add the size of the metadata we write to the total bytes written, yea?

parquet/pqarrow/file_writer_test.go
```diff
@@ -133,3 +133,78 @@ func TestFileWriterBuffered(t *testing.T) {
...
+
+	// Write second record, which creates a new row group
+	require.NoError(t, writer.Write(record2))
+
+	// Close the writer and verify final bytes
+	require.NoError(t, writer.Close())
+
+	// Verify total bytes & compressed bytes are greater than 0
+	assert.Greater(t, writer.TotalCompressedBytes(), int64(0))
+	assert.Greater(t, writer.TotalBytesWritten(), int64(0))
```
zeroshade: since we're creating the test and know the settings, we can probably know the exact size to test for and have the test be explicit on the size we expect so that we know if we break anything in the future. Make sense?

parquet/pqarrow/file_writer.go
DuanWeiFan: makes a lot of sense! I've added that to Close() so we correctly track the last row group's bytes written

parquet/pqarrow/file_writer_test.go
DuanWeiFan: good point! updated the test cases

parquet/pqarrow/file_writer.go
zeroshade: I was referring to the `Close` method in `file/file_writer.go` specifically, inside of `FlushWithFooter`. Currently when we close the file, we'll cover all of the bytes *except* for the metadata footer which is written at the end of the file. Make sense?

zeroshade: Sorry for the delay here, I was travelling for a conference last week

DuanWeiFan:
sorry for the delay I lost track of this.
I think if we are considering the metadata (header/footer..), instead of rowGroup level, I think we would need to move
total bytes written to file/file_writer.go so it has the complete picture.

PR #551, opened 2025-10-27
fix(parquet/pqarrow): decoding Parquet with Arrow dict in schema

### LINKED ISSUE #546: Cannot read Parquet files with multiple dictionary pages per column chunk

The `pqarrow.NewFileReader` API fails to read valid Parquet files that contain multiple dictionary pages within a single column chunk. This prevents reading files created by Apache DataFusion for instance.

Error:
```
parquet: column chunk cannot have more than one dictionary.
````

Version: v18

### Component(s)

Parquet

### PR DESCRIPTION
**Pull request by freakyzoidberg**
Fix spurious `parquet: column chunk cannot have more than one dictionary.` with specific parquet file
Resolve #546

Parquet with 
* Arrow Dict column
* Arrow Schema serialied in Parquet Metadata
* ColumnChunks with 1 dict page + at least 2 Data page

# Bug
When maybeWriteNewDictionary() resets `newDictionary = false` at line 965, it causes the next call to readDictionary() to try to read the dictionary page again from the pager, which then calls configureDict() again, which throws the "cannot have more than one dictionary" error!

  The sequence is:
  1. Read DICTIONARY_PAGE → newDictionary = true
  2. Read DATA_PAGE_1 → calls maybeWriteNewDictionary() → resets newDictionary = false
  3. Read DATA_PAGE_2 → calls readDictionary() → since newDictionary = false, tries to get dictionary page again → calls configureDict() → ERROR because decoder already exists

# Fix 
  Added DictionaryState enum (column_reader.go):
  - DictNotRead: Dictionary page hasn't been read yet
  - DictReadNotInserted: Dictionary page read and decoder configured, but not inserted into Arrow builder
  - DictFullyProcessed: Dictionary fully processed (read + inserted into builder)

### DIFF
--- a/parquet/file/column_reader.go
+++ b/parquet/file/column_reader.go
@@ -38,17 +38,17 @@ const (
 	defaultPageHeaderSize = 16 * 1024
 )
 
-// DictionaryState tracks the lifecycle of dictionary handling for a column chunk
-type DictionaryState int
+// dictionaryState tracks the lifecycle of dictionary handling for a column chunk
+type dictionaryState int
 
 const (
-	// DictNotRead: Dictionary page has not been read yet
-	DictNotRead DictionaryState = iota
-	// DictReadNotInserted: Dictionary page has been read and decoder configured,
+	// dictNotRead: Dictionary page has not been read yet
+	dictNotRead dictionaryState = iota
+	// dictReadNotInserted: Dictionary page has been read and decoder configured,
 	// but not yet inserted into Arrow builder (for Arrow Dictionary types only)
-	DictReadNotInserted
-	// DictFullyProcessed: Dictionary has been read, configured, and inserted into builder
-	DictFullyProcessed
+	dictReadNotInserted
+	// dictFullyProcessed: Dictionary has been read, configured, and inserted into builder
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ahmed-mez** [comment]
Thanks for the fix and reproducer @freakyzoidberg !

I confirm that running your branch on the file I mentioned in https://github.com/apache/arrow-go/issues/546 now doesn't return the `parquet: column chunk cannot have more than one dictionary.` error any more.

File: parquet/file/column_reader.go [hunk 1]
```diff
@@ -38,6 +38,19 @@ const (
...
+
+const (
+	// DictNotRead: Dictionary page has not been read yet
+	DictNotRead DictionaryState = iota
+	// DictReadNotInserted: Dictionary page has been read and decoder configured,
+	// but not yet inserted into Arrow builder (for Arrow Dictionary types only)
+	DictReadNotInserted
+	// DictFullyProcessed: Dictionary has been read, configured, and inserted into builder
+	DictFullyProcessed
+)
```
zeroshade: my only critique is that we shouldn't export these.

zeroshade (approved):

Issue #552, opened 2025-10-28
[pqarrow] Support chunked array


ahmed-mez:
### Describe the enhancement requested

Chunked arrays are currently not supported and it's a feature gap.

https://github.com/apache/arrow-go/blob/95b3f76a5c82529784402ff992c511fe9998ec6f/parquet/pqarrow/column_readers.go#L502-L503

Related to this, it's hard to identify the root cause of the error without tracing the arrow-go code itself. It'd be super useful to add context to the error message (currently it's a vague `not implemented`).

Happy to take a stab at this if the maintainers / people familiar with the project could share some pointers.

### Component(s)

Parquet

PR #553, opened 2025-10-28
feat: support conversion of chunked arrays

ahmed-mez:
### Rationale for this change

Fixes https://github.com/apache/arrow-go/issues/552

Make `chunksToSingle` concatenate multiple chunks into a single array using `array.Concatenate` instead of returning `ErrNotImplemented`

### Are these changes tested?

Unit tests + tested end-to-end locally on a parquet file that contains chunked arrays

Changes are backward compatible

### DIFF
--- a/parquet/pqarrow/column_readers_test.go
+++ b/parquet/pqarrow/column_readers_test.go
@@ -17,11 +17,17 @@
 package pqarrow
 
 import (
+	"bytes"
+	"context"
+	"io"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/compress"
+	"github.com/apache/arrow-go/v18/parquet/file"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


parquet/pqarrow/column_readers_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChunksToSingle(t *testing.T) {
+	mem := memory.NewGoAllocator()
```
zeroshade: Can we use a CheckedAllocator and assert the size to ensure we don't have any memory leak?

zeroshade (changes requested):
Looks Good! Just two things: the comment below and can we add a round trip test that actually writes a file using a chunked table?

parquet/pqarrow/column_readers_test.go
ahmed-mez: Done in https://github.com/apache/arrow-go/pull/553/commits/dc63c0951fd7212d5138f8a13db252b15dc5d473

ahmed-mez:
Thank you @zeroshade for the review!

I applied the `CheckedAllocator` suggestion and added a round trip test.

PR #555, opened 2025-10-30
format: regenerate internal/flatbuf from arrow repo and newer flatc

pixelherodev:
I plan on at least sending a PR to optimize these files, so regenerating them using upstream data as is without modifications should make such PRs easier to review.


pixelherodev:
Internal review noticed files added by mistake; force-pushed with those removed.

PR #556, opened 2025-10-30
Batch of small optimizations

pixelherodev:
This primarily consists of reducing allocations.

e.g. 03be6d3f adds a specific case to TypeEqual for ListView types, which avoids reflect.DeepEqual. l.elem.Equal does not allocate; reflect.DeepEqual does, so this reduces allocations when comparing a ListViewType.

8c7729fc changes from referencing flatbug internal types by reference to by-value, which avoids letting them escape to the heap.

e3406582 switches from reflect.DeepEqual to slices.Equal when comparing the child IDs of a Union type, which is functionally equivalent for slices and, again, does not allocate.

9eba6d2c does similar for an arrowflight cookie test, replacing reflect.DeepEquals with maps.Equal; I did a grep for reflect.DeepEqual and noticed it.

7557d15c is probably the most objectionable; it replaces a *float64 with float64 directly for minSpaceSavings, since the current nil value has identical behavior to zero anyways. It's a tiny cleanup IMO, but doesn't really have any practical value.

2d3593f6 is a bit ugly; it inlines a copy of NewReaderFromMessageReader and a copy of NewMessageReader into NewReader - by implementing the two directly, it's able to avoid allocating two copies of the `config` structure, and to avoid looping over and invoking config functions twice. It's kinda ugly though, and the `// types: make(dictTypeMap)` which is copied over suggests that there's further room for improvement.

I didn't drop the two APIs, as they are exposed to users, but IMO the performance win is probably worth it, and it can be cleaned up later if needed.

0273ffe8 is a small optimization to the allocator: the Go allocator banches on each allocation, but both branches are functionally equivalent, so I just merged the two. I doubt this would have a _major_ impact, since the cost of a branch is negligible compared to the cost of a heap allocation, but performance is often won with a lot of small impacts IMO :)

Lastly, for now at least, f1a6a331 statically allocates 4 bytes per IPC messageReader (I did _not_ touch the ArrowFlight implementation), and uses it on every message read, rather than heap-allocating 4 bytes each time Message is called.

There's more to be done, but this should be a much more practically sized chunk to review, and I can add more later :)

### DIFF
--- a/arrow/flight/basic_auth_flight_test.go
+++ b/arrow/flight/basic_auth_flight_test.go
@@ -81,7 +81,7 @@ func TestErrorAuths(t *testing.T) {
 
 	go func() {
 		if err := s.Serve(); err != nil {
-			t.Fatal(err)
+			panic(err)
 		}
 	}()
 	defer s.Shutdown()


pixelherodev:
There's a lot of variance in the run times for the existing benchmarks for me, but the before and after for allocations are consistent:

```bash
# branch 'main'
goos: linux
goarch: amd64
pkg: github.com/apache/arrow-go/v18/arrow/ipc
cpu: AMD Ryzen 9 7900X3D 12-Core Processor          
BenchmarkIPC/Writer/codec=plain-24         	 207901	     5824 ns/op	   8160 B/op	     91 allocs/op
BenchmarkIPC/Reader/codec=plain-24         	 247482	     4812 ns/op	   6132 B/op	     83 allocs/op
BenchmarkIPC/Writer/codec=zstd-24          	   3618	   332506 ns/op	2338754 B/op	    150 allocs/op
BenchmarkIPC/Reader/codec=zstd-24          	  40870	    24626 ns/op	  25887 B/op	    194 allocs/op
BenchmarkIPC/Writer/codec=lz4-24           	 112738	    10490 ns/op	  12864 B/op	    134 allocs/op
BenchmarkIPC/Reader/codec=lz4-24           	 133892	     9154 ns/op	  10115 B/op	    124 allocs/op


# branch 'small-optimizations'
BenchmarkIPC/Writer/codec=plain-24         	 168954	     7469 ns/op	   8160 B/op	     91 allocs/op
BenchmarkIPC/Reader/codec=plain-24         	 259921	     4970 ns/op	   5993 B/op	     76 allocs/op
BenchmarkIPC/Writer/codec=zstd-24          	   3633	   327594 ns/op	2338740 B/op	    150 allocs/op
BenchmarkIPC/Reader/codec=zstd-24          	  40442	    26658 ns/op	  25739 B/op	    187 allocs/op
BenchmarkIPC/Writer/codec=lz4-24           	 113894	    13151 ns/op	  13098 B/op	    134 allocs/op
BenchmarkIPC/Reader/codec=lz4-24           	 134320	     8952 ns/op	   9951 B/op	    117 allocs/op
```

pixelherodev:
I ran `go test -bench . -benchmem -benchtime 30s` before and after to highlight the variance better without running the test 20 times. 

Before:

```
BenchmarkIPC/Writer/codec=plain-24         	6277569	     6240 ns/op	   8160 B/op	     91 allocs/op
BenchmarkIPC/Reader/codec=plain-24         	6876433	     5247 ns/op	   6131 B/op	     83 allocs/op
BenchmarkIPC/Writer/codec=zstd-24          	 114764	   324831 ns/op	2338736 B/op	    150 allocs/op
BenchmarkIPC/Reader/codec=zstd-24          	1310571	    28743 ns/op	  25869 B/op	    194 allocs/op
BenchmarkIPC/Writer/codec=lz4-24           	2930214	    12186 ns/op	  12905 B/op	    134 allocs/op
BenchmarkIPC/Reader/codec=lz4-24           	3445965	    10790 ns/op	  10174 B/op	    124 allocs/op
```


After:

```
BenchmarkIPC/Writer/codec=plain-24         	6302840	     5482 ns/op	   8160 B/op	     91 allocs/op
BenchmarkIPC/Reader/codec=plain-24         	7418485	     4689 ns/op	   5995 B/op	     76 allocs/op
BenchmarkIPC/Writer/codec=zstd-24          	 102850	   346718 ns/op	2338731 B/op	    150 allocs/op
BenchmarkIPC/Reader/codec=zstd-24          	1422286	    25704 ns/op	  25742 B/op	    187 allocs/op
BenchmarkIPC/Writer/codec=lz4-24           	2921695	    11956 ns/op	  12880 B/op	    134 allocs/op
BenchmarkIPC/Reader/codec=lz4-24           	3545550	    10179 ns/op	   9986 B/op	    117 allocs/op
```

Both numbers are a bit worse like this, but more consistent; readers look to be consistently ~5-10% faster.

pixelherodev:
Yeah, the ubuntu one passed on the second run with no changes, and the macos image is failing an entirely different one.

The macOS memory leak looks correct, by the way - in that it's leaking intentionally, because the memory is pooled. When returning the memory to the pool, `buffer.ResizeNoShrink` is used; since shrink=false is specified to resize, Free() is not called.

pixelherodev:
...the test runs runtime.GC() manually twice, but AFAIK there is _no guarantee_ that the GC will run the finalizers on the same pass that frees them.

pixelherodev:
Pushed a small fix to the arrow flight tests; I didn't have `localhost` configured locally, so the test failed with `s.lis == nil; s.lis.Addr()`.

With localhost configured correctly locally, those tests are passing...

zeroshade:
> I'm not even sure how that middleware test can even be flaky tbh.

Yea that's the issue, I've looked at it several times and still haven't been able to figure out what is flakey with it.

zeroshade:
🤞

PR #557, opened 2025-10-30
perf: optimize compute.Take for fewer memory allocations

hamilton-earthscope:
### Rationale for this change

When writing data to partitioned Iceberg tables using `iceberg-go`'s new partitioned write capability, I found that Arrow's `compute.Take` operation was causing the writes to be significantly slower than unpartitioned writes.

Causes: 
- the `VarBinaryImpl` function here in `arrow-go` was pre-allocating the data buffer with only `meanValueLen` (size for just one string/buffer), not the total size needed for all strings. For, my use case (writing 100k rows at a time) this was causing 15x buffer reallocations as the buffer incrementally scales up. For my iceberg table with 20+ string/binary cols of various size this is significant overhead.
- when `VarBinaryImpl` allocates additional space for new items, it would only add the exact amount needed for the new value. This caused O(n) reallocations.

1. Pre-allocate upfront with a better estimate of the necessary buffer size to eliminate repeated reallocations.

    I somewhat arbitrarily chose a cap of 16MB as a guess at what would be effective at reducing the number of allocations but also trying to be cognizant of the library being used in many bespoke scenarios and not wanting to make massive memory spikes. For my use case, I never hit this 16MB threshold and it could be smaller.

    I am curious for your input on whether there should be a cap at all or what a reasonable cap would be.

2. Use exponential growth for additional allocations for O(log n) total reallocations.

### Are these changes tested?

No unit tests.

However, with a dedicated reproducing script that mimics my use case with different write sizes (on a macbook pro m3 max + 64GB ram) average of 3 `table.Append()` calls:

| Configuration | 100k rows | 500k rows | 1M rows | 2.5M rows | 10M rows |
|--------------|-----------|-----------|---------|-----------|-----------|
| **Before** | 4.1s | 53.3s | didn't try | didn't try | didn't try |
| **Change 1** | 336ms | 2.41s | 8.75s | cancelled after 5mins | didn't try |
| **Change 2** | 227ms | 897ms | 1.72s | 4.10s | 17.1s |

No; just more performant

### DIFF
--- a/arrow/compute/vector_selection_test.go
+++ b/arrow/compute/vector_selection_test.go
@@ -1835,7 +1835,7 @@ func BenchmarkTakeMultiColumn(b *testing.B) {
 	tsBldr.Release()
 
 	// Create record batch
-	batch := array.NewRecord(schema, arrays, numRows)
+	batch := array.NewRecordBatch(schema, arrays, numRows)
 	defer batch.Release()
 
 	// Create indices for partitioning pattern


arrow/compute/internal/kernels/vector_selection.go
```diff
@@ -1486,7 +1486,15 @@ func VarBinaryImpl[OffsetT int32 | int64](ctx *exec.KernelCtx, batch *exec.ExecS
...
 		dataLength := rawOffsets[values.Len] - rawOffsets[0]
 		meanValueLen := float64(dataLength) / float64(values.Len)
-		dataBuilder.reserve(int(meanValueLen))
+		estimatedTotalSize := int(meanValueLen * float64(outputLength))
+
+		// Cap the pre-allocation at a reasonable size
+		const maxPreAlloc = 16777216 // 16 MB
+		if estimatedTotalSize > maxPreAlloc {
+			estimatedTotalSize = maxPreAlloc
+		}
```
zeroshade:
```suggestion
		estimatedTotalSize = min(maxPreAlloc, estimatedTotalSize)
```

PR #558, opened 2025-10-30
Add `TimestampWithOffset` canonical extension type

serramatutu:
# Which issue does this PR close?

Implements the new `arrow.timestamp_with_offset` canonical extension type for arrow-go.

# Rationale for this change

Be compatible with the Arrow columnar spec's `arrow.timestamp_with_offset` canonical extension type.

# What changes are included in this PR?

Adds a `TimestampWithOffset` extension type. This type represents a timestamp column that stores a potentially different timezone offset per value: the timestamp is stored in UTC alongside the original timezone offset in minutes. The offset-in-minutes field can be primitive-, dictionary-, or run-end-encoded.

**This PR has been decoupled from the inner-nullability parity work it was previously based on.** Those comparison / JSON-marshaling changes now live in #918 and can be reviewed and merged independently. This PR is now scoped to just the extension type and touches only:

- `arrow/extensions/timestamp_with_offset.go` (new)
- `arrow/extensions/timestamp_with_offset_test.go` (new)
- `arrow/extensions/extensions.go` (register the canonical type)

To keep this PR self-contained, the JSON round-trip test asserts the values the type guarantees (per-row instant, timezone offset, and validity) rather than `array.RecordEqual`, which would compare inner non-nullable struct child validity bitmaps — the parity concern tracked separately in #918.

# Are these changes tested?

Yes — primitive/dictionary/run-end offset encodings, JSON marshal + round-trip, and IPC round-trip.

# Are there any user-facing changes?

Yes, this is a new canonical extension type.

### DIFF
--- a/arrow/extensions/timestamp_with_offset.go
+++ b/arrow/extensions/timestamp_with_offset.go
@@ -17,7 +17,7 @@
 package extensions
 
 import (
-	_ "bytes"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -35,19 +35,31 @@ type TimestampWithOffsetType struct {
 	arrow.ExtensionBase
 }
 
+func isOffsetTypeOk(offsetType arrow.DataType) bool {
+	switch offsetType := offsetType.(type) {
+	case *arrow.Int16Type:
+		return true
+	case *arrow.DictionaryType:
+		return arrow.IsInteger(offsetType.IndexType.ID()) && arrow.TypeEqual(offsetType.ValueType, arrow.PrimitiveTypes.Int16)
+	default:
+		return false
+	}
+}
+
 // Whether the storageType is compatible with TimestampWithOffset.
 //
-// Returns (time_unit, ok). If ok is false, time unit is garbage.
-func isDataTypeCompatible(storageType arrow.DataType) (arrow.TimeUnit, bool) {
+// Returns (time_unit, offset_type, ok). If ok is false, time_unit and offset_type are garbage.
+func isDataTypeCompatible(storageType arrow.DataType) (arrow.TimeUnit, arrow.DataType, bool) {
 	timeUnit := arrow.Second
+	offsetType := arrow.PrimitiveTypes.Int16
 	switch t := storageType.(type) {
 	case *arrow.StructType:
 		if t.NumFields() != 2 {
-			return timeUnit, false
+			return timeUnit, offsetType, false
 		}
 
-		maybeTimestamp := t.Field(0);
-		maybeOffset := t.Field(1);
+		maybeTimestamp := t.Field(0)
+		maybeOffset := t.Field(1)
 
 		timestampOk := false
 		switch timestampType := maybeTimestamp.Type.(type) {
@@ -59,49 +71,84 @@ func isDataTypeCompatible(storageType arrow.DataType) (arrow.TimeUnit, bool) {
 		default:
 		}
 
+		offsetOk := isOffsetTypeOk(maybeOffset.Type)
+
 		ok := maybeTimestamp.Name == "timestamp" &&
 			timestampOk &&
 			!maybeTimestamp.Nullable &&
 			maybeOffset.Name == "offset_minutes" &&
-			arrow.TypeEqual(maybeOffset.Type, arrow.PrimitiveTypes.Int16) &&
+			offsetOk &&
 			!maybeOffset.Nullable
 
-		return timeUnit, ok
+		return timeUnit, maybeOffset.Type, ok
 	default:
-		return timeUnit, false
+		return timeUnit, offsetType, false
 	}
 }
 
-
 // NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
-// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Int16), where T is any TimeUnit.
-func NewTimestampWithOffsetType(unit arrow.TimeUnit) *TimestampWithOffsetType {
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=O), where T is any TimeUnit and O is a valid offset type.
+//
+// The error will be populated if the data type is not a valid encoding of the offsets field.
+func NewTimestampWithOffsetType(unit arrow.TimeUnit, offsetType arrow.DataType) (*TimestampWithOffsetType, error) {
+	if !isOffsetTypeOk(offsetType) {
+		return nil, errors.New(fmt.Sprintf("Invalid offset type %s", offsetType))
+	}
+
 	return &TimestampWithOffsetType{
 		ExtensionBase: arrow.ExtensionBase{
 			Storage: arrow.StructOf(
 				arrow.Field{
 					Name: "timestamp",
 					Type: &arrow.TimestampType{
-						Unit: unit,
+						Unit:     unit,
 						TimeZone: "UTC",
 					},
 					Nullable: false,
 				},
 				arrow.Field{
-					Name: "offset_minutes",
-					Type: arrow.PrimitiveTypes.Int16,
+					Name:     "offset_minutes",
+					Type:     offsetType,
 					Nullable: false,
 				},
 			),
 		},
-	}
+	}, nil
 }
 
-func (b *TimestampWithOffsetType) ArrayType() reflect.Type { return reflect.TypeOf(TimestampWithOffsetArray{}) }
+
+// NewTimestampWithOffsetTypePrimitiveEncoded creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Int16), where T is any TimeUnit.
+func NewTimestampWithOffsetTypePrimitiveEncoded(unit arrow.TimeUnit) *TimestampWithOffsetType {
+	v, _ := NewTimestampWithOffsetType(unit, arrow.PrimitiveTypes.Int16)
+	// SAFETY: This should never error as Int16 is always a valid offset type
+
+	return v
+}
+
+// NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Dictionary(I, Int16)), where T is any TimeUnit and I is a
+// valid Dictionary index type.
+//
+// The error will be populated if the index is not a valid dictionary-encoding index type.
+func NewTimestampWithOffsetTypeDictionaryEncoded(unit arrow.TimeUnit, index arrow.DataType) (*TimestampWithOffsetType, error) {
+	offsetType := arrow.DictionaryType{
+		IndexType: index,
+		ValueType: arrow.PrimitiveTypes.Int16,
+		Ordered:   false,
+	}
+	return NewTimestampWithOffsetType(unit, &offsetType)
+}
+
+func (b *TimestampWithOffsetType) ArrayType() reflect.Type {
+	return reflect.TypeOf(TimestampWithOffsetArray{})
+}
 
 func (b *TimestampWithOffsetType) ExtensionName() string { return "arrow.timestamp_with_offset" }
 
-func (b *TimestampWithOffsetType) String() string { return fmt.Sprintf("extension<%s>", b.ExtensionName()) }
+func (b *TimestampWithOffsetType) String() string {
+	return fmt.Sprintf("extension<%s>", b.ExtensionName())
+}
 
 func (e *TimestampWithOffsetType) MarshalJSON() ([]byte, error) {
 	return []byte(fmt.Sprintf(`{"name":"%s","metadata":%s}`, e.ExtensionName(), e.Serialize())), nil
@@ -110,23 +157,30 @@ func (e *TimestampWithOffsetType) MarshalJSON() ([]byte, error) {
 func (b *TimestampWithOffsetType) Serialize() string { return "" }
 
 func (b *TimestampWithOffsetType) Deserialize(storageType arrow.DataType, data string) (arrow.ExtensionType, error) {
-	timeUnit, ok := isDataTypeCompatible(storageType)
+	timeUnit, offsetType, ok := isDataTypeCompatible(storageType)
 	if !ok {
 		return nil, fmt.Errorf("invalid storage type for TimestampWithOffsetType: %s", storageType.Name())
 	}
-	return NewTimestampWithOffsetType(timeUnit), nil
+
+	v, _ := NewTimestampWithOffsetType(timeUnit, offsetType)
+	// SAFETY: the offsetType has already been checked by isDataTypeCompatible, so we can ignore the error
+
+	return v, nil
 }
 
 func (b *TimestampWithOffsetType) ExtensionEquals(other arrow.ExtensionType) bool {
 	return b.ExtensionName() == other.ExtensionName()
 }
 
-func (b *TimestampWithOffsetType) TimeUnit() arrow.TimeUnit { 
+func (b *TimestampWithOffsetType) TimeUnit() arrow.TimeUnit {
 	return b.ExtensionBase.Storage.(*arrow.StructType).Fields()[0].Type.(*arrow.TimestampType).TimeUnit()
 }
 
 func (b *TimestampWithOffsetType) NewBuilder(mem memory.Allocator) array.Builder {
-	return NewTimestampWithOffsetBuilder(mem, b.TimeUnit())
+	v, _ := NewTimestampWithOffsetBuilder(mem, b.TimeUnit(), arrow.PrimitiveTypes.Int16)
+	// SAFETY: This will never error as Int16 is always a valid type for the offset field
+
+	return v
 }
 
 // TimestampWithOffsetArray is a simple array of struct
@@ -153,13 +207,13 @@ func (a *TimestampWithOffsetArray) String() string {
 }
 
 func timeFromFieldValues(utcTimestamp arrow.Timestamp, offsetMinutes int16, unit arrow.TimeUnit) time.Time {
-	hours := offsetMinutes / 60;
+	hours := offsetMinutes / 60
 	minutes := offsetMinutes % 60
 	if minutes < 0 {
 		minutes = -minutes
 	}
 
-	loc := time.FixedZone(fmt.Sprintf("UTC%+03d:%02d", hours, minutes), int(offsetMinutes) * 60)
+	loc := time.FixedZone(fmt.Sprintf("UTC%+03d:%02d", hours, minutes), int(offsetMinutes)*60)
 	return utcTimestamp.ToTime(unit).In(loc)
 }
 
@@ -182,11 +236,18 @@ func (a *TimestampWithOffsetArray) rawValueUnsafe(i int) (arrow.Timestamp, int16
 
 	timestampField := structs.Field(0)
 	timestamps := timestampField.(*array.Timestamp)
-	offsets := structs.Field(1).(*array.Int16)
 
 	timeUnit := timestampField.DataType().(*arrow.TimestampType).Unit
-	utcTimestamp := timestamps.Value(i)	
-	offsetMinutes := offsets.Value(i)
+	utcTimestamp := timestamps.Value(i)
+
+	var offsetMinutes int16
+
+	switch offsets := structs.Field(1).(type) {
+	case *array.Int16:
+		offsetMinutes = offsets.Value(i)
+	case *array.Dictionary:
+		offsetMinutes = offsets.Dictionary().(*array.Int16).Value(offsets.GetValueIndex(i))
+	}
 
 	return utcTimestamp, offsetMinutes, timeUnit
 }
@@ -243,18 +304,25 @@ type TimestampWithOffsetBuilder struct {
 	*array.ExtensionBuilder
 
 	// The layout used to parse any timestamps from strings. Defaults to time.RFC3339
-	Layout string
-	unit   arrow.TimeUnit
+	Layout     string
+	unit       arrow.TimeUnit
+	offsetType arrow.DataType
 }
 
 // NewTimestampWithOffsetBuilder creates a new TimestampWithOffsetBuilder, exposing a convenient and efficient interface
 // for writing time.Time values to the underlying storage array.
-func NewTimestampWithOffsetBuilder(mem memory.Allocator, unit arrow.TimeUnit) *TimestampWithOffsetBuilder {
-	return &TimestampWithOffsetBuilder{
-		unit: unit,
-		Layout: time.RFC3339,
-		ExtensionBuilder: array.NewExtensionBuilder(mem, NewTimestampWithOffsetType(unit)),
+func NewTimestampWithOffsetBuilder(mem memory.Allocator, unit arrow.TimeUnit, offsetType arrow.DataType) (*TimestampWithOffsetBuilder, error) {
+	dataType, err := NewTimestampWithOffsetType(unit, offsetType)
+	if err != nil {
+		return nil, err
 	}
+
+	return &TimestampWithOffsetBuilder{
+		unit:             unit,
+		offsetType: 	  offsetType,
+		Layout:           time.RFC3339,
+		ExtensionBuilder: array.NewExtensionBuilder(mem, dataType),
+	}, nil
 }
 
 func (b *TimestampWithOffsetBuilder) Append(v time.Time) {
@@ -263,7 +331,13 @@ func (b *TimestampWithOffsetBuilder) Append(v time.Time) {
 
 	structBuilder.Append(true)
 	structBuilder.FieldBuilder(0).(*array.TimestampBuilder).Append(timestamp)
-	structBuilder.FieldBuilder(1).(*array.Int16Builder).Append(int16(offsetMinutes))
+
+	switch offsets := structBuilder.FieldBuilder(1).(type) {
+	case *array.Int16Builder:
+		offsets.Append(int16(offsetMinutes))
+	case *array.Int16DictionaryBuilder:
+		offsets.Append(int16(offsetMinutes))
+	}
 }
 
 func (b *TimestampWithOffsetBuilder) UnsafeAppend(v time.Time) {
@@ -272,10 +346,16 @@ func (b *TimestampWithOffsetBuilder) UnsafeAppend(v time.Time) {
 
 	structBuilder.Append(true)
 	structBuilder.FieldBuilder(0).(*array.TimestampBuilder).UnsafeAppend(timestamp)
-	structBuilder.FieldBuilder(1).(*array.Int16Builder).UnsafeAppend(int16(offsetMinutes))
+
+	switch offsets := structBuilder.FieldBuilder(1).(type) {
+	case *array.Int16Builder:
+		offsets.UnsafeAppend(int16(offsetMinutes))
+	case *array.Int16DictionaryBuilder:
+		offsets.Append(int16(offsetMinutes))
+	}
 }
 
-// By default, this will try to parse the string using the RFC3339 layout. 
+// By default, this will try to parse the string using the RFC3339 layout.
 //
 // You can change the default layout by using builder.SetLayout()
 func (b *TimestampWithOffsetBuilder) AppendValueFromString(s string) error {
@@ -295,22 +375,36 @@ func (b *TimestampWithOffsetBuilder) AppendValueFromString(s string) error {
 
 func (b *TimestampWithOffsetBuilder) AppendValues(values []time.Time, valids []bool) {
 	structBuilder := b.ExtensionBuilder.Builder.(*array.StructBuilder)
-	timestamps := structBuilder.FieldBuilder(0).(*array.TimestampBuilder);
-	offsets := structBuilder.FieldBuilder(1).(*array.Int16Builder);
+	timestamps := structBuilder.FieldBuilder(0).(*array.TimestampBuilder)
 
 	structBuilder.AppendValues(valids)
+	// SAFETY: by this point we know all buffers have available space given the earlier
+	// call to structBuilder.AppendValues which calls Reserve internally, so it's OK to
+	// call UnsafeAppend on inner builders
 
-	for i, v := range values {
-		timestamp, offsetMinutes := fieldValuesFromTime(v, b.unit)
-
-		// SAFETY: by this point we know all buffers have available space given the earlier
-		// call to structBuilder.AppendValues which calls Reserve internally
-		if valids[i] {
-			timestamps.UnsafeAppend(timestamp)
-			offsets.UnsafeAppend(offsetMinutes)
-		} else {
-			timestamps.UnsafeAppendBoolToBitmap(false)
-			offsets.UnsafeAppendBoolToBitmap(false)
+	switch offsets := structBuilder.FieldBuilder(1).(type) {
+	case *array.Int16Builder:
+		for i, v := range values {
+			timestamp, offsetMinutes := fieldValuesFromTime(v, b.unit)
+			if valids[i] {
+				timestamps.UnsafeAppend(timestamp)
+				offsets.UnsafeAppend(offsetMinutes)
+			} else {
+				timestamps.UnsafeAppendBoolToBitmap(false)
+				offsets.UnsafeAppendBoolToBitmap(false)
+			}
+		}
+	case *array.Int16DictionaryBuilder:
+		for i, v := range values {
+			timestamp, offsetMinutes := fieldValuesFromTime(v, b.unit)
+			if valids[i] {
+				timestamps.UnsafeAppend(timestamp)
+				// TODO: I was here, this needs to be equivalent to UnsafeAppend
+				offsets.Append(offsetMinutes)
+			} else {
+				timestamps.UnsafeAppendBoolToBitmap(false)
+				offsets.UnsafeAppendBoolToBitmap(false)
+			}
 		}
 	}
 }
@@ -350,5 +444,5 @@ var (
 	_ arrow.ExtensionType          = (*TimestampWithOffsetType)(nil)
 	_ array.CustomExtensionBuilder = (*TimestampWithOffsetType)(nil)
 	_ array.ExtensionArray         = (*TimestampWithOffsetArray)(nil)
-	_ array.Builder 	       = (*TimestampWithOffsetBuilder)(nil)
+	_ array.Builder                = (*TimestampWithOffsetBuilder)(nil)
 )
--- a/arrow/extensions/timestamp_with_offset.go
+++ b/arrow/extensions/timestamp_with_offset.go
@@ -285,13 +285,70 @@ func (a *TimestampWithOffsetArray) Value(i int) time.Time {
 	return timeFromFieldValues(utcTimestamp, offsetMinutes, timeUnit)
 }
 
+// Iterates over the array and calls the callback with the timestamp at each position. If it's null,
+// the timestamp will be nil.
+//
+// This will iterate using the fastest method given the underlying storage array
+func (a* TimestampWithOffsetArray) iterValues(callback func(i int, utcTimestamp *time.Time)) {
+	structs := a.Storage().(*array.Struct)
+	offsets := structs.Field(1)
+	if reeOffsets, isRee := offsets.(*array.RunEndEncoded); isRee {
+		timestampField := structs.Field(0)
+		timeUnit := timestampField.DataType().(*arrow.TimestampType).Unit
+		timestamps := timestampField.(*array.Timestamp)
+
+		offsetValues := reeOffsets.Values().(*array.Int16)
+		offsetPhysicalIdx := 0
+
+		var getRunEnd (func(int) int)
+		switch arr := reeOffsets.RunEndsArr().(type) {
+		case *array.Int16:
+			getRunEnd = func(idx int) int { return int(arr.Value(idx)) }
+		case *array.Int32:
+			getRunEnd = func(idx int) int { return int(arr.Value(idx)) }
+		case *array.Int64:
+			getRunEnd = func(idx int) int { return int(arr.Value(idx)) }
+		}
+
+		for i := 0; i < a.Len(); i++ {
+			if i >= getRunEnd(offsetPhysicalIdx) {
+				offsetPhysicalIdx += 1
+			}
+
+			timestamp := (*time.Time)(nil)
+			if a.IsValid(i) {
+				utcTimestamp := timestamps.Value(i)
+				offsetMinutes := offsetValues.Value(offsetPhysicalIdx)
+				v := timeFromFieldValues(utcTimestamp, offsetMinutes, timeUnit)
+				timestamp = &v
+			} 
+
+			callback(i, timestamp)
+		}
+	} else {
+		for i := 0; i < a.Len(); i++ {
+			timestamp := (*time.Time)(nil)
+			if a.IsValid(i) {
+				utcTimestamp, offsetMinutes, timeUnit := a.rawValueUnsafe(i)
+				v := timeFromFieldValues(utcTimestamp, offsetMinutes, timeUnit)
+				timestamp = &v
+			} 
+
+			callback(i, timestamp)
+		}
+	}
+}
+
+
 func (a *TimestampWithOffsetArray) Values() []time.Time {
 	values := make([]time.Time, a.Len())
-	// TODO: optimize for run-end encoding
-	for i := range a.Len() {
-		val := a.Value(i)
-		values[i] = val
-	}
+	a.iterValues(func(i int, timestamp *time.Time) {
+		if timestamp == nil {
+			values[i] = time.Unix(0, 0)
+		} else {
+			values[i] = *timestamp
+		}
+	})
 	return values
 }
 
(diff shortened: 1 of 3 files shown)


arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+}
+
+// Whether the storageType is compatible with TimestampWithOffset.
+//
+// Returns (time_unit, offset_type, ok). If ok is false, time_unit and offset_type are garbage.
+func isDataTypeCompatible(storageType arrow.DataType) (arrow.TimeUnit, arrow.DataType, bool) {
+	timeUnit := arrow.Second
+	offsetType := arrow.PrimitiveTypes.Int16
+	switch t := storageType.(type) {
+	case *arrow.StructType:
```
zeroshade: This is a perfect case that would make sense to use named returns:


```suggestion
func isDataTypeCompatible(storageType arrow.DataType) (unit arrow.TimeUnit, offsetType arrow.DataType, ok bool) {
    st, compat := storageType.(*arrow.StructType)
    if !compat || st.NumFields() != 2 {
        return
    }                    
```

etc...

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+
+		timestampOk := false
+		switch timestampType := maybeTimestamp.Type.(type) {
+		case *arrow.TimestampType:
+			if timestampType.TimeZone == "UTC" {
+				timestampOk = true
+				timeUnit = timestampType.TimeUnit()
+			}
+		default:
+		}
```
zeroshade: We can simplify this with something like:

```go
    if ts, compat := t.Field(0).Type.(*arrow.TimestampType); compat && ts.TimeZone == "UTC" {        
        timeUnit = ts.TimeUnit()
    } else {
        return
    }

    maybeOffset := t.Field(1)
    offsetType = maybeOffset.Type
    ok = t.Field(0).Name == "timestamp" && !t.Field(0).Nullable &&
        maybeOffset.Name == "offset_minutes" && isOffsetTypeOk(offsetType) &&
        !maybeOffset.Nullable
    return
```

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+				},
+			),
+		},
+	}, nil
+}
+
+
+// NewTimestampWithOffsetTypePrimitiveEncoded creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Int16), where T is any TimeUnit.
+func NewTimestampWithOffsetTypePrimitiveEncoded(unit arrow.TimeUnit) *TimestampWithOffsetType {
```
zeroshade: my personal preference for naming would be the inverse of this. The default `NewTimestampWithOffsetType` should be this one, while the version that takes a custom offsetType should be `NewTimestampWithOffsetTypeCustomOffset` or something to that effect. The simplest/most common usage should have the shorter name.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+
+	return v
+}
+
+// NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Dictionary(I, Int16)), where T is any TimeUnit and I is a
+// valid Dictionary index type.
+//
+// The error will be populated if the index is not a valid dictionary-encoding index type.
+func NewTimestampWithOffsetTypeDictionaryEncoded(unit arrow.TimeUnit, index arrow.DataType) (*TimestampWithOffsetType, error) {
```
zeroshade: Another idea that might be interesting would be to use generics to avoid the need for the error output.

For example:

```go
func NewTimestampWithOffsetTypeDictEncoded[T arrow.IntType | arrow.UintType](unit arrow.TimeUnit) *TimestampWithOffsetType {
    offsetType := &arrow.DictionaryType{
        IndexType: arrow.GetDataType[T](),
        ValueType: arrow.PrimitiveTypes.Int16,
    }

    ...
```

This would be used like `NewTimestampWithOffsetTypeDictEncoded[int8](arrow.Second)`

or maybe:

```go
type IntegralType interface {
    *arrow.Int8Type | *arrow.Int16Type | *arrow.Int32Type | *arrow.Int64Type |
    *arrow.Uint8Type | *arrow.Uint16Type | *arrow.Uint32Type | *arrow.Uint64Type
}

func NewTimestampWithOffsetTypeDictEncoded[T IntegralType](unit arrow.TimeUnit, index T) *TimestampWithOffsetType {
    offsetType := &arrow.DictionaryType{
        IndexType: index,
        ValueType: arrow.PrimitiveTypes.Int16,
    }

    ...
```

This could be used like `NewTimestampWithOffsetTypeDictEncoded(arrow.Second, arrow.PrimitiveTypes.Int8)`

Leveraging the generics might make the usage simpler potentially by compile time enforcement of the proper types. This is just an idea, curious what you think

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+	}
+	return NewTimestampWithOffsetType(unit, &offsetType)
+}
+
+// NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=RunEndEncoded(E, Int16)), where T is any TimeUnit and E is a
+// valid run-ends type.
+//
+// The error will be populated if runEnds is not a valid run-end encoding run-ends type.
+func NewTimestampWithOffsetTypeRunEndEncoded(unit arrow.TimeUnit, runEnds arrow.DataType) (*TimestampWithOffsetType, error) {
```
zeroshade: See the previous comment about potentially leveraging generics for simplification

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+func (b *TimestampWithOffsetType) Deserialize(storageType arrow.DataType, data string) (arrow.ExtensionType, error) {
+	timeUnit, offsetType, ok := isDataTypeCompatible(storageType)
+	if !ok {
+		return nil, fmt.Errorf("invalid storage type for TimestampWithOffsetType: %s", storageType.Name())
+	}
+
+	v, _ := NewTimestampWithOffsetType(timeUnit, offsetType)
+	// SAFETY: the offsetType has already been checked by isDataTypeCompatible, so we can ignore the error
+
+	return v, nil
```
zeroshade: at this rate, just do `return NewTimestampWithOffsetType(timeUnit, offsetType)` since the function signature has an error, there's no reason to explicitly ignore it even if we know it'll always be valid.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+}
+
+func (b *TimestampWithOffsetType) ExtensionEquals(other arrow.ExtensionType) bool {
+	return b.ExtensionName() == other.ExtensionName()
+}
```
zeroshade, line 200: are we still equal even if the time unit and structure of the offset minutes is different?

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+
+func (b *TimestampWithOffsetType) TimeUnit() arrow.TimeUnit {
+	return b.ExtensionBase.Storage.(*arrow.StructType).Fields()[0].Type.(*arrow.TimestampType).TimeUnit()
```
zeroshade: use `.Field(0)` instead of `.Fields()[0]` ?

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+
+	loc := time.FixedZone(fmt.Sprintf("UTC%+03d:%02d", hours, minutes), int(offsetMinutes)*60)
+	return utcTimestamp.ToTime(unit).In(loc)
+}
+
+func fieldValuesFromTime(t time.Time, unit arrow.TimeUnit) (arrow.Timestamp, int16) {
+	// naive "bitwise" conversion to UTC, keeping the underlying date the same
+	utc := t.UTC()
+	naiveUtc := time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), time.UTC)
+	offsetMinutes := int16(naiveUtc.Sub(t).Minutes())
```
zeroshade: This would probably be simpler?

```go
_, offsetSeconds := t.Zone()
offsetMinutes = offsetSeconds / 60

ts, _ := arrow.TimestampFromTime(t.UTC(), unit)
return ts, offsetMinutes
```

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+	var offsetMinutes int16
+
+	switch offsets := structs.Field(1).(type) {
+	case *array.Int16:
+		offsetMinutes = offsets.Value(i)
+	case *array.Dictionary:
+		offsetMinutes = offsets.Dictionary().(*array.Int16).Value(offsets.GetValueIndex(i))
+	case *array.RunEndEncoded:
+		offsetMinutes = offsets.Values().(*array.Int16).Value(offsets.GetPhysicalIndex(i))
+	}
```
zeroshade, line 278: as a follow-up we might be able to improve this by initializing a `arrow.TypedArray[int16]` in the `TimestampWithOffsetArray`. That way this can just be simplified to `offsetMinutes := offsets.Value(i)`

For dictionaries we have `NewDictWrapper[int16](offsets)` which will wrap the `*arrow.Dictionary` and give the interface we want, but there isn't a generic wrapper for RunEndEncoded arrays yet that does the same. We'd need to create that before using this simplification, hence why I think this would be a good follow-up rather than doing this now.

Aside from simplifying this function, it would also allow us to avoid this switch on every index by just using an interface that provides `Value(int) int16`.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+	utcTimestamp, offsetMinutes, timeUnit := a.rawValueUnsafe(i)
+	return timeFromFieldValues(utcTimestamp, offsetMinutes, timeUnit)
+}
+
+// Iterates over the array and calls the callback with the timestamp at each position. If it's null,
+// the timestamp will be nil.
+//
+// This will iterate using the fastest method given the underlying storage array
+func (a* TimestampWithOffsetArray) iterValues(callback func(i int, utcTimestamp *time.Time)) {
```
zeroshade: instead, can we just use `iter.Seq`?

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+	values := make([]time.Time, a.Len())
+	a.iterValues(func(i int, timestamp *time.Time) {
+		if timestamp == nil {
+			values[i] = time.Unix(0, 0)
+		} else {
+			values[i] = *timestamp
+		}
+	})
+	return values
+}
```
zeroshade, line 366: if we change `iterValues` to return an `iter.Seq[time.Time]` then this becomes just `return slices.Collect(a.iterValues())`

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+		return a.Value(i).String()
+	}
+}
+
+func (a *TimestampWithOffsetArray) MarshalJSON() ([]byte, error) {
+	values := make([]interface{}, a.Len())
+	a.iterValues(func(i int, timestamp *time.Time) {
+		values[i] = timestamp
+	})
+	return json.Marshal(values)
```
zeroshade, line 388:
```suggestion
    return json.Marshal(a.Values())
```

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+	switch offsets := structBuilder.FieldBuilder(1).(type) {
+	case *array.Int16Builder:
+		for i, v := range values {
+			timestamp, offsetMinutes := fieldValuesFromTime(v, b.unit)
+			if valids[i] {
+				timestamps.UnsafeAppend(timestamp)
+				offsets.UnsafeAppend(offsetMinutes)
+			} else {
+				timestamps.UnsafeAppendBoolToBitmap(false)
+				offsets.UnsafeAppendBoolToBitmap(false)
```
zeroshade: these are non-nullable I thought? We should be pushing default 0,0 in the case where `!valids[i]`

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+		}
+	case *array.Int16DictionaryBuilder:
+		for i, v := range values {
+			timestamp, offsetMinutes := fieldValuesFromTime(v, b.unit)
+			if valids[i] {
+				timestamps.UnsafeAppend(timestamp)
+				offsets.Append(offsetMinutes)
+			} else {
+				timestamps.UnsafeAppendBoolToBitmap(false)
+				offsets.UnsafeAppendBoolToBitmap(false)
```
zeroshade: same as above, these are non-nullable according to the spec.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,568 @@
...
+				if offsetMinutes != b.lastOffset {
+					offsets.Append(1)
+					offsetValuesBuilder.Append(offsetMinutes16)
+				} else {
+					offsets.ContinueRun(1)
+				}
+				b.lastOffset = offsetMinutes16
+			} else {
+				timestamps.UnsafeAppendBoolToBitmap(false)
+				offsets.UnsafeAppendBoolToBitmap(false)
```
zeroshade: same as above. these are non-nullable according to the spec

arrow/extensions/timestamp_with_offset_test.go
```diff
@@ -0,0 +1,354 @@
...
+)
+
+var testTimeUnit = arrow.Microsecond
+
+var testDate1 = time.Date(2025, 01, 01, 00, 00, 00, 00, time.FixedZone("UTC+00:00", 0))
+
+var testZone1 = time.FixedZone("UTC-08:00", -8*60*60)
+var testDate2 = testDate1.In(testZone1)
+
+var testZone2 = time.FixedZone("UTC+06:00", +6*60*60)
```
zeroshade: make sure there's a test with a zone that isn't evenly divisible by 60, perhaps a zone with an offset of `5:30` or something to that effect.

Also, make sure we're testing a zone with an offset greater than 10 hours

zeroshade (changes requested):

arrow/extensions/timestamp_with_offset_test.go
serramatutu: Fixed here: https://github.com/apache/arrow-go/pull/558/changes/cee17a3e4e4677124c07cd71359f724f4fce5d50

arrow/extensions/timestamp_with_offset.go
serramatutu: I ended up not being able to do this because I changed `iterValues()` to return `time.Time` instead of `*time.Time`. So we need a check if `time.Unix() == 0`, meaning it should actually be serialized as `null` instead of `1970-01-01`.

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: This is not semantically equal (I tried, it messes up the values, but only sometimes for some reason?).

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/bd12287760927080eda309594b36dc9e2ec813ff

arrow/extensions/timestamp_with_offset.go
serramatutu: Oops, yea this was reminiscent of the old implementation before we all circled around to this being non-nullable again... Good catch :)

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/e36a78ffad1d8c650a75e9adf12312caec99a5c6

arrow/extensions/timestamp_with_offset.go
serramatutu: As we discussed in Slack, I found an issue when roundtripping to JSON. 

It looks like replacing these with `UnsafeAppend` results in an array with `nulls=0` in the inner (non-nullable) field.
If I write the same array to JSON and parse it back with `arrow.RecordFromJSON()`, the inner array has `nulls=1`, even if the field is not nullable.

This leads to `arrow.RecordEqual()` returning `false` due to `original.NullN() != roundtrip.NullN()`.

There is an ongoing thread on the mainling list about how to resolve this, so I'll hold off on fixing it for now.

arrow/extensions/timestamp_with_offset.go
serramatutu: I like it. I prefer compile-time errors over runtime errors :)

serramatutu: Commit: https://github.com/apache/arrow-go/pull/558/changes/cbf8f654f5cae86d1fe7cf382cf1b0a992f6cd7a

Ok, I fixed this but I'm still not sure I like the look of the API. Turns out `arrow.PrimitiveTypes.*` are all `DataTypes` and the compiler complains `DataType` is too broad to fit in the generic constraints.

You need to manually use the type struct like so:

`NewTimestampWithOffsetTypeDictEncoded(arrow.Second, &arrow.Int8Type{})`

The other disadvantage is that I can't just iterate over a list of `DataType` like I was doing in tests before, you need to specify the specialization of the function at compile-time. Which might not be a problem for users, but wanted to point that out...

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/2762d64b50221ac82ed0a22231eb8478427823ed

serramatutu: Unrelated, but while fixing this I found a bug on REE and dict encoding builders, where they were inheriting the `UnsafeAppendBoolToBitmap()` from the `Builder` interface. The default implementation does not bump the length of the inner `idxBuilder` or `runEndsBuilder`, which causes it to be in an inconsistent state where the parent has greater length than the index/run-ends builder. This causes a panic when instancing a record batch.

I added a fix here, but I should probably cherry pick it and submit it in a separate PR? Also not sure if my change is good enough since it's very easy to forget to override `UnsafeAppendBoolToBitmap()`.

 https://github.com/apache/arrow-go/pull/558/changes/eff57e650e1fd2ea25193f0f6c9acc37e7e31c1f

serramatutu: @felipecrv Tells me `UnsafeAppendBoolToBitmap()` should not be callable on dictionary/REE. So I should probably revert the commit I mentioned above and just call `AppendNulls()`.

Either way, I think it would be nice to either:
- immediately panic when this gets called on Dict/REE
- just make it call `AppendNulls()` and ignore the fact there is no bitmap

otherwise people will get very strange errors when creating a `RecordBatch` (like I did).

arrow/array/encoded.go
```diff
@@ -386,6 +386,15 @@ func (b *RunEndEncodedBuilder) ContinueRun(n uint64) {
 	b.addLength(n)
 }
 
+func (b *RunEndEncodedBuilder) UnsafeAppendBoolToBitmap(v bool) {
+	if !v {
+		b.finishRun()
+	}
```
zeroshade: this seems incorrect, it could be a run of nulls, couldn't it? And wouldn't the run itself be handled by the value append? I don't think we need to call `b.finishRun` here

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+type TimestampWithOffsetType struct {
+	arrow.ExtensionBase
+}
+
+func isOffsetTypeOk(offsetType arrow.DataType) bool {
+	switch offsetType := offsetType.(type) {
+	case *arrow.Int16Type:
+		return true
+	case *arrow.DictionaryType:
+		return arrow.IsInteger(offsetType.IndexType.ID()) && arrow.TypeEqual(offsetType.ValueType, arrow.PrimitiveTypes.Int16)
```
zeroshade: `arrow.IsInteger` is unnecessary here, by definition only integer types are allowed for the IndexTypes of a dictionary type. Anything else is invalid anyways, so there's no need to check that.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+	}
+	v, _ := NewTimestampWithOffsetTypeCustomOffset(unit, &offsetType)
+	// SAFETY: This should never error as DictIndexType is always a valid index type
+
+	return v
+}
+
+type TimestampWithOffsetRunEndsType interface {
+	*arrow.Int8Type | *arrow.Int16Type | *arrow.Int32Type | *arrow.Int64Type |
+		*arrow.Uint8Type | *arrow.Uint16Type | *arrow.Uint32Type | *arrow.Uint64Type
```
zeroshade: this and the `DictIndexType` are both identical, instead of duplicating this we should either create a single type constraint or embed one in the other:

```go
type TimestampWithOffsetRunEndsType interface {
     DictIndexType
}
```

My personal preference would be to have a single type though, but I'm not averse to having the two separate ones if necessary.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+// NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=RunEndEncoded(E, Int16)), where T is any TimeUnit and E is a
+// valid run-ends type.
+//
+// The error will be populated if runEnds is not a valid run-end encoding run-ends type.
```
zeroshade: Update the comment to remove this since the generics will affirm that we can't have this error

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+type DictIndexType interface {
+	*arrow.Int8Type | *arrow.Int16Type | *arrow.Int32Type | *arrow.Int64Type |
+		*arrow.Uint8Type | *arrow.Uint16Type | *arrow.Uint32Type | *arrow.Uint64Type
+}
+
+// NewTimestampWithOffsetType creates a new TimestampWithOffsetType with the underlying storage type set correctly to
+// Struct(timestamp=Timestamp(T, "UTC"), offset_minutes=Dictionary(I, Int16)), where T is any TimeUnit and I is a
+// valid Dictionary index type.
+//
+// The error will be populated if the index is not a valid dictionary-encoding index type.
```
zeroshade: remove this line from the comment since the generics prevent this from happening

arrow/extensions/timestamp_with_offset.go
zeroshade: I agree with this, we should update those builders to error / panic if called and add docstring comments to explain that.

arrow/extensions/timestamp_with_offset.go
zeroshade: can you give me an example where it messes up the values? 

Just to verify, the goal here is to take `t` and provide the same *instant of time* in UTC while figuring out what the offset was in minutes for the timezone, yes? i.e. given `2009-11-10 15:00:00 -0800 PST` this should return `2009-11-10 23:00:00 +0000 UTC` and `480`, correct?

arrow/extensions/timestamp_with_offset.go
zeroshade: Using `0` as a sentinel for `null` seems incorrect as that means you can't actually encode a non-null value of 1970-01-01 and get the correct output. Since we need to be able to represent the difference between a *value* of `0` and a `null`, then either `iterValues()` needs to return a sequence of `*time.Time` or it needs to be a sequence of `iter.Seq2[time.Time, bool]` where the bool indicates validity (if the bool is false, then it's null).

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+					ts = v
+				}
+
+				if !yield(ts) {
+					return
+				}
+			}
+		} else {
+			for i := 0; i < a.Len(); i++ {
+				ts := time.Unix(0, 0)
```
zeroshade: this should be equivalent to just doing `var ts time.Time`, except faster since it avoids the extra processing of calling the `Unix` function.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,577 @@
...
+	if a.IsNull(i) {
+		return time.Unix(0, 0)
+	}
+	utcTimestamp, offsetMinutes, timeUnit := a.rawValueUnsafe(i)
+	return timeFromFieldValues(utcTimestamp, offsetMinutes, timeUnit)
+}
+
+// Iterates over the array returning the timestamp at each position.
+//
+// If the timestamp is null, the returned time will be the unix epoch.
```
zeroshade: what happens if the value is not-null but is the unix epoch? this iteration would be unable to represent that case

arrow/extensions/timestamp_with_offset_test.go
```diff
@@ -0,0 +1,358 @@
...
+	// assert.False(t, extensions.NewTimestampWithOffsetType(arrow.Nanosecond).ExtensionEquals(extensions.NewTimestampWithOffsetType(arrow.Microsecond)))
+	// assert.False(t, extensions.NewTimestampWithOffsetType(arrow.Nanosecond).ExtensionEquals(extensions.NewTimestampWithOffsetType(arrow.Second)))
+	// assert.False(t, extensions.NewTimestampWithOffsetType(arrow.Microsecond).ExtensionEquals(extensions.NewTimestampWithOffsetType(arrow.Second)))
+	//
+	// // Different underlying storage type is not equal
+	// assert.False(t, extensions.NewTimestampWithOffsetType(arrow.Microsecond).ExtensionEquals(extensions.NewTimestampWithOffsetTypeDictionaryEncoded(arrow.Microsecond, &arrow.Int16Type{})))
+	// assert.False(t, extensions.NewTimestampWithOffsetType(arrow.Microsecond).ExtensionEquals(extensions.NewTimestampWithOffsetTypeRunEndEncoded(arrow.Microsecond, &arrow.Int16Type{})))
+	// assert.False(t, extensions.NewTimestampWithOffsetTypeDictionaryEncoded(arrow.Microsecond, &arrow.Int16Type{}).ExtensionEquals(extensions.NewTimestampWithOffsetTypeRunEndEncoded(arrow.Microsecond, &arrow.Int16Type{})))
+	//
+	// // Dict-encoding key type is not equal
```
zeroshade: why are these all commented out?

zeroshade (changes requested):

arrow/extensions/timestamp_with_offset_test.go
serramatutu: I was dumb. Fixed https://github.com/apache/arrow-go/pull/558/changes/02498067c12dfd04265234e715f8c8da34d22fbc

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/76e3ace8d45e9162706282203eff81766db8dd97

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/76e3ace8d45e9162706282203eff81766db8dd97

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/231cb1ec4211cf65d8e96fa3c5beec1a94bfc940

arrow/extensions/timestamp_with_offset.go
serramatutu: Fixed: https://github.com/apache/arrow-go/pull/558/changes/f42417b775645a0fd4b4ec2502acb1a327cc9222

arrow/extensions/timestamp_with_offset.go
serramatutu: Did this in a separate PR: https://github.com/apache/arrow-go/pull/758

arrow/extensions/timestamp_with_offset.go
serramatutu: Actually I had to revert. Spec says those types are different:

Dict index is either signed or unsigned int of 8-64 bits
> The index type of a Dictionary type can only be an integer type, preferably signed, with width 8 to 64 bits.

Run ends is a signed int of 16-64 bits
> The run end type of a Run-End Encoded type can only be a signed integer type with width 16 to 64 bits.

https://arrow.apache.org/docs/format/Columnar.html#data-types

arrow/extensions/timestamp_with_offset.go
serramatutu: https://github.com/apache/arrow-go/pull/558/changes/d955cdb46fe2d17c547e51f9f5e123b2edcb7d76

Fix here.

arrow/extensions/timestamp_with_offset.go
serramatutu: I did change this but now the IPC roundtrip tests are failing because the internal count of nulls is different between the thing before IPC and after IPC.

To make them pass we'll need to relax the `arrow.RecordEqual()` to ignore `NullN()` when the field is not nullable. Alternatively we can change it so that an array builder sets `nulls = 0` if the array is not nullable, regardless of what has been set before in calls to `UnsafeAppendBoolToBitmap()` etc

Related: https://lists.apache.org/thread/7gbqjwykh1ob3xbvwph3ljsdl5c7kxpd

arrow/extensions/timestamp_with_offset.go
serramatutu: Actually I'm a dumdum. Fixed: https://github.com/apache/arrow-go/pull/558/changes/3a9a8f2af3bb37058418a2ebb94ae0bdd4e7ed5e

arrow/array/compare.go
```diff
@@ -35,9 +39,17 @@ func RecordEqual(left, right arrow.RecordBatch) bool {
 	}
 
 	for i := range left.Columns() {
+		lf := left.Schema().Field(i)
+		rf := left.Schema().Field(i)
```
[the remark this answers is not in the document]
zeroshade: Addressed in the current head: `recordEqual` now takes `rf := right.Schema().Field(i)` and gates on `schemaFieldEqual(lf, rf)` (name + nullability), so left/right field differences are detected. Resolving.

arrow/array/compare.go
```diff
@@ -47,16 +59,26 @@ func RecordEqual(left, right arrow.RecordBatch) bool {
...
 		return false
 	case left.NumRows() != right.NumRows():
 		return false
 	}
 
-	opt := newEqualOption(opts...)
-
 	for i := range left.Columns() {
+		lf := left.Schema().Field(i)
+		rf := left.Schema().Field(i)
```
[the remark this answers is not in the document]
zeroshade: Addressed: `recordApproxEqual` now reads `rf := right.Schema().Field(i)` and checks `schemaFieldEqual(lf, rf)` before comparing, so nullability/name differences are caught. Resolving.

arrow/array/compare.go
```diff
@@ -160,21 +194,29 @@ func TableEqual(left, right arrow.Table) bool {
 	}
 
 	for i := 0; int64(i) < left.NumCols(); i++ {
-		lc := left.Column(i)
-		rc := right.Column(i)
-		if !lc.Field().Equal(rc.Field()) {
+		lf := left.Schema().Field(i)
+		rf := left.Schema().Field(i)
```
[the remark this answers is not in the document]
zeroshade: Addressed: `tableEqual` now uses `rf := right.Schema().Field(i)` with a `schemaFieldEqual(lf, rf)` guard. Resolving.

arrow/array/compare.go
```diff
@@ -183,27 +225,35 @@ func TableApproxEqual(left, right arrow.Table, opts ...EqualOption) bool {
...
```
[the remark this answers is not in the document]
zeroshade: Addressed: `tableApproxEqual` now uses `rf := right.Schema().Field(i)` with the same `schemaFieldEqual(lf, rf)` guard. Resolving.

arrow/array/record_test.go
```diff
@@ -511,14 +516,20 @@ func TestRecordBuilder(t *testing.T) {
 		}
 	}
 
+	err := b.UnmarshalJSON([]byte(`{"f1-i32": 6, "f2-f64-notnull": 6.6, "map": [{"key": "4": "value": "d"}]}`))
```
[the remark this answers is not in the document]
zeroshade: Fixed: the literal now reads `{"key": "4", "value": "d"}` (valid JSON) instead of the malformed colon, so the test exercises the intended path. Resolving.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,535 @@
...
+			timestamps.UnsafeAppend(timestamp)
+			offsetMinutes16 := int16(offsetMinutes)
+			// If value at i is null, simply continue the run to maximize compression
+			if valids[i] && offsetMinutes != b.lastOffset {
+				offsets.Append(1)
+				offsetValuesBuilder.Append(offsetMinutes16)
+			} else {
+				offsets.ContinueRun(1)
+			}
+			b.lastOffset = offsetMinutes16
```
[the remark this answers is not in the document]
zeroshade: Fixed in the builder rework. Both `Append` and `AppendValues` now start a run with `offsets.Append(1)` whenever none exists yet (`b.lastOffset == noLastOffset`), so `ContinueRun(1)` is never called without a matching value entry — including leading-null rows. `lastOffset` is only updated on the branch that starts a new run, so a null row can't split a run or shift a boundary based on a masked value, and `AppendNull`/`AppendNulls`/`NewArray` reset the tracker. There's now an inline comment documenting exactly this case. Resolving.

serramatutu (commented):
@zeroshade some comments:

Can you think of a way to decouple these 2 things (timestamp with offset and the inner nullability fixes)? So we can merge the extension type without having to wait on all the fixes for that and do that more carefully?

I'm sorry I left the branch based off of an old version of the inner nullability parity fixes. There are some differences with what I was doing in the new version of the [JSON PR](https://github.com/apache/arrow-go/pull/833):
- I ended up dropping the `equalOpts` because this is not what C++ does. It looks like they just validate the schemas on `Validate()` instead of making `Equal()` more lenient. I had not fixed that in that PR because there are some issues with the CSV reader (it calls `Validate()`) and it was becoming somewhat of a sprawl to fix it all. There are lots of behavior in `arrow-go` that never checks for the schema of structs.
- I did not do the `GetOneForMarshalNullable()` and instead I changed the struct/record builder to check for `IsValid()` before calling `GetOneForMarshal()`.

zeroshade:
@serramatutu updated this PR to be *only* the timestamp with offset extension type

zeroshade:
@serramatutu filed #918 to split out the nullability stuff

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,592 @@
...
+func timeFromFieldValues(utcTimestamp arrow.Timestamp, offsetMinutes int16, unit arrow.TimeUnit) time.Time {
+	hours := offsetMinutes / 60
+	minutes := offsetMinutes % 60
+	if minutes < 0 {
+		minutes = -minutes
+	}
+
+	loc := time.FixedZone(fmt.Sprintf("UTC%+03d:%02d", hours, minutes), int(offsetMinutes)*60)
+	return utcTimestamp.ToTime(unit).In(loc)
+}
```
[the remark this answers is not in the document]
zeroshade: Fixed in e7ef6d8d: the zone-name sign is now derived from the whole offset, so a sub-hour negative offset keeps its sign (e.g. -30 minutes now formats as "UTC-00:30" instead of "UTC+00:30"). Added TestTimestampWithOffsetSubHourNegativeZone as a regression test.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,592 @@
...
+	for ts, valid := range a.iterValues() {
+		if !valid {
+			values[i] = nil
+		} else {
+			values[i] = &ts
+		}
+		i += 1
+	}
+	return json.Marshal(values)
+}
```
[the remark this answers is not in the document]
zeroshade: Addressed in e7ef6d8d: MarshalJSON now stores the timestamp by value instead of the address of the range variable. (For the record, under this module's Go 1.25 per-iteration loop semantics it was not actually an aliasing bug — the existing exact-JSON roundtrip test asserts distinct per-row values — but marshaling by value is clearer.)

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,592 @@
...
+
+func (a *TimestampWithOffsetArray) Values() []time.Time {
+	return slices.Collect(func(yield func(time.Time) bool) {
+		for time := range a.iterValues() {
+			if !yield(time) {
+				return
+			}
+		}
+	})
+}
```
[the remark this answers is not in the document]
zeroshade: Fixed in e7ef6d8d: renamed the loop variable to `t` so it no longer shadows the `time` package.

arrow/extensions/timestamp_with_offset_test.go
```diff
@@ -0,0 +1,591 @@
...
+,{"timestamp_with_offset":"2024-12-31T15:30:00-08:30"}
+]`
+			require.Equal(t, expect, string(json))
+
+			// Record batch roundtrip to JSON should work
+			roundtripped, _, err := array.RecordFromJSON(memory.DefaultAllocator, schema, bytes.NewReader(json))
+			require.NoError(t, err)
+			defer roundtripped.Release()
+			require.Equal(t, schema, roundtripped.Schema())
+			// Avoid array.RecordEqual here: it compares the inner non-nullable
```
serramatutu: Can we add a FIXME here to revisit? Ideally the records SHOULD equal after roundtripping lol

zeroshade: done

arrow/extensions/timestamp_with_offset.go
```diff
@@ -0,0 +1,608 @@
...
+		return true
+	case *arrow.DictionaryType:
+		return arrow.TypeEqual(offsetType.ValueType, arrow.PrimitiveTypes.Int16)
+	case *arrow.RunEndEncodedType:
+		return offsetType.ValidRunEndsType(offsetType.RunEnds()) &&
+			arrow.TypeEqual(offsetType.Encoded(), arrow.PrimitiveTypes.Int16)
+		// FIXME: Technically this should be non-nullable, but a Arrow IPC does not deserialize
+		// ValueNullable properly, so enforcing this here would always fail when reading from an IPC
+		// stream
+		// !offsetType.ValueNullable
```
lidavidm, line 52: This is pretty major right? Can we link this to a tracking issue? (Does integration catch this?)

lidavidm (approved):

zeroshade (approved):

arrow/extensions/timestamp_with_offset.go
zeroshade: This is part of what was pulled out into #918 and friends

PR #561, opened 2025-11-07
doc(parquet): document arrow parquet mappings

### LINKED ISSUE #403: Document on Parquet to Arrow type mapping

Do you have any document on Parquet to Arrow type mapping for `pqarrow` package? Thanks

### Component(s)

Parquet

amoeba:
### Rationale for this change

Documents the Arrow to Parquet type conversions and documents which types have no conversion. Closes #403

I'm sneaking this into the doc.go file for the parquet package so this PR just adds text tehre.

### Are these changes tested?

Visually and partially with some unit tests (not included).

No

### DIFF
--- a/parquet/doc.go
+++ b/parquet/doc.go
@@ -79,7 +79,12 @@
 // # Arrow to Parquet Type Mappings
 //
 // When reading and writing Parquet, the parquet package converts between Arrow
-// and Parquet types in the following manner:
+// and Parquet types in the manner described in the table below.
+//
+// When converting a Parquet type where a large and non-large offset Arrow type
+// would work, the non-large variant is chosen. If the Parquet file is written
+// with `WithStoreSchema`, types will be preserved and dictionaries will be
+// restored when round-tripping.
 //
 //	Arrow Type              Parquet Physical Type     Parquet Logical Type
 //	----------              ---------------------     --------------------


parquet/doc.go
```diff
@@ -75,6 +75,60 @@
...
+//	EXTENSION               (depends on storage)      (may be custom)
+//
+// * FLBA means FixedLenByteArray
+//
+// Unsupported Arrow Types (will return arrow.ErrNotImplemented):
+//
+//	DURATION, INTERVAL_MONTHS, INTERVAL_DAY_TIME, INTERVAL_MONTH_DAY_NANO
+//	SPARSE_UNION, DENSE_UNION
+//	STRING_VIEW, BINARY_VIEW, LIST_VIEW, LARGE_LIST_VIEW
+//	LARGE_LIST, RUN_END_ENCODED
```
zeroshade, line 130: do we really error on LargeList here? I should probably fix that then :(

I need to get around to updating this to properly handle the view types.

zeroshade (approved):
Thanks for this!

parquet/doc.go
amoeba: I think so. This test passes,

```patch
diff --git a/parquet/pqarrow/schema_test.go b/parquet/pqarrow/schema_test.go
index 6f5d14c7..f9817492 100644
--- a/parquet/pqarrow/schema_test.go
+++ b/parquet/pqarrow/schema_test.go
@@ -439,6 +439,7 @@ func TestUnsupportedTypes(t *testing.T) {
 		{typ: &arrow.MonthDayNanoIntervalType{}},
 		{typ: &arrow.DenseUnionType{}},
 		{typ: &arrow.SparseUnionType{}},
+		{typ: arrow.LargeListOf(arrow.PrimitiveTypes.Int32)},
 	}
 	for _, tc := range unsupportedTypes {
 		t.Run(tc.typ.ID().String(), func(t *testing.T) {
```

PR #569, opened 2025-11-10
build: add pqarrow_read_only build tags to avoid building write related code

jacobromero:
### Rationale for this change

The current binary when built is 20mb. We have a use case in which we only need to read data from a parquet file and do not need to write any parquet data out. By Adding build tags to code that is only needed to write we can save ~12mb off the size of our binary compiled with arrow-go.

This PR adds the build tag `!pqarrow_read_only` to write related files in the `pqarrow` package.

### Are these changes tested?

Tested the build locally and against my read only use case.
Added test in `ci/scripts/build.sh`

No, this change is only for consumers who build `arrow-go` but do not need write functionality.


jacobromero:
Hi @zeroshade I think there was a build failure with my original change in `ci/scripts/build.sh` but that should be fixed now

zeroshade:
Looks good! Thanks for this!

PR #573, opened 2025-11-13
perf(compute): optimize Take kernel for list types

hamilton-earthscope:
### Rationale for this change

This PR implements the same allocation improvements as https://github.com/apache/arrow-go/pull/557 but for the Take list kernel.

### What changes are included in this PR?

- Pre-allocate upfront with a better estimate of the necessary buffer size to eliminate repeated reallocations.
- Use exponential growth for additional allocations for O(log n) total reallocations.

### Are these changes tested?

New benchmarks added for take list kernel

### Are there any user-facing changes?

No

### Performance Comparison

on `main`

```
goos: darwin
goarch: arm64
pkg: github.com/apache/arrow-go/v18/arrow/compute
cpu: Apple M3 Max
BenchmarkTakeList/SmallBatch_ShortLists-16         	    2320	     555889 ns/op	   1798923 rows/sec	     3355390 B/op	     396 allocs/op
BenchmarkTakeList/MediumBatch_ShortLists-16        	      31	   37600434 ns/op	    265955 rows/sec	   324238694 B/op	    3214 allocs/op
BenchmarkTakeList/LargeBatch_ShortLists-16         	       2	  653682896 ns/op	     76490 rows/sec	  7877124660 B/op	   15751 allocs/op
BenchmarkTakeList/XLargeBatch_ShortLists-16        	       1	 2019873959 ns/op	     49508 rows/sec	 31380868976 B/op	   31489 allocs/op
BenchmarkTakeList/SmallBatch_MediumLists-16        	     334	    3435205 ns/op	    291104 rows/sec	    43059334 B/op	    1086 allocs/op
BenchmarkTakeList/MediumBatch_MediumLists-16       	       5	  230308533 ns/op	     43420 rows/sec	  4041679715 B/op	   10098 allocs/op
BenchmarkTakeList/LargeBatch_MediumLists-16        	       1	 4600489959 ns/op	     10868 rows/sec	100213267960 B/op	   50328 allocs/op
BenchmarkTakeList/XLargeBatch_MediumLists-16       	       1	16462343792 ns/op	      6074 rows/sec	400427784144 B/op	  101032 allocs/op
BenchmarkTakeList/LargeBatch_ShortLists_Large-16   	       1	 1682293042 ns/op	     29721 rows/sec	 31379855376 B/op	   31422 allocs/op
BenchmarkTakeList/XLargeBatch_MediumLists_Large-16 	       1	33800353917 ns/op	      2959 rows/sec	800433389776 B/op	  102480 allocs/op
[...]

PR #574, opened 2025-11-14
feat(compute): Take kernel for Map type

hamilton-earthscope:
### Rationale for this change

Arrow Go is lacking a Take kernel for Map types which means from `iceberg-go` we cannot write to partitioned Iceberg tables containing columns with Map types.

- Adds a new Take kernel for Map types
  - the same allocation behavior from #557 and #573 is used

### Are these changes tested?

Yes.

* Can now use Take on Arrow schemas with Map columns.
* Can now write to partitioned Iceberg tables using `arrow-go`

### DIFF
--- a/arrow/compute/vector_selection_test.go
+++ b/arrow/compute/vector_selection_test.go
@@ -1264,23 +1264,23 @@ func (tk *TakeKernelMap) TestMapStringInt32() {
 
 	mapJSON := `[
 		{},
-		{"a": 1, "b": 2},
-		{"a": 2, "b": null},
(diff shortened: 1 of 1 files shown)


hamilton-earthscope:
Seems like CI failure was not related to these changes, but a test in arrow/flight?

zeroshade (changes requested):
This looks great, my only suggestion is to add a unit test where some of the elements in the map array have more and less key/value pairs from element to element.

zeroshade (approved):
Thanks!

PR #578, opened 2025-11-21
proof of concept: unify refcounting logic

pixelherodev:
A step towards the addcleanup work: we want to be able to turn refcounting off entirely. Rather than move all ~200 release+retain functions into files with build tags, IMO it makes sense to unify all refcounting logic into memory.Refcount, embed that everywhere else, and then we can just turn _that_ off with switching to AddCleanup.

This is a proof of concept using that with just ipc/Message to obtain feedback before finishing the work.

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @pixelherodev** · by `@zeroshade` · 2026-08-27 20:05 UTC
>
> Helpful heads-up from the maintainers — please address these [Pull Request quality criteria](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md#did-you-write-a-patch-that-fixes-a-bug-or-brings-an-improvement) items before this PR can be reviewed:
>
> - :x: **Merge conflicts** — the branch conflicts with `main`. See [docs](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md#did-you-write-a-patch-that-fixes-a-bug-or-brings-an-improvement).
> - :x: **Static checks** — failing: Lint. See [docs](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md#did-you-write-a-patch-that-fixes-a-bug-or-brings-an-improvement).
> - :x: **Failing CI checks** — failing: Archive. See [docs](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md#did-you-write-a-patch-that-fixes-a-bug-or-brings-an-improvement).
> - :x: **Unresolved review comments** — 3 thread(s) remain open ([1](https://github.com/apache/arrow-go/pull/578#discussion_r2558104886), [2](https://github.com/apache/arrow-go/pull/578#discussion_r2558106891), [3](https://github.com/apache/arrow-go/pull/578#discussion_r2558108594)). See [docs](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md#did-you-write-a-patch-that-fixes-a-bug-or-brings-an-improvement).
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>

### DIFF
--- a/arrow/ipc/message.go
+++ b/arrow/ipc/message.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"sync/atomic"
 
+	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/arrow/internal/flatbuf"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 )
@@ -157,7 +158,7 @@ func (r *messageReader) Retain() {
 // When the reference count goes to zero, the memory is freed.
 // Release may be called simultaneously from multiple goroutines.
 func (r *messageReader) Release() {
-	r.refCount.Load()
+	debug.Assert(r.refCount.Load() > 0, "too many releases")
 
 	if r.refCount.Add(-1) == 0 {
 		if r.msg != nil {
--- a/arrow/ipc/message.go
+++ b/arrow/ipc/message.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"sync/atomic"
+	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/arrow/internal/flatbuf"
@@ -86,7 +87,7 @@ func NewMessage(meta, body *memory.Buffer) *Message {
 		body: body,
 	}
 	m.Refcount.Buffers = []**memory.Buffer{&m.meta, &m.body}
-	m.Additional = func() { m.msg = nil }
+	m.Derived = []unsafe.Pointer{unsafe.Pointer(&m.msg)}
 	m.Retain()
 	return m
 }
@@ -102,7 +103,7 @@ func newMessageFromFB(meta *flatbuf.Message, body *memory.Buffer) *Message {
 		body: body,
 	}
 	m.Refcount.Buffers = []**memory.Buffer{&m.meta, &m.body}
-	m.Additional = func() { m.msg = nil }
+	m.Derived = []unsafe.Pointer{unsafe.Pointer(&m.msg)}
 	m.Retain()
 	return m
 }
--- a/arrow/ipc/message.go
+++ b/arrow/ipc/message.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"sync/atomic"
 
-	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/arrow/internal/flatbuf"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 )
@@ -66,10 +65,10 @@ func (m MessageType) String() string {
 
 // Message is an IPC message, including metadata and body.
 type Message struct {
-	refCount atomic.Int64
-	msg      *flatbuf.Message
-	meta     *memory.Buffer
-	body     *memory.Buffer
+	memory.Refcount
+	msg  *flatbuf.Message
+	meta *memory.Buffer
+	body *memory.Buffer
 }
 
 // NewMessage creates a new message from the metadata and body buffers.
@@ -85,7 +84,9 @@ func NewMessage(meta, body *memory.Buffer) *Message {
 		meta: meta,
 		body: body,
 	}
-	m.refCount.Add(1)
+	m.Refcount.Buffers = []**memory.Buffer{&m.meta, &m.body}
+	m.Additional = func() { m.msg = nil }
+	m.Retain()
 	return m
 }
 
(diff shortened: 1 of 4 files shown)


pixelherodev:
Tested that the derived pointers are nilled correctly by modifying the MessageReader test, definitely is working. 

It's also guaranteed to be safe:

>   (1) Conversion of a *T1 to Pointer to *T2.
> 
>     Provided that T2 is no larger than T1 and that the two share an equivalent
>     memory layout, this conversion allows reinterpreting data of one type as
>     data of another type. An example is the implementation of math.Float64bits:

pixelherodev:
In principle, doing this and then moving Refcount.(Retain|Release) into a disabled file is enough to switch off refcounting.

I'd also like to see if we can move the Refcount initialization into an inline-able function call, though - that would preserve this behavior when it's enabled, but also drop the cost of initializing the refcount information when it's not needed...

pixelherodev:
And, tested with MessageReader dependency on Message. Test passes, and if the explicit dependency is dropped, the test fails due to a memory leak.

pixelherodev:
`go fmt`ed. Forgot :upside_down_face:

zeroshade:
I'm not a fan of this approach. I dislike the users having to call *multiple* functions as opposed to just having `Retain` and `Release` (i.e. I don't like having to call the `Referenced` and keep track of Derived etc.)

In theory, a buffer need only keep track of its parent (if it has one) and doesn't need to keep track of any other buffers which were sliced off from it. I don't understand the need/desire for adding the pointers that you're doing as opposed to just having the `atomic.Int64` and essentially putting that behind a struct which can have a version that is empty for turning off the refcounts.

i.e. Why isn't it just something like:

```go
//go:build !norc

type RefCount struct {
    ref atomic.Int64
    cleanup func()
}

func (r *RefCount) Retain() {
    ...
}

func (r *RefCount) Release() {
    ...
}

/////////////////////
//go:build norc

type RefCount struct {}
func (*RefCount) Retain() {}
func (*RefCount) Release() {}

```

arrow/ipc/message.go
```diff
@@ -85,7 +84,9 @@ func NewMessage(meta, body *memory.Buffer) *Message {
 		meta: meta,
 		body: body,
 	}
-	m.refCount.Add(1)
+	m.ReferenceBuffer(&m.meta, &m.body)
+	m.ReferenceDerived(unsafe.Pointer(&m.msg))
```
zeroshade, line 88: we shouldn't need to track the `m.msg`, the point of the refcounting is to track allocations made by the `memory.Allocator`, the flatbuffer message object is never allocated by the memory.Allocator

arrow/memory/refcount.go
```diff
@@ -0,0 +1,75 @@
...
+	"sync/atomic"
+	"unsafe"
+
+	"github.com/apache/arrow-go/v18/arrow/internal/debug"
+)
+
+type Refcount struct {
+	count        atomic.Int64
+	dependencies []unsafe.Pointer
+	buffers      []**Buffer
```
zeroshade, line 15: why `**Buffer` instead of `*Buffer`?

arrow/memory/refcount.go
```diff
@@ -0,0 +1,75 @@
...
+func (r *Refcount) ReferenceDependency(d ...unsafe.Pointer) {
+	r.dependencies = d
+}
+
+// Must only be called once per object. Defines buffers that are referenced
+// by this object. When this object is unreferenced, all such buffers will
+// be deallocated immediately.
+func (r *Refcount) ReferenceBuffer(b ...**Buffer) {
+	r.buffers = b
+}
```
zeroshade, line 32: you mean "When this object is cleaned up" or "released"? Because Go is garbage collected and we don't have destructors, then just being `unreferenced` won't deallocate the buffers immediately, they'll get deallocated when the GC gets around to it

arrow/ipc/message.go
pixelherodev: The msg pointer is _derived_ from the `meta` Buffer, which is itself allocated by the `Allocator`. `ReferenceDerived` is not for refcounting; it basically means "when this object goes free, nil out the target to prevent use-after-free."

Without it, accessing the message after calling Release could potentially access other memory allocated by the `Allocator` - if integrating with C, this could _theoretically_ allow access to other heap allocations. Note that the previous code for Message.Release contains this:

```go
if msg.refCount.Add(-1) == 0 {
		msg.meta.Release()
		msg.body.Release()
		msg.msg = nil
		msg.meta = nil
		msg.body = nil
	}
```

it's releasing the two subresources, but also nilling the one that depends on one. This is just to replace the `msg = nil` line, basically.

arrow/memory/refcount.go
pixelherodev: no, this is for the last call to `Refcount.Release()`: that call will deallocate the buffers immediately - or, rather, it will call Allocator.Free() immediately. IMO these should be seen as the same thing.

arrow/memory/refcount.go
pixelherodev: I'll explain in a dedicated comment going over the whole design

pixelherodev:
> ```
>     cleanup func()
> ```
> 
> Using a cleanup function (or closure) may well be the best approach. I can play around with that instead and see if we like that more? It means not needing to manually track the graph at all, because the function is invoked on release time but defined at creation time; it can take a closure of the object itself, and check whether fields needs unreferenced without any of the typing shenanigans..

One worry I have: I'll need to check if the compiler can optimize out the _construction of the closure_, since it won't actually be used. The current approach generates zero code when disabled, I want to make sure we can maintain that property...

pixelherodev:
Hey,

I left the employer I was working with when I was experimenting with this. I'm still down to work on it, but my house caught fire last month, so it's going to be a little bit before I can take a look :upside_down_face:

PR #591, opened 2025-12-02
fix: correctly initialize SchemaField.ColIndex

JamesGuthrie:
### Rationale for this change

The documentation for SchemaField says:

> ColIndex is only populated (not -1) when it is a leaf column.

But the implementation never actually initializes the value to -1, so it is incorrectly set to 0 on non-leaf nodes.

This commit adds a helper function to correctly initialize SchemaField, and comprehensive test cases to cover all existing branches.

### Are these changes tested?

Yes, see test cases.


zeroshade (approved):
thanks for this!

PR #596, opened 2025-12-08
fix(parquet): decryption of V2 data pages

daniel-adam-tfs:
### Rationale for this change
Fixing the decryption of V2 pages.

This PR fixes the decryption of V2 pages.

### Are these changes tested?
Yes, tests cover various scenarios for both V1 and V2 page encryption and decryption.

No.

### DIFF
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -515,7 +515,7 @@ func (p *serializedPageReader) decompress(rd io.Reader, lenCompressed int, buf [
 	return p.codec.Decode(buf, data), nil
 }
 
-func (p *serializedPageReader) decryptV2(rd io.Reader, lenCompressed int, levelsBytelen int, compressed bool, buf []byte) error {
+func (p *serializedPageReader) readV2Encrypted(rd io.Reader, lenCompressed int, levelsBytelen int, compressed bool, buf []byte) error {
 	// if encrypted, we need to decrypt before decompressing
 	p.decompressBuffer.ResizeNoShrink(lenCompressed)
 	b := bytes.NewBuffer(p.decompressBuffer.Bytes()[:0])
@@ -538,7 +538,7 @@ func (p *serializedPageReader) decryptV2(rd io.Reader, lenCompressed int, levels
 	return nil
 }
 
-func (p *serializedPageReader) decompressV2(rd io.Reader, lenCompressed int, levelsBytelen int, compressed bool, buf []byte) error {
+func (p *serializedPageReader) readV2Unencrypted(rd io.Reader, lenCompressed int, levelsBytelen int, compressed bool, buf []byte) error {
 	if !compressed {
 		// uncompressed, just read into the buffer
 		if _, err := io.ReadFull(rd, buf); err != nil {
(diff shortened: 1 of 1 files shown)


parquet/encryption_write_config_test.go
```diff
@@ -135,20 +136,18 @@ func (en *EncryptionConfigTestSuite) encryptFile(configs *parquet.FileEncryption
 
 		// write the int64 column, each row repeats twice
 		int64Writer := nextColumn().(*file.Int64ColumnChunkWriter)
-		for i := 0; i < 2*en.rowsPerRG; i++ {
+		for i := 0; i < en.rowsPerRG; i++ {
```
daniel-adam-tfs, line 139: I was getting "batch writing for V2 data pages must start at a row boundary" errors for V2 pages with the original version

parquet/file/file_reader_test.go
```diff
@@ -207,7 +207,7 @@ func (p *PageSerdeSuite) TestDataPageV2() {
 
 	p.dataPageHdrV2.Statistics = getDummyStats(statsSize, true)
 	p.dataPageHdrV2.NumValues = nrows
-	p.WriteDataPageHeaderV2(1024, 20, 10)
+	p.WriteDataPageHeaderV2(1024, 0, 0)
```
daniel-adam-tfs, line 210: The original version of serializedPageReader.Next() ignored error checking of `io.ReadFull`. Was that deliberate? I've added the error checking and now the `p.pageReader.Next()` returns false if the compressed and uncompressed values are not 0, so I modified those.

parquet/file/page_reader.go
```diff
@@ -812,15 +855,16 @@ func (p *serializedPageReader) Next() bool {
...
+					p.err = err
 					return false
 				}
 			} else {
-				io.ReadFull(p.r, buf.Bytes())
+				if err := p.readV2Unencrypted(p.r, lenCompressed, levelsBytelen, compressed, buf.Bytes()); err != nil {
+					p.err = err
+					return false
+				}
 			}
```
zeroshade, line 868: So the issue is simply that we're not passing uncompressed v2 through the decryptor right?

Wouldn't a simpler solution here just be to add the decryption check to the non-compressed branch?

i.e. since the call to `decompress` will already decrypt the data, we just need to replace the `io.ReadFull(p.r, buf.Bytes())` with code to check if we need to decrypt and then decrypt it here, otherwise just leave the `io.ReadFull` right?

daniel-adam-tfs: Not quite, the compressed branch had an issue as well. It must decrypt the whole buffer with data first, then it can read the levels and decompress the values.

zeroshade: Ah, gotcha. Okay, that makes more sense.

parquet/encryption_write_config_test.go
zeroshade: I'm confused why it would encounter that issue when it wasn't already encountering it. What is changing that is causing this to occur?

parquet/file/file_reader_test.go
zeroshade: Why does it return false if the compressed and uncompressed values are not 0? Is it because this is a test case and we aren't actually giving it the raw bytes?

parquet/encryption_write_config_test.go
daniel-adam-tfs: The issue is only with files with V2 pages. Originally, encryption was tested only for V1 pages, but I've added files with V2 pages, so the issue started occurring (see modifications to TestUniformEncryption and other test cases in the EncryptionConfigTestSuite)

parquet/file/file_reader_test.go
daniel-adam-tfs: yep, there are no raw bytes so the ReadFull returns EOF error, which was ignored previously, but for which I now check for. I don't see a reason why the error should be ignored, but maybe I'm missing something...?

Issue #613, opened 2025-12-19
[Go] An uncaught error caused a subsequent panic


xiaocai2333:
### Describe the bug, including details regarding any error messages, version, and platform.

1. If an error occurs here, it may result in a nil child.
<img width="893" height="226" alt="Image" src="https://github.com/user-attachments/assets/cb1825f0-fcb5-44b1-a8c5-8a43c7dede93" />

2. Subsequent construction of StructData will trigger a panic due to a nil pointer dereference.
<img width="1104" height="354" alt="Image" src="https://github.com/user-attachments/assets/38596da4-1505-4c9d-8486-c2a5c3d475ad" />

### Component(s)

Parquet

PR #614, opened 2025-12-19
fix(arrow/cdata): Handle errors to prevent panic

xiaocai2333:
### Rationale for this change
Fixes: #613 

Return error correctly.

### Are these changes tested?
Yes

No


xiaocai2333:
@zeroshade please review it

xiaocai2333:
I have resubmitted the code based on the error messages, and the tests passed successfully on my local mechine.
@zeroshade  Could you please approve it again?

zeroshade:
@xiaocai2333 looks like there's a memory leak found by the leak sanitizer that needs to be addressed. I can poke at it tomorrow if you need help tracking it down.

Specifically, the 48 bytes leaks are likely ArrowSchema objects that aren't getting freed (probably children in a nested test)

PR #615, opened 2025-12-20
fix(flight): make StreamChunksFromReader ctx aware and cancellation-safe

arnoldwakim:
### Rationale for this change

`StreamChunksFromReader` previously did not observe context cancellation. As a result, if a client disconnected early, the reader could continue producing data indefinitely, potentially blocking on channel sends, leaking `RecordBatch` objects, leaking the reader, and consuming unbounded memory and CPU (this observation triggered this PR).

This fix ensures that data streaming promptly stops when the client disconnects.

- `StreamChunksFromReader` now accepts a `context.Context`.
- Tiny change was made to `DoGet`, to ensure it continues to work with the context-aware `StreamChunksFromReader`.

### Are these changes tested?

- To be removed from description: the tests are bit tricky to write, similar to that of #437. Maybe @zeroshade has suggestions?

- `StreamChunksFromReader` now accepts a `context.Context`.

### DIFF
--- a/arrow/flight/flight_test.go
+++ b/arrow/flight/flight_test.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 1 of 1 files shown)


arnoldwakim:
I have used a callbackRecord reader to test that the new behaviour is okay. 
Kindly let me know if this is what you expected.

PR #645, opened 2026-01-20
perf(arrow): Reduce the amount of allocated objects

spiridonov:
### Rationale for this change

`apache/arrow-go` is used in the new query engine in Grafana Loki and we are actively working on improving its performance. Here are few low hanging fruits that reduce allocations by many gigabytes when used on the hot path.

This PR reduces amount of allocated objects.

* `arrow/datatype_binary.go`: straightforward, those `Layout` objects are allocated on each call currently.
* `arrow/schema.go`: `Schema.Fields()` does not return a clone anymore and `NewSchemaWithEndian()` does not clone inputs anymore. All call sites are func-signature compatible, but should not modify those args/results anymore. In our opinion this is a reasonable trade-off for higher performance. I am happy to discuss other ways to implement that.

### Are these changes tested?

* Tested with the current `apache/arrow-go` unit test suite.
* Tested with `grafana/loki` test suite for the query engine.

Yes. Explained above.

### DIFF
--- a/arrow/schema_test.go
+++ b/arrow/schema_test.go
@@ -275,6 +275,13 @@ func TestSchema(t *testing.T) {
 				t.Fatalf("invalid field: got=%#v, want=%#v", got, want)
 			}
 
+			fields := s.Fields()
+			fields[0].Name = "other"
+			// check that the fields are copied and not shared
+			if got, want := s.Field(0), tc.fields[0]; !got.Equal(want) {
+				t.Fatalf("invalid field: got=%#v, want=%#v", got, want)
+			}
+
 			if got, want := s.HasMetadata(), tc.md != nil; got != want {
 				t.Fatalf("invalid metadata: got=%v, want=%v", got, want)
 			}


arrow/schema.go
```diff
@@ -170,18 +170,17 @@ func NewSchema(fields []Field, metadata *Metadata) *Schema {
 
 func NewSchemaWithEndian(fields []Field, metadata *Metadata, e endian.Endianness) *Schema {
 	sc := &Schema{
-		fields:     make([]Field, 0, len(fields)),
+		fields:     fields,
```
zeroshade, line 186: The only problem I see with this is that currently we explicitly do the copy to ensure that if the caller maintains a reference to the slice they can't modify the fields after calling this.

By just putting the slice in there, then technically a caller could modify the slice afterwards and possibly screw up the schema. 

Is copying the fields really causing that much memory allocations and leakage?

arrow/schema.go
```diff
@@ -170,18 +170,17 @@ func NewSchema(fields []Field, metadata *Metadata) *Schema {
...
 		index:      make(map[string][]int, len(fields)),
 		endianness: e,
 	}
 	if metadata != nil {
-		sc.meta = metadata.clone()
+		sc.meta = *metadata
```
zeroshade, line 191: same as above, since the members of the `Metadata` object are slices (pointer types) just copying the metadata struct like this means that the caller could potentially modify the metadata after constructing the schema. This is less concerning than possibly modifying the fields slice so I'm more inclined to not see this as an issue here.

arrow/schema.go
```diff
@@ -194,11 +193,10 @@ func (sc *Schema) WithEndianness(e endian.Endianness) *Schema {
...
 func (sc *Schema) Metadata() Metadata            { return sc.meta }
-func (sc *Schema) Fields() []Field {
-	fields := make([]Field, len(sc.fields))
-	copy(fields, sc.fields)
-	return fields
-}
+
+// Fields returns the fields of the schema.
+// The result is not a clone of the schema's fields and therefore should not be modified by the caller.
+func (sc *Schema) Fields() []Field   { return sc.fields }
```
zeroshade: A while back we explicitly converted this into being a clone operation. 

Instead of changing this, can we introduce a function that returns `iter.Seq[Field]`? that way we can avoid the slice copy while still preventing a user from being able to modify the slice.

arrow/schema.go
```diff
@@ -207,11 +205,17 @@ func (sc *Schema) FieldsByName(n string) ([]Field, bool) {
...
+		return sc.fields[indices[0] : indices[0]+1], ok
+	} else if len(indices) > 1 {
+		fields := make([]Field, 0, len(indices))
+		for _, v := range indices {
+			fields = append(fields, sc.fields[v])
+		}
+		return fields, ok
+	} else {
+		return nil, false
 	}
```
zeroshade:
```suggestion
    }
    
	return nil, false
}
```

spiridonov:
That is a great question! I also was confused in the beginning, because I saw those objects being allocated in a memory profile report. I did some research and I think that it happens _sometimes_ because the compiler is conservative. If it is not 100% sure in the result of escape analysis (the value is indeed passed into another func that passes it into another func) or a structure is too large it falls back to allocate it on the heap instead of stack. I did build a sandbox example https://godbolt.org/z/h19PnWM7T which shows `runtime.newobject` clearly. Compare that with a simpler type https://godbolt.org/z/KMznfPxMx which does not allocate on the heap.

spiridonov:
@zeroshade ^

PR #654, opened 2026-02-02
[Parquet] Add SIMD-accelerated byte-stream-split decoding

daniel-adam-tfs:
### Rationale for this change
The byte-stream-split encoding is commonly used in Parquet for floating-point data, as it improves compression ratios by grouping similar bytes together. However, the existing Go implementation uses a simple scalar loop which is inefficient for large datasets. By leveraging SIMD instructions (AVX2 on x86 and NEON on ARM), we can significantly accelerate the decoding process and improve overall Parquet read performance.
 
Optimized implementation of byte-stream split decoding algorithm.

Added SIMD-accelerated implementations:
AVX2 implementation for amd64 architecture using 256-bit vectors processing 32 values per block
NEON implementation for arm64 architecture using 128-bit vectors processing 16 values per block
Both use 2-stage byte unpacking hierarchy following the same algorithm structure
Implemented runtime CPU feature detection with automatic dispatch to the best available implementation (SIMD vs scalar fallback)
Added proper build tags and file suffixes for cross-platform compatibility
Included an optimized V2 scalar implementation using unsafe pointer casting as a fallback

### Are these changes tested?
Yes. Various tests were added:

- Correctness tests covering various input sizes (1, 2, 7, 8, 31, 32, 33, 63, 64, 65, 127, 128, 129, 255, 256, 512, 1024) to validate all implementations (Reference, V2, AVX2, NEON)
- Edge case tests including exact block boundaries, single values, all-zero data, and all-ones data
- Benchmark suite with multiple data sizes (8, 64, 512, 4096, 32768, 262144 values) comparing all implementations

No user-facing API changes. This is a performance optimization that maintains full backward compatibility. Users will automatically benefit from faster Parquet decoding when reading files with byte-stream-split encoded floating-point columns, with no code changes required.


daniel-adam-tfs:
@zeroshade One of our departments has integrated byte-stream-split encoding/decoding into the currently used proprietary format that is used to store data. We did some comparisons and they were getting faster decoding so looked into their code and they were using SIMD implementation with `VPUNPCKLBW` in C#. I took their fallback and SIMD implementations and fed them to Claude (haven't seen much of assembler since college myself) and it gave me these implementation which are really fast.
The fallback is pretty fast, faster than the current implementation so I'd replace the current implementation with it. And I've copied the file names and build tags for the existing assembly in the repo, so it should be OK to add them. 

I've actually wrote a C code first and tried c2goasm with AppleClang and clang21.0 and I couldn't get that to generate me a compilable code. 
And I've also tried the new https://pkg.go.dev/simd/archsimd package that we're getting in go1.26, but it doesn't have VPUNPCKLBW wrapper there, so I couldn't get it to be as fast as the Claude generated code.


Anyway, most of our data is float32s or float64s, so the bss decoding function was at the top. After trying processing some files with this change it fell to like to 10th place, `memmove` and `LevelDecoder.Decode` are the top 2 now. I think I can do something with both. (I see potential improvements to the level decoding for our case and I should get rid of some memmoves if I at some point figure out the other PR with the buffers. 😆 )

daniel-adam-tfs:
> @daniel-adam-tfs any further work needed here to make it ready for review?

Yeah, let me add the equivalent version for float64 and check how complex would it be do add SIMD when doing encoding.

- [x] Add the float64 version
(- [ ] Add SIMD to byte-stream split encoder, skip this for now)

I've also tried using SIMD for the `[]parquet.FixedLenByteArray` 2/4/8byte versions (decodeByteStreamSplitBatchFLBAWidth2, decodeByteStreamSplitBatchFLBAWidth4, decodeByteStreamSplitBatchFLBAWidth8). The problem is that for SIMD to be usable all of the `parquet.FixedLenByteArray` (=[]byte) elements need to share the backing array and be in order, then SIMD can be used. I'm not sure whether we can ensure that.

daniel-adam-tfs:
@zeroshade Do we care about big-endian architectures? I see that there are couple of files with s390x build tags, but when I use docker to emulate a s390x and run tests on main, I get dozens of failures. I'll fix the bss failures if BE should be supported.

parquet/internal/encoding/byte_stream_split.go
```diff
@@ -30,7 +30,7 @@ import (
 // encodeByteStreamSplit encodes the raw bytes provided by 'in' into the output buffer 'data' using BYTE_STREAM_SPLIT encoding.
 // 'data' must have space for at least len(in) bytes.
 func encodeByteStreamSplit(data []byte, in []byte, width int) {
-	debug.Assert(len(data) >= len(in), fmt.Sprintf("not enough space in destination buffer for encoding, dest: %d bytes, src: %d bytes", len(data), len(in)))
+	debug.Assert(len(data) >= len(in), "not enough space in destination buffer for encoding")
```
Danielius1922, line 33: the allocations by fmt.Sprintf messed with the timings, I think the shorter messages are fine for the asserts

parquet/internal/encoding/byte_stream_split_big_endian.go
```diff
@@ -19,43 +19,98 @@
...
+		// Big-endian machine: put s0 as MSB, s7 as LSB
+		out64[i] = uint64(s7[i])<<56 | uint64(s6[i])<<48 | uint64(s5[i])<<40 | uint64(s4[i])<<32 |
+			uint64(s3[i])<<24 | uint64(s2[i])<<16 | uint64(s1[i])<<8 | uint64(s0[i])
+	}
+}
+
+// decodeByteStreamSplitBatchFLBAWidth2 decodes the batch of nValues FixedLenByteArrays of length 2 provided by 'data',
+// into the output slice 'out' using BYTE_STREAM_SPLIT encoding.
+// 'out' must have space for at least nValues slices.
+func decodeByteStreamSplitBatchFLBAWidth2(data []byte, nValues, stride int, out []parquet.FixedLenByteArray) {
```
Danielius1922, line 77: @zeroshade what is the interpretation of BSS encoding when applied to a byte array? Since everything is little-endian in parquet, data is little endian. But what about the out array, should the bytes be reordered on big-endian?

zeroshade: I missed this, I believe that the output array would need the bytes to get re-ordered on big-endian yes.

zeroshade: we can look at other implementations to compare though

daniel-adam-tfs:
> Anything else needed here before this is ready for review @daniel-adam-tfs ?

Let me check where we are with this. (I was away for Easter too, but I'll be available to finish this).

daniel-adam-tfs:
Couple of thoughts: 

1. if we could guarantee that []parquet.FixedLenByteArray (=[][]byte) is contiguous (ie if in memory we have a single backing []byte array) we could use SIMD implementation for this case. The output buffer is given to the Decode function by the user, so we cannot guarantee it. I guess we could change []parquet.FixedLenByteArray to a struct with private []byte field and which we would use for all FLBA items, but that would mean changing the public API.

2. I wanted to implement also encoding using SIMD, but I don't know when I'll have the time for it, so lets do it outside of this PR.

3) We should add some big endian architecture in CI builds. I've used emulation in Docker to emulate s390x and ran all unit tests, they all pass for me currently. 

Got nothing else, so review away and I'll fix anything necessary, so we can this merged.

parquet/internal/encoding/byte_stream_split_big_endian.go
daniel-adam-tfs: There is TestByteStreamSplitEncodingFileRead that tests reading of ByteStreamSplit encoded FLBA. The test uses a file that has plain columns and byte stream split columns. The pairs of plain+bss columns should have the same values. The test reads the values for corresponding columns and checks that the values are the same for the plain and for the BSS column. 
Reading of the plain encoded byte array columns should be correct on BE, and given that the test passes when I run it on s390x emulator, I think we can conclude that we have the correct implementation for reading of BSS encoded FBLA columns... hopefully.

daniel-adam-tfs:
@zeroshade Bump. Any review notes? I'll fix anything problematic so we can get this merged.

zeroshade (approved):
Looks good! I'll merge when the CI finishes running if everything is green

Issue #622, opened 2025-07-28
[Go][Parquet] slice bounds out of range - panic in writeDenseArrow


thoro:
### Describe the bug, including details regarding any error messages, version, and platform.

I'm getting this error when writing large amounts of strings:

```
> panic: runtime error: slice bounds out of range [:-2147360192]

  goroutine 186374 [running]:
  github.com/apache/arrow/go/v17/parquet/pqarrow.writeDenseArrow(0xddae04a9b0, {0x20e96a8, 0xc4ccf9ef00}, {0x20e5320, 0xc4d1a6e5f0}, {0x0, 0x0, 0x0},
  {0x0, 0x0, ...}, ...)
      /go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/parquet/pqarrow/encode_arrow.go:491 +0x2d2d
  github.com/apache/arrow/go/v17/parquet/pqarrow.WriteArrowToColumn({0x20c7298, 0xdeb4627020}, {0x20e96a8, 0xc4ccf9ef00}, {0x20e5320, 0xc4d1a6e5f0},
  {0x0, 0x0, 0x0}, {0x0, ...}, ...)
      /go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/parquet/pqarrow/encode_arrow.go:232 +0x325
  github.com/apache/arrow/go/v17/parquet/pqarrow.(*arrowColumnWriter).Write(0xc4fb951710, {0x20c7298, 0xdeb4627020})
      /go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/parquet/pqarrow/encode_arrow.go:193 +0x533
  github.com/apache/arrow/go/v17/parquet/pqarrow.(*FileWriter).WriteColumnChunked(0xdeb4634a10, 0x31b3220?, 0xc4fb9517d8?, 0x1?)
      /go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/parquet/pqarrow/file_writer.go:330 +0xe5
  github.com/apache/arrow/go/v17/parquet/pqarrow.(*FileWriter).WriteColumnData(0xdeb4634a10, {0x20e5320, 0xc1c0963360})
      /go/pkg/mod/github.com/apache/arrow/go/v17@v17.0.0/parquet/pqarrow/file_writer.go:339 +0xc5
[...]

PR #655, opened 2026-02-06
fix(parquet/file): write large string values

zeroshade:
### Rationale for this change
Writing large byte array values (e.g., 50 values x 50MB each = 2.5GB total) caused a panic due to exceeding max page size. 

This happened because the writer accumulates the values in batches *before* checking the page size limits:

1. `WriteBatch()` calls `writeValues()` which adds ALL values to the encoder buffer
2. `commitWriteAndCheckPageLimit()` checks if the buffer exceeds the limit
3. **PROBLEM**: At this point, the buffer alraedy contains > 2GB of data if we hit the limit
4. `FlushCurrentPage()` attempts to do `int32(values.Len())` which overflows: `2,500,000,000 -> -1,794,967,296`
5. `bytes.Buffer.Grow(-1,794,967,296)` panics

See https://github.com/apache/arrow-go/issues/622#issuecomment-3822818048

Modified `writeValues()` and `writeValuesSpaced()` for `ByteArray` and `FixedLenByteArray` types to check the buffer size *beore* adding the values and proactively flush when approaching the 2GB limit (parquet uses an int32 for page size).

### Are these changes tested?
Yes, new tests are added, including some benchmarks to ensure that the new changes don't cause any performance impacts.

## Performance Impact

**TL;DR: <1% overhead for typical workloads, 0% for fixed-size types**

### Benchmarks

```
Benchmark                            Time       Data        Throughput
─────────────────────────────────────────────────────────────────────
WriteSmallByteArrayValues (100B)    2.19 ms    1 MB        457 MB/s
WriteMediumByteArrayValues (10KB)   18.0 ms   10 MB        556 MB/s
WriteLargeByteArrayValues (1MB)      137 ms  100 MB        730 MB/s
WriteInt32Values (control)          0.15 ms 0.04 MB        267 MB/s (unchanged)
```

### Impact by Data Type

| Data Type | Overhead | Notes |
|-----------|----------|-------|
| Int32, Int64, Float, Boolean | **0%** | Unchanged code paths |
| ByteArray (small, <1KB) | **<1%** | Batched processing |
| ByteArray (large, >1MB) | **<0.01%** | I/O dominates, checking negligible |

### Per-Value Overhead

| Value Size | Encoding Time | Added Overhead | % Impact |
|------------|--------------|----------------|----------|
| 100 bytes | 200 ns | ~10 ns | ~5% |
| 1 KB | 2,000 ns | ~10 ns | ~0.5% |
| 100 KB | 200,000 ns | ~10 ns | ~0.005% |
| 1 MB+ | 2,000,000 ns | ~120 ns | ~0.006% |

Only the fix to the previous situation that would panic.

### DIFF
--- a/parquet/file/large_value_test.go
+++ b/parquet/file/large_value_test.go
@@ -129,8 +129,8 @@ func TestLargeStringArrayWithArrow(t *testing.T) {
 	// Each batch: 10 values × 10MB = 100MB
 	// Total: 11 batches = 1.1GB written (only 100MB memory at a time!)
 	const valueSize = 10 * 1024 * 1024 // 10MB per string (realistic large blob)
-	const valuesPerBatch = 10           // 10 values per batch
-	const numBatches = 11               // 11 batches = 1.1GB total
+	const valuesPerBatch = 10          // 10 values per batch
+	const numBatches = 11              // 11 batches = 1.1GB total
 
 	largeStr := string(make([]byte, valueSize))
 
@@ -142,7 +142,7 @@ func TestLargeStringArrayWithArrow(t *testing.T) {
 		}
 		arr := builder.NewArray()
 
-		rec := array.NewRecord(arrowSchema, []arrow.Array{arr}, int64(valuesPerBatch))
+		rec := array.NewRecordBatch(arrowSchema, []arrow.Array{arr}, int64(valuesPerBatch))
 
 		// Write batch - this should NOT panic with int32 overflow
 		err = pqw.Write(rec)


parquet/file/large_value_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestLargeByteArrayValuesDoNotOverflowInt32 tests that writing large byte array
+// values that would exceed 2GB when accumulated in a single page does not cause
+// an int32 overflow panic.
+//
+// This is the RED phase - the test should FAIL because the bug exists.
+// Expected failure: panic when trying to cast >2GB buffer size to int32
```
lidavidm: The test is enabled, presumably this no longer fails?

zeroshade: yea, i just updated that

PR #668, opened 2026-02-19
fix(parquet/pqarrow): selective column reading of complex map column

zeroshade:
### Rationale for this change
Upstream fix for the issue identified in https://github.com/apache/iceberg-go/issues/737. When reading maps with nested values using column indices for selective column reading, if the child fields of the map weren't in the list of indices there was a problem:

- Maps are represented in Parquet as a list of key-value structs (`list<struct<key, value>>`
- The struct *MUST* have exactly 2 fields (key and value) to be converted into a proper Arrow typed Map column
- When applying the column filtering, if only the key *OR* value field (but not both) were in the list of columns, the resulting child struct would only have 1 field
- As a result, the `Map.validateData()` method would fail with a panic of `arrow/array: map array child array should have two fields`.

In pqarrow/file_reader.go leaf filtering is disabled when reading a map's key-value struct. This will ensure both the key and value columns are always read together, maintaining the required 2-field structure for  map array.

### Are these changes tested?
Yes a test case is added for the change.

This only affects map type reading with column filter selection ensuring correctness. The only change is that a failure mode has been eliminated.

### DIFF
--- a/parquet/pqarrow/file_reader_test.go
+++ b/parquet/pqarrow/file_reader_test.go
@@ -732,7 +732,7 @@ func TestMapColumnWithFilters(t *testing.T) {
 	// Read the record batch
 	require.True(t, rr.Next())
 	result := rr.RecordBatch()
-	defer result.Release()
+	// Note: Don't release result manually - the record reader owns it
 
 	// Verify schema - should have only 2 fields (id and properties)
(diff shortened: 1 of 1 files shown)


alexandre-normand:
Thanks for fixing this quickly, @zeroshade! This is very much appreciated.

lidavidm (approved):

PR #683, opened 2026-03-04
chore(dev/release): use release/KEYS instead of dev/KEYS

### LINKED ISSUE #677: [Release] Use release/KEYS not dev/KEYS in verification

On the main repository we moved to use the KEYS file from:
ARROW_KEYS_URL="https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS"

See:
- https://github.com/apache/arrow/issues/47084

They might not be synchronized.

We should use the release/ version:

- Users use the release/ version not dev/ version when they verify our artifacts' signature
-  https://dist.apache.org/ may reject our request when we request many times by CI

So https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS is better than https://dist.apache.org/repos/dist/dev/arrow/KEYS .

I've noticed because I did update the release version of the KEYS file in the last weeks but I didn't update the dev/ version as I thought we were not using it anymore and didn't see my new KEY on the verification.

### Component(s)

Developer Tools

zeroshade:
fixes #677

### DIFF
--- a/dev/release/verify_rc.sh
+++ b/dev/release/verify_rc.sh
@@ -83,7 +83,7 @@ download_rc_file() {
 
 import_gpg_keys() {
   if [ "${VERIFY_SIGN}" -gt 0 ]; then
-    download "${ARROW_DIST_URL}"
+    download "${ARROW_KEYS_URL}"
     gpg --import KEYS
   fi
 }


dev/release/verify_rc.sh
```diff
@@ -34,7 +34,7 @@ set -x
 VERSION="$1"
 RC="$2"
 
-ARROW_DIST_BASE_URL="https://dist.apache.org/repos/dist/dev/arrow"
+ARROW_DIST_URL="https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS"
```
raulcd:
```suggestion
ARROW_KEYS_URL="https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS"
```

dev/release/verify_rc.sh
```diff
@@ -83,7 +83,7 @@ download_rc_file() {
 
 import_gpg_keys() {
   if [ "${VERIFY_SIGN}" -gt 0 ]; then
-    download "${ARROW_DIST_BASE_URL}/KEYS"
+    download "${ARROW_DIST_URL}"
```
raulcd:
```suggestion
    download "${ARROW_KEYS_URL}"
```

PR #704, opened 2026-03-11
[WIP] Parquet Go ALP implementation support

ArnavBalyan:
cc @prtkgaur @julienledem @alamb 

### Rationale for this change
 - Implements ALP (Adaptive Lossless floating point) encoding for float and double columns as per Prateek's spec. Related to https://github.com/apache/parquet-format/pull/548.
 - ALP converts floating point values to integers via decimal scaling, then applies For and bit packing. Values that don't round trip exactly are stored as exceptions.
 - The encoder is incremental and flushes per vector as values arrive. The decoder is lazy with vectors decoded on demand via offset array.

 - Wired ALP implementation through encoder.go and decoder.go based on the original [spec](https://docs.google.com/document/d/1xz2cudDpN2Y1ImFcTXh15s-3fPtD_aWt/edit). 
 - Registered a new encoding as Encodings.ALP.
 - Added unit tests/cross compat tests

### Are these changes tested?
 - Yes new unit tests were added, all of which pass in addition to the existing tests.
 - Cross compat tests were added,  the external encoded file can be provided as env var which triggers arrow-go ALP decoding.

 - Yes

CoAuthored by: @prtkgaur

Related to #646

### DIFF
--- a/parquet/file/column_writer_test.go
+++ b/parquet/file/column_writer_test.go
@@ -502,6 +502,15 @@ func (p *PrimitiveWriterTestSuite) TestRequiredByteStreamSplit() {
 	}
 }
 
+func (p *PrimitiveWriterTestSuite) TestRequiredAlp() {
+	switch p.Typ {
+	case reflect.TypeOf(float32(0)), reflect.TypeOf(float64(0)):
+		p.testRequiredWithEncoding(parquet.Encodings.ALP)
+	default:
+		p.Panics(func() { p.testRequiredWithEncoding(parquet.Encodings.ALP) })
+	}
+}
+
 func (p *PrimitiveWriterTestSuite) TestRequiredDictionary() {
 	p.testRequiredWithEncoding(parquet.Encodings.PlainDict)
 }
--- a/parquet/file/file_reader_test.go
+++ b/parquet/file/file_reader_test.go
@@ -24,6 +24,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path"
 	"testing"
@@ -939,5 +940,62 @@ func TestListColumns(t *testing.T) {
 				require.Equal(t, records[j][i], vals.ValueStr(j))
 			}
 		}
+	}
+}
+
+func TestAlpEncodingFileRead(t *testing.T) {
+	testFile := os.Getenv("ALP_TEST_FILE")
+	if testFile == "" {
+		t.Skip("ALP_TEST_FILE not set, skipping")
+	}
+	if _, err := os.Stat(testFile); os.IsNotExist(err) {
+		t.Skipf("ALP_TEST_FILE not found: %s", testFile)
+	}
+
+	props := parquet.NewReaderProperties(memory.DefaultAllocator)
+	fileReader, err := file.OpenParquetFile(testFile, false, file.WithReadProps(props))
+	require.NoError(t, err)
+	defer fileReader.Close()
+
+	nRows := 1024
+	require.Equal(t, 1, fileReader.NumRowGroups())
+	meta := fileReader.MetaData()
+	require.EqualValues(t, nRows, meta.GetNumRows())
+	require.Equal(t, 1, meta.Schema.NumColumns())
+	require.Equal(t, "value_f64", meta.Schema.Column(0).Name())
+	require.Equal(t, parquet.Types.Double, meta.Schema.Column(0).PhysicalType())
+
+	rgr := fileReader.RowGroup(0)
+	require.EqualValues(t, nRows, rgr.NumRows())
+
+	rdr, err := rgr.Column(0)
+	require.NoError(t, err)
+
+	f64Reader, ok := rdr.(*file.Float64ColumnChunkReader)
+	require.True(t, ok)
+
+	values := make([]float64, nRows)
+	total, read, err := f64Reader.ReadBatch(int64(nRows), values, nil, nil)
+	require.NoError(t, err)
+	require.EqualValues(t, nRows, total)
+	require.EqualValues(t, nRows, read)
+
+	md, err := rgr.MetaData().ColumnChunk(0)
+	require.NoError(t, err)
+	encodings := md.Encodings()
+	found := false
+	for _, enc := range encodings {
+		if enc == parquet.Encodings.ALP {
+			found = true
+			break
+		}
+	}
+	require.True(t, found, "expected ALP encoding in column chunk metadata, got %v", encodings)
+
+	for i := 0; i < nRows; i++ {
+		expected := 0.125 + float64(i)*0.25
+		assert.Equal(t, math.Float64bits(expected), math.Float64bits(values[i]),
+			"value[%d]: got %v (bits %016x), want %v (bits %016x)",
+			i, values[i], math.Float64bits(values[i]), expected, math.Float64bits(expected))
 	}
 }
--- a/parquet/internal/encoding/alp_encoding.go
+++ b/parquet/internal/encoding/alp_encoding.go
@@ -1,0 +1,966 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package encoding
+
+// ALP (Adaptive Lossless floating-Point) encoding for FLOAT and DOUBLE types.
+// Reference: https://dl.acm.org/doi/10.1145/3626717
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+	"math/bits"
+	"sort"
+
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	format "github.com/apache/arrow-go/v18/parquet/internal/gen-go/parquet"
+	"github.com/apache/arrow-go/v18/parquet/schema"
+	"golang.org/x/xerrors"
+)
+
+const (
+	alpVersion            = 1
+	alpCompressionMode    = 0 // ALP
+	alpIntegerEncodingFOR = 0 // kForBitPack
+	alpHeaderSize         = 8
+	alpDefaultVectorSize  = 1024
+	alpDefaultLogVector   = 10
+	alpMaxLogVectorSize   = 15
+	alpMinLogVectorSize   = 3
+
+	alpInfoSize       = 4 // exponent(1) + factor(1) + num_exceptions(2)
+	alpFloatForInfo   = 5 // frame_of_reference(4) + bit_width(1)
+	alpDoubleForInfo  = 9 // frame_of_reference(8) + bit_width(1)
+
+	alpFloatMaxExponent  = 10
+	alpDoubleMaxExponent = 18
+
+	alpSamplerVectors        = 8
+	alpMaxPresetCombinations = 5
+
+	alpMagicFloat32 = float32(12582912.0)     // 2^22 + 2^23
+	alpMagicFloat64 = float64(6755399441055744.0) // 2^51 + 2^52
+
+	alpNegZeroFloat32Bits = uint32(0x80000000)
+	alpNegZeroFloat64Bits = uint64(0x8000000000000000)
+)
+
+var (
+	alpFloatPow10 = [11]float32{
+		1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9, 1e10,
+	}
+	alpDoublePow10 = [19]float64{
+		1e0, 1e1, 1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9,
+		1e10, 1e11, 1e12, 1e13, 1e14, 1e15, 1e16, 1e17, 1e18,
+	}
+)
+func alpFloatMultiplier(exponent, factor int) float32 {
+	m := alpFloatPow10[exponent]
+	if factor > 0 {
+		m /= alpFloatPow10[factor]
+	}
+	return m
+}
+
+func alpDoubleMultiplier(exponent, factor int) float64 {
+	m := alpDoublePow10[exponent]
+	if factor > 0 {
+		m /= alpDoublePow10[factor]
+	}
+	return m
+}
+
+func alpFastRoundFloat32(v float32) int32 {
+	if v >= 0 {
+		return int32((v + alpMagicFloat32) - alpMagicFloat32)
+	}
+	return int32((v - alpMagicFloat32) + alpMagicFloat32)
+}
+
+func alpFastRoundFloat64(v float64) int64 {
+	if v >= 0 {
+		return int64((v + alpMagicFloat64) - alpMagicFloat64)
+	}
+	return int64((v - alpMagicFloat64) + alpMagicFloat64)
+}
+
+func alpIsBasicFloatException(v float32) bool {
+	return math.IsNaN(float64(v)) || math.IsInf(float64(v), 0) ||
+		math.Float32bits(v) == alpNegZeroFloat32Bits
+}
+
+func alpIsBasicDoubleException(v float64) bool {
+	return math.IsNaN(v) || math.IsInf(v, 0) ||
+		math.Float64bits(v) == alpNegZeroFloat64Bits
+}
+
+func alpIsFloatException(v float32, exponent, factor int) bool {
+	if alpIsBasicFloatException(v) {
+		return true
+	}
+	m := alpFloatMultiplier(exponent, factor)
+	scaled := float64(v) * float64(m)
+	if scaled > math.MaxInt32 || scaled < math.MinInt32 {
+		return true
+	}
+	encoded := alpFastRoundFloat32(v * m)
+	decoded := float32(encoded) / m
+	return math.Float32bits(v) != math.Float32bits(decoded)
+}
+
+func alpIsDoubleException(v float64, exponent, factor int) bool {
+	if alpIsBasicDoubleException(v) {
+		return true
+	}
+	m := alpDoubleMultiplier(exponent, factor)
+	scaled := v * m
+	if scaled > math.MaxInt64 || scaled < math.MinInt64 {
+		return true
+	}
+	encoded := alpFastRoundFloat64(v * m)
+	decoded := float64(encoded) / m
+	return math.Float64bits(v) != math.Float64bits(decoded)
+}
+
+func alpEncodeFloat32(v float32, exponent, factor int) int32 {
+	return alpFastRoundFloat32(v * alpFloatMultiplier(exponent, factor))
+}
+
+func alpDecodeFloat32(encoded int32, exponent, factor int) float32 {
+	return float32(encoded) / alpFloatMultiplier(exponent, factor)
+}
+
+func alpEncodeFloat64(v float64, exponent, factor int) int64 {
+	return alpFastRoundFloat64(v * alpDoubleMultiplier(exponent, factor))
+}
+
+func alpDecodeFloat64(encoded int64, exponent, factor int) float64 {
+	return float64(encoded) / alpDoubleMultiplier(exponent, factor)
+}
+
+func alpBitWidth32(v uint32) int {
+	if v == 0 {
+		return 0
+	}
+	return bits.Len32(v)
+}
+
+func alpBitWidth64(v uint64) int {
+	if v == 0 {
+		return 0
+	}
+	return bits.Len64(v)
+}
+func alpPackBits32(values []uint32, count, bitWidth int) []byte {
+	if bitWidth == 0 || count == 0 {
+		return nil
+	}
+	packedSize := (count*bitWidth + 7) / 8
+	out := make([]byte, packedSize)
+
+	var buf uint64
+	bitsInBuf := 0
+	byteIdx := 0
+
+	for i := 0; i < count; i++ {
+		buf |= uint64(values[i]) << uint(bitsInBuf)
+		bitsInBuf += bitWidth
+
+		for bitsInBuf >= 8 {
+			out[byteIdx] = byte(buf)
+			buf >>= 8
+			bitsInBuf -= 8
+			byteIdx++
+		}
+	}
+
+	if bitsInBuf > 0 {
+		out[byteIdx] = byte(buf)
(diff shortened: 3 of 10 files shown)


parquet/file/file_reader_test.go
```diff
@@ -941,3 +942,60 @@ func TestListColumns(t *testing.T) {
...
 }
+
+func TestAlpEncodingFileRead(t *testing.T) {
+	testFile := os.Getenv("ALP_TEST_FILE")
+	if testFile == "" {
+		t.Skip("ALP_TEST_FILE not set, skipping")
+	}
+	if _, err := os.Stat(testFile); os.IsNotExist(err) {
+		t.Skipf("ALP_TEST_FILE not found: %s", testFile)
+	}
```
zeroshade, line 953: can the ALP_TEST_FILE get added to the `parquet-testing` repo rather than needing a separate env var to specify it?

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+
+	alpMagicFloat32 = float32(12582912.0)         // 2^22 + 2^23
+	alpMagicFloat64 = float64(6755399441055744.0) // 2^51 + 2^52
+
+	alpNegZeroFloat32Bits = uint32(0x80000000)
+	alpNegZeroFloat64Bits = uint64(0x8000000000000000)
+)
+
+var (
+	alpFloatPow10 = [11]float32{
```
zeroshade, line 64: is it possible that the sizes of these Arrays may change at any point? Should we use `...` instead of hard-coding the 11 and 19?

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	return m
+}
+
+func alpDoubleMultiplier(exponent, factor int) float64 {
+	m := alpDoublePow10[exponent]
+	if factor > 0 {
+		m /= alpDoublePow10[factor]
+	}
+	return m
+}
```
zeroshade, line 87: if we create base generic functions that do something like the following for the magic numbers and the pow10s (the compiler *should* optimize appropriately):

```go
func alpMagic[T float32 | float64]() T {
    var z T
    switch any(z).(type) {
    case float32:
        return T(alpMagicFloat32)
    case float64:
        return T(alpMagicFloat64)
    default:
        panic("should never happen")
    }
}
```

then we can simplify the rest of the code with things like:

```go
func alpFastRound[T float32 | float64](v T) int32 {
    if v >= 0 {
        return int32((v + alpMagic[T]()) - alpMagic[T]())
    }
    return int32((v - alpMagic[T]()) + alpMagic[T]())
}
```

and so on rather than having duplicate functions for everything for float32/float64.

Not sure if it's necessarily worth it, but given the sheer volume of duplication here, it might be.

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+			byteIdx++
+		}
+	}
+
+	if bitsInBuf > 0 {
+		out[byteIdx] = byte(buf)
+	}
+
+	return out
+}
```
zeroshade, line 198: we have a `BitReader` and `BitWriter` which have optimized methods for packing and unpacking integers into bytes/bits. Is there a reason you wrote these methods instead of reusing the existing `BitReader`/`BitWriter` in `internal/utils`?

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+type alpEncodingParams struct {
+	exponent      int
+	factor        int
+	numExceptions int
+}
+
+func alpFindBestFloat32Params(values []float32, offset, length int) alpEncodingParams {
+	best := alpEncodingParams{exponent: 0, factor: 0, numExceptions: length}
+
+	for e := 0; e <= alpFloatMaxExponent; e++ {
```
zeroshade, line 301:
```suggestion
	for e := range alpFloatMaxExponent {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		for f := 0; f <= e; f++ {
```
zeroshade, line 302:
```suggestion
		for f := range e {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+			exceptions := 0
+			for i := 0; i < length; i++ {
```
zeroshade, line 304:
```suggestion
			for i := range length {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	return out
+}
+
+type alpEncodingParams struct {
+	exponent      int
+	factor        int
+	numExceptions int
+}
+
+func alpFindBestFloat32Params(values []float32, offset, length int) alpEncodingParams {
```
zeroshade, line 298: could the caller just pass `values[offset:offset+length]`? That would simplify this function to just take `values []float32` and operate solely on that slice.

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	return best
+}
+
+func alpFindBestFloat32ParamsWithPresets(values []float32, offset, length int, presets [][2]int) alpEncodingParams {
+	best := alpEncodingParams{exponent: presets[0][0], factor: presets[0][1], numExceptions: length}
+
+	for _, p := range presets {
+		e, f := p[0], p[1]
+		exceptions := 0
+		for i := 0; i < length; i++ {
```
zeroshade, line 326:
```suggestion
		for i := range length {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+				if best.numExceptions == 0 {
+					return best
+				}
+			}
+		}
+	}
+	return best
+}
+
+func alpFindBestFloat32ParamsWithPresets(values []float32, offset, length int, presets [][2]int) alpEncodingParams {
```
zeroshade, line 320: same comment as above. the caller could call this with `values[offset:offset+length]` and then this function gets simplified into operating solely on the `[]float32` slice.

The loop below could then be just 

```go
for _, v := range values {
    if alpIsFloatException(v, e, f) {
        exceptions++
    }
}
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+			best = alpEncodingParams{exponent: e, factor: f, numExceptions: exceptions}
+			if best.numExceptions == 0 {
+				return best
+			}
+		}
+	}
+	return best
+}
+
+func alpFindBestFloat64Params(values []float64, offset, length int) alpEncodingParams {
```
zeroshade, line 341: another case of my earlier comment that could benefit from using generics to avoid the duplication. otherwise all the same comments as in the float32 version of this

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+				if best.numExceptions == 0 {
+					return best
+				}
+			}
+		}
+	}
+	return best
+}
+
+func alpFindBestFloat64ParamsWithPresets(values []float64, offset, length int, presets [][2]int) alpEncodingParams {
```
zeroshade, line 363: same comments as above

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	panic("unreachable")
+}
+
+func (enc *alpEncoder[T]) Put(in []T) {
+	for i := 0; i < len(in); {
+		space := enc.vectorSize - enc.bufCount
+		toCopy := len(in) - i
+		if toCopy > space {
+			toCopy = space
+		}
```
zeroshade, line 429:
```suggestion
        toCopy := min(len(in) - i, enc.vectorSize - enc.bufCount)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		copy(enc.vectorBuf[enc.bufCount:], in[i:i+toCopy])
```
zeroshade, line 430: copy returns the number of elements copied. So you might be able to further simplify this:

```go
copied := copy(enc.vectorBuf[enc.bufCount:], in[i:])
enc.bufCount += copied
enc.totalCount += copied
i += copied
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	case float64:
+		enc.encodeFloat64Vector(vectorLen)
+	}
+}
+
+func (enc *alpEncoder[T]) encodeFloat32Vector(vectorLen int) {
+	values := make([]float32, vectorLen)
+	for i := 0; i < vectorLen; i++ {
+		values[i] = float32(any(enc.vectorBuf[i]).(float32))
+	}
```
zeroshade, line 462: why do we have to copy this again? we already copied from the input slice to the buf, we shouldn't need to copy the values again, can't we just operate on the vectorBuf as is?

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+
+	var params alpEncodingParams
+	if enc.cachedPresets != nil {
+		params = alpFindBestFloat32ParamsWithPresets(values, 0, vectorLen, enc.cachedPresets)
```
zeroshade, line 466: `values[:vectorLen]` would be sufficient, or (even better) just `enc.vectorBuf[:vectorLen]`

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	if enc.cachedPresets == nil && enc.vectorsProcessed >= alpSamplerVectors {
+		enc.buildPresetCache()
+	}
+
+	encoded := make([]int32, vectorLen)
+	excPositions := make([]uint16, 0, params.numExceptions)
+	excValues := make([]float32, 0, params.numExceptions)
+
+	var placeholder int32
+	for i := 0; i < vectorLen; i++ {
```
zeroshade, line 483: Use the range syntax and then just use `v` in the lines below instead of `values[i]`

```suggestion
	for _, v := range enc.vectorBuf[:vectorLen] {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		if !alpIsFloatException(values[i], params.exponent, params.factor) {
+			placeholder = alpEncodeFloat32(values[i], params.exponent, params.factor)
+			break
+		}
+	}
+
+	minValue := int32(math.MaxInt32)
+	for i := 0; i < vectorLen; i++ {
```
zeroshade, line 491: same comment as above

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		if alpIsFloatException(values[i], params.exponent, params.factor) {
+			excPositions = append(excPositions, uint16(i))
+			excValues = append(excValues, values[i])
+			encoded[i] = placeholder
+		} else {
+			encoded[i] = alpEncodeFloat32(values[i], params.exponent, params.factor)
+		}
+		if encoded[i] < minValue {
+			minValue = encoded[i]
+		}
```
zeroshade, line 501:
```suggestion
        minValue = min(encoded[i], minValue)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	}
+
+	var maxDelta uint32
+	deltas := make([]uint32, vectorLen)
+	for i := 0; i < vectorLen; i++ {
```
zeroshade, line 506: same comment as above

```suggestion
	for i, e := range encoded {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		deltas[i] = uint32(encoded[i] - minValue) // wrapping subtraction, result is unsigned
+		if deltas[i] > maxDelta {
+			maxDelta = deltas[i]
+		}
```
zeroshade, line 510:
```suggestion
        maxDelta = max(deltas[i], maxDelta)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	}
+
+	bitWidth := alpBitWidth32(maxDelta)
+
+	startLen := len(enc.encodedBuf)
+
+	var alpInfo [alpInfoSize]byte
+	alpInfo[0] = byte(params.exponent)
+	alpInfo[1] = byte(params.factor)
+	binary.LittleEndian.PutUint16(alpInfo[2:], uint16(params.numExceptions))
```
zeroshade, line 520: why bother with the `alpInfo` intermediary here? why not just do something like 

```go
enc.encodedBuf = append(enc.encodedBuf, byte(params.exponent), byte(params.factor))
enc.encodedBuf = binary.LittleEndian.AppendUint16(enc.encodedBuf, uint16(params.numExceptions))
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	enc.encodedBuf = append(enc.encodedBuf, alpInfo[:]...)
+
+	var forInfo [alpFloatForInfo]byte
+	binary.LittleEndian.PutUint32(forInfo[0:], uint32(minValue))
+	forInfo[4] = byte(bitWidth)
+	enc.encodedBuf = append(enc.encodedBuf, forInfo[:]...)
```
zeroshade, line 526: same comment as above

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	if bitWidth > 0 {
+		packed := alpPackBits32(deltas, vectorLen, bitWidth)
+		enc.encodedBuf = append(enc.encodedBuf, packed...)
+	}
+
+	if params.numExceptions > 0 {
+		for _, pos := range excPositions {
+			var buf [2]byte
+			binary.LittleEndian.PutUint16(buf[:], pos)
+			enc.encodedBuf = append(enc.encodedBuf, buf[:]...)
```
zeroshade, line 537: `enc.encodedBuf = binary.LittleEndian.AppendUint16(enc.encodedBuf, pos)`

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		}
+		for _, val := range excValues {
+			var buf [4]byte
+			binary.LittleEndian.PutUint32(buf[:], math.Float32bits(val))
+			enc.encodedBuf = append(enc.encodedBuf, buf[:]...)
```
zeroshade, line 542: same as above, `enc.encodedBuf = binary.LittleEndian.AppendUint32(enc.encodedBuf, math.Float32bits(val))`

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	}
+
+	enc.vectorSizes = append(enc.vectorSizes, len(enc.encodedBuf)-startLen)
+}
+
+func (enc *alpEncoder[T]) encodeFloat64Vector(vectorLen int) {
+	values := make([]float64, vectorLen)
+	for i := 0; i < vectorLen; i++ {
+		values[i] = float64(any(enc.vectorBuf[i]).(float64))
+	}
```
zeroshade, line 553: can we just leverage generics to have a single `encodeVector` method using the techniques I mentioned above? otherwise, all of my comments from the float32 version apply here too

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		sorted = append(sorted, kv{k, v})
+	}
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i].count > sorted[j].count
+	})
+
+	numPresets := len(sorted)
+	if numPresets > alpMaxPresetCombinations {
+		numPresets = alpMaxPresetCombinations
+	}
```
zeroshade, line 656:
```suggestion
	numPresets := min(len(sorted), alpMaxPresetCombinations)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	numVectors := len(enc.vectorSizes)
+	offsetArraySize := numVectors * 4
+	totalSize := alpHeaderSize + offsetArraySize + len(enc.encodedBuf)
+	enc.sink.Reserve(totalSize)
+
+	var header [alpHeaderSize]byte
+	header[0] = alpVersion
+	header[1] = alpCompressionMode
+	header[2] = alpIntegerEncodingFOR
+	header[3] = byte(enc.logVectorSize)
```
zeroshade, line 687: you could write these first, and then just do `binary.Write(enc.sink, binary.LittleEndian, uint32(enc.totalCount))` without needing the intermediary

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	binary.LittleEndian.PutUint32(header[4:], uint32(enc.totalCount))
+	enc.sink.Write(header[:])
+
+	currentOffset := uint32(offsetArraySize)
+	for _, size := range enc.vectorSizes {
+		var buf [4]byte
+		binary.LittleEndian.PutUint32(buf[:], currentOffset)
+		enc.sink.Write(buf[:])
```
zeroshade, line 695:
```suggestion
		binary.Write(enc.sink, binary.LittleEndian, currentOffset)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	enc.encodedBuf = enc.encodedBuf[:0]
+	enc.vectorSizes = enc.vectorSizes[:0]
+	enc.vectorsProcessed = 0
+	enc.cachedPresets = nil
+	enc.presetCounts = make(map[uint32]int)
+}
+
+func (enc *alpEncoder[T]) Release() {
+	enc.encoder.Release()
+}
```
zeroshade, line 717: this is unnecessary since you're already embedding `encoder` it'll inherit this method without needing to explicitly do this

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+
+func (dec *alpDecoder[T]) decodeVector(vectorIdx int) {
+	var z T
+	switch any(z).(type) {
+	case float32:
+		dec.decodeFloat32Vector(vectorIdx)
+	case float64:
+		dec.decodeFloat64Vector(vectorIdx)
+	}
+}
```
zeroshade, line 822: same comment as elsewhere, see let's try to have a single `decodeVector` generic method instead if possible, or at least significantly shrink the duplicated code between the two

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+	if bitWidth > 0 {
+		packedSize := (vectorLen*bitWidth + 7) / 8
+		deltas = alpUnpackBits32(dec.bodyData[pos:pos+packedSize], vectorLen, bitWidth)
+		pos += packedSize
+	} else {
+		deltas = make([]uint32, vectorLen)
+	}
+
+	outSlice := any(&dec.decodedValues).(*[]float32)
+	for i := 0; i < vectorLen; i++ {
```
zeroshade, line 847: as above

```suggestion
	for i, d := range deltas {
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		encoded := int32(deltas[i]) + frameOfRef
+		(*outSlice)[i] = alpDecodeFloat32(encoded, exponent, factor)
+	}
+
+	if numExceptions > 0 {
+		excPositions := make([]uint16, numExceptions)
+		for e := 0; e < numExceptions; e++ {
+			excPositions[e] = binary.LittleEndian.Uint16(dec.bodyData[pos:])
+			pos += 2
+		}
```
zeroshade, line 857: if we're running on a little endian system, we can avoid the copy entirely using
 
```go
        excPositions := unsafe.Slice((*uint16)(unsafe.Pointer(&dec.bodyData[pos])), numExceptions)
```

That's essentially equivalent to doing a `reinterpret_cast` in C++. 

For big-endian, you should use `for e := range numExceptions {`

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		for e := 0; e < numExceptions; e++ {
+			bits := binary.LittleEndian.Uint32(dec.bodyData[pos:])
+			(*outSlice)[excPositions[e]] = math.Float32frombits(bits)
+			pos += 4
+		}
```
zeroshade, line 862: same comment as above

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+			pos += 8
+		}
+	}
+}
+
+func (dec *alpDecoder[T]) Decode(out []T) (int, error) {
+	toRead := len(out)
+	if toRead > dec.nvals {
+		toRead = dec.nvals
+	}
```
zeroshade, line 912:
```suggestion
	toRead := min(len(out), dec.nvals)
```

parquet/internal/encoding/alp_encoding.go
```diff
@@ -0,0 +1,971 @@
...
+		// Copy from decoded buffer
+		indexInVector := dec.currentIndex % dec.vectorSize
+		vectorLen := dec.getVectorLength(vectorIdx)
+		available := vectorLen - indexInVector
+		toCopy := toRead - read
+		if toCopy > available {
+			toCopy = available
+		}
+
+		copy(out[read:], dec.decodedValues[indexInVector:indexInVector+toCopy])
```
zeroshade, line 937: As before, `copy` will automatically stop when it hits the end of either the src or dst and returns how many values were copied. So you can simplify this with

```go
     copied := copy(out[read:], dec.decodedValues[indexInVector:])
```

This should work provided the actual length of `dec.decodedValues` matches `vectorLen` etc.

zeroshade (changes requested):

parquet/file/column_writer_test.go
```diff
@@ -502,6 +502,15 @@ func (p *PrimitiveWriterTestSuite) TestRequiredByteStreamSplit() {
 	}
 }
 
+func (p *PrimitiveWriterTestSuite) TestRequiredAlp() {
```
prtkgaur, line 505: Hi @ArnavBalyan ,
Thanks a lot for working on this change :).
Wanted to check before I start digging deeper.
Were we able to test the code on some of the test files in https://github.com/apache/parquet-testing/pull/100?

ArnavBalyan: Hey Prateek, thanks for taking a look. The code was tested with cross compat tests where any external file can be decoded using the Go implementation. We may have to the test files above for arrow-go (don't recall if this file was used). I should have bandwidth to take this fwd early August. 
Please feel free to use this change and take it forward if you have time before thanks!

PR #712, opened 2026-03-15
fix(compute): fix data race and memory leak in concurrent is_in kernel

laskoviymishka:
by cleaning up SetLookupState via per-invocation ctx.State instead of shared kernel.Data

fix: apache/iceberg-go/issues/489

### Rationale for this change

fix: apache/iceberg-go/issues/489 (flaky test in iceberg-go)

fix data race and memory leak in concurrent is_in kernel

### Are these changes tested?

yes, added a test to repro memory leak (3/10 runs fails)


zeroshade (approved):
AH, *that's* the cause of that flakiness.... Thanks!!

PR #707, opened 2026-03-12
perf(parquet/file): avoid double bool bitmap conversion

zeroshade:
### Rationale for this change
Boolean columns currently get double converted when transferring between Arrow and Parquet

### What changes are included in this PR?

**1. Arrow bitutil (`arrow/bitutil/bitmaps.go`)**
- Added `AppendBitmap()` method to `BitmapWriter`
- Directly copies bits from source bitmap using efficient `CopyBitmap()`

**2. Parquet encoder (`parquet/internal/encoding/boolean_encoder.go`)**
- Added `PutBitmap()` method to `PlainBooleanEncoder`
- Writes bitmap data directly without bool slice conversion

**3. Parquet decoder (`parquet/internal/encoding/boolean_decoder.go`)**
- Added `DecodeToBitmap()` method to `PlainBooleanDecoder`
- Reads directly into output bitmap
- Optimized fast path for byte-aligned cases

**4. Column writer (`parquet/file/column_writer_types.gen.go`)**
- Added `WriteBitmapBatch()` for non-nullable boolean columns
- Added `WriteBitmapBatchSpaced()` for nullable boolean columns
- Internal helper methods `writeBitmapValues()` and `writeBitmapValuesSpaced()`

**5. Arrow-Parquet bridge (`parquet/pqarrow/encode_arrow.go`)**
- Modified `writeDenseArrow()` to detect boolean arrays
- Uses bitmap methods when available
- Falls back to original `[]bool` path if needed (backward compatible)

### Are these changes tested?

Yes, and new benchmarks are added as appropriate

### Are there any user-facing changes?

Just performance:

### Non-Nullable Boolean Columns
```
BenchmarkBooleanBitmapWrite/1K-16          314847    19126 ns/op    6.54 MB/s    36057 B/op    237 allocs/op
BenchmarkBooleanBitmapWrite/10K-16         174715    33985 ns/op   36.78 MB/s    53266 B/op    247 allocs/op
BenchmarkBooleanBitmapWrite/100K-16         34099   175655 ns/op   71.16 MB/s   218866 B/op    340 allocs/op
BenchmarkBooleanBitmapWrite/1M-16            3778  1568818 ns/op   79.68 MB/s  1763712 B/op   1237 allocs/op
```

### Nullable Boolean Columns (10% null rate)
```
BenchmarkBooleanBitmapWriteNullable/1K-16   214921    28002 ns/op    4.46 MB/s    39706 B/op    249 allocs/op
[...]

PR #715, opened 2026-03-16
perf(parquet): optimize stats and bloom filters for bool columns

zeroshade:
### Rationale for this change
The initial bool optimization (#707) added direct bitmap encoding with `WriteBitmapBatch` and `WriteBitmapBatchSpaced` functions to reduce allocations during encoding and decoding. However, this didn't implement the same optimization for statistics updated, bloom filter inserts or spaced encoding.

1. Update the bloom filter handling to compute hashes directly from bitmaps, including for spaced values and reuse slice allocation across iterations.
2. Update statistics handling to update min/max directly from bitmaps for spaced and non-spaced scenarios.
3. Update the encoder interface for the boolean encoder to have a `PutSpacedBitmap` and compress bitmaps with validity buffers. 
4. Add `writeBitmapValues` and `writeBitmapValuesSpaced` for the boolean column writer with fallback to `[]bool` conversions if the encoder doesn't implement the interface.

### Are these changes tested?
Yes, unit tests are added for everything

No user-facing API changes, this is a pure internal optimization change for boolean columns.

*Benchmark Results*

*Statistics Update*
```
  BEFORE (with bitmap → []bool conversion):
  BenchmarkBooleanStatisticsWithConversion-16
    153,398 ns/op    109,278 B/op (107 KB)    6 allocs/op

  AFTER (direct bitmap operations):
  BenchmarkBooleanStatisticsDirectBitmap-16
        393 ns/op      2,698 B/op (2.6 KB)    5 allocs/op
```

* 390x faster 
* 97.5% less memory
* 1 fewer allocation

*Bloom filter hashing*
```
  BEFORE (with bitmap → []bool conversion):
  BenchmarkBloomFilterHashingWithConversion-16
    1,084,001 ns/op    3,309,593 B/op (3.23 MB)    3 allocs/op

  AFTER (direct bitmap operations):
  BenchmarkBloomFilterHashingDirectBitmap-16
      448,882 ns/op      802,821 B/op (784 KB)    2 allocs/op
```

* 2.4x faster
* 76% less memory
* 2.5 MB saved/operation
* 1 fewer allocation

*Full write path (Stats + Bloom Filter)*
```
  BEFORE (with bitmap → []bool conversion):
  BenchmarkFullWritePathWithConversion-16
    1,211,525 ns/op    3,315,566 B/op (3.24 MB)    15 allocs/op

  AFTER (direct bitmap operations):
  BenchmarkFullWritePathDirectBitmap-16
      580,934 ns/op      807,640 B/op (789 KB)    13 allocs/op
```

* 2.1x faster
* 76% less memory
* 2.5 MB saved per 100k bools written
* 2 fewer allocations

### DIFF
--- a/parquet/metadata/bloom_filter.go
+++ b/parquet/metadata/bloom_filter.go
@@ -155,6 +155,37 @@ func GetHashesFromBitmap(h Hasher, bitmap []byte, bitmapOffset int64, numValues 
 	return out
 }
 
+// GetSpacedHashesFromBitmap computes hashes for boolean values from a bitmap with validity information,
+// only hashing valid (non-null) values. Returns hashes for numValid values.
+// This avoids the 8x memory overhead of converting to []bool.
+func GetSpacedHashesFromBitmap(h Hasher, numValid int64, bitmap []byte, bitmapOffset int64, numValues int64, validBits []byte, validBitsOffset int64) []uint64 {
+	if numValid == 0 {
+		return []uint64{}
+	}
+
+	out := make([]uint64, 0, numValid)
+
+	// Use SetBitRunReader to efficiently iterate over valid values
+	setReader := bitutils.NewSetBitRunReader(validBits, validBitsOffset, numValues)
+	for {
+		run := setReader.NextRun()
+		if run.Length == 0 {
+			break
+		}
+
+		// Hash each valid bit in this run
+		for i := int64(0); i < run.Length; i++ {
+			val := bitutil.BitIsSet(bitmap, int(bitmapOffset+run.Pos+i))
+			var b [1]byte
+			if val {
+				b[0] = 1
+			}
+			out = append(out, h.Sum64(b[:]))
+		}
+	}
+	return out
+}
+
 func getBytes[T parquet.ColumnTypes](v T) []byte {
 	switch v := any(v).(type) {
 	case parquet.ByteArray:
--- a/parquet/metadata/bloom_filter.go
+++ b/parquet/metadata/bloom_filter.go
@@ -145,7 +145,7 @@ func GetHashesFromBitmap(h Hasher, bitmap []byte, bitmapOffset int64, numValues 
 	// Reuse a single-byte slice to avoid allocating per value
 	out := make([]uint64, numValues)
 	b := []byte{0}
-	for i := int64(0); i < numValues; i++ {
+	for i := range numValues {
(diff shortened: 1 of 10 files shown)


parquet/metadata/bloom_filter.go
```diff
@@ -134,6 +134,64 @@ func GetSpacedHashes[T parquet.ColumnTypes](h Hasher, numValid int64, vals []T,
...
+func GetHashesFromBitmap(h Hasher, bitmap []byte, bitmapOffset int64, numValues int64) []uint64 {
+	if numValues == 0 {
+		return []uint64{}
+	}
+
+	// Convert each bool bit to []byte for hashing
+	// Reuse a single-byte slice to avoid allocating per value
+	out := make([]uint64, numValues)
+	b := []byte{0}
+	for i := int64(0); i < numValues; i++ {
```
lidavidm: (nit: use the new range syntax? maybe enforce `go fix` in CI?)

parquet/metadata/bloom_filter.go
```diff
@@ -134,6 +134,64 @@ func GetSpacedHashes[T parquet.ColumnTypes](h Hasher, numValid int64, vals []T,
...
+	setReader := bitutils.NewSetBitRunReader(validBits, validBitsOffset, numValues)
+	for {
+		run := setReader.NextRun()
+		if run.Length == 0 {
+			break
+		}
+
+		// Hash each valid bit in this run
+		for i := int64(0); i < run.Length; i++ {
+			val := bitutil.BitIsSet(bitmap, int(bitmapOffset+run.Pos+i))
```
lidavidm, line 183: Can't you hoist this out of the loop? The point of a bit run is that the bits are all the same in the run, right?

parquet/metadata/bloom_filter.go
zeroshade: good point, looks like the go fix is gonna be a big change honestly so I'll hold off on that for the moment and just fix this one spot for this PR

parquet/metadata/bloom_filter.go
zeroshade: You're right that the point of a bit run is that the bits are all the same in the run, but we're doing runs of the validity bitmap here and then using the values from the value bitmap buffer. The loop is over a run of non-null bits in the bitmap, which means we have to actually check BitIsSet for each one. That said, I can use the range syntax for the loop

lidavidm (approved):

Issue #714, opened 2026-03-16
Support dictionary encoded arrays for Filter, Take


lesam:
### Describe the enhancement requested

A RecordBatch cannot have a 'filter' or 'take' selection applied if it contains a dictionary encoded array.

The filter can be basic- just filter on the index column of the dictionary and shallow-copy the dictionary itself.

I'm happy to have a try at sending a PR.

### Component(s)

Other

PR #716, opened 2026-03-16
feat: filter and take for dictionary arrays

lesam:
Add 'take' and 'filter' implementations for dictionary arrays. These implementations simply filter/take on the index column and shallow-copy the dictionary.

https://github.com/apache/arrow-go/issues/714

### Rationale for this change

It would be nice for 'take' and 'filter' to work for all array encodings, including dictionary

Just 'take' and 'filter' implementations and an addition to the tests

### Are these changes tested?

Unit tests added

These will now work instead of failing with an error like:

```
not implemented: function 'array_filter' has no kernel matching input types (dictionary<values=utf8, indices=int8, ordered=false>, bool)
```

### DIFF
--- a/arrow/compute/selection.go
+++ b/arrow/compute/selection.go
@@ -485,6 +485,40 @@ func structFilter(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResul
 	return nil
 }
 
+
+// dictionaryFilter is a special case for filtering a dictionary array
+//
+// The implementation uses the 'dictionaryTake' implementation.
+func dictionaryFilter(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResult) error {
+	// convert the filter (boolean array) to indices to take from the dictionary array.
+	indices, err := kernels.GetTakeIndices(exec.GetAllocator(ctx.Ctx),
+		&batch.Values[1].Array, ctx.State.(kernels.FilterState).NullSelection)
+	if err != nil {
+		return err
+	}
+	defer indices.Release()
+
+	filter := NewDatum(indices)
+	defer filter.Release()
+
+	valData := batch.Values[0].Array.MakeData()
+	defer valData.Release()
+
+	vals := NewDatum(valData)
+	defer vals.Release()
+
+	// run 'take' on the dictionary array, which will call dictionaryTake.
+	// we know the bounds are good because the indices were just created by GetTakeIndices
+	result, err := Take(ctx.Ctx, kernels.TakeOptions{BoundsCheck: false}, vals, filter)
+	if err != nil {
+		return err
+	}
+	defer result.Release()
+
+	out.TakeOwnership(result.(*ArrayDatum).Value)
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


zeroshade:
@lesam is this still a draft? Or is it ready for review?

lesam:
@zeroshade I've updated the PR with some comments, it's ready for review.

arrow/compute/vector_selection_test.go
```diff
@@ -1826,6 +1851,7 @@ func TestFilterKernels(t *testing.T) {
 	for _, dt := range baseBinaryTypes {
 		suite.Run(t, &FilterKernelWithString{dt: dt})
 	}
+	suite.Run(t, &FilterKernelWithString{dt: &arrow.DictionaryType{IndexType: arrow.PrimitiveTypes.Int8, ValueType: arrow.BinaryTypes.String}})
```
zeroshade, line 2025: Can we add tests for a few more combinations of different index/value types? Perhaps add a numeric dict test or a complex type like list/struct? Just to add more type coverage to the tests for this.

arrow/compute/selection.go
```diff
@@ -485,6 +485,40 @@ func structFilter(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResul
 	return nil
 }
 
+
+// dictionaryFilter is a special case for filtering a dictionary array
+//
+// The implementation uses the 'dictionaryTake' implementation.
+func dictionaryFilter(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResult) error {
```
zeroshade, line 492: Possible improvement, you could update this to mirror the `extensionFilterImpl` and just filter the indices array directly and then rewrap as a dictionary instead of doing bool->indices->Take.

It's a nitpick to eliminate a bit of overhead, but could also be done as a follow-up improvement. Not critical.

zeroshade (commented):
LGTM! Very nice, just a couple nitpicks so let me know what you think. 

Also, can we just add a single line documentation note explaining that the dictionary itself isn't compacted or modified? i.e. after execution of the filter, the dictionary itself would still contain all the same values. This is intentional, but might be non-obvious.

arrow/compute/selection.go
lesam: If you don't mind I think I will leave this to a follow-up - FWIW I was mirroring structFilter that also does bool -> indices -> Take

arrow/compute/vector_selection_test.go
lesam: I've added a commit with more tests

lesam:
> Also, can we just add a single line documentation note explaining that the dictionary itself isn't compacted or modified? i.e. after execution of the filter, the dictionary itself would still contain all the same values. This is intentional, but might be non-obvious.

I've added this on dictionaryTake and dictionaryFilter - did you want it on the public generic Take / Filter doc comment as well? My feeling is not to add it to the public method contract, to give more flexibility to change it in the future.

zeroshade (approved):
LGTM. Will merge once CI finishes

PR #726, opened 2026-03-24
Test and fix for bug in selectMapimpl

wwarner-inf:
### Rationale for this change

Maps of Nullables produce corrupted parquet.

The existing tests for `TakeKernelMap` are made runnable by correcting the test json to be suitable input for `array.FromJSON()`. Added a condition in `assertTakeArrays()` that checks that the data type has been preserved. Using the C++ implementation as a reference.

### Are these changes tested?
Yes, but not deeply.

The api is unchanged, however parquet output will change slightly for those files containing maps with optional values.

### DIFF
--- a/arrow/compute/selection.go
+++ b/arrow/compute/selection.go
@@ -26,7 +26,6 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/compute/exec"
 	"github.com/apache/arrow-go/v18/arrow/compute/internal/kernels"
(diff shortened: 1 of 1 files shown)


wwarner-inf:
Not an accident -- I thought maybe you wanted a clean demonstration of the problem. It's [here](https://github.com/apache/arrow-go/pull/726/commits/21f7a96a93b733d3d39d837de675ce761c4c52b2#diff-772e3a52d660a14b0c56730d0edb708a51b7b5097088869ba574346185e4ddd4) and sure I can put it back if that suits you!

PR #732, opened 2026-03-27
feat(arrow/flight/sql): Add is_update field to ActionCreatePreparedStatementResult

### LINKED ISSUE #705: [FlightSQL] - Add new is_update field to PreparedStatement

Following the [discussion in the mailing list](https://lists.apache.org/thread/88msflnwkkw8t81czs2ndqhkn1fb1pxd), the new field `is_update` sent in the `ActionCreatePreparedStatementResult` should be made available (if available) in the `PreparedStatement` type.

### Component(s)

Other

ennuite:
### Rationale for this change
Following the [discussion in the mailing list
](https://lists.apache.org/thread/88msflnwkkw8t81czs2ndqhkn1fb1pxd), this PR introduces a new field in `ActionCreatePreparedStatementResult`, `optional boolean is_update`, that allows servers to indicate to clients the correct execution path for prepared statements.

The new field `is_update` is added to `PreparedStatement`, as well as serialization/deserialization and getters.

`FlightSql.pb.go` was regenerated in order to include the changes in https://github.com/apache/arrow/commit/2cd5ad233fab33e693b4e764db0d45ca52e93ea0, with `protoc-gen-go v1.36.11` and `protoc v7.34.0`.

### Are these changes tested?
Yes. Unit tests for the new field were added to `client_test.go`, that complement the already existing unit tests for prepared statements queries (there were no tests for prepared statement updates). Integration tests for the new field were added to `server_test.go`. These are also the first integration tests for prepared statements.

Yes. 
There are breaking changes to unkeyed literals for the exported structs `flightsql.ActionCreatePreparedStatementResult` and `flightsql.CreatePreparedStatementResult`. 

There are also user-facing but non breaking changes: an extension to prepared statements was added to allow for reading/writing to the new field.

### AI Disclaimer
This change was created with AI assistance (Augment Code and Claude Code and Codex). All lines were manually reviewed by a human. The output is not copyrightable subject matter.

Closes #705


zeroshade:
I rebased this onto main which should solve the CI issues, I'll also re-review it

arrow/flight/flightsql/server_test.go
```diff
@@ -133,6 +133,78 @@ func (*testServer) DoGetStatement(ctx context.Context, ticket flightsql.Statemen
...
+	case "prepared query":
+		isUpdate := false
+		result.IsUpdate = &isUpdate
+	case "prepared update":
+		isUpdate := true
+		result.IsUpdate = &isUpdate
+	default:
+		err = fmt.Errorf("unknown query: %s", query)
+	}
+	if isUpdate != nil {
```
zeroshade: The `isUpdate := …` declarations inside the two cases above shadow the outer `var isUpdate *bool`, so this `if isUpdate != nil` always sees nil — it's dead code. The cases already set `result.IsUpdate` directly, so the outer `var isUpdate *bool` and this block can both be removed.

arrow/flight/flightsql/client.go
```diff
@@ -1348,6 +1351,11 @@ func (p *PreparedStatement) DatasetSchema() *arrow.Schema { return p.datasetSche
 // the prepared statement.
 func (p *PreparedStatement) ParameterSchema() *arrow.Schema { return p.paramSchema }
 
+// IsUpdate returns the server's hint about whether this prepared statement
+// should be executed as an update (true) or query (false). If nil, the server
+// did not provide a hint and the client can choose how to execute the statement.
+func (p *PreparedStatement) IsUpdate() *bool { return p.isUpdate }
```
zeroshade: Returning the internal `*bool` leaks mutable state (a caller can do `*ps.IsUpdate() = false` and mutate the statement) and isn't idiomatic for an optional. Prefer `IsUpdate() (val bool, ok bool)`, or return a copy.

zeroshade (changes requested):
Thanks for the PR and the thorough tests — the feature looks good.

_(Edit: withdrawing my earlier concern about the regenerated `FlightSql.pb.go` toolchain — that's fine as-is, no need to regenerate.)_

One thing to tidy before merge — see the inline comment on `server_test.go`: the `isUpdate := …` declarations inside the cases shadow the outer `var isUpdate *bool`, so the trailing `if isUpdate != nil` block is dead code. Dropping the outer var and that block keeps the helper correct and clear.

I also left an optional, non-blocking note on the `IsUpdate() *bool` getter shape — take it or leave it.

Happy to approve once the test helper is cleaned up.

zeroshade:
@ennuite are you still working on this?

ennuite:
@zeroshade Sorry, I will resume work on this now and will address the comments this week.

arrow/flight/flightsql/server_test.go
ennuite: Thank you for this review. 
Addressed in https://github.com/apache/arrow-go/pull/732/commits/346696955e501cdce0f575f2bd42fda894c621e4

arrow/flight/flightsql/server_test.go
```diff
@@ -133,6 +133,78 @@ func (*testServer) DoGetStatement(ctx context.Context, ticket flightsql.Statemen
...
+		result.IsUpdate = isUpdate
+	}
+	return
```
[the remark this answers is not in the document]
ennuite: Good catch. This was also highlighted by Matt and I addressed it in https://github.com/apache/arrow-go/pull/732#discussion_r3517096791

ennuite:
@zeroshade I addressed your review comments, and refactored some other poor patterns in the tests. 

I will put this as ready for review again once I update https://github.com/apache/arrow-adbc/pull/4161 with the changes I did now and ran end-to-end tests to ensure that everything is ok.

arrow/flight/flightsql/client.go
ennuite: Thank you for pointing this out. I took the idiomatic approach with the double boolean return, see https://github.com/apache/arrow-go/pull/732/changes/2c8c152988cb68a3d622d90bddbd3af962e27901

arrow/flight/flightsql/client.go
```diff
@@ -1105,13 +1107,14 @@ func (tx *Txn) RollbackSavepoint(ctx context.Context, sp Savepoint, opts ...grpc
 // prepared statement handle.
 //
 // If the server returned the Dataset Schema or Parameter Binding schemas
-// at creation, they will also be accessible from this object. Close
+// or isUpdate at creation, they will also be accessible from this object. Close
 // should be called when no longer needed.
```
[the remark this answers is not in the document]
ennuite: Addressed in https://github.com/apache/arrow-go/pull/732/changes/028867928ff59e7f4a86eda023b8576f63b0c8c6

arrow/flight/flightsql/client_test.go
```diff
@@ -423,6 +423,123 @@ func (s *FlightSqlClientSuite) TestPreparedStatementExecute() {
 	s.Equal(&emptyFlightInfo, info)
 }
 
+func (s *FlightSqlClientSuite) TestPreparedStatementExecuteWithIsUpdateFalse() {
```
zeroshade, line 431: **Please add a test for the unset/`nil` case.** This is my only blocking request.

Both new client tests here, and both new server tests, assert `ok == true`. Nothing anywhere asserts `ok == false`. That's the path taken by every server that doesn't implement the new field — i.e. the backward-compatibility contract — and it's currently the only untested branch of the getter.

`TestPreparedStatementExecute` above (line 368) already builds an `ActionCreatePreparedStatementResult` with no `IsUpdate`, so this can be as small as adding two lines there:

```go
val, ok := prepared.IsUpdate()
s.False(ok)
s.False(val)
```

Worth locking down explicitly because a future refactor that switched `isUpdate *bool` to a plain `bool` would silently turn "no hint" into "is a query" — which is a real behavior change for clients that branch on this — and no existing test would catch it.

arrow/flight/flightsql/client.go
```diff
@@ -566,6 +566,7 @@ func (c *Client) LoadPreparedStatementFromResult(result *CreatePreparedStatement
 		handle:        result.PreparedStatementHandle,
 		datasetSchema: dsSchema,
 		paramSchema:   paramSchema,
+		isUpdate:      result.IsUpdate,
```
zeroshade, line 569: Two small things on this propagation site.

1. **Untested.** `TestPreparedStatementLoadFromResult` (client_test.go:971) is the only coverage for `LoadPreparedStatementFromResult`, and it doesn't assert `IsUpdate`. Since that test already constructs the result and checks the other two propagated fields (`ParameterSchema`, `DatasetSchema`), adding an `IsUpdate` assertion there would be consistent and nearly free. (Couldn't leave this as an inline comment on that line — it's outside the diff.)

2. **Nit, no change required:** this is the one place that stores a `*bool` owned by the *caller*, so a caller mutating `result.IsUpdate` after this call would change the statement's view. `Prepare`/`PrepareSubstrait` don't have this property since they point at their own unmarshaled local. Not worth a defensive copy unless you'd want it for `handle` too — just noting it so the asymmetry is a conscious choice.

arrow/flight/flightsql/client.go
```diff
@@ -1365,6 +1368,17 @@ func (p *PreparedStatement) DatasetSchema() *arrow.Schema { return p.datasetSche
 // the prepared statement.
 func (p *PreparedStatement) ParameterSchema() *arrow.Schema { return p.paramSchema }
 
+// IsUpdate reports the server's hint for how to execute this prepared statement.
+// val is true if the server indicated an update, false if it indicated a query.
+// ok is false if the server did not provide a hint, in which case the client
+// can choose how to execute the statement.
+func (p *PreparedStatement) IsUpdate() (val bool, ok bool) {
```
zeroshade, line 1375: **Endorsing this API shape explicitly**, since it has no precedent elsewhere in `arrow/flight` and I'd rather it not get churned in a later review pass.

`(val, ok bool)` is the right call here:
- It's the direct analogue of the Java client's nullable `Boolean isUpdate()` (guarded by `hasIsUpdate()`), so we're consistent cross-language.
- Returning `*bool` would hand out internal state and invite mutation; a tri-state enum is overkill for one optional bool.
- It's plain idiomatic Go (`v, ok := m[k]`).

Also endorsing that this is **informational only** and that `Execute`/`ExecuteUpdate` don't consult it. That matches the spec ("If not set, the client can choose") and mirrors upstream layering: in Java it's the *JDBC adapter* that consumes the hint to pick `StatementType`, not `FlightSqlClient` itself. Our analogous consumer is arrow-adbc, outside this repo. Enforcing or warning on mismatch here would be a behavior change nobody asked for.

The doc comment is clear about all three states. Nice.

arrow/flight/flightsql/server.go
```diff
@@ -155,6 +155,9 @@ type ActionCreatePreparedStatementResult struct {
 	Handle          []byte
 	DatasetSchema   *arrow.Schema
 	ParameterSchema *arrow.Schema
+	// IsUpdate indicates whether the prepared statement should be executed
+	// as an update (true) or query (false). If nil, the client can choose.
+	IsUpdate *bool
```
zeroshade, line 160: Not a defect, but this needs a **release note**, because the public API impact is wider than this struct.

`types.go:882` has:

```go
type CreatePreparedStatementResult = pb.ActionCreatePreparedStatementResult
```

Being a type *alias*, the regenerated proto field lands on a public `flightsql` symbol too. So this PR adds an exported field to two publicly-reachable structs. Keyed composite literals are unaffected (and that's what I found in the downstream usage I spot-checked), but any downstream **unkeyed** literal like:

```go
return flightsql.ActionCreatePreparedStatementResult{handle, ds, ps}, nil
```

will stop compiling. That's permitted under the Go 1 compatibility rules — unkeyed literals of imported struct types are explicitly not covered — and `go vet`'s `composites` check flags them, so real-world exposure should be low. I don't think it should block the PR, but it's worth a line in the changelog rather than letting someone discover it at upgrade time.

The unconditional `result.IsUpdate = output.IsUpdate` on both the statement and Substrait branches is correct, by the way — both sides are `*bool`, so nil flows through and no presence check is needed. Good that the Substrait path wasn't forgotten.

arrow/flight/gen/flight/FlightSql.pb.go
```diff
@@ -17,19 +17,19 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
-// 	protoc        v4.25.3
+// 	protoc-gen-go v1.36.11
```
zeroshade, line 20: This regeneration jumped the toolchain: `protoc-gen-go v1.31.0 -> v1.36.11` and `protoc v4.25.3 -> v7.34.0`, while the sibling `Flight.pb.go` and `Flight_grpc.pb.go` stay on the old versions. That's what produces the `+765/-1633` diff on a one-field change.

**I am not asking you to revert it, and I want to record that I verified it's safe**, so this doesn't get re-litigated:
- I diffed every `protobuf:"..."` wire tag, every exported generated method, every generated type, and every enum constant against the true merge-base. The only deltas are the `is_update` tag and `GetIsUpdate()`. Zero other semantic change.
- Both the old and new files carry the same `protoimpl.EnforceVersion(20 ...)` guards, so this does **not** raise the protobuf runtime floor.
- `go.mod` already requires `google.golang.org/protobuf v1.36.11`, so the new codegen actually *matches* our pinned runtime — `Flight.pb.go` is the stale one, not this file.

Two requests, both cheap:
1. Please note the exact `protoc` / `protoc-gen-go` versions you used in the PR description, and if it's not already the case, keep the regeneration as its own commit so the feature commit stays reviewable in isolation.
2. Please **don't** also regenerate `Flight.pb.go` here — that belongs in a separate cleanup PR.

The root cause isn't yours: `arrow/flight/gen.go` generates from `-I../../../format` (the proto isn't vendored) and we have no CI check that generated files are current, so version drift like this is invited. I'll open a follow-up issue for pinning the generator and adding a regeneration check — no action needed from you.

arrow/flight/flightsql/server_test.go
```diff
@@ -133,6 +133,74 @@ func (*testServer) DoGetStatement(ctx context.Context, ticket flightsql.Statemen
 	return
 }
 
+func (*testServer) CreatePreparedStatement(ctx context.Context, req flightsql.ActionCreatePreparedStatementRequest) (result flightsql.ActionCreatePreparedStatementResult, err error) {
```
zeroshade, line 136: Since you're already switching on the query string here, adding a third case with **no** hint set would give end-to-end coverage of the `ok == false` path through the real server/client round-trip (which is stronger than the mock-level test I asked for in `client_test.go`):

```go
case "prepared no hint":
    // leave result.IsUpdate nil - server provides no hint
```

Then a short test asserting `IsUpdate()` returns `(false, false)`. Pairs naturally with the two tests you added below.

Thanks for wiring up `CreatePreparedStatement`/`DoGetPreparedStatement`/`DoPutPreparedStatementUpdate` on `testServer` — I confirmed this doesn't disturb the existing `TestDoGet/DoGetPreparedStatement` expectations in the unimplemented-server suite, and the whole `./arrow/flight/...` suite still passes.

arrow/flight/flightsql/server_test.go
```diff
@@ -287,6 +355,69 @@ func (s *FlightSqlServerSuite) TestExecuteChunkError() {
...
+			case 1:
+				s.False(data.IsNull(i))
+				s.Assert().Equal(int16(1), data.Value(i))
+			}
+			n++
+		}
+	}
+}
+
+func (s *FlightSqlServerSuite) TestExecutePreparedStatementUpdate() {
```
zeroshade, line 434: Nit / optional: the Substrait path (`CreatePreparedSubstraitPlan` on the server, `PrepareSubstrait` on the client) also carries the hint now but has no coverage. Residual risk is low since it shares `parsePreparedStatementResponse` with the path you did test, so I wouldn't hold the PR on it — just flagging in case you'd rather close the loop while you're in here.

Separately: the `s.Assert().Equal(true, x)` -> `s.True(x)` cleanups in this file are unrelated to the feature. They're strict improvements so I'm not going to ask you to strip them, but ideally they'd be a separate commit.

zeroshade (changes requested):
Thanks for pushing this through — and thanks especially for the thorough test additions; the prepared-statement integration tests are a genuine gap this fills independently of the feature itself.

I verified the change fairly aggressively and the **feature itself is correct**. Summary of what I checked:

| Check | Result |
|---|---|
| Spec conformance vs upstream `apache/arrow` `format/FlightSql.proto` | `optional bool is_update = 4`; doc comments match the spec verbatim |
| Upstream status | apache/arrow#49498, merged 2026-06-09 — canonical, not a proposal |
| Generated-file semantic delta (all wire tags / exported methods / types / enum constants, vs true merge-base) | Exactly **+1 wire tag, +1 getter, 0 type changes, 0 enum changes** |
| Runtime descriptor | `#4 is_update kind=bool hasPresence=true` |
| Presence round-trip | `true`, `false`, and unset all round-trip correctly — importantly **`false` survives the wire**, confirming real presence semantics |
| `go build` / `go vet` / `go test ./arrow/flight/...` | All clean; the 4 new tests confirmed executing |

So I want to be clear that the `-868` net lines in `FlightSql.pb.go` is **not** hiding anything — I checked that specifically.

I'm marking this `request changes` for one substantive reason only: **the unset/`nil` case is never asserted anywhere.** Every new test asserts `ok == true`, but `ok == false` is the backward-compatibility contract with the (currently overwhelming majority of) servers that don't send the hint. That's the one path most likely to regress silently and it's the cheapest possible test to add.

Everything else below is either a release-note item or a nit. Two design choices I want to explicitly **endorse** so they don't get second-guessed in later review rounds: the `(val, ok bool)` getter shape, and keeping the hint purely informational. Details inline.

arrow/flight/gen/flight/FlightSql.pb.go
zeroshade: Filed the toolchain follow-up as #1029 — pinning the generators, documenting where `format/*.proto` comes from, and adding a CI check that regenerates and fails on diff. Nothing there is asked of this PR; the two requests in this thread still stand as-is (note the versions you used, keep the regen in its own commit).

Also captured there: please **don't** regenerate `Flight.pb.go`/`Flight_grpc.pb.go` here — normalizing all four files is tracked as its own mechanical PR in #1029.

arrow/flight/flightsql/client_test.go
ennuite: Do we want the `s.False(val)` check? If `s.False(ok)`, then `val` is meaningless.

ennuite:
Thanks for the very thorough review! 

I'm going over the comments one by one, I might be a bit slow as I'm ramping up on some of the Go concepts, and your feedback has been a great learning experience.

I appreciate your patience as I improve the PR. I will continue working on it in the next few days, and at the end I'll try to tackle https://github.com/apache/arrow-go/issues/1029 as well.

arrow/flight/flightsql/client_test.go
ennuite: I addressed this in https://github.com/apache/arrow-go/pull/732/changes/eb160883859a2843fdfa2a76598393efac2eb9b1

I do not think it makes sense to have the check in `val`: it being `False` seems like an implementation detail, and not a part of the contract of the `IsUpdate() `method. Let me know if you disagree: I can change it to look exactly like your suggestion.

arrow/flight/flightsql/client.go
ennuite: 1. I addressed this with 2 tests. In `TestPreparedStatementLoadFromResult` I let things as they are, and tested with the new field unset, see https://github.com/apache/arrow-go/pull/732/changes/8ba207b4402f60ce27cbad99282eebb8c649e50e.  Note that this test has the same nuance about `val` that I pointed out in https://github.com/apache/arrow-go/pull/732#discussion_r3670708880

I added a new test that checks that the field is properly loaded also when set, see https://github.com/apache/arrow-go/pull/732/changes/a334b86f6c7dbc7e1fc33acc014e5b70b5201e23

2. For that one I purposefully followed the example set by `handle`, but I will admit I had my own reservations while doing it. 

I will not change this in this PR unless you ask me to, but in your opinion is it worth a follow-up PR to have a defensive copy for both `handle` and `isUpdate`? If so I can tackle that after this one

arrow/flight/flightsql/server_test.go
ennuite: Addressed this in https://github.com/apache/arrow-go/pull/732/changes/786adc73394a5d1f1987f3c122a6def3d28d1823 

I purposefully do not execute the query: since the hint is not set, it is up to the client to decide whether to execute with `DoGet` or `DoPut`. I do not want the test to misguide someone to think that one way is preferred.

The same nuance about checking `val` when `ok` is `False` is also present in this test (see https://github.com/apache/arrow-go/pull/732#discussion_r3670708880)

zeroshade:
sounds good! Just ping me here when you're ready for another review! feel free to ask any questions if you run into trouble

arrow/flight/flightsql/server_test.go
ennuite: 1. About the Substrait path: I added server-side tests at 4816b1b718db17f018f94f57d0f84ab1c64252ff 
I don't actually execute the queries because, apart from Prepared Statement creation, Substrait and SQL have the exact same execution path in Flight SQL, the difference is only in how the query is sent. Executing the queries would just be exercising the same code path unnecessarily.

I note that I didn't find any client-side tests for Substrait, and I think if we need them then they should go in a separate issue/PR. Do you agree?

2. Yeah, the cleanups outside my own tests were a Find->Replace mistake: I was trying to fix my own tests and did not realize I changed other ones. I'll be more careful next time.

ennuite:
Still not ready for a review as I still need to look into the protobuf stuff and the release notes, but a question popped in my mind (which is probably nitpicky).

What is your opinion on `s.Require.False(`) vs `s.False()`? I haven't found a canonical way to handle these (although I understand in how their behavior differs).

For the server tests, I'm always using `Require()` for every check of values returned by `PreparedStatement.IsUpdate()`, because the code after those checks (if there is any) just execute the query, and without proper values returned by IsUpdate we don't know how to properly execute the query (DoPut vs DoGet)

For the client tests:

1. In my new tests (`TestPreparedStatementExecuteWithIsUpdateFalse` and `TestPreparedStatementExecuteUpdateWithIsUpdateTrue`) I also use `Require()` using the same logic as above. 
2. For `TestPreparedStatementLoadFromResultWithIsUpdate`: given that we are just testing the loading and not execution, it would be interesting to see if `ok` was loaded incorrectly and `val` correctly, so I don't use `Require()`
3. For the pre-existing tests I don't use `Require` because there are interesting things to test even if these fields are broken (the tests were not created solely for this feature).

zeroshade:
> What is your opinion on s.Require.False() vs s.False()? I haven't found a canonical way to handle these (although I understand in how their behavior differs).

Generally my way of looking at it is pretty much what you describe, whether I find the subsequent tests still interesting/useful if the condition I'm asserting fails. Essentially if I'm testing a condition that if it fails I know that every subsequent assertion in the test will fail, then I just use `Require` so that it bails out early on the first failure. If I just want to assert something but it's still useful for diagnostic (or other test cases) for the test to continue, then I don't use `Require`

zeroshade:
@ennuite Converting to **draft** — this PR doesn't yet meet our [Pull Request quality criteria](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md).

- :x: **Unaddressed Copilot review** — an unresolved review thread from Copilot has been open since 2026-06-08 with no reply ([thread](https://github.com/apache/arrow-go/pull/732#discussion_r3374798100)). See [docs](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md).

> **Note:** Your branch is **109 commits behind `main`**. Please rebase and push again to get up-to-date CI results.

See the linked criteria for how to fix each item, then mark the PR "Ready for review". This is **not** a rejection — just an invitation to bring the PR up to standard. No rush.

---

_Note: This comment was drafted by an AI-assisted triage tool run by a maintainer, and may contain mistakes. Once you have addressed the points above, an Apache Arrow Go maintainer — a real person — will take the next look at your PR. If anything here looks wrong, say so on the PR and a maintainer will follow up. See [CONTRIBUTING.md](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md) for the project's contribution conventions._

arrow/flight/flightsql/server.go
```diff
@@ -1286,6 +1290,7 @@ func (f *flightSqlServer) DoAction(cmd *flight.Action, stream flight.FlightServi
 		if output.ParameterSchema != nil {
 			result.ParameterSchema = flight.SerializeSchema(output.ParameterSchema, f.mem)
 		}
+		// is_update is not relevant for prepared substrait plans
```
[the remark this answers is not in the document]
ennuite: Addressed in fcde3da8de20361c11ece62b2ba914d49cfd906f

arrow/flight/flightsql/server.go
ennuite: The comment mentions 2 distinct public structs. For clarity, you are asking about release notes for both of them, right? 

I updated the PR description with the breaking change. I looked into how release notes are generated, and it seems like at release time Github will use the PR title in the release notes, but not the description. Given that the description is the place where the breaking change is highlighted, do I need to add release notes elsewhere?

 **Explanation of the 2 structs** 
 
 If you agree with my PR description you don't need to read this detail, I leave it for clarity in case I misunderstood your ask.

The type alias refers to `flightsql.CreatePreparedStatementResult`. This is the raw proto version, what goes in the wire, and I think it's there so that users of the `flightsql` package don't need to import the proto package directly. In this codebase, this is only used in `LoadPreparedStatementFromResult()`

But the line about the unkeyed literal is referring to a different struct `flightsql.ActionCreatePreparedStatementResult` that is unrelated to the alias. It is defined in `server.go`, and it is a developer-friendly type with the `*arrow.Schema` objects decoded from wire format. It is used in quite a few places in the codebase.

Both of them fall under the situation of a public struct where unkeyed literals will break, so I put both in the PR description.

Thank you for pointing out the compatibility rules, for reference I found them at https://go.dev/doc/go1compat: 

> For the addition of features in later point releases, it may be necessary to add fields to exported structs in the API. Code that uses unkeyed struct literals (such as pkg.T{3, "x"}) to create values of these types would fail to compile after such a change. However, code that uses keyed literals (pkg.T{A: 3, B: "x"}) will continue to compile after such a change. We will update such data structures in a way that allows keyed struct literals to remain compatible, although unkeyed literals may fail to compile.

arrow/flight/gen/flight/FlightSql.pb.go
ennuite: The first commit is the only one that touches the regenerated proto file: https://github.com/apache/arrow-go/pull/732/changes/09e063c34af39c7f1349859e839aa19802b8bb87 It doesn't touch anything else.

I added the protobuf tools versions to the PR description.

ennuite:
Addressed all comments and rebased @zeroshade. Let me know what you think!

arrow/flight/flightsql/server_test.go
zeroshade: Yes, a separate issue/PR is fine.

zeroshade (approved):
LGTM, thanks for all the work here and getting this in!

arrow/flight/flightsql/server_test.go
ennuite: I opened a separate issue at https://github.com/apache/arrow-go/issues/1267 just to make sure it is tracked.

PR #746, opened 2026-04-03
fix(parquet/pqarrow): normalize the element name in the stored ARROW:schema

zeroshade:
### Rationale for this change
closes #744

Normalizes the element name of list fields in the stored ARROW:schema of a parquet file to avoid mismatches. 

### Are these changes tested?
Yes, a new unit test is added.

Yes, should be a bug fix to ensure consistency.

### DIFF
--- a/parquet/pqarrow/file_writer.go
+++ b/parquet/pqarrow/file_writer.go
@@ -38,41 +38,67 @@ import (
 // Specifically, list element field names are set to "element" because
 // ListOfWithName (used by fieldToNode) always names the Parquet element group
 // "element", regardless of the original Arrow element field name.
-func normalizeFieldForParquet(f arrow.Field) arrow.Field {
+func normalizeFieldForParquet(f arrow.Field) (arrow.Field, error) {
 	switch dt := f.Type.(type) {
 	case *arrow.ListType:
-		elem := normalizeFieldForParquet(dt.ElemField())
+		elem, err := normalizeFieldForParquet(dt.ElemField())
+		if err != nil {
+			return arrow.Field{}, err
+		}
 		elem.Name = "element"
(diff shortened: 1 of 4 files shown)


parquet/pqarrow/file_writer.go
```diff
@@ -33,6 +33,48 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// normalizeFieldForParquet recursively normalizes an Arrow field so that its
+// type matches the Parquet column structure that fieldToNode would produce.
+// Specifically, list element field names are set to "element" because
+// ListOfWithName (used by fieldToNode) always names the Parquet element group
+// "element", regardless of the original Arrow element field name.
+func normalizeFieldForParquet(f arrow.Field) arrow.Field {
```
lidavidm: Do we need to recurse down into run-end-encoded arrays and listviews and such too? (Should we use NestedType so we don't have to hardcode all the nested types every time we want to do structural recursion like this?)

zeroshade: We haven't implemented handling for run-end-encoded or list-views yet for pqarrow, so I'm okay putting that off to add here until we add that support.

For `NestedType`: it gives us `Fields()`/`NumFields()` for recursion but no generic reconstruction path. We'd still need type-specific cases to rebuild each type (e.g. `ListOfField` vs `FixedSizeListOfField` vs `StructOf`). So the switch here is likely better than trying to use `NestedType` and then having to switch on which nested type anyways.

lidavidm: Fair enough. Can we explicitly bail out so there's no chance of forgetting to update this later on? (I think we might consider a generic traverser API, this need comes up often enough, but apart from arrow-java this is something lacking from other major Arrow implementations too AFAIK :sweat_smile:)

zeroshade: Updated to explicitly bail

lidavidm (approved):

PR #749, opened 2026-04-04
feat(arrow/compute): sort support

### LINKED ISSUE #66: [Go][Table] Implement sort function

Seems the python library has a sort functionality for table - https://arrow.apache.org/cookbook/py/data.html#sort-a-table

Will it make sense to implement the same for the Go library (unless it is already implemented and I missed it)?

### Component(s)

Go

hamilton-earthscope:
## Summary

Implements stable `sort_indices` (and `sort` via `take`) for arrays, chunked arrays, record batches, and tables using logical row indices over `Chunked` data **without concatenating chunks**. The control flow and ordering rules are modeled on Apache Arrow C++ **`vector_sort.cc` / `vector_sort_internal.h`**, with a few **Go- and performance-driven** differences called out below.

## Parity with Arrow C++ (`vector_sort.cc` / `vector_sort_internal.h`)

**Same overall structure**

- **Single sort key, one column**
  - **Multiple chunks:** per-chunk sort then **pairwise merge** of sorted spans (C++ **ChunkedArraySorter** / **ChunkedMergeImpl** idea).
  - **Single chunk, no validity nulls and no null-likes:** direct stable sort on indices (C++ skips null partitioning when `null_count == 0` and there are no null-likes).
  - Otherwise: **partition validity nulls**, **partition float null-likes (NaN)**, stable sort of finite values, then **VisitConstantRanges**-style handling of ties (`vector_sort_internal.go`).

- **Multiple sort keys**
  - **`len(keys) <= kMaxRadixSortKeys` (8):** **MSD radix** path per record-batch range (`radixRecordBatchSortRange` ↔ **ConcreteRecordBatchColumnSorter::SortRange**).
  - **More than 8 keys:** **MultipleKeyRecordBatchSorter**-style global stable sort with lexicographic compare across keys (`multipleKeyRecordBatchSortRange`).
  - **Aligned chunk boundaries** across all keyed columns (typical table): sort **each chunk slice** with the same strategy, then **merge spans** like C++ **TableSorter** batch merge.

**Same ordering semantics (intended match to C++)**

- Per-key **ascending / descending** and **null placement** (including **NaN** as null-like for floats).
- **Stable** ordering: merge and `slices.SortStableFunc` are used so tie-breaking matches the C++ “left before right” stable merge behavior where documented in code.

**Same “column comparator” role**

- Go **`columnComparator`** interface ↔ C++ **`ColumnComparator`**: `compareRowsForKey`, null / null-like metadata, `columnHasValidityNulls` (skip **PartitionNullsOnly** when there are no validity nulls).

**Physical types**

- One **monomorphic** comparator type per supported physical pattern in **`vector_sort_physical.go`**, analogous to C++ **`ConcreteColumnComparator<T>`** (concrete `*array.T` + direct `Value` / `Cmp` / special cases for bool and intervals).

## Intentional differences and rationale

| Area                                           | C++                                                                                   | This Go port                                                                                                                                                                                                                                                                                                                              |
| ---------------------------------------------- | ------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Resolving logical row → (chunk, offset)**    | Chunk / resolver machinery in C++                                                     | **Dense `logicalRowMap`**: one `rowMapCell{chunk, local}` per logical row when `len(chunks) > 1`; **`pair(i,j)`** resolves two rows in one shot. **Why:** random compares during sort/merge need O(1) resolution; a flat table + co-located fields beats repeated resolver work and improves locality vs separate `chunk`/`local` slices. |
| **`physicalColumnBase` methods**               | N/A (different language)                                                              | **Pointer receivers** on `pair` / `isNullAtGlobal` / `cell`. **Why:** value receivers would copy slice headers (and map state) on every compare.                                                                                                                                                                                          |
| **Stable sort primitive**                      | `std::stable_sort`                                                                    | **`slices.SortStableFunc`** (Go 1.21+). **Why:** library primitive; semantics aligned with stable weak ordering used elsewhere in the port.                                                                                                                                                                                               |
| **Column dispatch at runtime**                 | Templates + virtuals                                                                  | **`columnComparator` interface** for “which column” in multi-key and merge loops. **Why:** idiomatic Go; per-type work stays in concrete `compareRowsForKey` implementations.                                                                                                                                                             |
| **Chunked merge with null-likes (e.g. float)** | C++ can **split** merge for null-like vs non-null-like regions (**ChunkedMergeImpl**) | **Single `less` over full row order** after per-chunk partitioning/sort. **Why:** simpler merge while preserving order as long as per-chunk phases match C++; documented in `vector_sort.go` comments.                                                                                                                                    |
| **Generics for physical columns**              | Templates instantiate fully                                                           | **Explicit monomorphs only** for the hot compare path. **Why:** measured regression vs Go generics on this hot path (inlining / assertions); verbosity traded for performance.                                                                                                                                                            |

## File Layout

- `arrow/compute/vector_sort.go` — `sort_indices` / `sort` registration and datum dispatch.
- `arrow/compute/vector_sort_test.go` — functional tests.
- `arrow/compute/internal/kernels/vector_sort.go` — orchestration, merge, `SortIndices` kernel.
- `arrow/compute/internal/kernels/vector_sort_internal.go` — null partitions, radix / multi-key batch sort.
- `arrow/compute/internal/kernels/vector_sort_support.go` — `logicalRowMap` and ordering helpers.
- `arrow/compute/internal/kernels/vector_sort_physical.go` — per-type column comparators.
- `arrow/compute/internal/kernels/vector_sort_bench_test.go` — benchmarks.

## Testing

- `go test ./arrow/compute -run TestSort -count=1`
- Benchmarks: `go test ./arrow/compute/internal/kernels -bench=BenchmarkSortIndices -benchmem` .

## References

- Arrow C++: `cpp/src/arrow/compute/kernels/vector_sort.cc` and `vector_sort_internal.h` (and related comparators).
  - https://github.com/apache/arrow/blob/main/cpp/src/arrow/compute/kernels/vector_sort.cc
  - https://github.com/apache/arrow/blob/main/cpp/src/arrow/compute/kernels/vector_sort_internal.h
  
## Related Issues
- Closes #66

### DIFF
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -383,8 +383,10 @@ func sortIndicesMultiColumnAlignedChunks(indices []uint64, offs []int, comparato
 }
 
 // SortIndices returns a stable permutation of 0..n-1 that would lexicographically sort the given
-// columns. Each *arrow.Chunked is used via its .Chunks() only—no concatenate. columns[i] pairs with
-// keys[i] for order and null placement on that column.
+// columns. Each *arrow.Chunked is used via its .Chunks() only—no concatenate.
+//
+// Important: columns[i] pairs with keys[i] for order and null placement on that column.
+// This kernel expects the public API to have already extracted the relevant columns from the input batch/table.
 func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) (*exec.ExecResult, error) {
 	if len(columns) == 0 || len(keys) == 0 {
 		return nil, fmt.Errorf("%w: must have at least one column and one sort key", arrow.ErrInvalid)
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -128,6 +128,9 @@ func extensionStorageFixedSizeBinaryChunks(chunks []arrow.Array) ([]arrow.Array,
 			return nil, fmt.Errorf("%w: extension column must implement array.ExtensionArray", arrow.ErrInvalid)
 		}
 		st := ext.Storage()
+
+		// TODO: allow individual extension types to sort themselves properly
+
 		if st.DataType().ID() != arrow.FIXED_SIZE_BINARY {
 			return nil, fmt.Errorf("%w: sorting extension columns is only supported when storage is fixed_size_binary (got %s)",
 				arrow.ErrNotImplemented, st.DataType())
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build go1.18
+//go:build go1.22
 
 package kernels
 
@@ -146,8 +146,8 @@ func newFixedSizeBinaryComparator(chunks []arrow.Array, numRows int, vn bool) (c
 		return nil, fmt.Errorf("%w: expected *array.FixedSizeBinary chunk", arrow.ErrInvalid)
 	}
 	w := f0.DataType().(*arrow.FixedSizeBinaryType).ByteWidth
-	for i := 1; i < len(chunks); i++ {
-		fi, ok := chunks[i].(*array.FixedSizeBinary)
+	for _, chunk := range chunks[1:] {
+		fi, ok := chunk.(*array.FixedSizeBinary)
 		if !ok {
 			return nil, fmt.Errorf("%w: expected *array.FixedSizeBinary chunk", arrow.ErrInvalid)
 		}
@@ -315,7 +315,7 @@ func alignedChunkBoundaries(columns []*arrow.Chunked) ([]int, bool) {
 		return nil, false
 	}
 	offs := make([]int, n+1)
-	for i := 0; i < n; i++ {
+	for i := range n {
 		chunkLength := ch0[i].Len()
 		for _, col := range columns[1:] {
 			cj := col.Chunks()
@@ -363,7 +363,7 @@ func sortIndicesSingleColumnChunked(indices []uint64, chunks []arrow.Array, comp
 func sortIndicesMultiColumnAlignedChunks(indices []uint64, offs []int, comparators []columnComparator, keys []SortKey, multiComp *multiColumnComparator, tmp []uint64, spanScratch []chunkIndexSpan) {
 	nChunks := len(offs) - 1
 	useRadix := len(keys) <= maxRadixSortKeys
-	for c := 0; c < nChunks; c++ {
+	for c := range nChunks {
 		lo, hi := offs[c], offs[c+1]
 		if useRadix {
 			radixRecordBatchSortRange(indices, tmp, comparators, keys, 0, lo, hi)
@@ -376,7 +376,7 @@ func sortIndicesMultiColumnAlignedChunks(indices []uint64, offs []int, comparato
 	}
 	less := func(a, b uint64) bool { return multiComp.compare(a, b) < 0 }
 	spans := make([]chunkIndexSpan, nChunks)
-	for c := 0; c < nChunks; c++ {
+	for c := range nChunks {
 		spans[c] = chunkIndexSpan{offs[c], offs[c+1]}
 	}
 	pairwiseMergeSortedSpans(indices, tmp, spans, less, spanScratch)
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 
 	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/compute/exec"
 )
 
@@ -119,6 +120,43 @@ func (m *multiColumnComparator) compare(idxA, idxB uint64) int {
 	return 0
 }
 
+func extensionStorageFixedSizeBinaryChunks(chunks []arrow.Array) ([]arrow.Array, error) {
+	out := make([]arrow.Array, len(chunks))
+	for i, ch := range chunks {
+		ext, ok := ch.(array.ExtensionArray)
+		if !ok {
+			return nil, fmt.Errorf("%w: extension column must implement array.ExtensionArray", arrow.ErrInvalid)
+		}
+		st := ext.Storage()
+		if st.DataType().ID() != arrow.FIXED_SIZE_BINARY {
+			return nil, fmt.Errorf("%w: sorting extension columns is only supported when storage is fixed_size_binary (got %s)",
+				arrow.ErrNotImplemented, st.DataType())
+		}
+		out[i] = st
+	}
+	return out, nil
+}
+
+func newFixedSizeBinaryComparator(chunks []arrow.Array, numRows int, vn bool) (columnComparator, error) {
+	f0, ok := chunks[0].(*array.FixedSizeBinary)
+	if !ok {
+		return nil, fmt.Errorf("%w: expected *array.FixedSizeBinary chunk", arrow.ErrInvalid)
+	}
+	w := f0.DataType().(*arrow.FixedSizeBinaryType).ByteWidth
+	for i := 1; i < len(chunks); i++ {
+		fi, ok := chunks[i].(*array.FixedSizeBinary)
+		if !ok {
+			return nil, fmt.Errorf("%w: expected *array.FixedSizeBinary chunk", arrow.ErrInvalid)
+		}
+		wi := fi.DataType().(*arrow.FixedSizeBinaryType).ByteWidth
+		if wi != w {
+			return nil, fmt.Errorf("%w: fixed_size_binary chunks must have the same byte width (%d vs %d)",
+				arrow.ErrInvalid, w, wi)
+		}
+	}
+	return newPhysicalSortFixedSizeBinaryColumn(chunks, numRows, vn), nil
+}
+
 // createChunkedComparator builds a column comparator for these chunks (one Arrow type for all chunks).
 func createChunkedComparator(chunks []arrow.Array, numRows int) (columnComparator, error) {
 	if len(chunks) == 0 {
@@ -189,6 +227,14 @@ func createChunkedComparator(chunks []arrow.Array, numRows int) (columnComparato
 		return newPhysicalSortBinaryColumn(chunks, numRows, validityNulls), nil
 	case arrow.LARGE_BINARY:
 		return newPhysicalSortLargeBinaryColumn(chunks, numRows, validityNulls), nil
+	case arrow.FIXED_SIZE_BINARY:
+		return newFixedSizeBinaryComparator(chunks, numRows, validityNulls)
+	case arrow.EXTENSION:
+		storageChunks, err := extensionStorageFixedSizeBinaryChunks(chunks)
+		if err != nil {
+			return nil, err
+		}
+		return newFixedSizeBinaryComparator(storageChunks, numRows, validityNulls)
 	default:
 		return nil, fmt.Errorf("%w: sorting not supported for type %s", arrow.ErrNotImplemented, typeID)
 	}
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -423,7 +423,7 @@ func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) 
 	out.Nulls = 0
 
 	buf := ctx.Allocate(int(length) * arrow.Uint64SizeBytes)
-	indices := arrow.Uint64Traits.CastFromBytes(buf.Buf())[:length]
+	indices := arrow.GetData[uint64](buf.Buf())[:length]
 
 	for i := range indices {
 		indices[i] = uint64(i)
@@ -433,7 +433,7 @@ func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) 
 		chunks := columns[0].Chunks()
 		if len(chunks) > 1 {
 			tmpBuf := ctx.Allocate(nRows * arrow.Uint64SizeBytes)
-			tmp := arrow.Uint64Traits.CastFromBytes(tmpBuf.Buf())[:nRows]
+			tmp := arrow.GetData[uint64](tmpBuf.Buf())[:nRows]
 			spanScratch := make([]chunkIndexSpan, len(chunks))
 			sortIndicesSingleColumnChunked(indices, chunks, comparators[0], keys[0], tmp, spanScratch)
 		} else {
@@ -443,7 +443,7 @@ func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) 
 				slices.SortStableFunc(indices, func(a, b uint64) int { return c0.compareRowsForKey(a, b, k0) })
 			} else {
 				tmpBuf := ctx.Allocate(nRows * arrow.Uint64SizeBytes)
-				tmp := arrow.Uint64Traits.CastFromBytes(tmpBuf.Buf())[:nRows]
+				tmp := arrow.GetData[uint64](tmpBuf.Buf())[:nRows]
 				arraySortOneColumnRange(indices, tmp, c0, k0, 0, nRows)
 			}
 		}
@@ -457,7 +457,7 @@ func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) 
 		multiChunkMerge := aligned && nSeg > 1
 
 		tmpBuf := ctx.Allocate(nRows * arrow.Uint64SizeBytes)
-		tmp := arrow.Uint64Traits.CastFromBytes(tmpBuf.Buf())[:nRows]
+		tmp := arrow.GetData[uint64](tmpBuf.Buf())[:nRows]
 
 		var spanScratch []chunkIndexSpan
 		if multiChunkMerge {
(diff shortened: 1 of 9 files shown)


arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+const (
+	NullsAtEnd NullPlacement = iota
+	NullsAtStart
+)
+
+// SortOptions defines options for the sort_indices function.
+type SortOptions struct {
+	Order         SortOrder     `compute:"order"`
+	NullPlacement NullPlacement `compute:"null_placement"`
+}
```
zeroshade, line 50: the `compute` struct tags are unnecessary as there is no longer anything that expects them or will use them.

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+
+func extensionStorageFixedSizeBinaryChunks(chunks []arrow.Array) ([]arrow.Array, error) {
+	out := make([]arrow.Array, len(chunks))
+	for i, ch := range chunks {
+		ext, ok := ch.(array.ExtensionArray)
+		if !ok {
+			return nil, fmt.Errorf("%w: extension column must implement array.ExtensionArray", arrow.ErrInvalid)
+		}
+		st := ext.Storage()
+		if st.DataType().ID() != arrow.FIXED_SIZE_BINARY {
```
zeroshade, line 134: this is fine for now, but we might want to enable individual extension types to allow specifying how to sort themselves properly, let's leave a TODO here.

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+	if len(columns) == 0 {
+		return nil, false
+	}
+	ch0 := columns[0].Chunks()
+	n := len(ch0)
+	if n == 0 {
+		return nil, false
+	}
+	offs := make([]int, n+1)
+	for i := 0; i < n; i++ {
```
zeroshade:
```suggestion
	for i := range n {
```

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		L := ch0[i].Len()
```
zeroshade: please pick a better variable name than just `L`

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		for j := 1; j < len(columns); j++ {
+			cj := columns[j].Chunks()
```
zeroshade:
```suggestion
		for _, col := range columns[1:] {
		    cj := col.Chunks()
```

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+	pairwiseMergeSortedSpans(indices, tmp, spans, less, spanScratch)
+}
+
+// sortIndicesMultiColumnAlignedChunks sorts each aligned chunk (C++ RadixRecordBatchSorter or
+// MultipleKeyRecordBatchSorter), then merges like Arrow C++ TableSorter.
+func sortIndicesMultiColumnAlignedChunks(indices []uint64, offs []int, comparators []columnComparator, keys []SortKey, multiComp *multiColumnComparator, tmp []uint64, spanScratch []chunkIndexSpan) {
+	nChunks := len(offs) - 1
+	useRadix := len(keys) <= maxRadixSortKeys
+	for c := 0; c < nChunks; c++ {
```
zeroshade:
```suggestion
	for c := range nChunks {
```

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		} else {
+			multipleKeyRecordBatchSortRange(indices, tmp, comparators, keys, lo, hi, makeTailComparator(comparators, keys, 1))
+		}
+	}
+	if nChunks <= 1 {
+		return
+	}
+	less := func(a, b uint64) bool { return multiComp.compare(a, b) < 0 }
+	spans := make([]chunkIndexSpan, nChunks)
+	for c := 0; c < nChunks; c++ {
```
zeroshade:
```suggestion
	for c := range nChunks {
```

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+// keys[i] for order and null placement on that column.
+func SortIndices(ctx *exec.KernelCtx, columns []*arrow.Chunked, keys []SortKey) (*exec.ExecResult, error) {
+	if len(columns) == 0 || len(keys) == 0 {
+		return nil, fmt.Errorf("%w: must have at least one column and one sort key", arrow.ErrInvalid)
+	}
+
+	if len(columns) != len(keys) {
+		return nil, fmt.Errorf("%w: number of columns (%d) must match number of sort keys (%d)",
+			arrow.ErrInvalid, len(columns), len(keys))
+	}
```
zeroshade, line 398: why must it match the number of sort keys? Shouldn't it just be `len(columns) >= len(keys)`?

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+
+	length := int64(columns[0].Len())
+	for _, col := range columns {
+		if int64(col.Len()) != length {
+			return nil, fmt.Errorf("%w: all columns must have the same length", arrow.ErrInvalid)
+		}
+	}
+
+	comparators := make([]columnComparator, len(columns))
```
zeroshade, line 407: `len(keys)`?

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		keys:    keys,
+	}
+
+	out := &exec.ExecResult{}
+	out.Len = length
+	out.Type = arrow.PrimitiveTypes.Uint64
+	out.Nulls = 0
+
+	buf := ctx.Allocate(int(length) * arrow.Uint64SizeBytes)
+	indices := arrow.Uint64Traits.CastFromBytes(buf.Buf())[:length]
```
zeroshade: use `arrow.GetData[uint64](buf.Buf())[:length]` now instead.

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+
+	for i := range indices {
+		indices[i] = uint64(i)
+	}
+
+	if len(keys) == 1 {
+		chunks := columns[0].Chunks()
```
zeroshade, line 435: shouldn't this use keys[0].ColumnIndex instead of 0?

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		if len(chunks) > 1 {
+			tmpBuf := ctx.Allocate(nRows * arrow.Uint64SizeBytes)
+			tmp := arrow.Uint64Traits.CastFromBytes(tmpBuf.Buf())[:nRows]
```
zeroshade: same as above, use `arrow.GetData[uint64](tmpBuf.Buf())[:nRows]`

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+			spanScratch := make([]chunkIndexSpan, len(chunks))
+			sortIndicesSingleColumnChunked(indices, chunks, comparators[0], keys[0], tmp, spanScratch)
```
zeroshade, line 440: same as above, shouldn't it be `comparators[keys[0].ColumnIndex]`?

arrow/compute/internal/kernels/vector_sort.go
```diff
@@ -0,0 +1,476 @@
...
+		} else {
+			k0 := keys[0]
+			c0 := comparators[0]
```
zeroshade, line 443: same comment as above

arrow/compute/internal/kernels/vector_sort_internal.go
```diff
@@ -0,0 +1,280 @@
...
+	return nullPartitionIndices{lo, finEnd, finEnd, hi}
+}
+
+// nullPartitionNoValidityNulls is PartitionNullsOnly when null_count == 0 (vector_sort_internal.h).
+func nullPartitionNoValidityNulls(lo, hi int, nullPlacement NullPlacement) nullPartitionIndices {
+	if nullPlacement == NullsAtStart {
+		return nullPartitionIndices{lo, hi, lo, lo}
+	}
+	return nullPartitionIndices{lo, hi, hi, hi}
+}
```
zeroshade, line 148: this is identical to some conditions above, can we just call this there instead of duplicating the lines? like lines 45, 95, 108 etc

arrow/compute/internal/kernels/vector_sort_internal.go
```diff
@@ -0,0 +1,280 @@
...
+
+// visitConstantRanges mirrors VisitConstantRanges in vector_sort.cc.
+func visitConstantRanges(indices []uint64, offsetLo, offsetHi int, key SortKey, comp columnComparator, visit func(rs, re int)) {
```
zeroshade: couldn't we simplify this by just passing `indices[offsetLo:offsetHi]` into this instead of having to do all the complex logic below?

arrow/compute/internal/kernels/vector_sort_internal.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// multipleKeyRecordBatchSortRange mirrors MultipleKeyRecordBatchSorter::SortInternal in vector_sort.cc:
+// partition nulls / null-likes on the first key, stable_sort non-null finites with tail comparator.
+func multipleKeyRecordBatchSortRange(indices []uint64, scratch []uint64, comparators []columnComparator, keys []SortKey, lo, hi int, tail func(a, b uint64) int) {
+	if hi-lo <= 1 {
+		return
+	}
+	key := keys[0]
+	comp := comparators[0]
```
zeroshade, line 213: same comment as before, shouldn't we be using `key.ColumnIndex`?

arrow/compute/internal/kernels/vector_sort_physical.go
```diff
@@ -0,0 +1,873 @@
...
+	}
+	return compareOrdered(key.Order, a.Value(li), b.Value(lj))
+}
+
+func (c *physicalSortInt8Column) isNullAt(row uint64) bool { return c.base.isNullAtGlobal(row) }
+func (c *physicalSortInt8Column) hasNullLikeValues() bool  { return false }
+func (c *physicalSortInt8Column) isNullLikeAt(uint64) bool { return false }
+func (c *physicalSortInt8Column) columnHasValidityNulls() bool {
+	return c.base.columnHasValidityNulls()
+}
```
zeroshade, line 105: can't we use go generics here and just embed the base?

i.e.

```go
type physicalSortColumn[T arrow.ValueTypes] struct { physicalColumnBase }

...

func (c *physicalSortColumn[T]) compareRowsForKey(i, j uint64, key SortKey) int {
	ai, aj, li, lj := c.pair(i, j)
	a := ai.(arrow.TypedArray[T])
	b := aj.(arrow.TypedArray[T])
	if c.validityNulls {
		if v, stop := compareKeyedNulls(a.IsNull(li), b.IsNull(lj), key); stop {
			return v
		}
	}
	return compareOrdered(key.Order, a.Value(li), b.Value(lj))
}

func (c *physicalSortColumn[T]) isNullAt(row uint64) bool { return c.isNullAtGlobal(row) }
```

etc.

arrow/compute/internal/kernels/vector_sort_support.go
```diff
@@ -0,0 +1,211 @@
...
+
+func compareCmperOrdered[T interface{ Cmp(T) int }](order SortOrder, vi, vj T) int {
+	c := vi.Cmp(vj)
+	if order == Descending {
+		return -c
+	}
+	return c
+}
+
+func compareDayTimeOrdered(order SortOrder, vi, vj arrow.DayTimeInterval) int {
```
zeroshade: would it make more sense to add a `Cmp` method to these types instead of defining them separately?

arrow/compute/vector_sort.go
```diff
@@ -0,0 +1,391 @@
...
+		key := inputSortKeys[0]
+		key.ColumnIndex = 0
+		sortKeys = []kernels.SortKey{key}
+
+	case KindRecord:
+		batch := input.(*RecordDatum).Value
+
+		sortColumns = make([]*arrow.Chunked, len(inputSortKeys))
+		needsRelease = make([]bool, len(inputSortKeys))
+		for i, key := range inputSortKeys {
```
zeroshade, line 159: if we're pre-ordering by the sort keys ColumnIndex fields, we should at least document that in the internal kernel implementation that this is assumed to have happened already.

arrow/compute/vector_sort.go
```diff
@@ -0,0 +1,391 @@
...
+		for i, shouldRelease := range needsRelease {
+			if shouldRelease {
+				sortColumns[i].Release()
+			}
+		}
+	}()
+
+	allocator := exec.GetAllocator(ctx)
+	// Kernel expects exec allocator on context; avoid canceling user ctx mid-sort.
+	execCtx := &exec.KernelCtx{Ctx: exec.WithAllocator(context.Background(), allocator)}
```
zeroshade: i'm confused, why would we cancel the user ctx mid-sort? Doing this means that if the input ctx gets cancelled on us, we won't honor it and cancel the operation. That's a problem

zeroshade (changes requested):

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: The tests will not compile the suggested syntax without moving the build constraint at the top of the file to golang 1.22+. (note I'm using golang 1.25 to develop)

Most (all?) of the source files in `arrow-go` specify a build constraint of `//go:build go1.18` at the top of the file. If I leave the constraint as `//go:build go1.18` then I can't compile. If I change the constraint to `//go:build go1.22` then it will.

How do you suggest we proceed? I'm unsure of the implications of having this one file specify `//go:build go1.22` while the remainder still specify `//go:build go1.18`. Would that leave undefined symbols for golang versions <1.22?

Seems like we should either
1. upgrade all the source files to build constraint 1.22 (outside the scope of this MR) or
2. leave the legacy loop syntax

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: This kernel function does not receive the whole table/record with all of its columns - it expects `keys` and `columns` to be parallel arrays containing only the columns that are being sorted, where `keys[0]` sorts `columns[0]` and so on.

The public methods plumb through `sortIndicesImpl` which extracts the columns being sorted and passes to `kernels.SortIndices` (this method).

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: they are equivalent based on answer [here](https://github.com/apache/arrow-go/pull/749#discussion_r3048254469)

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: they are equivalent based on answer [here](https://github.com/apache/arrow-go/pull/749#discussion_r3048254469)

arrow/compute/internal/kernels/vector_sort.go
zeroshade: It's fine to update the build constraint on these files to //go:build go1.22

zeroshade: As long as all the files that reference the symbols used are also `//go:build go1.22` then we're fine. i.e. every new file you created here.

arrow/compute/internal/kernels/vector_sort.go
zeroshade: update the documentation so that this is obvious that the `ColumnIndex` is only utilized in pre-processing *before* calling the kernel

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: Does that mean `arrow/compute/registry.go` as well?

arrow/compute/vector_sort.go
hamilton-earthscope: done. let me know if the docstring on the kernel method is insufficient.

arrow/compute/internal/kernels/vector_sort_support.go
hamilton-earthscope: much cleaner!

arrow/compute/internal/kernels/vector_sort_physical.go
hamilton-earthscope: We certainly can and I made several attempts at a generic solution to avoid code duplication. Correctness was straightforward given the test suite, but I failed to match the performance of these verbose single-use classes via go's generics. The closest I got was ~25% slower. I opted for the performance over maintenance thinking this would be in the hot path for some critical operations (sorted iceberg writes/maintenance activities).

When I say performance, I mean the following benchmark test suite. It is rather limited in that it's only testing int64 and string, but the results were consistent across the types.

`go test ./arrow/compute/internal/kernels -bench=BenchmarkSortIndices -benchmem`

arrow/compute/internal/kernels/vector_sort.go
zeroshade: I think that's fine personally.

arrow/compute/internal/kernels/vector_sort_physical.go
zeroshade: Interesting, I'm surprised that there was that much of a slowdown using generics. I'll take a look at this but for now I agree that the performance is definitely a higher priority. Maybe we set this up like elsewhere that we do codegen? Just so that we can prevent future bugs if we change this logic

arrow/compute/internal/kernels/vector_sort.go
hamilton-earthscope: bumped `arrow/compute/registry.go` build constraint to 1.22 as well.

zeroshade:
@hamilton-earthscope  Can you please rebase this from main?

zeroshade (approved):
This is awesome, thanks much!

PR #758, opened 2026-04-10
`UnsafeAppendBoolToBitmap` for dictionary and REE builders

serramatutu:
### Rationale for this change

For more context: https://github.com/apache/arrow-go/pull/558/changes#r2758798540

The Builders for REE and dict-encoded arrays were inheriting the `UnsafeAppendBoolToBitmap()` and `UnsafeAppend()` from the `Builder` interface defined in `arrow/array/builder.go`. The default implementation does not bump the length of the inner `idxBuilder` or `runEndsBuilder`, which causes it to be in an inconsistent state where the parent has greater length than the index/run-ends builder. This causes a panic when instancing a record batch.

This PR makes `UnsafeAppendBoolToBitmap()` for `RunEndEncodedBuilder` panic as there is not a good semantic meaning for this method in case of REE.

It also implements `UnsafeAppendBoolToBitmap()` and `UnsafeAppend()`  for the Dictionary builders. This allows users to use `bldr.Reserve(n)` followed by `n` calls to `UnsafeAppend()` or `UnsafeAppendBoolToBitmap()` without the need to check if the array needs to grow everytime.

### Are these changes tested?

Yes.

No. The API was already there (inherited from `Builder`), but the behavior was incorrect.

### DIFF
--- a/arrow/array/dictionary.go
+++ b/arrow/array/dictionary.go
@@ -314,7 +314,8 @@ func arrayApproxEqualDict(l, r *Dictionary, opt equalOption) bool {
 // helper for building the properly typed indices of the dictionary builder
 type IndexBuilder struct {
 	Builder
-	Append func(int)
+	Append       func(int)
+	UnsafeAppend func(int)
 }
 
 func createIndexBuilder(mem memory.Allocator, dt arrow.FixedWidthDataType) (ret IndexBuilder, err error) {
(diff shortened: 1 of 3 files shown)


zeroshade:
Can you fix the lint issues?

zeroshade:
You might also need to rebase from main

serramatutu:
@zeroshade just rebased and fixed all the errors. 

Lint is still failing due to unformatted import on main: https://github.com/apache/arrow-go/blob/176940535f1335a6dba98f8830bde49232ef32fb/arrow/datatype_fixedwidth.go#L20

I can change it here if you want me to :)

arrow/array/dictionary.go
```diff
@@ -990,6 +1030,26 @@ type dictBuilder[T arrow.ValueType] struct {
 	dictionaryBuilder
 }
 
+func (b *dictBuilder[T]) UnsafeAppend(v T) error {
```
serramatutu: NOTE: Since `Append()` returns `error`, I had done the same thing for and `UnsafeAppend()`. However, I noticed the only thing that could return an error from `appendValue`, `appendBytes` or `unsafeAppendValue` is `b.memoTable.GetOrInsert(val)` since `val` is `interface{}` and the type ID could be wrong.

Since `v` is typed explicitly with `T` in `Append()`/`UnsafeAppend()`, it is safe to assume there should never be any error, unless there is an implementation bug, in which case a panic is better. 

In the latest commit, I changed `UnsafeAppend()` to be infallible, and added `TODO` comments for all the other `Append()`, as that would be a breaking change to the public API.

What do you think?

zeroshade:
> Lint is still failing due to unformatted import on main:

I actually just merged that fix. So you can just rebase it to fix the lint :)

zeroshade (approved):
I rebased with the lint fix, once CI passes this is good and I'll merge

PR #763, opened 2026-04-13
chore: replace golang.org/x/xerrors for stdlib errors

thaJeztah:
relates to:

- https://github.com/apache/arrow/pull/5997
- https://github.com/apache/arrow/issues/16992

The golang.org/x/xerrors module is a transitional module for what became go1.13 errors. Most of its functionality is now available in stdlib errors, with the exception of stdlib not providing a stack-trace (and some utility functions related to that).

Looking at history to see if that functionality was essential for using this package, I found that the golang.org/x/errors module was introduced in [apache/arrow@7126fdb], to address [ARROW-7357].

> we should migrate away from `pkg/errors` to `golang.org/x/xerrors` to
> ensure better error handling (and one that is Go-1.13 compatible).

Based on the above, it looks like the intent was to provide compatibility with go1.13+ errors, which are now mainline, so stdlib errors should be preferred for this.

[ARROW-7357]: https://issues.apache.org/jira/browse/ARROW-7357
[apache/arrow@7126fdb]: https://github.com/apache/arrow/commit/7126fdb5d6d2f7dd94fdd3d43ab9256825ff3d44


zeroshade (approved):

zeroshade:
Can you run `pre-commit` and fix the linting issues?

thaJeztah:
Doh! I saw some of those errors having punctuation, but thought I'd leave them as they were already there and to keep scope minimal, but forgot that linters tend to ignore them for `xerrors` but will flag them for stdlib 😂 

I'll have a look at fixing the linting issues.

thaJeztah:
I think it should be good now 👍

PR #774, opened 2026-04-20
fix(parquet): clarify docstring on Close method of FileWriter, proper err output

### LINKED ISSUE #703: FileWriter.Close() unexpectedly closes the caller-provided io.Writer and always returns nil
Version: github.com/apache/arrow-go/v18 v18.5.2

pqarrow.NewFileWriter accepts an io.Writer, but FileWriter.Close() internally calls TellWrapper.Close() which performs a runtime type assertion and calls Close() on the underlying writer if it satisfies io.WriteCloser.

This violates the established Go convention that a writer accepting io.Writer does not own the sink's lifetime. Both gzip.Writer and bufio.Writer write their format trailers in Close() but explicitly leave the underlying io.Writer open — the caller who created the sink is responsible for closing it.

Suggested fix: Remove the Close call from TellWrapper.Close(). If closing the underlying writer is intentional, NewFileWriter should accept io.WriteCloser explicitly so the contract is visible to callers.

### Component(s)

Parquet

zeroshade:
### Rationale for this change
Fixes #703

Clarifying the docstrings on the Close methods of the FileWriter and pqarrow.FileWriter

### Are these changes tested?
Documentation changes, no need to test

No

### DIFF
--- a/parquet/file/file_writer.go
+++ b/parquet/file/file_writer.go
@@ -228,7 +228,8 @@ func (fw *Writer) AppendKeyValueMetadata(key string, value string) error {
 }
 
 // Close closes any open row group writer and writes the file footer. Subsequent
-// calls to close will have no effect.
+// calls to close will have no effect. If the underlying writer implements [io.Closer],
+// then this will also call Close on the underlying writer.
 func (fw *Writer) Close() (err error) {
 	if fw.open {
 		// if any functions here panic, we set open to be false so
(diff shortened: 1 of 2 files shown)


zeroshade:
@tilinna is this what you were thinking?

lidavidm (approved):

tilinna:
LGTM, although I would have classified this as fix(parquet) (there's a small code-change too)

PR #781, opened 2026-04-24
fix(release): keys download

zeroshade:
### Rationale for this change
fixes #780

### What changes are included in this PR?
Adds `--output` argument to the `download` function in `verify_rc.sh` to fix the verification script.

PR #782, opened 2026-04-24
release: fix syntax error in verify_rc.sh

amoeba:
The previous commit didn't fix the issue because \ needs to be the final
character, or at least it does on my system.

Ref: https://github.com/apache/arrow-go/pull/781#issuecomment-4314898385

### DIFF
--- a/dev/release/verify_rc.sh
+++ b/dev/release/verify_rc.sh
@@ -66,7 +66,7 @@ setup_tmpdir() {
 download() {
   curl \
     --fail \
-    --location \    
+    --location \
     --show-error \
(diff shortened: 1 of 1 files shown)


amoeba:
Without this fix, I get,

```
+ download 'https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/KEYS' KEYS
+ curl --fail --location ' '
curl: (3) URL rejected: Malformed input to a URL function
```

zeroshade:
beat me to it, i noticed this when i saw the lint failure. *sigh* my bad on that one

PR #793, opened 2026-04-30
fix(arrow/cdata): make nativeCRecordBatchReader deterministic

zeroshade:
### Rationale for this change
Instead of relying on a finalizer, make the `nativeCRecordBatchReader` use atomic Retain and Release to make releasing deterministic.

Remove the finalizer, call C.ArrowArrayRelease and C.ArrowArrayStreamRelease based on refcount.

### Are these changes tested?
Yes, tests cover this already.

Only that retain and release now properly control the determinism of releasing the C memory instead of relying on a finalizer.

### DIFF
--- a/arrow/cdata/cdata.go
+++ b/arrow/cdata/cdata.go
@@ -953,9 +953,10 @@ func (n *nativeCRecordBatchReader) Retain() {
 }
 
 func (n *nativeCRecordBatchReader) Release() {
-	debug.Assert(n.refCount.Load() > 0, "too many releases")
+	rc := n.refCount.Add(-1)
+	debug.Assert(rc >= 0, "too many releases")
 
-	if n.refCount.Add(-1) == 0 {
+	if rc == 0 {
 		n.cleanUps[0].Stop()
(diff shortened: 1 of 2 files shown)


arrow/cdata/cdata.go
```diff
@@ -940,12 +943,31 @@ type nativeCRecordBatchReader struct {
...
 }
 
-// No need to implement retain and release here as we used runtime.SetFinalizer when constructing
-// the reader to free up the ArrowArrayStream memory when the garbage collector cleans it up.
-func (n *nativeCRecordBatchReader) Retain()  {}
-func (n *nativeCRecordBatchReader) Release() {}
+func (n *nativeCRecordBatchReader) Release() {
+	debug.Assert(n.refCount.Load() > 0, "too many releases")
+
+	if n.refCount.Add(-1) == 0 {
```
lidavidm: nit, but instead of two atomics you could just do the add, assert that it's >= 0, then check for 0

lidavidm (approved):

PR #802, opened 2026-05-05
feat(compute): support casts between binary/string and binary_view/string_view

### LINKED ISSUE #184: unsupported cast to string_view from utf8 in v18

Casting between string <-> string_view types has been added in C++ and Python implementations of arrow in v18: https://github.com/apache/arrow/pull/43302

I tested this in Python and it works with the latest (18.0.0) pyarrow release:

```python
[ins] In [1]: import pyarrow as pa

[ins] In [2]: import pyarrow.compute as pc

[ins] In [3]: a = pa.array(["one", "two", "three"])

[ins] In [4]: a
Out[4]: 
<pyarrow.lib.StringArray object at 0x111e3b340>
[
  "one",
  "two",
  "three"
]

[ins] In [5]: a.cast(pa.string_view())
Out[5]: 
<pyarrow.lib.StringViewArray object at 0x111a3be80>
[
  "one",
  "two",
  "three"
]
```

It doesn't work with the v18 release of arrow-go:

```go
package main

import (
	"context"
	"fmt"

)

func main() {
	alloc := memory.NewGoAllocator()
	b := array.NewStringBuilder(alloc)
	defer b.Release()
	b.AppendValues([]string{"a", "b", "c"}, nil)
	arr := b.NewArray()
	fmt.Println(arr)

	newArr, err := compute.CastArray(context.Background(), arr, &compute.CastOptions{
		ToType: arrow.BinaryTypes.StringView,
	})
	if err != nil {
		panic(err)
	}
	fmt.Println(newArr)
}

// Output:
// ["a" "b" "c"]
// panic: not implemented: unsupported cast to string_view from utf8
// ...
// exit status 2
```

I'd be happy to work on the PR to add it if someone can give me a pointer with description of what would it take - high-level - to add support for this in Go.

### Component(s)

Release

zeroshade:
### Rationale for this change

Fixes #184. `compute.CastArray` currently fails with `unsupported cast to string_view from utf8` (and symmetrically for `binary_view`, inverse directions, and FixedSizeBinary) because no cast kernels are registered for the view variants. PyArrow has supported this since 18.0.0 (apache/arrow#43302); this PR brings arrow-go to parity.

**New cast kernels** in `arrow/compute/internal/kernels/binary_view_casts.go`:

| Direction | Kernel | Notes |
|---|---|---|
| `binary`, `large_binary`, `string`, `large_string`, `fixed_size_binary` → `binary_view`/`string_view` | `CastBinaryToBinaryView` | Uses `BinaryViewBuilder`/`StringViewBuilder`. UTF-8 validated when destination is `string_view` and source is non-utf8, unless `CastOptions.AllowInvalidUtf8` is set. |
| `binary_view`, `string_view` → `binary`, `large_binary`, `string`, `large_string` | `CastBinaryViewToBinary[OutOffsetT]` | Materializes into a contiguous data buffer. UTF-8 validated when casting `binary_view` → `string`/`large_string`, unless `AllowInvalidUtf8` is set. Checks `int32` overflow when targeting `string`/`binary`. |
| `binary_view` ↔ `string_view` (and identity) | `CastBinaryViewToBinaryView` | Zero-copy via `ZeroCopyCastExec`; UTF-8 validated on `binary_view` → `string_view` direction. |

These are wired into the dispatch table via `cast_binary_view` and `cast_string_view` entries in `cast.go`, and `addToBinaryKernels` now also registers `view → base-binary` kernels so the existing `cast_binary`/`cast_string`/etc. paths accept view inputs.

**Related executor/array fixes** (required to get view types through the generic dispatch):

* `arrow/compute/exec.getNumBuffers` now returns 3 for `binary_view`/`string_view` (bitmap + view headers + one overflow data buffer), matching `ArraySpan`'s fixed `[3]BufferSpan`. Previously this fell through to the `default: return 2` case, so the data buffer was dropped when `ArraySpan.MakeData` sliced to `bufs[:NumBuffers()]`, which crashed any cast that read non-inline values via `input.MakeArray()`.
* `arrow/array.getMaxBufferLen` and the `nullArrayFactory` now handle `BinaryViewDataType`, so `MakeArrayOfNull` works for view outputs. Without this, casting a zero-length input array to a view type panicked with `arrayofnull not implemented for type string_view`.

### Are these changes tested?

Yes. Added four new `CastSuite` tests in `arrow/compute/cast_test.go`:

* `TestBinaryLikeToBinaryView` - all base-binary → view, FSB → view, short (inline) + long (out-of-line) values + nulls, UTF-8 validation + opt-out via `AllowInvalidUtf8`.
* `TestBinaryViewToBinaryLike` - view → all base-binary, UTF-8 validation for `binary_view` → `string`.
* `TestBinaryViewToBinaryView` - cross-view casts with UTF-8 validation.
* `TestCanCastViewTypes` - `CanCast` coverage for the new paths.

These use a simple array-level comparison (`checkCastArrayOnly`) rather than the existing `checkCast` because the latter iterates via `scalar.GetScalar`, which does not yet support view types (separate limitation, outside scope). Full `arrow/compute/...` and `arrow/array/...` test suites pass; `arrow/compute` also passes with `-race`.

The exact reproducer from #184 now succeeds:

```
input: ["a" "b" "c" "this is a longer string beyond 12 bytes"]
cast to string_view: ["a" "b" "c" "this is a longer string beyond 12 bytes"]
type: string_view
```

Yes - previously-unsupported cast paths now work:

* `compute.CastArray(ctx, arr, {ToType: BinaryTypes.StringView})` and inverse.
* `compute.CastArray(ctx, arr, {ToType: BinaryTypes.BinaryView})` and inverse.
* `binary_view` ↔ `string_view` casts.
* FSB → `binary_view`/`string_view`.
* `compute.CanCast` advertises all of the above.

No existing behavior changes. The change to `getNumBuffers` could in principle affect other compute paths that manipulate view arrays, but the previous value of `2` was incorrect for views (causing data-buffer loss), so this is a bug fix rather than a behavioral change.

### DIFF
--- a/arrow/compute/exec/span.go
+++ b/arrow/compute/exec/span.go
@@ -579,6 +579,13 @@ func getNumBuffers(dt arrow.DataType) int {
 		return 1
 	case arrow.BINARY, arrow.LARGE_BINARY, arrow.STRING, arrow.LARGE_STRING, arrow.DENSE_UNION:
 		return 3
+	case arrow.BINARY_VIEW, arrow.STRING_VIEW:
+		// bitmap + view-header buffer + a single overflow data buffer.
+		// ArraySpan's fixed buffer array limits additional data buffers,
+		// so callers producing multi-buffer views must keep their data
+		// within a single block (the default 32KB allocation in the
+		// builder is sufficient for most use cases).
+		return 3
 	case arrow.EXTENSION:
 		return getNumBuffers(dt.(arrow.ExtensionType).StorageType())
 	default:
--- a/arrow/array/util.go
+++ b/arrow/array/util.go
@@ -447,6 +447,8 @@ func getMaxBufferLen(dt arrow.DataType, length int) int {
 		return bufferLen
 	case arrow.OffsetsDataType:
 		return maxOf(dt.OffsetTypeTraits().BytesRequired(length + 1))
+	case arrow.BinaryViewDataType:
+		return maxOf(arrow.ViewHeaderSizeBytes * length)
 	case *arrow.FixedSizeListType:
 		return maxOf(getMaxBufferLen(dt.Elem(), int(dt.Len())*length))
 	case arrow.ExtensionType:
@@ -495,6 +497,8 @@ func (n *nullArrayFactory) create() *Data {
 		defer arr.Release()
 		dictData = arr.Data()
 	case arrow.FixedWidthDataType:
+		bufs = append(bufs, n.buf)
+	case arrow.BinaryViewDataType:
 		bufs = append(bufs, n.buf)
 	case arrow.BinaryDataType:
 		bufs = append(bufs, n.buf, n.buf)
--- a/arrow/compute/cast.go
+++ b/arrow/compute/cast.go
@@ -345,9 +345,10 @@ func unpackDictionary(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecR
 
 // unpackViewDictionary materializes a dictionary whose values are a view
 // type (string_view or binary_view) into a flat view array of the same
-// type, preserving per-element nulls from the dictionary indices. This
-// exists because array_take has no view kernel; the cast meta function
-// still needs to expand dictionaries as part of DICTIONARY -> X casts.
+// type, preserving per-element nulls from the dictionary indices and
+// per-slot nulls from the dictionary values themselves. This exists
+// because array_take has no view kernel; the cast meta function still
+// needs to expand dictionaries as part of DICTIONARY -> X casts.
 func unpackViewDictionary(mem memory.Allocator, dictArr *array.Dictionary) (arrow.Array, error) {
 	switch vals := dictArr.Dictionary().(type) {
 	case *array.StringView:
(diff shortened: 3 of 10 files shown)


arrow/compute/exec/span.go
```diff
@@ -579,6 +579,13 @@ func getNumBuffers(dt arrow.DataType) int {
 		return 1
 	case arrow.BINARY, arrow.LARGE_BINARY, arrow.STRING, arrow.LARGE_STRING, arrow.DENSE_UNION:
 		return 3
+	case arrow.BINARY_VIEW, arrow.STRING_VIEW:
+		// bitmap + view-header buffer + a single overflow data buffer.
+		// ArraySpan's fixed buffer array limits additional data buffers,
+		// so callers producing multi-buffer views must keep their data
+		// within a single block (the default 32KB allocation in the
+		// builder is sufficient for most use cases).
```
lidavidm, line 587: Is it? It seems we need to address this at some point

lidavidm (approved):

PR #805, opened 2026-05-08
feat(arrow/array): add Resize method to RecordBuilder

### LINKED ISSUE #796: Helper method for RecordBuilder to drop last, incomplete "rows"

When populating `RecordBuilder` batches from row based data, we frequently end up in a situation, whereupon some sort of error prevents processing on an entire incoming data row. This will result in some field values (first few) being added to field arrays, whereupon some remaining field arrays will remain shorter, as incoming row processing was aborted.

Of course, `NewRecordBatch` method has a check to prevent such record batches from actually going somewhere.

But what will be really useful, is to have a method which will truncate the extant field arrays within `RecordBuilder` to the length of the shortest one amongst them, effectively discarding the incomplete rows.

This simple method will simplify non-critical error processing substantially, without forcing users to build their own, potentially buggy methods doing the same with individual `Builder.Resize` methods.

### Component(s)

Other

oakad:
Presently added Resize method addresses two, relatively common needs:

1. All fields within the Builder may need to be resized to the same length.
2. As part of error recovery process, fields may need to be truncated to the length of the shortest one, effectively discarding incomplete "rows".

Fixes #796

### Rationale for this change
Easier error handling when only partial data is available due to error.

Additional method for RecordBuilder.

### Are these changes tested?

Unit test is provided.

Additional method in RecordBuilder.

### DIFF
--- a/arrow/array/json_reader_test.go
+++ b/arrow/array/json_reader_test.go
@@ -17,6 +17,7 @@
 package array_test
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"strings"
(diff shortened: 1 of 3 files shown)


zeroshade (changes requested):
Generally looks fine, but needs unit tests and to reference the relevant issue in the description.

oakad:
It says "Fixes #796" right there in the commit message. What sort of additional reference can I add?

Will see regarding the unit test tomorrow.

zeroshade:
> It says "Fixes #796" right there in the commit message.

Ah, I missed that. My fault there.

So yea, just needs tests and to fix the linting issue

oakad:
I have added a unit test. I also had to alter the `NewRecordBatch`s semantics slightly, as I could not get rid of memory leak otherwise - after the panic inside `NewRecordBatch` was raised, the remaining memory state was somehow non-obviously inconsistent.

Also, it feels that current approach (check row counts ahead of any array move action)  is generally safer.

zeroshade (approved):

PR #807, opened 2026-05-10
fix(parquet/metadata): do not pre-set hasDistinctCount in stat factories

### LINKED ISSUE #806: [Bug] NewIntXStatistics factories unconditionally set hasDistinctCount=true, causing distinct_count=0 to always appear in Parquet output

metadata/statistics_types.gen.go factories (NewInt32Statistics, NewInt64Statistics, NewFloat32Statistics, NewFloat64Statistics, etc.) initialize hasDistinctCount: true at construction time, even though no distinct count has been computed. The result: every Statistics object emitted by arrow-go's parquet writer carries distinct_count = 0 in the thrift output — both in ColumnChunk.metadata.statistics (footer) and in DataPageHeader.statistics (column-data page header).

The Parquet thrift Statistics.distinct_count field is OPTIONAL. Per spec semantics, the field MUST be absent when the value is unknown — emitting 0 conflates "no distinct values" with "we don't know."

Comparison to parquet-cpp (pyarrow)
parquet-cpp (pyarrow) does NOT exhibit this behavior: distinct_count is set only when actually computed. For the same input table, pyarrow's footer + page headers omit the distinct_count field entirely; arrow-go's emit distinct_count: 0. This is a real cross-implementation byte-asymmetry for any consumer who needs byte-identical Parquet output across language implementations.

Reproducer (minimal Go)
package main

import (
    "bytes"
    "fmt"
    "github.com/apache/arrow-go/v18/parquet/file"
)

func main() {
    pool := memory.NewGoAllocator()
    schema := arrow.NewSchema([]arrow.Field{{Name: "id", Type: arrow.PrimitiveTypes.Int64, Nullable: false}}, nil)
    builder := array.NewInt64Builder(pool)
    builder.AppendValues([]int64{1, 2, 3}, nil)
    arr := builder.NewArray()
    rec := array.NewRecord(schema, []arrow.Array{arr}, 3)

    var buf bytes.Buffer
    props := parquet.NewWriterProperties(parquet.WithStats(true), parquet.WithCompression(0)) // Uncompressed for clarity
    arrowProps := pqarrow.NewArrowWriterProperties()
    w, _ := pqarrow.NewFileWriter(schema, &buf, props, arrowProps)
    w.Write(rec)
    w.Close()

    rdr, _ := file.OpenParquetReader(bytes.NewReader(buf.Bytes()), nil)
    md := rdr.MetaData()
    cm, _ := md.RowGroup(0).ColumnChunk(0)
    stats, _ := cm.Statistics()
    fmt.Printf("DistinctCount set? %v (value=%d)\n", stats.HasDistinctCount(), stats.DistinctCount())
    // Expected output (after fix): false (value=0)
    // Actual output (today):       true  (value=0)
}
Root cause
metadata/statistics_types.gen.go lines 61-62 (NewInt32Statistics), 362-363 (NewInt64Statistics), 663-664, 942-943, 1235-1236 — each factory sets:

statistics: statistics{
    descr:            descr,
    hasNullCount:     true,
    hasDistinctCount: true,
    ...
},
The hasDistinctCount: true flag should NOT default to true; it should flip only when a caller explicitly invokes SetDistinctCount(...) with a known value. (hasNullCount: true may need similar review — see "Out of scope" below.)

Proposed fix
In each numeric NewXxxStatistics factory in metadata/statistics_types.gen.go, change hasDistinctCount: true to hasDistinctCount: false. The flag flips to true only via explicit SetDistinctCount() (or IncDistinct() after a real distinct-count calculation).

Backward compatibility
Output bytes change (smaller — distinct_count field removed when unknown).
All readers tolerate the optional-field-absent case (pyarrow already omits, and all readers handle pyarrow output).
arrow-go's own IsSetDistinctCount() bool method already exists and behaves correctly given the fix.
The behavior of consumers that explicitly call SetDistinctCount(N) remains unchanged.
Test
Add a regression test: write a table without computing distinct_count; read back; assert IsSetDistinctCount() == false.
Cross-impl conformance: produce a Parquet file via arrow-go and via parquet-cpp on the same input + settings; assert footer thrift bytes match (currently they don't because of this).
Out of scope (related but separate)
hasNullCount: true at the same construction sites — same factory bug pattern. Whether to fix in the same PR is a design call. arrow-go and pyarrow may align on null_count (it's typically computed during the write); flag for separate review.
pyarrow's missing LogicalType emission knob for INT64 — separate cross-impl byte-identity issue; tracked at the consumer's spec level (sealed-variant pattern).
Adjacent precedent
Similar prior fix accepted upstream: apache/arrow-go#210 "fix(parquet/metadata): fix default unsigned int statistics" (merged 2024-12-10). Same code area, same kind of factory-default correction.

PRs welcome from our side if maintainers prefer; happy to follow up.

### Component(s)

Parquet

SAY-5:
### Rationale for this change

Fixes #806. The `New*Statistics` factories in `parquet/metadata/statistics_types.gen.go` initialize `hasDistinctCount: true` at construction time, so every fresh stat object reports a distinct count of 0. `Encode()` then unconditionally calls `enc.SetDistinctCount(0)`, which makes the writer emit `distinct_count: 0` in both `ColumnChunk.metadata.statistics` and `DataPageHeader.statistics` for every column, even when no distinct count has been computed.

The Parquet thrift `Statistics.distinct_count` field is OPTIONAL. Per spec semantics, it must be absent when unknown. Emitting 0 conflates "no distinct values" with "we don't know" and causes byte-asymmetry against parquet-cpp (pyarrow), which leaves the field unset for the same input.

`IncDistinct()` is the only call site that should flip `hasDistinctCount` to true, and grepping the repo confirms it is never invoked from the writer paths, so flipping the default is safe.

- Remove `hasDistinctCount: true` from the factory initializer in `statistics_types.gen.go.tmpl` and regenerate `statistics_types.gen.go`. `hasNullCount: true` is kept as-is (its semantics differ: `IncNulls` is called per page and `null_count: 0` is meaningful for a writer that observes zero nulls).
- Add `TestNewStatisticsDistinctCountUnset` covering Int32/Int64/Float32/Float64/Boolean/ByteArray, asserting both `HasDistinctCount()` and `Encode().HasDistinctCount` start false and flip to true after `IncDistinct`.

### Are these changes tested?

Yes. The new test fails on master (HasDistinctCount returns true on a fresh stat object) and passes after the fix. `go test ./parquet/metadata/...` is green.

Yes, writers that previously emitted `distinct_count: 0` for every column will now omit the field, matching the spec and parquet-cpp. Readers already tolerate both, so this is a wire-format normalization, not a breaking change for downstream consumers.

### DIFF
--- a/parquet/metadata/statistics_test.go
+++ b/parquet/metadata/statistics_test.go
@@ -603,3 +603,40 @@ func TestBooleanStatisticsUpdateFromBitmapSpaced(t *testing.T) {
 		assert.Equal(t, int64(8), stats.NullCount())
 	})
 }
+
+func TestNewStatisticsDistinctCountUnset(t *testing.T) {
+	// Issue #806: factories must not pre-set hasDistinctCount, otherwise
+	// Encode() emits distinct_count=0 for every column instead of leaving the
+	// optional Thrift field absent. IncDistinct() is responsible for marking
+	// it set when an actual distinct count is computed.
+	mk := func(n *schema.PrimitiveNode) metadata.TypedStatistics {
+		col := schema.NewColumn(n, 0, 0)
+		return metadata.NewStatistics(col, memory.DefaultAllocator)
+	}
+	cases := []struct {
+		name string
+		s    metadata.TypedStatistics
+	}{
+		{"Int32", mk(schema.NewInt32Node("i32", parquet.Repetitions.Required, -1))},
+		{"Int64", mk(schema.NewInt64Node("i64", parquet.Repetitions.Required, -1))},
+		{"Float32", mk(schema.NewFloat32Node("f32", parquet.Repetitions.Required, -1))},
+		{"Float64", mk(schema.NewFloat64Node("f64", parquet.Repetitions.Required, -1))},
+		{"Boolean", mk(schema.NewBooleanNode("b", parquet.Repetitions.Required, -1))},
+		{"ByteArray", mk(schema.NewByteArrayNode("ba", parquet.Repetitions.Required, -1))},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			assert.False(t, c.s.HasDistinctCount(), "fresh stats must not advertise a distinct count")
+			enc, err := c.s.Encode()
+			require.NoError(t, err)
+			assert.False(t, enc.HasDistinctCount, "encoded stats must leave distinct_count unset until IncDistinct is called")
+
+			c.s.IncDistinct(3)
+			assert.True(t, c.s.HasDistinctCount())
+			enc, err = c.s.Encode()
+			require.NoError(t, err)
+			assert.True(t, enc.HasDistinctCount)
+			assert.Equal(t, int64(3), enc.DistinctCount)
+		})
+	}
+}
--- a/parquet/metadata/statistics_types.gen.go
+++ b/parquet/metadata/statistics_types.gen.go
@@ -57,12 +57,11 @@ func NewInt32Statistics(descr *schema.Column, mem memory.Allocator) *Int32Statis
 
 	return &Int32Statistics{
 		statistics: statistics{
-			descr:            descr,
-			hasNullCount:     true,
-			hasDistinctCount: true,
-			order:            descr.SortOrder(),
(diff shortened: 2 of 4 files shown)


zeroshade (approved):
Good catch!

zeroshade (approved):

PR #808, opened 2026-05-10
fix(avro): error on complex (non-nullable) unions instead of silently dropping (#777)

### LINKED ISSUE #777: [Avro] Silently ignores complex unions

https://github.com/apache/arrow-go/blob/c5f0943277155a00439e082cbe3a44af84ea4f46/arrow/avro/schema.go#L142

Only processes simple nullable unions and does not process complex unions, but silently ignores them.
Expectation:
1. Limitation to simple nullables is documented
2. By default, schema with complex unions errors out during conversion unless an optional flag to override this behavior and drop complex union types is passed.

### Component(s)

Other

SAY-5:
### Rationale for this change
Fixes #777. `ArrowSchemaFromAvro` previously silently dropped non-nullable Avro unions (e.g. `["int","string"]`) in both `arrowSchemafromAvro` and `iterateFields`. Output schemas were missing fields with no warning to the caller.

Panic with a clear error message at the two `case "union"` / `case *avro.UnionSchema` arms when the union is not a simple nullable. The package's existing `recover()` in `ArrowSchemaFromAvro` converts this into a returned error.

### Are these changes tested?
Yes, `TestComplexUnionReportsError` exercises a record with `["int","string"]` and asserts a non-nil error mentioning `union`.

Callers that previously received an incomplete schema for a complex-union field now get an explicit error.

### DIFF
--- a/arrow/avro/schema.go
+++ b/arrow/avro/schema.go
@@ -139,13 +139,16 @@ func arrowSchemafromAvro(n *schemaNode) {
 		arrowSchemafromAvro(c)
 		n.arrowField = buildArrowField(n, arrow.MapOf(arrow.BinaryTypes.String, c.arrowField.Type), c.arrowField.Metadata)
 	case "union":
-		if n.schema.(*avro.UnionSchema).Nullable() {
-			if len(n.schema.(*avro.UnionSchema).Types()) > 1 {
-				n.schema = n.schema.(*avro.UnionSchema).Types()[1]
+		us := n.schema.(*avro.UnionSchema)
+		if us.Nullable() {
+			if len(us.Types()) > 1 {
+				n.schema = us.Types()[1]
 				n.union = true
 				n.nullable = true
 				arrowSchemafromAvro(n)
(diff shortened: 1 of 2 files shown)


arrow/avro/schema.go
```diff
@@ -139,13 +139,16 @@ func arrowSchemafromAvro(n *schemaNode) {
...
+		us := n.schema.(*avro.UnionSchema)
+		if us.Nullable() {
+			if len(us.Types()) > 1 {
+				n.schema = us.Types()[1]
 				n.union = true
 				n.nullable = true
 				arrowSchemafromAvro(n)
 			}
+		} else {
+			panic(fmt.Errorf("complex (non-nullable) avro union at '%v' is not supported", n.schemaPath()))
```
zeroshade, line 151: Should we add support and use the arrow union type? We could also do this as a follow up instead.

zeroshade (approved):
LGTM, thanks for this! Just had one question

PR #809, opened 2026-05-11
fix(ipc): preserve map child field metadata

dimakuz:
### Rationale for this change

IPC deserialization of `MapType` reconstructed map key/value children from only their data types, which dropped field-level metadata on the key and value fields. This could silently lose metadata such as field IDs after an IPC schema round-trip.

The IPC metadata reader now reconstructs `MapType` using `arrow.MapOfFields(...)` with the already-deserialized key and value fields, preserving their metadata and nullability.

A regression test was added to verify that map key/value field metadata, item nullability, and `KeysSorted` survive an IPC schema round-trip.

### Are these changes tested?

Yes - unit test included.

Yes - IPC schema deserialization now preserves field-level metadata on MapType key and value children.


zeroshade (approved):
Nice catch!

PR #816, opened 2026-05-14
fix(arrow/array): preserve large integer precision in JSON decoding

### LINKED ISSUE #804: [Go] RecordFromJSON does not handle integer values not representable by double

This limits the usefulness of RecordFromJSON for testing, and implies we may have been writing incorrect tests.

You silently get wrong data:

```go
package main

import (
	"bytes"
	"fmt"
	"log"

)

func main() {
	mem := memory.DefaultAllocator
	dt := arrow.FixedWidthTypes.Duration_s
	record, _, err := array.FromJSON(mem, dt, bytes.NewReader([]byte("[9223372036854775807]")))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(record)
}
```

Result:

```
[-9223372036854775808]
```

Commonly, large integers are represented as strings in JSON to get around this. That doesn't work:

```go
package main

import (
	"bytes"
	"fmt"
	"log"

)

func main() {
	mem := memory.DefaultAllocator
	dt := arrow.FixedWidthTypes.Duration_s
	record, _, err := array.FromJSON(mem, dt, bytes.NewReader([]byte(`["9223372036854775807"]`)))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(record)
}
```

Result:

```
2026/05/07 15:36:14 json: cannot unmarshal 9223372036854775807 into Go value of type arrow.Duration
exit status 1
```

### Component(s)

Other

zeroshade:
## Summary

Fixes #804.

JSON decoding in `arrow/array` silently corrupted integer values larger than 2^53 because Go's `encoding/json` decodes numeric tokens as `float64` by default, losing precision for very large integers. This PR makes `UseNumber` the unconditional default across `FromJSON`, `RecordFromJSON`, and `NewJSONReader` so full `int64` precision is preserved without requiring callers to opt in.

## Changes

- **`RecordBuilder` gains `UnmarshalOne` and `Unmarshal`** so caller-configured `json.Decoder` options propagate through nested field decoding. `RecordFromJSON` now calls `bldr.UnmarshalOne(dec)` instead of `dec.Decode(bldr)`, avoiding the bytes-roundtrip through `json.Unmarshaler` that previously dropped the outer decoder configuration (root cause noted in [this comment](https://github.com/apache/arrow-go/issues/804#issuecomment-4454244067)).

- **All builder `UnmarshalJSON([]byte)` methods now enable `UseNumber`** on their internal decoder so large integers also survive the standard `json.Unmarshal(data, &builder)` path.

- **String-encoded integers now work for time-based types.** `Duration`, `Timestamp`, `Time32`, `Time64`, `Date32`, and `Date64` builders accept stringified integers (e.g. `"9223372036854775807"`) as a fallback before their format-specific parsers — addressing the second snippet from the issue. This matches how `Int64Builder` and the decimal builders already behaved.

- **`WithUseNumber` and `WithUseNumberJSONReader` are deprecated** (marked with `Deprecated:` per pkg.go.dev convention) since `UseNumber` is now always enabled. Both options remain functional no-ops for backward compatibility.

## Test Plan

15 new tests added in `arrow/array/util_test.go` and `arrow/array/json_reader_test.go` covering:

- `RecordFromJSON` and `JSONReader` round-tripping `MaxInt64`/`MinInt64` with and without the deprecated options.
- `RecordBuilder.UnmarshalJSON` and `UnmarshalOne` directly preserving large integers.
- Stringified integer round-trip for `Duration`, `Timestamp`, `Time32`, `Time64`, `Date32`, `Date64`.
- Regression: invalid duration strings (`"abc"`) still error; format-style duration strings (`"3h2m0.5s"`) still parse.

Verified passing locally:

- `./arrow/array/...` (full suite)
- `./arrow/compute/...` (largest blast-radius consumer of `WithUseNumber` — 49 test sites)
- `./arrow/ipc/...`, `./arrow/scalar/...`, `./arrow/extensions/...`, `./arrow/encoded/...`, `./arrow/util/...`
- `./parquet/file/...` (the `./parquet/pqarrow/...` package fails identically on `main` due to a missing `PARQUET_TEST_DATA` env var — pre-existing, unrelated)

All `lsp_diagnostics` clean across the 23 changed files.

## Behavior Change Notes

The only observable behavior change for existing callers is: large integer values that previously decoded to a wrong `int64` (via float64 round-trip) now decode to the correct value. Existing tests in `arrow/compute` continue to pass, including those that explicitly opt in via `WithUseNumber` and the ones that don't. No production callers in this repo break (verified across `flight`, `compute`, `parquet`, `ipc`).

### DIFF
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -243,6 +243,7 @@ func (b *BooleanBuilder) Unmarshal(dec *json.Decoder) error {
 func (b *BooleanBuilder) UnmarshalJSON(data []byte) error {
 	dec := json.NewDecoder(bytes.NewReader(data))
 	dec.UseNumber()
+	dec.UseNumber()
 	t, err := dec.Token()
 	if err != nil {
 		return err
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -243,7 +243,6 @@ func (b *BooleanBuilder) Unmarshal(dec *json.Decoder) error {
 func (b *BooleanBuilder) UnmarshalJSON(data []byte) error {
 	dec := json.NewDecoder(bytes.NewReader(data))
 	dec.UseNumber()
-	dec.UseNumber()
 	t, err := dec.Token()
 	if err != nil {
 		return err
--- a/arrow/array/json_reader.go
+++ b/arrow/array/json_reader.go
@@ -62,6 +62,24 @@ func WithAllocator(mem memory.Allocator) Option {
 	}
 }
 
+// WithUseNumberJSONReader previously enabled UseNumber on the JSONReader's
+// internal json.Decoder. As of issue #804, NewJSONReader unconditionally
+// enables UseNumber so integer values too large to fit in float64 are
+// preserved. This option is now a no-op and is retained for backward
+// compatibility.
+//
+// Deprecated: UseNumber is now always enabled; this option has no effect.
+func WithUseNumberJSONReader() Option {
+	return func(cfg config) {
+		switch cfg := cfg.(type) {
+		case *JSONReader:
+			cfg.useNumber = true
+		default:
+			panic(fmt.Errorf("arrow/json): unknown config type %T", cfg))
+		}
+	}
+}
+
 // JSONReader is a json reader that meets the RecordReader interface definition.
 //
 // To read in an array of objects as a record, you can use RecordFromJSON
@@ -78,8 +96,9 @@ type JSONReader struct {
 	cur  arrow.RecordBatch
 	err  error
 
-	chunk int
-	done  bool
+	chunk     int
+	useNumber bool
+	done      bool
 
 	mem  memory.Allocator
 	next func() bool
@@ -102,6 +121,8 @@ func NewJSONReader(r io.Reader, schema *arrow.Schema, opts ...Option) *JSONReade
 	for _, o := range opts {
 		o(rr)
 	}
+
+	rr.r.UseNumber()
 
 	if rr.mem == nil {
 		rr.mem = memory.DefaultAllocator
@@ -166,7 +187,7 @@ func (r *JSONReader) Next() bool {
 }
 
 func (r *JSONReader) readNext() bool {
-	r.err = r.r.Decode(r.bldr)
+	r.err = r.bldr.UnmarshalOne(r.r)
 	if r.err != nil {
 		r.done = true
 		if errors.Is(r.err, io.EOF) {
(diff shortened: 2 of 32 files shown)


arrow/array/booleanbuilder.go
```diff
@@ -243,6 +243,7 @@ func (b *BooleanBuilder) Unmarshal(dec *json.Decoder) error {
 func (b *BooleanBuilder) UnmarshalJSON(data []byte) error {
 	dec := json.NewDecoder(bytes.NewReader(data))
 	dec.UseNumber()
+	dec.UseNumber()
```
lidavidm: Er, there's a duplicate call...

arrow/array/json_reader.go
```diff
@@ -62,6 +62,24 @@ func WithAllocator(mem memory.Allocator) Option {
...
 }
 
+// WithUseNumberJSONReader previously enabled UseNumber on the JSONReader's
+// internal json.Decoder. As of issue #804, NewJSONReader unconditionally
+// enables UseNumber so integer values too large to fit in float64 are
+// preserved. This option is now a no-op and is retained for backward
+// compatibility.
+//
+// Deprecated: UseNumber is now always enabled; this option has no effect.
+func WithUseNumberJSONReader() Option {
```
lidavidm: Er, why are we adding a new, immediately deprecated, option? There's no such option currently.

arrow/array/json_reader.go
```diff
@@ -78,8 +96,9 @@ type JSONReader struct {
 	cur  arrow.RecordBatch
 	err  error
 
-	chunk int
-	done  bool
+	chunk     int
+	useNumber bool
```
lidavidm: This seems to be unused?

arrow/array/booleanbuilder.go
zeroshade: oy, fixed

arrow/array/json_reader.go
zeroshade: good catch, that was silly. removed

arrow/array/json_reader.go
zeroshade: removed

arrow/array/numericbuilder.gen_test.go.tmpl
lidavidm, line 1: I suppose it was originally like this but why does the Float branch not test float values?

lidavidm (approved):

arrow/array/numericbuilder.gen_test.go.tmpl
zeroshade: I honestly don't remember, but it's easy enough to add some cases for that. I'll do so before merging

lidavidm (approved):

PR #817, opened 2026-05-18
perf(parquet/pqarrow): cap RecordReader batch size to actual row count

paveon:
### Rationale for this change
 `GetRecordReader` passes `BatchSize` directly to the internal `recordReader`
without capping it to the actual number of rows. When `BatchSize` is configured
to a large value (e.g. 131072) but the file or requested row groups contain
few rows (e.g. 10), `leafReader.LoadBatch` calls `Reserve(131072)` which
pre-allocates definition/repetition level buffers and value buffers sized for
the full batch. For a 200-column int64 table with 10 rows this wastes ~250 MB
of allocations. 

Cap `batchSize` to `NextPowerOf2(nrows)` when a `BatchSize` is explicitly
configured. The power-of-2 rounding keeps allocations aligned with the
downstream `updateCapacity` logic that already rounds to powers of two,
avoiding a redundant reallocation on the first read.

### Are these changes tested?
Existing tests pass. The change is on the allocation-sizing path only —
read correctness is unaffected since `LoadBatch` already stops reading
when rows are exhausted.

No

### DIFF
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -517,7 +517,7 @@ func (fr *FileReader) GetRecordReader(ctx context.Context, colIndices, rowGroups
 		nrows += fr.rdr.MetaData().RowGroup(rg).NumRows()
 	}
 
-	batchSize := fr.Props.BatchSize
+	var batchSize int64
 	if fr.Props.BatchSize <= 0 {
 		batchSize = nrows
 	} else {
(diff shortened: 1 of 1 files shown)


zeroshade:
Can you fix the linting issue? Otherwise this looks good

paveon:
Should be fixed now

zeroshade (approved):

PR #824, opened 2026-05-22
fix(parquet): return error instead of panicking on first-write failure

### LINKED ISSUE #820: parquet/file: NewParquetWriter panics on transient sink.Write errors during file initialization

`(*file.Writer).startFile()` panics with `"failed to write magic number"` when the underlying sink's first `Write` returns an error or short-writes. Because `NewParquetWriter` calls `startFile` synchronously inside the constructor — and `NewParquetWriter` has no error return — there is no way for a consumer to recover via the standard Go error handling pattern.

**Code** ([`parquet/file/file_writer.go:201-204`](https://github.com/apache/arrow-go/blob/main/parquet/file/file_writer.go#L201-L204)):

```go
n, err := fw.sink.Write(magic)
if n != 4 || err != nil {
    panic("failed to write magic number")
}
```

**Constructor signature** ([`parquet/file/file_writer.go:72`](https://github.com/apache/arrow-go/blob/main/parquet/file/file_writer.go#L72)):

```go
func NewParquetWriter(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) *Writer
```

The higher-level [`pqarrow.NewFileWriter`](https://github.com/apache/arrow-go/blob/main/parquet/pqarrow/file_writer.go#L135) does have an `error` return, but it propagates the panic from `file.NewParquetWriter` through unconverted — so callers of either constructor must wrap the call in `defer recover()` to handle a flaky sink.

### Why this matters in practice

When the `sink` is backed by a network-attached writer — e.g., a `*storage.Writer` from `cloud.google.com/go/storage`, a Cloud Storage resumable upload, an HDFS client, an S3 multipart upload — a transient network blip on the first 4-byte write reliably crashes the consumer process.

Concrete example: a Kubernetes-deployed Go service that writes ~thousands of Parquet files per day to GCS sees ~1-2 pod crashes per 5 pod-hours under normal cluster conditions, all from this panic. Each crash:

1. Loses the entire worker process (not just the failing call)
2. Forces a container restart
3. Leaves any in-flight resumable upload session orphaned in GCS until session timeout
4. Requires application-level orphan-recovery machinery to reclaim work

The panic stack consistently looks like:

```
panic: failed to write magic number

goroutine N [running]:
github.com/apache/arrow-go/v18/parquet/file.(*Writer).startFile(...)
	github.com/apache/arrow-go/v18@v18.6.0/parquet/file/file_writer.go:203 +0x257
github.com/apache/arrow-go/v18/parquet/file.NewParquetWriter(...)
	github.com/apache/arrow-go/v18@v18.6.0/parquet/file/file_writer.go:90 +0x2d4
github.com/apache/arrow-go/v18/parquet/pqarrow.NewFileWriter(...)
	github.com/apache/arrow-go/v18@v18.6.0/parquet/pqarrow/file_writer.go:166 +0x34d
<consumer code calling pqarrow.NewFileWriter>
```

Workarounds exist (`defer recover()` at the call site) but they're brittle:
- Tied to an undocumented panic-string format
- Forced on every consumer of the public API
- Leave any partial sink state (open upload session, buffered bytes) for the caller to clean up

### Expected behavior

Per Go's errors-as-values convention and the contract suggested by `pqarrow.NewFileWriter`'s `(*FileWriter, error)` signature, a failure to write the magic header should surface as an error to the caller, not a panic.

### Suggested fixes (smallest change first)

1. **Add a sibling constructor** that returns `(*Writer, error)`, e.g.:

   ```go
   func NewParquetWriterWithError(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) (*Writer, error)
   ```

   Non-breaking. Existing callers keep current panic semantics; new callers opt into clean error handling. `pqarrow.NewFileWriter` switches to the new constructor internally so its `error` return becomes meaningful for this failure mode.

2. **Defer `startFile()` until first append.** Move the magic-header write to lazy execution on the first `AppendRowGroup` / `AppendBufferedRowGroup` call — both of which already have error paths. Slightly larger change but doesn't add a new public symbol.

3. **Lazy error pattern.** Have `startFile` store its error in an unexported `Writer.initErr` field, and have every other public method (`AppendRowGroup`, `Close`, `WritePageIndex`, …) short-circuit on it. Non-breaking but touches every public method.

### Reproduction sketch

```go
// Synthetic flaky sink that fails the first Write.
type flakySink struct{ writes int }

func (f *flakySink) Write(p []byte) (int, error) {
    f.writes++
    if f.writes == 1 {
        return 0, errors.New("synthesized network blip")
    }
    return len(p), nil
}

func main() {
    // ... build a trivial schema.GroupNode ...
    defer func() {
        if r := recover(); r != nil {
            log.Printf("recovered: %v", r) // prints "failed to write magic number"
        }
    }()
    _ = file.NewParquetWriter(&flakySink{}, schemaNode) // panics here
}
```

### Component(s)

Parquet

### Version

`github.com/apache/arrow-go/v18@v18.6.0` (also reproducible on current `main`; the [code at line 201-203](https://github.com/apache/arrow-go/blob/main/parquet/file/file_writer.go#L201-L204) is unchanged).

zeroshade:
### Rationale for this change

Closes #820.

`(*file.Writer).startFile()` panics with the literal string `"failed to write magic number"` when the underlying sink's first `Write` call returns an error or short-writes. Because `file.NewParquetWriter` calls `startFile` synchronously inside the constructor and has no `error` return, callers had no idiomatic Go recovery path. The higher-level `pqarrow.NewFileWriter` already returns `(*FileWriter, error)`, but the panic propagated through it unchanged, so its `error` return was never reached for this failure mode.

In production this manifests when the sink is a network-attached writer (`*storage.Writer` from GCS, an S3 multipart upload, an HDFS client, etc.). A transient network blip on the first 4-byte magic-header write reliably crashes the worker process, orphans any in-flight upload session, and forces a container restart.

1. **`(*Writer).startFile()` now returns `error`** instead of panicking on sink-write failures. The configuration-validation panic for "encrypted column not found in file schema" is preserved — that path is a programmer error, not an I/O failure.
2. **New public constructor `file.NewParquetWriterWithError(w, sc, opts...) (*Writer, error)`** is the preferred entry point for callers whose sink may transiently fail.
3. **`file.NewParquetWriter` is now a thin wrapper** over `NewParquetWriterWithError` that panics with the *exact same string* (`"failed to write magic number"`) on init failure. Any consumer with a `recover()` block that string-matches the panic value continues to work unchanged.
4. **`pqarrow.NewFileWriter` switches to `NewParquetWriterWithError`** internally. Its public signature is unchanged; its `error` return now meaningfully covers transient sink failures, with no API change for any of its callers.

### Are these changes tested?

Yes. Six new tests in `parquet/file/file_writer_test.go` exercise the new behavior using a `flakyMagicSink` test helper modeled on the existing `errCloseWriter` pattern in the same file:

- `TestNewParquetWriterWithError_Success` — happy path on a `bytes.Buffer`.
- `TestNewParquetWriterWithError_FirstWriteFails` — first `Write` returns `io.ErrUnexpectedEOF`; constructor returns a wrapped error and does not panic.
- `TestNewParquetWriterWithError_FirstWriteShortWrites` — first `Write` returns `n=len(p)-1, nil`; constructor returns an error mentioning the short write.
- `TestNewParquetWriter_PreservesPanicMessage` — the legacy constructor still panics, the panic value is still a `string`, and the message is still the literal `"failed to write magic number"` (back-compat guard).
- `TestPqarrowNewFileWriter_PropagatesInitError` — `pqarrow.NewFileWriter` returns the wrapped sink error rather than panicking.
- `TestPqarrowNewFileWriter_PropagatesShortWrite` — same for the short-write case.

The existing `TestCloseError`, full `parquet/file/...` and `parquet/pqarrow/...` test suites all pass (with `PARQUET_TEST_DATA` pointed at `parquet-testing/data`).

**No breaking changes.**

- `file.NewParquetWriter(w, sc, opts...) *Writer` — signature unchanged, still panics on first-write failure with the identical literal string `"failed to write magic number"`. Existing `recover()` workarounds keep working.
- `file.NewParquetWriterWithError(w, sc, opts...) (*Writer, error)` — new exported symbol. Adding an exported function is non-breaking in Go.
- `pqarrow.NewFileWriter(...) (*FileWriter, error)` — signature unchanged. Its `error` return now also covers the magic-header-write failure that previously escaped as a panic. Callers who already check `err` get strictly better behavior.

The new constructor's docstring is the recommended path for users on cloud-storage upload sinks. The legacy constructor's docstring now documents the panic behavior the original issue called out as undocumented.

### Out of scope (separate follow-up)

While investigating, I noticed `WriteBatchSpaced` in `parquet/file/column_writer_types.gen.go` lacks the `defer recover()` that `WriteBatch` has, so an I/O error during a spaced write panics out of the calling goroutine instead of returning an error. Same class of bug as #820 but a different code path; happy to file a separate issue.

### DIFF
--- a/parquet/file/file_writer.go
+++ b/parquet/file/file_writer.go
@@ -70,12 +70,12 @@ func WithWriteMetadata(meta metadata.KeyValueMetadata) WriteOption {
 // If props is nil, then the default Writer Properties will be used. If the key value metadata is not nil,
 // it will be added to the file.
 //
-// This constructor panics with "failed to write magic number" if the
-// initial 4-byte write of the parquet magic header to the underlying sink
-// fails. Callers using a sink that may transiently fail (for example, a
-// cloud-storage upload writer or other network-attached writer) should
-// prefer [NewParquetWriterWithError], which surfaces the failure as a
-// returned error instead.
+// This constructor panics with the literal string "failed to write magic
+// number" if the initial write of the parquet magic header to the
+// underlying sink fails. The behavior is preserved for backward
+// compatibility with callers that string-match the panic value in a
+// recover() block; new code should prefer [NewParquetWriterWithError],
+// which returns the failure as an error instead.
 func NewParquetWriter(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) *Writer {
 	fw, err := NewParquetWriterWithError(w, sc, opts...)
 	if err != nil {
@@ -86,17 +86,16 @@ func NewParquetWriter(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) *W
 	return fw
 }
 
-// NewParquetWriterWithError is identical to [NewParquetWriter] except that
-// it returns an error to the caller if the initial write of the parquet
-// magic header to the underlying sink fails or short-writes. This is the
-// preferred constructor when the sink is backed by a writer that may
-// transiently fail (for example, a Google Cloud Storage resumable upload,
-// an S3 multipart upload, an HDFS client, or any other network-attached
-// writer) where a transient first-write failure should not panic the
-// calling process.
+// NewParquetWriterWithError returns a Writer that writes to the provided
+// io.Writer with the given schema.
 //
 // If props is nil, then the default Writer Properties will be used. If the
 // key value metadata is not nil, it will be added to the file.
+//
+// An error is returned if the initial write of the parquet magic header to
+// the underlying sink fails or short-writes, which can happen when the
+// sink is a flaky or network-attached writer (for example, a cloud-storage
+// upload writer).
 func NewParquetWriterWithError(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) (*Writer, error) {
 	config := &writerConfig{}
 	for _, o := range opts {
--- a/parquet/file/file_writer.go
+++ b/parquet/file/file_writer.go
@@ -65,11 +65,39 @@ func WithWriteMetadata(meta metadata.KeyValueMetadata) WriteOption {
 	}
 }
 
-// NewParquetWriter returns a Writer that writes to the provided WriteSeeker with the given schema.
+// NewParquetWriter returns a Writer that writes to the provided io.Writer with the given schema.
 //
 // If props is nil, then the default Writer Properties will be used. If the key value metadata is not nil,
 // it will be added to the file.
+//
+// This constructor panics with "failed to write magic number" if the
+// initial 4-byte write of the parquet magic header to the underlying sink
+// fails. Callers using a sink that may transiently fail (for example, a
+// cloud-storage upload writer or other network-attached writer) should
+// prefer [NewParquetWriterWithError], which surfaces the failure as a
+// returned error instead.
 func NewParquetWriter(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) *Writer {
+	fw, err := NewParquetWriterWithError(w, sc, opts...)
+	if err != nil {
+		// Preserve the historical panic value verbatim so any consumer
+		// performing a string-match in a recover() block continues to work.
+		panic("failed to write magic number")
+	}
+	return fw
+}
+
+// NewParquetWriterWithError is identical to [NewParquetWriter] except that
+// it returns an error to the caller if the initial write of the parquet
+// magic header to the underlying sink fails or short-writes. This is the
+// preferred constructor when the sink is backed by a writer that may
+// transiently fail (for example, a Google Cloud Storage resumable upload,
+// an S3 multipart upload, an HDFS client, or any other network-attached
+// writer) where a transient first-write failure should not panic the
+// calling process.
+//
+// If props is nil, then the default Writer Properties will be used. If the
+// key value metadata is not nil, it will be added to the file.
+func NewParquetWriterWithError(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) (*Writer, error) {
 	config := &writerConfig{}
 	for _, o := range opts {
 		o(config)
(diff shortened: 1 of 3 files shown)


lidavidm:
> While investigating, I noticed WriteBatchSpaced in parquet/file/column_writer_types.gen.go lacks the defer recover() that WriteBatch has, so an I/O error during a spaced write panics out of the calling goroutine instead of returning an error. Same class of bug as https://github.com/apache/arrow-go/issues/820 but a different code path; happy to file a separate issue.

Uhh, so @zeroshade do you (or your agent) want to file that issue? :sweat_smile:

zeroshade:
Filed #826 🫤

parquet/file/file_writer.go
```diff
@@ -65,11 +65,39 @@ func WithWriteMetadata(meta metadata.KeyValueMetadata) WriteOption {
...
 // If props is nil, then the default Writer Properties will be used. If the key value metadata is not nil,
 // it will be added to the file.
+//
+// This constructor panics with "failed to write magic number" if the
+// initial 4-byte write of the parquet magic header to the underlying sink
+// fails. Callers using a sink that may transiently fail (for example, a
+// cloud-storage upload writer or other network-attached writer) should
+// prefer [NewParquetWriterWithError], which surfaces the failure as a
+// returned error instead.
 func NewParquetWriter(w io.Writer, sc *schema.GroupNode, opts ...WriteOption) *Writer {
```
lidavidm, line 79: Should we deprecate this as well?

parquet/file/file_writer.go
lidavidm, line 1: For someone coming to the codebase _without_ being a maintainer, these doc comments are too focused on historical knowledge. I think NewParquetWriterWithError should be documented without reference to NewParquetWriter and simply mention that it may fail if it cannot write the header (e.g. because of a flaky writer), and NewParquetWriter can document the history if needed.

parquet/file/file_writer.go
zeroshade: My preference would be to not deprecate this unless we provide a similar short-named function, The new `NewParquetWriterWithError` function is a bit long to be the canonical function to use. Given the only difference is just having an error return instead of a panic, we could either rename the new function or keep both as relevant.

parquet/file/file_writer.go
zeroshade: updated

lidavidm (approved):

PR #825, opened 2026-05-22
feat(arrow/compute/exprs): support expr.IntervalYearToMonthLiteral in literalToDatum

### LINKED ISSUE #815: Support `expr.IntervalYearToMonthLiteral` in `literalToDatum`
`arrow/compute/exprs.literalToDatum` handles interval year/month literals through the `*expr.ProtoLiteral` path, but not the raw `expr.IntervalYearToMonthLiteral` type.

Let's add support for `expr.IntervalYearToMonthLiteral`.

Relevant code:

- `literalToDatum` type switch: https://github.com/apache/arrow-go/blob/v18.5.1/arrow/compute/exprs/exec.go#L182-L415
- Existing interval year/month handling in the `*expr.ProtoLiteral` path: https://github.com/apache/arrow-go/blob/v18.5.1/arrow/compute/exprs/exec.go#L355-L366
- `expr.IntervalYearToMonthLiteral`: https://github.com/substrait-io/substrait-go/blob/v7.5.0/expr/interval_year_to_month.go#L11-L68

zeroshade:
### Rationale for this change

Closes #815.

`arrow/compute/exprs.literalToDatum` handled interval year/month literals only through the `*expr.ProtoLiteral` path. The substrait-go library also surfaces these as the dedicated `expr.IntervalYearToMonthLiteral` type — that is the value [`intervalYearToMonthLiteralFromProto`](https://github.com/substrait-io/substrait-go/blob/v8.1.1/expr/interval_year_to_month.go#L41-L47) returns when deserializing protobuf — and any caller reaching `literalToDatum` with this type previously fell through to `arrow.ErrNotImplemented`.

1. **New type-switch cases for `expr.IntervalYearToMonthLiteral`** (value form, the canonical form returned by substrait-go's deserializer) **and `*expr.IntervalYearToMonthLiteral`** (pointer form, used in parts of substrait-go itself, e.g. `expr/string_test.go:86`). Both produce the same `FixedSizeList<int32>[2]` extension scalar that the existing `*expr.ProtoLiteral` path already produces, so callers reaching either form get a Datum that is `compute.Equals` to the one they would have gotten via the protobuf path.
2. **New helper `intervalYearToMonthDatum(mem, years, months)`**. The existing `*types.IntervalYearToMonthType` branch inside the `*expr.ProtoLiteral` case now calls this helper instead of inlining the builder logic, so the two paths cannot drift.
3. **Allocator threading**. The helper takes the `mem memory.Allocator` parameter that `literalToDatum` already receives, instead of hardcoding `memory.DefaultAllocator` (which the original inlined branch did). All three call sites forward `mem`.

### Are these changes tested?

Yes. New `TestLiteralToDatumIntervalYearToMonth` in `arrow/compute/exprs/exec_internal_test.go` constructs a `*expr.ProtoLiteral` baseline and a parallel `expr.IntervalYearToMonthLiteral` (value) and `*expr.IntervalYearToMonthLiteral` (pointer), then asserts the resulting Datums are `compute.Equals`. Existing `parquet/file/...`, `parquet/pqarrow/...`, and `arrow/compute/exprs/...` test suites all continue to pass; full repo `go build ./...` is clean.

The test deliberately uses `memory.DefaultAllocator` rather than `memory.NewCheckedAllocator` with `AssertSize(t, 0)`. `*scalar.Extension` does not implement `Release()` (see `arrow/scalar/scalar.go:397-466`), so an extension scalar's underlying storage is never released even when the wrapping Datum is — adding a checked-allocator assertion would chase a pre-existing leak that is unrelated to this issue. (Filing that as a separate follow-up issue is straightforward and out of scope here.)

Yes, but **purely additive**:

- Callers passing an `expr.IntervalYearToMonthLiteral` (or pointer to one) to `literalToDatum`, `ExecuteScalarExpression`, or related entry points used to receive `arrow.ErrNotImplemented`; they now receive a properly-formed extension Datum identical to what the `*expr.ProtoLiteral` path produces.
- No public API signatures change.
- No existing call site behavior changes (the `*expr.ProtoLiteral` path produces a byte-identical Datum to before).

### Out of scope (separate follow-up)

`*scalar.Extension` does not implement `Release()`, so extension scalars (including the ones produced by this code path, the existing `*types.IntervalYearToMonthType` path, the `*types.IntervalDayType` path, and `intervalDay()` / `intervalYear()` extension types in general) leak their storage through `compute.ScalarDatum.Release()`. Worth a separate issue.

### DIFF
--- a/arrow/compute/exprs/exec_internal_test.go
+++ b/arrow/compute/exprs/exec_internal_test.go
@@ -121,6 +121,7 @@ func TestLiteralToDatumIntervalYearToMonth(t *testing.T) {
 	// *scalar.Extension does not implement Release() (see
 	// arrow/scalar/scalar.go), so an extension scalar's underlying
 	// storage is never released even when the wrapping Datum is.
+	// This will be fixable once https://github.com/apache/arrow-go/issues/827 is addressed.
 	mem := memory.DefaultAllocator
 
 	extSet := NewExtensionSetDefault(
--- a/arrow/compute/exprs/exec_internal_test.go
+++ b/arrow/compute/exprs/exec_internal_test.go
@@ -29,6 +29,9 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/substrait-io/substrait-go/v8/expr"
+	"github.com/substrait-io/substrait-go/v8/extensions"
+	"github.com/substrait-io/substrait-go/v8/types"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


lidavidm:
@zeroshade same here.

> *scalar.Extension does not implement Release(), so extension scalars (including the ones produced by this code path, the existing *types.IntervalYearToMonthType path, the *types.IntervalDayType path, and intervalDay() / intervalYear() extension types in general) leak their storage through compute.ScalarDatum.Release(). Worth a separate issue.

zeroshade:
Filed #827

arrow/compute/exprs/exec_internal_test.go
```diff
@@ -112,3 +115,47 @@ func TestMakeExecBatch(t *testing.T) {
 		})
 	}
 }
+
+func TestLiteralToDatumIntervalYearToMonth(t *testing.T) {
+	// memory.NewCheckedAllocator with AssertSize would fail here:
```
lidavidm, line 120: This should be linked to the new issue and fixed when the new issue is fixed.

lidavidm (approved):

arrow/compute/exprs/exec_internal_test.go
zeroshade: added

PR #830, opened 2026-05-26
refactor(arrow/avro): migrate from hamba/avro to twmb/avro

### LINKED ISSUE #813: [Avro] hamba/avro is abandoned

https://github.com/hamba/avro is abandoned since January 2026.
The iceberg-go project has already migrated:
https://github.com/apache/iceberg-go/commit/e175604d1d0582a7942e9402bd3f49e2acee5c0b

### Component(s)

Other

prochac:
### Rationale for this change

Closes https://github.com/apache/arrow-go/issues/813

`arrow/avro` moves from `hamba/avro/v2` to `twmb/avro`.

The parsing side is simplified: `twmb/avro` hands union values to the decoder
directly instead of wrapping them in a `map[string]any` tagged with the branch
name, so the reader no longer has to unwrap `{"long.timestamp-millis": v}`
shapes on every appender.

Also included:

- `timestamp-nanos` and `local-timestamp-nanos` logical types are now supported
  and map to `timestamp[ns, tz=UTC]` / `timestamp[ns]`.
- A `fixed(16)` field annotated `logicalType: "uuid"` now keeps its logical
  type and decodes to `extensions.UUIDType` instead of falling back to
  `fixed_size_binary[16]`.
- Named-type references are resolved against fully-qualified names. A bare
  short name that is defined under more than one namespace is now rejected;
  previously the hamba schema cache was keyed on the short name alone, so such
  a reference silently resolved to whichever definition was parsed last.
- The top-level schema must be a record. OCF requires this, and anything else
  previously produced a degenerate or empty Arrow schema.
- Appender errors raised for map values and list items now propagate out of
  `loadDatum` instead of being dropped.

### Are these changes tested?

Yes. All existing tests pass, and the suite gained coverage for:

- `timestamp-nanos` / `local-timestamp-nanos`, in the `alltypes.avsc` fixture
  and end to end
- nullable logical timestamps, with `time.Local` pinned to a non-UTC zone so a
  machine's TZ cannot shift decoded values
- rejection of every union shape other than `["null", T]`, including nested
  ones
- the `uuid` logical type, on both `string` and `fixed(16)`
- ambiguous bare named-type references
- `Reuse()`, both the accept path and the schema-mismatch rejection
- positive `WithChunk(n)` sizes, including the short final batch
- records made up only of enum columns

Yes, three.

**`ArrowSchemaFromAvro` is deprecated.** Its signature takes a
`hamba/avro/v2.Schema`, which couples callers to a third-party Avro module.
Use `ArrowSchemaFromAvroJSON(string)` instead. The deprecated function still
works — it marshals the hamba schema back to JSON and forwards.

**Custom decoders registered with hamba are no longer applied.** Anyone who
called [`hamba/avro/v2.Register`](https://pkg.go.dev/github.com/hamba/avro/v2#Register)
to map an Avro type onto their own Go type was, in effect, changing what the
appenders received. Those registrations are invisible to `twmb/avro`, so values
that previously arrived as a custom type now arrive as the decoder's native Go
type — and where a registration produced something the appenders did not expect,
the reader now returns `avro: unsupported value of type X for ... column` rather
than silently accepting it. There is no equivalent hook to carry these
registrations across, since the global registry was never part of the arrow-go
API. Callers that relied on it need to do the conversion after reading.

**Ambiguous bare named-type references now error.** See the note above: a
schema that defines `Foo` in two namespaces and then refers to a bare
`{"type": "Foo"}` used to resolve to one of them arbitrarily and now returns an
error asking for a fully-qualified name. This only affects schemas that were
already ambiguous.


arrow/avro/schema.go
```diff
@@ -56,33 +67,84 @@ func (node *schemaNode) schemaPath() string {
...
+// so later references like {"type": "Address"} or {"type": "ns.Address"}
+// resolve back to the original definition.
+func (node *schemaNode) rememberNamed(s avro.SchemaNode) {
+	if s.Name == "" {
+		return
+	}
+	node.namedCache[s.Name] = s
+	if s.Namespace != "" {
+		node.namedCache[s.Namespace+"."+s.Name] = s
+	}
```
zeroshade: It's worth noting that the hamba SchemaCache keyed on full name only, so for schemas with two records named `Foo` in different namespaces a bare `{"type": "Foo"}` will resolve to whichever was defined latest as opposed to being an error (the previous behavior under hamba).

We should either restrict to full-name lookups or document the change in behavior here that where a malformed schema may now silently produce something where hamba would have errored.

arrow/avro/reader_types.go
```diff
@@ -574,126 +539,108 @@ func mapFieldBuilders(b array.Builder, field arrow.Field, parent *fieldPos) {
...
-			err := binary.Read(buf, binary.BigEndian, &intData)
-			if err != nil {
-				return err
-			}
-			b.Append(decimal128.FromI64(intData))
-		} else {
-			var bigIntData big.Int
-			b.Append(decimal128.FromBigInt(bigIntData.SetBytes(buf.Bytes())))
-		}
-	case map[string]any:
```
zeroshade, line 727: If anyone was using custom-decoders that emitted `[]byte` or `map[string]any` etc. they are now going to see "unsupported value of type X" errors where old code would have accepted the input. Let's update the description on the PR (and therefore the resulting changelog later) to note this change in behavior for anyone who was using custom-decoders.

arrow/avro/reader_types.go
```diff
@@ -574,126 +539,108 @@ func mapFieldBuilders(b array.Builder, field arrow.Field, parent *fieldPos) {
...
+	case string:
+		return b.AppendValueFromString(dt)
+	case [16]byte:
+		b.AppendBytes(dt)
+	case []byte:
+		switch len(dt) {
+		case 16:
+			b.AppendBytes([16]byte(dt))
+		case 36:
+			return b.AppendValueFromString(string(dt))
```
zeroshade: let's add a comment here explaining why we have a case for 36 (i guess, hex-dash UUID string?)

arrow/avro/schema.go
```diff
@@ -277,22 +335,46 @@ func iterateFields(n *schemaNode) {
...
+// Heterogeneous non-nullable unions (e.g. ["null", "int", "string"] or
+// ["int", "string"]) are not supported and callers panic on them rather
+// than silently picking one arm.
+func nullableBranch(s avro.SchemaNode) (avro.SchemaNode, bool) {
+	if s.Type != "union" || len(s.Branches) < 2 {
+		return avro.SchemaNode{}, false
+	}
+	var nonNull *avro.SchemaNode
+	for i := range s.Branches {
+		b := s.Branches[i]
```
zeroshade:
```suggestion
	for _, b := range s.Branches {
```

arrow/avro/schema.go
```diff
@@ -56,33 +67,84 @@ func (node *schemaNode) schemaPath() string {
...
 			s = nil
 			err = utils.FormatRecoveredError("invalid avro schema", r)
 		}
 	}()
+	root := schema.Root()
 	n := newSchemaNode()
-	n.schema = schema
-	c := n.newChild(n.schema.(avro.NamedSchema).Name(), n.schema)
+	n.node = root
+	c := n.newChild(root.Name, root)
```
zeroshade, line 187: OCF requires a named record at the top, we should probably add a guard or explicit "must be named record" check here (the hamba code would have panic'd).

zeroshade (changes requested):
Can you rebase and resolve the conflicts please?

zeroshade:
@prochac are you still working on this?

arrow/avro/reader_types.go
prochac: This was possible only if they used the global type registry. The `map[string]any` shouldn't affect people  who were using only the arrow-go API. But those who used https://pkg.go.dev/github.com/hamba/avro/v2#Register ... There is no good way to handle that :/

arrow/avro/reader_types.go
```diff
@@ -92,21 +90,10 @@ func (d *dataLoader) drawTree(field *fieldPos) {
...
-// filterNullStruct drops ErrNullStructData, which signals a null struct
-// whose sub-fields can be skipped rather than a failure.
-func filterNullStruct(err error) error {
-	if err == ErrNullStructData {
-		return nil
-	}
-	return err
-}
-
 func (d *dataLoader) loadDatum(data any) error {
```
zeroshade, line 93: this currently propagates errors for top-level scalar fields, but discards them everywhere else:

- mapValue struct loader (line 96)
- recursive calls to a child's `loadDatum` (lines 137, 155, 160, 162, 164, 195, 198, 222-223)
- list append and items: lines 172, 174, 177, 203, 206, 227-228
- map branch append/key/value: lines 234, 237, 239, 241, 243

The original test being replaced asserted that the map-value and list-item type mismatches would have surfaced as errors. Can we re-wrap the nested appendFunc/loadDatum calls with the same error/`ErrNullStructData` handling used in the scalar loops? And can we restore a `TestLoadDatumPropagatesNestedAppendErrors` unit test (using `twmb` of course)

arrow/avro/reader_types.go
zeroshade: as far as I'm aware, I don't think `twmb` ever yields a 36-byte `[]byte` for a uuid case, so what scenario is this testing? If this is unreachable can we just remove it?

arrow/avro/reader.go
zeroshade: nit: should we change this to doing an actual schema check instead? just so that we don't end up rejecting cases with different whitespace/formatting

zeroshade (changes requested):
Looks great! Only the loadDatum comment is blocking, the rest are nitpicks.

Also can we add some test scenarios for:

- timestamp-nanos/local-timestamp-nanos: you added them but they aren't in the `alltypes.avsc` or `testdata.go` fixtures.
- non-utc local-timestamp: a small unit test to assert the local-timestamp value in a non-utc TZ
- heterogeneous-union rejection: something like `["int", "string"], ["null", A, B]` will properly panic -> error, but the test `TestComplexUnionReportsError` was removed and not replaced, let's add a test for it and assert the new "unsupported avro union" message
- We need tests for `Reuse()`, a positive `WithChunk(n)` and enum-only columns. The code looks correct, but there's no tests for them currently.

arrow/avro/reader.go
prochac: This is a good catch. The `String()` method gave me a not-just-one headache :D 
by [the docs](https://pkg.go.dev/github.com/twmb/avro#Schema.String), `twmb/avro` preserves parsing data, meanwhile `hamba/avro` gives the "reduced" canonical form.

https://pkg.go.dev/github.com/twmb/avro#Schema.Canonical mentions fingerprinting, so I'm going to listen to the docs.

zeroshade (approved):
Thanks for working through all of this — the nested `loadDatum` error propagation is exactly what I was after, and the added coverage (nanos fixtures, non-UTC local-timestamp, union rejection, `Reuse`, `WithChunk`, enum-only) all looks great.

I also verified the local-timestamp change: `twmb/avro` decodes both the global and local timestamp logical types through `time.Unix…().UTC()` (local types map to the same UTC deserializers), so dropping the old hamba wall-clock reinterpretation is correct — and the non-UTC `time.Local` test confirms the value doesn't shift. Build, `go test -race`, and `go vet` are all green for me, and it merges cleanly onto current `main`.

LGTM. Two small **non-blocking** follow-ups that are fine to handle in a later PR:

1. `Reuse()` compares `Canonical()` (Parsing Canonical Form), which by spec drops `logicalType`. So a second file whose schema differs *only* by a logical-type annotation would be accepted for reuse. It fails safe today (you'd hit a decode-time type mismatch rather than silent corruption), so it's not blocking — just noting it in case we want a stricter check later.

2. Dropping the `golang/snappy` indirect dep relies on `twmb/avro/ocf` handling snappy-compressed OCF internally. We don't currently have a test that reads a snappy-compressed file — worth adding one so we don't silently regress OCF codec support.

Approving; will squash-merge once CI is green.

PR #831, opened 2026-05-30
perf(parquet/metadata): avoid constructing ColumnChunkMetaData in page index range determination

paveon:
### Rationale for this change

  `determinePageIndexRangesInRowGroup` called `RowGroupMetaData.ColumnChunk()` for every column in a row group, constructing a full `ColumnChunkMetaData` (allocating the struct, copying encoding slices, etc.) just to read two offset/length
  pairs. In a compaction workload processing hundreds of small Parquet files, this path alone accounted for 18.3 GB of allocation churn (36% of total allocations). The objects were immediately discarded after reading two int fields.

  Added lightweight `ColumnIndexLocation()` and `OffsetIndexLocation()` methods on `RowGroupMetaData` that read index locations directly from the underlying thrift struct with zero heap allocations. Updated `determinePageIndexRangesInRowGroup`
  to use these instead of constructing full `ColumnChunkMetaData` objects.

  ### Are these changes tested?

  Yes — existing `parquet/metadata` and `parquet/file` tests pass. Verified zero heap allocations on the hot path via escape analysis (`go build -gcflags='-m -m'`). Confirmed with heap profiles that `NewColumnChunkMetaData` allocations through
  this path dropped from 18.3 GB to 106 MB (99.4% reduction).

  Two new public methods on `RowGroupMetaData`: `ColumnIndexLocation(i int) (IndexLocation, bool)` and `OffsetIndexLocation(i int) (IndexLocation, bool)`. No breaking changes.

### DIFF
--- a/parquet/metadata/page_index.go
+++ b/parquet/metadata/page_index.go
@@ -443,7 +443,6 @@ func determinePageIndexRangesInRowGroup(rgMeta *RowGroupMetaData, cols []int32) 
 		return nil
 	}
 
-	var colChunk *ColumnChunkMetaData
 	if len(cols) == 0 {
 		cols = make([]int32, rgMeta.NumColumns())
 		for i := 0; i < rgMeta.NumColumns(); i++ {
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Nice! Thanks for this. Do any of the existing benchmark's show the reduced allocations? If not, can we add a benchmark that hits this path to keep track of it?

paveon:
Thanks! I added new benchmark test and also regular test to enforce that allocs/op is not scaling linearly with the number of columns. Let me know what you think.

zeroshade (approved):

PR #832, opened 2026-06-01
feat(avro): support local-timestamp logical types

abir-eon:
### Rationale for this change

`arrow/avro/schema.go` had the Avro `local-timestamp-millis` and `local-timestamp-micros` logical types commented out with a note that they were "not implemented in hamba/avro". As of the `github.com/hamba/avro/v2` version this module depends on, both types are fully supported (defined as logical types, parsed, and encoded/decoded to `time.Time`), so the note is stale and the types can now be mapped.

### What changes are included in this PR

- Map `local-timestamp-millis` / `local-timestamp-micros` to a **zone-less** `arrow.TimestampType` (`Millisecond` / `Microsecond` unit, empty `TimeZone`), reflecting their wall-clock semantics — as opposed to the global `timestamp-millis` / `timestamp-micros` types which map to UTC-zoned types.
- Fix value conversion in `reader_types.go`: hamba decodes a `local-timestamp` into a `time.Time` whose wall-clock fields are correct but whose location is the reader's machine zone. Because `arrow.TimestampFromTime` derives the stored value from the instant (`Unix()`), the machine's zone offset would otherwise leak into the stored value. The wall-clock fields are reinterpreted in UTC before conversion so the stored value is zone-independent.
- Extend the all-types schema and roundtrip tests (`schema_test.go`, `reader_test.go`, `testdata/`) to cover both new types. The roundtrip test was verified to fail without the conversion fix under a non-UTC timezone (e.g. `TZ=Asia/Kolkata`) and pass with it.

### Are these changes tested?

Yes — `go test ./arrow/avro/...` passes, including under non-UTC timezones.

Yes. Avro schemas using `local-timestamp-millis` / `local-timestamp-micros` were previously unsupported; they now convert to zone-less `arrow.TimestampType` columns.

### DIFF
--- a/arrow/avro/reader_test.go
+++ b/arrow/avro/reader_test.go
@@ -144,6 +144,14 @@ func TestReader(t *testing.T) {
 					Type: arrow.FixedWidthTypes.Timestamp_us,
 				},
 				{
+					Name: "localtimestampmillis",
+					Type: &arrow.TimestampType{Unit: arrow.Millisecond},
+				},
+				{
+					Name: "localtimestampmicros",
+					Type: &arrow.TimestampType{Unit: arrow.Microsecond},
+				},
+				{
 					Name: "duration",
 					Type: arrow.FixedWidthTypes.MonthDayNanoInterval,
 				},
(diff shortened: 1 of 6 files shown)


zeroshade (approved):

zeroshade:
LGTM! Thanks for this, I'll merge once the CI completes if all is green

PR #833, opened 2026-06-01
fix(arrow/array): Parity with Arrow C++ when reading/writing `null` to/from JSON

serramatutu:
### Rationale for this change

I raised an issue at the Arrow Community meeting regarding how Arrow Go allows the user to construct invalid record batches by reading JSON with nulls in it even if the schema says it's not null. It is also possible to construct invalid batches by calling `AppendNull()` in the builders, and currently there is no way to validate that. I was instructed to look at how Arrow C++/PyArrow do it, and replicate it here.

TL;DR:
1. Arrow C++ raises errors when reading JSON data that does not conform to the schema
2. Arrow C++ never JSON-encodes values as `null` if the parent field is non-nullable

Here's a link to my full investigation, including the code so you can run it yourself: https://github.com/serramatutu/arrow-internal-nulls

Rebased on `main`, and reworked to address @zeroshade's review.

**Reading (review point 4, plus the original issue 1).** Nullability is now validated recursively, in one place, before anything is appended:

- `arrow/array/nullability.go` adds `validateJSONNullability`, which walks a buffered row against the `arrow.Field` tree. It covers struct fields, the element field of lists / large lists / list views / fixed-size lists, map key and item fields, union children, dictionary values, run-end-encoded values and extension storage types. So `ListOfNonNullable(int32)` now rejects `[1, null]`, and a missing non-nullable field is rejected at any depth.
- `RecordBuilder` and `StructBuilder` keep upstream's streaming append path untouched. They only buffer the row and validate it when the schema declares a non-nullable field somewhere in its type tree (`typeHasNonNullableField`), so fully nullable schemas take exactly the same path as before this PR.
- Buffering the row also means a rejected row leaves the decoder positioned at the *next* row, so callers can keep reading. That is not achievable with streaming decode: an error raised inside a nested builder leaves the decoder at an unknown depth, which cannot be drained reliably.
- Roots that are not records or structs are validated in `FromJSON`, so `array.FromJSON(mem, arrow.ListOfNonNullable(...), ...)` is checked too.

**Rollback (review points 1 and 2).** #1113 landed while this PR was open and provides exactly the pre-row checkpoint the review asked for, so the `Resize(-1)` rollbacks are gone:

- `RecordBuilder` uses upstream's `builderCheckpoint` capture/restore as-is.
- `StructBuilder` now builds its own reusable `builderCheckpoint`, because it is the root builder for `array.FromJSON` on a struct type and had no rollback of its own. `Resize(-1)` is no longer used, and the panic the review reported (`columnLenRange` returning `-1` into child `Resize`) is gone with it.

**Writing (review point 3, plus the original issue 2).** Writers now fail instead of emitting a null that this PR's reader would reject:

- `RecordToJSON` and `Struct.MarshalJSON` (and therefore `RecordBatch.MarshalJSON`, which goes through `RecordToStructArray`) return `arrow.ErrInvalid` when a non-nullable field holds a null.
- The check honors parent-validity priority, so a null child under a null struct parent is *not* an error — that matches what the encoders actually emit, since they stop descending at a null parent.
- The previous `dtype.Field(i).Nullable && IsNull(i)` branches were no-ops (`GetOneForMarshal` returns `nil` for a null slot either way), so they were reverted.

I had to change a bunch of tests that implicitly depended on fields being nullable, even if they didn't declare the fields as such.

### Scope

`MarshalJSON` on non-struct array types (a list array marshalled on its own, say) still emits nulls for non-nullable element fields. Reading is validated for those roots; writing is not. Happy to extend it if you'd like it in this PR.

### Are these changes tested?

Yes — `arrow/array/nullability_test.go` covers nested list/map/union/struct element nullability, root-level `FromJSON`, writer errors, parent-validity priority, decoding the row after a rejected row, and a JSON round trip over a schema mixing nullable and non-nullable fields. There is also a regression test for the reported `StructBuilder` panic (a malformed nested list value followed by a valid row).

`go test ./arrow/... ./parquet/...` passes, with the `arrow_json_stdlib` backend as well.

### Performance

Buffering the row turned out to be a large *speed-up*, not a cost. `go test ./arrow/array -run '^$' -bench 'BenchmarkRecordFromJSON/Size_1000$' -benchtime=3x -count=2 -benchmem` (that benchmark's schema declares all five fields non-nullable, so it takes the new path):

| | ns/op | B/op | allocs/op |
| --- | --- | --- | --- |
| `main` (`b3dacd2a`) | 7,769,030,778 / 7,836,212,958 | 6,141,082 | 16,130 |
| this branch | 26,246,764 / 25,082,431 | 29,142,202 | 53,071 |

That is ~305x faster for ~4.7x the allocated bytes. A CPU profile of the old path attributes 98% of samples to `runtime.memmove` under `goccy/go-json`'s `decodeEscapeString` ← `(*Stream).Token`: decoding string tokens one at a time straight off the document decoder is quadratic in the remaining buffer, so it gets worse the larger the document. Decoding each row from its own small `json.RawMessage` avoids that entirely.

The flip side: a schema with no non-nullable field anywhere still takes the streaming path and so keeps the quadratic behaviour (~7.8 s/op on the same data). I kept the fast path opt-in by nullability to keep this PR's behaviour change scoped — an unconditional row buffer would fix it for every schema, but it would also force `UseNumber` on caller-supplied decoders in `UnmarshalOne`. Happy to do that here, or file it separately, whichever you prefer.

Yes, two breaking changes:

1. The JSON decoder is stricter: `null` or a missing value for a non-nullable field is now an error, at any nesting depth. This breaks callers relying on the decoder's leniency.
2. The JSON encoders now return an error for arrays that hold a null in a non-nullable field, instead of silently writing `null`.

### DIFF
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -418,6 +418,9 @@ func (b *RecordBuilder) NewRecord() arrow.Record {
 // UnmarshalOne reads one row (a JSON object) from the supplied decoder and
 // appends a value to each field in the RecordBuilder.
 //
+// Missing nullable fields get nulls. If an error is found while decoding
+// any field, the entire row gets discarded and the error is returned.
+//
 // Unlike UnmarshalJSON, this method receives an already-configured
 // json.Decoder, so options such as UseNumber set by the caller are honored
 // for nested field decoding. This is critical for preserving large integer
@@ -478,21 +481,20 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
 		}
 	}
 
-	// at this point we know there are no integrity errors, append values to field builders
-	for key, val := range keylist {
+	// At this point we know there are no integrity errors, so append values to the
+	// field builders in schema order.
+	for i := 0; i < b.schema.NumFields(); i++ {
+		val, ok := keylist[b.schema.Field(i).Name]
+		if !ok {
+			b.fields[i].AppendNull()
+			continue
+		}
+
 		valDec := json.NewDecoder(bytes.NewReader(val))
 		valDec.UseNumber()
-
-		indices := b.schema.FieldIndices(key)
-		if err := b.fields[indices[0]].UnmarshalOne(valDec); err != nil {
+		if err := b.fields[i].UnmarshalOne(valDec); err != nil {
+			b.Resize(-1)
 			return err
-		}
-	}
-
-	// append nulls to nullable fields if values were not present
-	for i := 0; i < b.schema.NumFields(); i++ {
-		if _, ok := keylist[b.schema.Field(i).Name]; !ok {
-			b.fields[i].AppendNull()
 		}
 	}
 
@@ -501,7 +503,7 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
 
 // Unmarshal reads multiple rows from the decoder, calling UnmarshalOne in a
 // loop until dec.More() reports there are no more values. Like UnmarshalOne,
-// this honors decoder configuration such as UseNumber set by the caller.
+// field values are always re-decoded with UseNumber enabled.
 func (b *RecordBuilder) Unmarshal(dec *json.Decoder) error {
 	for dec.More() {
 		if err := b.UnmarshalOne(dec); err != nil {
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -416,8 +416,7 @@ func (b *RecordBuilder) NewRecord() arrow.Record {
 }
 
 // UnmarshalOne reads one row (a JSON object) from the supplied decoder and
-// appends a value to each field in the RecordBuilder. Missing fields are
-// appended as nulls and unrecognized keys are silently ignored.
+// appends a value to each field in the RecordBuilder.
 //
 // Unlike UnmarshalJSON, this method receives an already-configured
 // json.Decoder, so options such as UseNumber set by the caller are honored
@@ -434,7 +433,8 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
 		return fmt.Errorf("record should start with '{', not %s", t)
 	}
 
-	keylist := make(map[string]bool)
+	// consume one row checking for duplicates and nulls
+	keylist := make(map[string]json.RawMessage)
 	for dec.More() {
 		keyTok, err := dec.Token()
 		if err != nil {
@@ -442,23 +442,27 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
 		}
 
 		key := keyTok.(string)
-		if keylist[key] {
+		if _, ok := keylist[key]; ok {
 			return fmt.Errorf("key %s shows up twice in row to be decoded", key)
 		}
-		keylist[key] = true
+
+		var val json.RawMessage
+		if err := dec.Decode(&val); err != nil {
+			return err
+		}
 
 		indices := b.schema.FieldIndices(key)
 		if len(indices) == 0 {
-			var extra interface{}
-			if err := dec.Decode(&extra); err != nil {
-				return err
-			}
 			continue
 		}
 
-		if err := b.fields[indices[0]].UnmarshalOne(dec); err != nil {
-			return err
+		idx := indices[0]
+
+		if bytes.Equal(val, []byte("null")) && !b.schema.Field(idx).Nullable {
+			return fmt.Errorf("field '%s' is non-nullable but got null", key)
 		}
+
+		keylist[key] = val
 	}
 
 	// consume the closing '}'
@@ -466,11 +470,32 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
 		return err
 	}
 
+	// check that all non-nullable fields were specified
 	for i := 0; i < b.schema.NumFields(); i++ {
-		if !keylist[b.schema.Field(i).Name] {
+		f := b.schema.Field(i)
+		if _, ok := keylist[f.Name]; !ok && !f.Nullable {
+			return fmt.Errorf("field '%s' is required but no value was given", f.Name)
+		}
+	}
+
+	// at this point we know there are no integrity errors, append values to field builders
+	for key, val := range keylist {
+		valDec := json.NewDecoder(bytes.NewReader(val))
+		valDec.UseNumber()
+
+		indices := b.schema.FieldIndices(key)
+		if err := b.fields[indices[0]].UnmarshalOne(valDec); err != nil {
+			return err
+		}
+	}
+
+	// append nulls to nullable fields if values were not present
+	for i := 0; i < b.schema.NumFields(); i++ {
+		if _, ok := keylist[b.schema.Field(i).Name]; !ok {
 			b.fields[i].AppendNull()
 		}
 	}
+
 	return nil
 }
 
--- a/arrow/array/record_test.go
+++ b/arrow/array/record_test.go
@@ -522,7 +522,7 @@ func TestRecordBuilder(t *testing.T) {
 	err = b.UnmarshalJSON([]byte(`{"f1-i32": null, "map": null}`))
 	assert.Contains(t, err.Error(), "field 'f2-f64-notnull' is required but no value was given")
 
-	err = b.UnmarshalJSON([]byte(`{"f1-i32": 6, "f2-f64-notnull": 6.6, "map": [{"key": "4": "value": "d"}]}`))
+	err = b.UnmarshalJSON([]byte(`{"f1-i32": 6, "f2-f64-notnull": 6.6, "map": [{"key": "4", "value": "d"}]}`))
 	assert.NoError(t, err)
 
 	rec := b.NewRecordBatch()
--- a/arrow/array/record_test.go
+++ b/arrow/array/record_test.go
@@ -564,6 +564,46 @@ func TestRecordBuilder(t *testing.T) {
 	assert.Truef(t, array.Equal(arr, roundtripped), "JSON round trip returns different array: got=%q, want=%d", arr, roundtripped)
 }
 
+func TestRecordBuilderUnmarshalOneFieldErrorKeepsColumnsConsistent(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	schema := arrow.NewSchema(
+		[]arrow.Field{
+			{Name: "a", Type: arrow.PrimitiveTypes.Int32, Nullable: true},
+			{Name: "b", Type: arrow.PrimitiveTypes.Int32, Nullable: true},
+		},
+		nil,
+	)
+
+	b := array.NewRecordBuilder(mem, schema)
+	defer b.Release()
+
+	// Field "a" decodes fine and is appended, then field "b" fails to decode
+	// (a string is not a valid int32). The partially-appended row must be rolled
+	// back so all columns keep the same length.
+	err := b.UnmarshalJSON([]byte(`{"a": 1, "b": "not-an-int"}`))
+	assert.Error(t, err)
+
+	// A subsequent valid row and NewRecordBatch must not panic due to unequal
+	// column lengths.
+	err = b.UnmarshalJSON([]byte(`{"a": 2, "b": 3}`))
+	assert.NoError(t, err)
+
+	rec := b.NewRecordBatch()
+	defer rec.Release()
+
+	if got, want := rec.NumRows(), int64(1); got != want {
+		t.Fatalf("invalid number of rows: got=%d, want=%d", got, want)
+	}
+	if got, want := rec.Column(0).String(), `[2]`; got != want {
+		t.Fatalf("invalid column values: got=%q, want=%q", got, want)
+	}
+	if got, want := rec.Column(1).String(), `[3]`; got != want {
+		t.Fatalf("invalid column values: got=%q, want=%q", got, want)
+	}
+}
+
 func TestRecordBuilderResize(t *testing.T) {
 	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
 	defer mem.AssertSize(t, 0)
--- a/arrow/array/record_test.go
+++ b/arrow/array/record_test.go
@@ -17,6 +17,7 @@
 package array_test
 
 import (
+	"bytes"
 	"fmt"
 	"reflect"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/internal/json"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -485,9 +487,9 @@ func TestRecordBuilder(t *testing.T) {
 	mapDt.SetItemNullable(false)
 	schema := arrow.NewSchema(
 		[]arrow.Field{
-			{Name: "f1-i32", Type: arrow.PrimitiveTypes.Int32},
-			{Name: "f2-f64", Type: arrow.PrimitiveTypes.Float64},
-			{Name: "map", Type: mapDt},
+			{Name: "f1-i32", Type: arrow.PrimitiveTypes.Int32, Nullable: true},
+			{Name: "f2-f64-notnull", Type: arrow.PrimitiveTypes.Float64, Nullable: false},
+			{Name: "map", Type: mapDt, Nullable: true},
 		},
 		nil,
 	)
@@ -498,11 +500,14 @@ func TestRecordBuilder(t *testing.T) {
 	b.Retain()
 	b.Release()
 
-	b.Field(0).(*array.Int32Builder).AppendValues([]int32{1, 2, 3}, nil)
+	b.Field(0).(*array.Int32Builder).AppendNull()
+	b.Field(0).(*array.Int32Builder).AppendValues([]int32{2, 3}, nil)
 	b.Field(0).(*array.Int32Builder).AppendValues([]int32{4, 5}, nil)
-	b.Field(1).(*array.Float64Builder).AppendValues([]float64{1, 2, 3, 4, 5}, nil)
+
+	b.Field(1).(*array.Float64Builder).AppendValues([]float64{1.1, 2.2, 3.3, 4.4, 5.5}, nil)
+
 	mb := b.Field(2).(*array.MapBuilder)
-	for i := 0; i < 5; i++ {
+	for i := range 5 {
 		mb.Append(true)
 
 		if i%3 == 0 {
@@ -511,6 +516,15 @@ func TestRecordBuilder(t *testing.T) {
 		}
 	}
 
+	err := b.UnmarshalJSON([]byte(`{"f1-i32": null, "f2-f64-notnull": null, "map": null}`))
+	assert.Contains(t, err.Error(), "field 'f2-f64-notnull' is non-nullable but got null")
+
+	err = b.UnmarshalJSON([]byte(`{"f1-i32": null, "map": null}`))
+	assert.Contains(t, err.Error(), "field 'f2-f64-notnull' is required but no value was given")
+
+	err = b.UnmarshalJSON([]byte(`{"f1-i32": 6, "f2-f64-notnull": 6.6, "map": [{"key": "4": "value": "d"}]}`))
+	assert.NoError(t, err)
+
 	rec := b.NewRecordBatch()
 	defer rec.Release()
 
@@ -518,7 +532,7 @@ func TestRecordBuilder(t *testing.T) {
 		t.Fatalf("invalid schema: got=%#v, want=%#v", got, want)
 	}
 
-	if got, want := rec.NumRows(), int64(5); got != want {
+	if got, want := rec.NumRows(), int64(6); got != want {
 		t.Fatalf("invalid number of rows: got=%d, want=%d", got, want)
 	}
 	if got, want := rec.NumCols(), int64(3); got != want {
@@ -527,9 +541,27 @@ func TestRecordBuilder(t *testing.T) {
 	if got, want := rec.ColumnName(0), schema.Field(0).Name; got != want {
 		t.Fatalf("invalid column name: got=%q, want=%q", got, want)
 	}
-	if got, want := rec.Column(2).String(), `[{["0" "2" "3"] ["a" "b" "c"]} {[] []} {[] []} {["3" "2" "3"] ["a" "b" "c"]} {[] []}]`; got != want {
-		t.Fatalf("invalid column name: got=%q, want=%q", got, want)
+
+	if got, want := rec.Column(0).String(), `[(null) 2 3 4 5 6]`; got != want {
+		t.Fatalf("invalid column values: got=%q, want=%q", got, want)
 	}
+	if got, want := rec.Column(1).String(), `[1.1 2.2 3.3 4.4 5.5 6.6]`; got != want {
+		t.Fatalf("invalid column values: got=%q, want=%q", got, want)
+	}
+	if got, want := rec.Column(2).String(), `[{["0" "2" "3"] ["a" "b" "c"]} {[] []} {[] []} {["3" "2" "3"] ["a" "b" "c"]} {[] []} {["4"] ["d"]}]`; got != want {
+		t.Fatalf("invalid column values: got=%q, want=%q", got, want)
+	}
+
+	// roundtripping from JSON with array.FromJSON should work
+	arr := array.RecordToStructArray(rec)
+	defer arr.Release()
+	jsonStr, err := json.Marshal(arr)
+	assert.NoError(t, err)
+
+	roundtripped, _, err := array.FromJSON(mem, arr.DataType(), bytes.NewReader(jsonStr))
+	defer roundtripped.Release()
+	assert.NoError(t, err)
+	assert.Truef(t, array.Equal(arr, roundtripped), "JSON round trip returns different array: got=%q, want=%d", arr, roundtripped)
 }
 
 func TestRecordBuilderResize(t *testing.T) {
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -497,19 +497,27 @@ func (b *StructBuilder) UnmarshalOne(dec *json.Decoder) error {
 			if keylist[key] {
 				return fmt.Errorf("key %s is specified twice", key)
 			}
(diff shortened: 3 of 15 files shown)


zeroshade:
Is it possible for us to do this without a breaking change? I'd rather avoid the breaking change of changing a public API method if at all possible.

zeroshade:
@serramatutu are you still working on this?

serramatutu:
@zeroshade I haven't had the time to push this forward since I last worked on it. Currently on PTO, will be back in July.

serramatutu:
@zeroshade I addressed all the issues in new commits and rebased onto main. I don't have permissions to request another Copilot review. I did run a Claude Code review locally and it says it's good.

There is one issue that I found where `b.Resize(-1)` can leak memory which is causing a test to break. This was a pre-existing bug. [I'm making a PR to fix that here.](https://github.com/apache/arrow-go/pull/995) Will need to rebase onto that to make that test pass :)

serramatutu:
@zeroshade I've rebased this onto the latest main and the tests are now passing :)

arrow/array/record.go
```diff
@@ -434,49 +436,74 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
...
+		val, ok := keylist[b.schema.Field(i).Name]
+		if !ok {
 			b.fields[i].AppendNull()
+			continue
+		}
+
+		valDec := json.NewDecoder(bytes.NewReader(val))
+		valDec.UseNumber()
+		if err := b.fields[i].UnmarshalOne(valDec); err != nil {
+			b.Resize(-1)
```
zeroshade: `Resize(-1)` does not reliably discard this row. Nested builders can append their outer entry before a child fails, leaving every top-level field at the same length. For `{"a":1,"b":[2,"bad"]}`, the error leaves both columns at length 1, so this call is a no-op and the rejected row contaminates later output.

Please restore every builder to an explicit pre-row checkpoint rather than inferring rollback from top-level length differences. Add a malformed nested-value test followed by a valid row.

arrow/array/struct.go
```diff
@@ -494,40 +525,59 @@ func (b *StructBuilder) UnmarshalOne(dec *json.Decoder) error {
...
 				continue
 			}
-			idx, _ := b.dtype.(*arrow.StructType).FieldIdx(field.Name)
-			if _, hasKey := keylist[field.Name]; !hasKey {
-				b.fields[idx].AppendNull()
+
+			valDec := json.NewDecoder(bytes.NewReader(next))
+			valDec.UseNumber()
+			if err := b.fields[i].UnmarshalOne(valDec); err != nil {
+				b.Resize(-1)
```
zeroshade: This rollback panics when all child builders are equally advanced. In that case `columnLenRange` leaves `n == -1`, which is passed to child `Resize` methods and eventually causes an out-of-range panic in `CountSetBits`.

A malformed list field such as `[2,"bad"]` reproduces this. Please restore the parent and all descendants to their exact pre-row state and test that the decode returns an error without retaining the row or panicking.

arrow/array/util.go
```diff
@@ -296,7 +296,11 @@ func RecordToJSON(rec arrow.RecordBatch, w io.Writer) error {
 	cols := make(map[string]interface{})
 	for i := 0; int64(i) < rec.NumRows(); i++ {
 		for j, c := range rec.Columns() {
-			cols[fields[j].Name] = c.GetOneForMarshal(i)
+			if rec.Schema().Field(j).Nullable && c.IsNull(i) {
```
zeroshade, line 299: This still serializes a null in a non-nullable column: when the condition is false, the column’s `GetOneForMarshal` returns `nil`. The resulting `{"x":null}` is then rejected by this PR’s reader, breaking writer/reader round trips.

Please define and implement the intended behavior for invalid non-nullable data—likely return an encoding error—and add a round-trip regression. The equivalent struct path has the same issue.

arrow/array/record.go
```diff
@@ -434,49 +436,74 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
...
+		keylist[key] = val
 	}
 
 	// consume the closing '}'
 	if _, err := dec.Token(); err != nil {
 		return err
 	}
 
+	// check that all non-nullable fields were specified
 	for i := 0; i < b.schema.NumFields(); i++ {
```
zeroshade: This validates only top-level schema fields. Nested field nullability is still ignored; for example, a `ListOfNonNullable(int32)` accepts `[1,null]` without error.

Please enforce nested field metadata consistently, or narrow the stated compatibility scope if recursive list-element validation is intentionally deferred. Add coverage for non-nullable list elements and corresponding nested types.

zeroshade (changes requested):
The stricter top-level nullability checks are directionally correct, but four correctness issues remain:

1. `RecordBuilder` rollback can retain rejected nested rows when top-level lengths remain equal.
2. `StructBuilder` rollback can pass `-1` into child builders and panic.
3. Record and struct writers still emit non-nullable nulls that the new readers reject, breaking round trips.
4. Nested field metadata remains unenforced; for example, non-nullable list elements still accept `null`.

See the inline comments on `record.go`, `struct.go`, and `util.go`.

The existing array tests pass, as do focused race tests and all 23 CI checks. The PR is currently conflicting with `main` and will also require a rebase.

---

>

arrow/array/record.go
serramatutu: Fixed. `Resize(-1)` is gone from `RecordBuilder` — #1113 landed while this PR was open and gives exactly the pre-row checkpoint you asked for, so `UnmarshalOne` now relies on upstream's `builderCheckpoint` capture/restore. Nullability is also validated before anything is appended now, so the `{"a":1,"b":[2,"bad"]}` case is covered by upstream's rollback tests plus `TestRecordBuilderRejectsNullInNestedNonNullableField` and `TestRecordBuilderKeepsDecodingAfterRejectedRow` (malformed nested value, then a valid row).

arrow/array/struct.go
serramatutu: Fixed, and thanks for the precise diagnosis. `StructBuilder` no longer calls `Resize(-1)`; it builds its own reusable `builderCheckpoint` (it needs one because it is the root builder for `array.FromJSON` on a struct type, which #1113 does not cover) and restores it on error. `TestStructBuilderRollsBackRowAfterNestedDecodeError` decodes `{"a":1,"b":[2,"bad"]}` followed by a valid row and asserts no panic, no retained row, and correct output.

arrow/array/util.go
serramatutu: Defined it as an encoding error, as you suggested. `RecordToJSON` and `Struct.MarshalJSON` (so `RecordBatch.MarshalJSON` too, via `RecordToStructArray`) now return `arrow.ErrInvalid` when a non-nullable field holds a null, recursively. You were right that the old branch was pointless — `GetOneForMarshal` returns `nil` either way — so both no-op checks are reverted. The check honors parent-validity priority, so a null child under a null struct parent is not an error, matching what the encoders actually emit. Covered by `TestRecordToJSONRejectsNullInNonNullableField`, `TestStructMarshalJSONAllowsNullChildUnderNullParent` and a round-trip test.

One scope note: `MarshalJSON` on non-struct array roots (a bare list array, say) still writes nulls for non-nullable element fields. Reading validates those roots; writing does not. Happy to extend it here if you want it in this PR.

arrow/array/record.go
serramatutu: Enforced recursively now rather than narrowing the scope. `arrow/array/nullability.go` walks a buffered value against the `arrow.Field` tree, covering struct fields, the element field of list / large list / list view / fixed-size list, map key and item fields, union children, dictionary values, run-end-encoded values and extension storage. `ListOfNonNullable(int32)` rejects `[1,null]`, and missing non-nullable fields are rejected at any depth. Roots that are not records or structs are validated in `FromJSON`, so `array.FromJSON(mem, arrow.ListOfNonNullable(...), ...)` is checked too. `arrow/array/nullability_test.go` has the coverage.

Unexpected bonus: buffering the row made `BenchmarkRecordFromJSON/Size_1000` ~305x faster (7.77 s/op -> 25 ms/op), because decoding string tokens straight off the document decoder is quadratic in goccy's `(*Stream).Token`. Details in the PR description.

arrow/array/struct.go
```diff
@@ -207,9 +207,14 @@ func (a *Struct) GetOneForMarshal(i int) interface{} {
...
+	dtype := a.data.dtype.(*arrow.StructType)
+	fieldList := dtype.Fields()
 	for j, d := range a.fields {
-		tmp[fieldList[j].Name] = d.GetOneForMarshal(i)
+		if dtype.Field(j).Nullable && a.IsNull(i) {
+			tmp[fieldList[j].Name] = nil
+		} else {
+			tmp[fieldList[j].Name] = d.GetOneForMarshal(i)
+		}
 	}
```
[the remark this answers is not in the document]
serramatutu: Reverted this branch entirely — it was indeed unreachable, and even reachable it would have been a no-op since `GetOneForMarshal` returns `nil` for a null slot anyway. Non-nullable nulls are now rejected by `Struct.MarshalJSON` instead of being re-encoded.

arrow/array/struct.go
```diff
@@ -467,19 +472,27 @@ func (b *StructBuilder) UnmarshalOne(dec *json.Decoder) error {
...
 
-			if err := b.fields[idx].UnmarshalOne(dec); err != nil {
+			if bytes.Equal(next, []byte("null")) && !dtype.Field(idx).Nullable {
+				return fmt.Errorf("field '%s' is non-nullable but got null", dtype.Field(idx).Name)
+			}
+
+			valDec := json.NewDecoder(bytes.NewReader(next))
+			valDec.UseNumber()
+			if err := b.fields[idx].UnmarshalOne(valDec); err != nil {
 				return err
```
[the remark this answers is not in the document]
serramatutu: Fixed. Validation now runs on the buffered value before anything is appended, and `StructBuilder` restores a pre-row `builderCheckpoint` if a child decode fails, so the builder is never left advanced after an error.

arrow/array/record_test.go
```diff
@@ -511,14 +516,23 @@ func TestRecordBuilder(t *testing.T) {
...
 	}
 
+	err := b.UnmarshalJSON([]byte(`{"f1-i32": null, "f2-f64-notnull": null, "map": null}`))
+	assert.Contains(t, err.Error(), "field 'f2-f64-notnull' is non-nullable but got null")
+
+	err = b.UnmarshalJSON([]byte(`{"f1-i32": null, "map": null}`))
+	assert.Contains(t, err.Error(), "field 'f2-f64-notnull' is required but no value was given")
+
+	err = b.UnmarshalJSON([]byte(`{"f1-i32": 6, "f2-f64-notnull": 6.6, "map": [{"key": "4": "value": "d"}]}`))
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
serramatutu: Fixed — the missing comma is corrected.

arrow/array/record_test.go
```diff
@@ -527,9 +541,27 @@ func TestRecordBuilder(t *testing.T) {
...
+	// roundtripping from JSON with array.FromJSON should work
+	arr := array.RecordToStructArray(rec)
+	defer arr.Release()
+	jsonStr, err := json.Marshal(arr)
+	assert.NoError(t, err)
+
+	roundtripped, _, err := array.FromJSON(mem, arr.DataType(), bytes.NewReader(jsonStr))
+	defer roundtripped.Release()
+	assert.NoError(t, err)
+	assert.Truef(t, array.Equal(arr, roundtripped), "JSON round trip returns different array: got=%q, want=%d", arr, roundtripped)
```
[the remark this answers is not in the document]
serramatutu: Fixed. `require.NoError` now runs before `defer roundtripped.Release()`, and the format verb is `%q` for both arguments.

arrow/array/struct_test.go
```diff
@@ -472,52 +474,48 @@ func TestStructArrayUnmarshalJSONMissingFields(t *testing.T) {
...
+					} else if e != nil && tc.panicErr != nil && fmt.Errorf("%s", e).Error() != tc.panicErr.Error() {
+						t.Fatalf("invalid error. got=%v, want=%v", e, tc.panicErr.Error())
+					}
+				}()
 
 				err := sb.UnmarshalJSON([]byte(tc.jsonInput))
 				if err != nil {
-					t.Fatal(err)
+					panic(err)
 				}
```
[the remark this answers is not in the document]
serramatutu: Fixed. The test no longer converts errors into panics: it asserts the returned error with `require.ErrorContains` (and `main` has since rewritten this test to drop the panic-recover harness altogether).

arrow/array/record.go
```diff
@@ -434,43 +433,69 @@ func (b *RecordBuilder) UnmarshalOne(dec *json.Decoder) error {
...
+	// at this point we know there are no integrity errors, append values to field builders
+	for key, val := range keylist {
+		valDec := json.NewDecoder(bytes.NewReader(val))
+		valDec.UseNumber()
+
+		indices := b.schema.FieldIndices(key)
+		if err := b.fields[indices[0]].UnmarshalOne(valDec); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
serramatutu: Fixed. `unmarshalOne` validates the whole buffered row up front, and any error from a field builder is rolled back through the per-row `builderCheckpoint` added in #1113, so column lengths can no longer diverge.

arrow/array/record.go
serramatutu: I realized a new `builderCheckpoint` waas just introduced. Will use that to rollback instead.

arrow/array/struct.go
serramatutu: Using the builder checkpoint for this now. I added `[2,"bad"]` as a test case.

arrow/array/util.go
serramatutu: Per this: https://github.com/serramatutu/arrow-internal-nulls

In C++, it looks like the general approach is "validate all data in", then assume it's OK/consistent when "writing out". I.e it is an invariant that the data is correct when writing out.

If the user has messed around with the data:
- by tweaking memory directly: they need to ensure their code respects invariants OR call ValidateFull
- by using a public API: the API will always leave the array in a consistent internal state

This is my rationale for calling `GetOneForMarshal()` here: we assume the data is correct, if it's not then it's UB.

If we're very pedantic about this, we should just revert this change and use `c.GetOneForMarshal(i)` as we should assume `IsNull(i)` will always return false if the data is consistent with the schema. This is what C++ does actually, so here we're doing even more validation by checking `rec.Schema().Field(j).Nullable`. If this is writing wrong data that is failing validation upstream, I think we should fix the thing that allows the bad data to be constructed in the first place, not validate here while writing.

Sources:
- `JsonWriter::WriteArray` uses an `ArrayWriter`: https://github.com/apache/arrow/blob/b12755a1cc41ee5e9b1874ece3bcc92d6d3e3e78/cpp/src/arrow/integration/json_internal.cc#L2079
- `ArrayWriter::WriteDataValues` does not check for the schema at all, it just calls `IsValid()` directly: https://github.com/apache/arrow/blob/b12755a1cc41ee5e9b1874ece3bcc92d6d3e3e78/cpp/src/arrow/integration/json_internal.cc#L505

arrow/array/record.go
```diff
@@ -599,49 +601,73 @@ func (b *RecordBuilder) unmarshalOne(dec *json.Decoder) (err error) {
...
+	for i := 0; i < b.schema.NumFields(); i++ {
+		val, ok := keylist[b.schema.Field(i).Name]
+		if !ok {
 			b.fields[i].AppendNull()
+			continue
+		}
+
+		valDec := json.NewDecoder(bytes.NewReader(val))
+		valDec.UseNumber()
+		if err := b.fields[i].UnmarshalOne(valDec); err != nil {
```
zeroshade: **Blocking:** This still delegates the nested value directly to the child builder, which has no access to the containing nested `arrow.Field.Nullable` metadata. I reproduced `RecordFromJSON` accepting `{"x":[1,null]}` for a schema whose field type is `ListOfNonNullable(int32)`. Please enforce nested nullability recursively (including corresponding nested container types) and add a regression that expects this input to fail.

zeroshade (changes requested):
The checkpoint-based rollback changes resolve the prior record/struct contamination and panic findings. Two previously reported correctness gaps remain: nested nullability metadata is still not enforced, and record/struct serialization still emits nulls for non-nullable fields that the updated readers reject. Targeted, race, repeated rollback, impacted-package, formatting, vet, and diff checks pass. The PR is also still conflicting with main and currently has no reported CI checks.

arrow/array/record.go
serramatutu: Sorry, I was still fixing this yesterday but it got late... 

I just fixed it for list, map, union and REE, and added tests for all of those. I also made all of them have individual checkpointing so if users use e.g a standalone `ListBuilder` with a non-nullable field, it'll still keep a consistent state.

Note that I had to add a new `NewListBuilderWithField` (also for list view and large lists) so that the builder can know the schema it's trying to build.

arrow/array/record.go
```diff
@@ -599,49 +610,73 @@ func (b *RecordBuilder) unmarshalOne(dec *json.Decoder) (err error) {
 		return fmt.Errorf("record should start with '{', not %s", t)
 	}
 
-	keylist := make(map[string]bool)
+	// consume one row checking for duplicates and nulls
+	keylist := make(map[string]json.RawMessage)
```
zeroshade: This removes the streaming path for every schema, not only schemas containing non-nullable fields. A 1,000-row fully nullable benchmark regressed from ~320 us / 345 KB / 12,091 allocs to ~1.03 ms / 3.29 MB / 49,126 allocs. Please retain the original streaming path when recursive nullability validation is unnecessary, as described in the PR, or otherwise avoid buffering and re-decoding every field.

zeroshade (changes requested):
Thanks for the follow-up. The new commits fix the nested-nullability and rollback issue; the focused reproducer and array tests now pass.

1 blocker remains:

The current implementation buffers and re-decodes every record field, including fully nullable schemas. A direct benchmark against the PR base showed approximately 3.2x slower decoding, 9.5x more allocated bytes, and 4x more allocations. The nullable fast path described in the PR is not present in the pushed code.

The branch also currently conflicts with `main`, and no CI runs exist for the current head.

arrow/array/union_test.go
```diff
@@ -204,8 +204,8 @@ func TestUnionBuilderUnmarshalOnePreservesDecoderConfiguration(t *testing.T) {
 				name string
 				json string
 			}{
-				{name: "integer first", json: `{"i":1.5,"u":[0,1]}`},
-				{name: "union first", json: `{"u":[0,1],"i":1.5}`},
+				{name: "integer first", json: `{"i":2,"u":[0,1]}`},
```
serramatutu, line 207: Worth noting: there was a recent change that introduced this test, which was silently decoding `1.5` into an `Int32` array and casting it to `1`. I assume this is a correctness behavior, and since this PR is about validating schemas, the JSON decoder now raises an error for that case. I changed the expected value to `2` so it's the appropriate type.

arrow/array/builder.go
```diff
@@ -388,6 +390,170 @@ func (b *builder) UnsafeAppendBoolToBitmap(isValid bool) {
...
+	return child.UnmarshalOne(valDec)
+}
+
+// rowDecoder decodes a value out of a reused copy of that value, so that
+// unescaping a string costs O(value) instead of O(remaining document).
+//
+// NOTE: goccy/go-json unescapes in place and shifts every byte after the escape,
+// which is a big performance cost for large JSON documents:
+// https://github.com/goccy/go-json/blob/v0.10.6/internal/decoder/string.go#L190
+type rowDecoder struct {
```
serramatutu, line 438: I had to add this due to bad performance from goccy when escaping newlines...

arrow/array/builder.go
```diff
@@ -388,6 +390,170 @@ func (b *builder) UnsafeAppendBoolToBitmap(isValid bool) {
...
+
+// fieldContainer is what nestedJSONDecoder needs out of *arrow.Schema and
+// *arrow.StructType.
+type fieldContainer interface {
+	NumFields() int
+	Field(i int) arrow.Field
+}
+
+// nestedJSONDecoder decodes JSON objects into one builder per field.
+type nestedJSONDecoder struct {
```
serramatutu, line 467: I added this as a way to reuse code between `struct.go` and `builder.go` since they were more or less doing the same logic when unmarshaling JSON objects into a list of field builders.

zeroshade (changes requested):
Three issues remain on the current head:

- `arrow/array/builder.go:451` unconditionally enables `UseNumber()` while replaying buffered rows, overriding the caller's decoder semantics.
- Fully nullable records still take the buffering/re-decoding path, with roughly 1.5× slower decoding and 2× allocated bytes than the base implementation in a 1,000-row probe.
- `arrow/array/builder.go:515-521` skips unknown fields before recording seen keys, so duplicate unknown keys are silently accepted.

Please preserve caller decoder behavior, retain an efficient nullable-schema path, and track every encountered key for duplicate detection.

---

>

serramatutu:
@zeroshade regarding the previous comments:

## What I fixed

> Fully nullable records still take the buffering/re-decoding path, with roughly 1.5× slower decoding and 2× allocated bytes than the base implementation in a 1,000-row probe.

I have fixed the perf issue, now the row buffering only happens at the top-level struct, and we avoid re-decoding nested rows. Had to add some plumbing to `Builder` for that. Look at commit `Don't buffer rows on child struct builders`

## What I didn't fix (on purpose)

> arrow/array/builder.go:451 unconditionally enables UseNumber() while replaying buffered rows, overriding the caller's decoder semantics.

I kept it like this as I assume this is intender behavior for 2 reasons:
- `WithUseNumber()` is deprecated and the docs state it's a no-op and always enabled by default. So I'm always enabling it.
- Not enabling it makes the JSON decoder lenient as it can decode floats like `1.5` into integers by truncating, silently violating the schema. There was a test that asserted that lenient behavior (`TestUnionBuilderUnmarshalOnePreservesDecoderConfiguration`), which I have fixed (see my other comment above on that test)

> arrow/array/builder.go:515-521 skips unknown fields before recording seen keys, so duplicate unknown keys are silently accepted.

Also intended behavior. I added a comment that the decoder behaves like the default C++ decoder, with [`ParseOptions::Ignore`](https://github.com/apache/arrow/blob/05c1a4bf545a560f2b118a2e6e6d3d6fa36a2716/cpp/src/arrow/json/options.h#L35), which ignores unknown keys. I added it as a TODO comment to possibly support `ParseOptions` with `Error` and `InferType` as well for strict/lenient parsing.

zeroshade:
> I have fixed the perf issue, now the row buffering only happens at the top-level struct, and we avoid re-decoding nested rows. Had to add some plumbing to Builder for that. Look at commit Don't buffer rows on child struct builders

Can we add a benchmark for this (and post the main vs this PR result of the benchmark just to confirm)?

> What I didn't fix (on purpose)

That's fair, this logic is sound and makes sense to me. I'm good with it. 

> Also intended behavior. I added a comment that the decoder behaves like the default C++ decoder, with ParseOptions::Ignore, which ignores unknown keys.

Gotcha, this is fine then. Following the default behavior of the c++ decoder is more desirable anyways. Thanks!

So you just gotta fix the linting issue and then we're all good here :) 

Thanks for putting up with the reviews!

serramatutu:
# Benchmark

Results for `BenchmarkJSONReaderRowShape` and `BenchmarkJSONReaderEscapedString` in `arrow/array/json_reader_test.go`, comparing this branch against `main` (`227f6e3e`).

## Method

I ran it with Go 1.27.0, on an Apple M3 Pro. Every figure below is the median of six
runs of 20 iterations for `RowShape` and 100 iterations for `EscapedString`. 

```
go test ./arrow/array/ -run XXX -bench BenchmarkJSONReaderRowShape -benchtime 20x -count 6
go test ./arrow/array/ -run XXX -bench BenchmarkJSONReaderEscapedString -benchtime 100x -count 6

```

## The test cases

### `BenchmarkJSONReaderRowShape`

| shape | rows | fields per row | document size |
|---|---|---|---|
| small | 10,000 | 3 | ≈404 KB |
| medium | 1,000 | 10 | ≈119 KB |
| large | 1,000 | 100 | ≈1.3 MB |

All three shapes have 10,000 rows, so fields per row is the only variable. Every shape runs twice, once with all fields non-nullable and once with all
fields nullable.


### `BenchmarkJSONReaderEscapedString`

10,000 rows of one string field each. The field contains one `\n` escaped character. Runs twice, once with nullable and the other non-nullable.

## Results

### `BenchmarkJSONReaderRowShape`

TL;DR: the new encoder is slower on small field counts, but it's faster for lots of fields. Nullability has no impact on performance.

Non-nullable fields:

| shape | | ns/op | B/op | allocs/op |
|---|---|---|---|---|
| small | base | 4,707,431 | 3,210,491 | 100,111 |
| | branch | 7,080,244 | 10,011,996 | 170,054 |
| | | 1.50× slower | 3.12× | 1.70× |
| medium | base | 18,533,913 | 14,962,940 | 340,306 |
| | branch | 19,983,839 | 16,638,869 | 379,430 |
| | | 1.08× slower | 1.11× | 1.11× |
| large | base | 205,269,110 | 160,635,683 | 3,092,727 |
| | branch | 185,001,977 | 141,320,030 | 3,097,500 |
| | | 1.11× faster | 0.88× | 1.00× |

Nullable fields:

| shape | | ns/op | B/op | allocs/op |
|---|---|---|---|---|
| small | base | 4,697,328 | 3,210,491 | 100,111 |
| | branch | 6,720,561 | 10,011,997 | 170,054 |
| | | 1.43× slower | 3.12× | 1.70× |
| medium | base | 18,532,919 | 14,962,933 | 340,306 |
| | branch | 18,884,480 | 16,638,866 | 379,430 |
| | | 1.02× slower | 1.11× | 1.11× |
| large | base | 206,265,585 | 160,635,687 | 3,092,727 |
| | branch | 176,193,939 | 141,320,040 | 3,097,500 |
| | | 1.17× faster | 0.88× | 1.00× |

### `BenchmarkJSONReaderEscapedString`

TL;DR: the row buffering approach makes it faster because we're avoiding the big shift-left of the entire JSON document that goccy does when it finds an escaped character.

| | ns/op | MB/s | B/op | allocs/op |
|---|---|---|---|---|
| main, non-nullable | 9,751,840 | 19.37 | 1,532,712 | 50,064 |
| branch, non-nullable | 3,830,202 | 49.34 | 8,087,729 | 110,000 |
| | 2.55× faster | | 5.28× | 2.20× |
| main, nullable | 9,752,524 | 19.38 | 1,532,712 | 50,064 |
| branch, nullable | 3,758,762 | 50.28 | 8,087,729 | 110,000 |
| | 2.59× faster | | 5.28× | 2.20× |

PR #838, opened 2026-06-03
feat(parquet/pqarrow): support writing LARGE_LIST types

lidavidm:
### Rationale for this change

We can't write large list to a Parquet file.

Implement support for large list in pqarrow.

### Are these changes tested?

Yes

No

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/parquet/pqarrow/column_readers.go
+++ b/parquet/pqarrow/column_readers.go
@@ -452,12 +452,26 @@ func (lr *listReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
 		validityIO.ValidBits = validityBuffer.Bytes()
 	}
 	offsetsBuffer := memory.NewResizableBuffer(lr.rctx.mem)
-	offsetsBuffer.Resize(arrow.Int32Traits.BytesRequired(int(lenBound) + 1))
+	if lr.field.Type.ID() == arrow.LARGE_LIST {
+		offsetsBuffer.Resize(arrow.Int64Traits.BytesRequired(int(lenBound) + 1))
+	} else {
+		offsetsBuffer.Resize(arrow.Int32Traits.BytesRequired(int(lenBound) + 1))
+	}
 	defer offsetsBuffer.Release()
 
-	offsetData := arrow.Int32Traits.CastFromBytes(offsetsBuffer.Bytes())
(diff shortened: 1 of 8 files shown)


parquet/pqarrow/column_readers.go
```diff
@@ -488,6 +488,18 @@ func (lr *listReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
 		buffers[0] = validityBuffer
 	}
 
+	if lr.field.Type.ID() == arrow.LARGE_LIST {
```
zeroshade: Could we modify DefRepLevelsToListInfo to be generic and take []int64 directly to avoid having to first allocate and create the []int32 and then allocate again and copy everything over?

lidavidm: Done, though there's a slightly awkward construction to get the max value of the offset type; I see there's a helper inside the kernels code, but not sure if there's a good place to pull that out so it's just inlined for now.

parquet/pqarrow/encode_arrow_test.go
```diff
@@ -2027,7 +2027,9 @@ func (ps *ParquetIOTestSuite) TestLargeListRoundTrip() {
 	cnk := arrow.NewChunked(field.Type, []arrow.Array{arr})
 	defer arr.Release()
 
-	tbl := array.NewTable(arrow.NewSchema([]arrow.Field{field}, nil), []arrow.Column{*arrow.NewColumn(field, cnk)}, -1)
+	col := arrow.NewColumn(field, cnk)
+	defer col.Release()
+	tbl := array.NewTable(arrow.NewSchema([]arrow.Field{field}, nil), []arrow.Column{*col}, -1)
 	defer cnk.Release()
 	defer tbl.Release()
```
zeroshade: This can be simplified with a helper that we already have:

https://pkg.go.dev/github.com/apache/arrow-go/v18@v18.6.0/arrow/array#NewTableFromSlice

```go
tbl := array.NewTableFromSlice(arrow.NewSchema([]arrow.Field{field}, nil), [][]arrow.Array{{arr}})
defer tbl.Release()
```

That helper was added specifically to avoid having to separately do the intermediate steps of Array -> Chunk -> Column in consumer code, simplifying the work and avoiding the flood of defers here.

lidavidm: Updated

zeroshade (approved):

PR #841, opened 2026-06-07
fix(parquet/variant): correct binary search bounds in ObjectValue.ValueByKey

### LINKED ISSUE #842: [Go][Parquet] ObjectValue.ValueByKey binary search skips elements at mid - 1

The `ObjectValue.ValueByKey` function in `parquet/variant/variant.go` implements a custom binary search on the `v.value` array of field IDs. It initializes the search bounds using `i, j := 0, n` (an exclusive upper bound).
However, when the target key is smaller than the key at `mid`, it updates `j` to `mid - 1`:

```go
case 1:
    j = mid - 1
```

Because `j` is an exclusive upper bound (the search space is `[i, j)`), setting `j = mid - 1` excludes the element at index `mid - 1` from the search space entirely. If the target key is located at `mid - 1`, the lookup will fail and return `arrow.ErrNotFound`.

This issue occurs for objects containing >= 32 elements, where the binary search is active.

To verify with an example, you can paste the following into an online Go playground:
```go
package main

import (
	"fmt"
	"strings"
)

// BuggySearch replicates the binary search logic in ObjectValue.ValueByKey
func BuggySearch(keys []string, target string) int {
	n := len(keys)
	i, j := 0, n // Exclusive upper bound: search space is [i, j)
	for i < j {
		mid := (i + j) >> 1
		k := keys[mid]

		switch strings.Compare(k, target) {
		case -1:
			i = mid + 1
		case 0:
			return mid
		case 1:
			j = mid - 1 // Buggy: excludes mid - 1 from the semi-open interval [i, j)
		}
	}
	return -1 // Not found
}

func main() {
	// Create a sorted list of 32 keys (the threshold where binary search activates)
	keys := make([]string, 32)
	for i := 0; i < 32; i++ {
		keys[i] = fmt.Sprintf("key%02d", i)
	}

	// Try to search for "key15" (which sits at index 15)
	index := BuggySearch(keys, "key15")
	
	fmt.Printf("Searching for 'key15' (index 15):\n")
	fmt.Printf("  -> Result: index = %d\n", index)
	if index == -1 {
		fmt.Printf("  -> Status: ❌ BUG! 'key15' was skipped and not found.\n")
	} else {
		fmt.Printf("  -> Status: Correct\n")
	}
}
```

On Go Playground, the above code prints:

```text
Searching for 'key15' (index 15):
  -> Result: index = -1
  -> Status: ❌ BUG! 'key15' was skipped and not found.
```
Why it fails:
1. In the first division, `mid` is calculated as `(0 + 32) >> 1 = 16`.
2. `keys[16] ("key16")` is greater than `"key15"`, so the code executes `case 1`.
3. It sets `j = mid - 1 = 15`.
4. Because `j` is the exclusive upper bound of the semi-open interval `[i, j)`, index `15` is excluded from all subsequent checks, resulting in a silent search failure.

### Component(s)

Parquet

qzyu999:
### Rationale for this change

Closes #842

The `ObjectValue.ValueByKey` function implements a custom binary search on the `v.value` array of field IDs to look up fields by key. It initializes its boundaries using an exclusive upper bound (`i, j := 0, n`).
However, when the target key is smaller than the key at the `mid` index, it updates the upper bound with:
`j = mid - 1`

Since `j` is exclusive, setting it to `mid - 1` excludes the index `mid - 1` from the search space entirely. If the target key is located at `mid - 1`, the lookup will fail and return `arrow.ErrNotFound`. This results in missing fields or lookups failing on objects containing >= 32 fields.

- Correct the exclusive upper bound update in `ObjectValue.ValueByKey` from `j = mid - 1` to `j = mid`.

### Are these changes tested?

- Added `TestObjectBinarySearch` in `parquet/variant/variant_test.go` which constructs a 32-field object and verifies that all 32 lookups using `ValueByKey` succeed.

- No API changes. This is a correctness bug fix.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -405,7 +405,7 @@ func (v ObjectValue) ValueByKey(key string) (ObjectField, error) {
 				Key:   key,
 				Value: Value{value: v.value[v.dataStart+offset:], meta: v.meta}}, nil
 		case 1:
-			j = mid - 1
+			j = mid
 		}
 	}
 
--- a/parquet/variant/variant_test.go
+++ b/parquet/variant/variant_test.go
@@ -18,6 +18,7 @@ package variant_test
 
 import (
 	"encoding/json"
+	"fmt"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


qzyu999:
Hi @zeroshade, removed an empty line that was breaking the linter in 9a16c3b.

zeroshade (approved):

PR #852, opened 2026-06-11
fix(parquet): add WriteBatchSpacedWithError to surface spaced-write failures

### LINKED ISSUE #826: parquet/file: WriteBatchSpaced panics escape the API and silently discards commit-write errors

**Discovered while reviewing #824 (which fixes the related panic in `file.NewParquetWriter`).**

Across all 8 generated `*ColumnChunkWriter` types in [`parquet/file/column_writer_types.gen.go`](https://github.com/apache/arrow-go/blob/main/parquet/file/column_writer_types.gen.go), `WriteBatch` and `WriteBatchSpaced` are an asymmetric pair:

- `WriteBatch` returns `(valueOffset int64, err error)` and opens with a `defer recover()` that converts any panic into a returned `error` via `utils.FormatRecoveredError`.
- `WriteBatchSpaced` returns nothing (`void` signature), has no `defer recover()`, and silently discards the `error` returned by `commitWriteAndCheckPageLimit`.

This is the same class of "panic escapes the public API" bug as #820 (fixed in #824) but on a different code path.

#### Code

The asymmetry is most visible in `Int32ColumnChunkWriter` ([`column_writer_types.gen.go:66-143`](https://github.com/apache/arrow-go/blob/main/parquet/file/column_writer_types.gen.go#L66-L143)):

`WriteBatch` (line 66) — recovers panics, surfaces as `err`:

```go
func (w *Int32ColumnChunkWriter) WriteBatch(values []int32, defLevels, repLevels []int16) (valueOffset int64, err error) {
    defer func() {
        if r := recover(); r != nil {
            err = utils.FormatRecoveredError("unknown error type", r)
        }
    }()
    ...
    w.doBatches(n, repLevels, func(offset, batch int64) {
        ...
        if err := w.commitWriteAndCheckPageLimit(batch, toWrite); err != nil {
            panic(err) // caught above
        }
        ...
    })
    return
}
```

`WriteBatchSpaced` (line 118) — no recover, no error return, silently swallows the `commitWriteAndCheckPageLimit` error:

```go
func (w *Int32ColumnChunkWriter) WriteBatchSpaced(values []int32, defLevels, repLevels []int16, validBits []byte, validBitsOffset int64) {
    ...
    doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
        ...
        w.commitWriteAndCheckPageLimit(batch, info.numSpaced()) // <-- error ignored
        ...
    })
}
```

`commitWriteAndCheckPageLimit` is defined in [`parquet/file/column_writer.go:264`](https://github.com/apache/arrow-go/blob/main/parquet/file/column_writer.go#L264) as `func (w *columnWriter) commitWriteAndCheckPageLimit(numLevels, numValues int64) error` — the return value is meaningful.

The pattern is identical for the other 7 types — `Int64`, `Int96`, `Float32`, `Float64`, `Boolean`, `ByteArray`, `FixedLenByteArray`. The third method on each writer, `WriteDictIndices`, mirrors `WriteBatch`: it both returns `error` and uses `defer recover()`. So `WriteBatchSpaced` is the lone outlier.

#### Why this matters in practice

`WriteBatchSpaced` is reachable through public API surface — for example, `(*pqarrow.FileWriter).WriteBuffered` and the `pqarrow` array-encoder paths drive it for nullable columns. Any of the panic sources reachable from the inner closure (`maybeCalculateValidityBits`, `writeLevelsSpaced`, `writeValuesSpaced`, `commitWriteAndCheckPageLimit` returning a non-nil error in a context where the caller chose to panic, dictionary-fallback paths that `panic(err)` on bookkeeping failures) escapes the goroutine instead of surfacing as an `error`.

Concrete failure modes already in the package:

- `commitWriteAndCheckPageLimit` returns an error → ignored entirely (the simpler bug)
- Any helper called from the closure that panics on its own internal invariant (e.g. dictionary fallback's `panic(err)` calls in `FallbackToPlain`, or the `panic(err)` paths inside `WriteDictionaryPage`/`drainBufferedDataPages`) → propagates out of `WriteBatchSpaced`, past any caller that expected `WriteBatch`-style error semantics, and crashes the goroutine.

For consumers writing nullable columns through `pqarrow` to a network-attached sink (the same scenario described in #820 — GCS, S3 multipart, HDFS, etc.), this means a transient downstream failure on a spaced-write path crashes the worker process rather than returning an error.

#### Source of truth

Both methods are generated from [`parquet/file/column_writer_types.gen.go.tmpl`](https://github.com/apache/arrow-go/blob/main/parquet/file/column_writer_types.gen.go.tmpl):

- `WriteBatch` template body: lines 71–172 (returns `(valueOffset int64, err error)`, has `defer recover()`).
- `WriteBatchSpaced` template body: lines 189 onward (no error return, no `defer recover()`).

A fix lands in one place and regenerates 8 method bodies plus the equivalent dictionary-write paths.

#### Expected behavior

`WriteBatchSpaced` should match `WriteBatch`:

1. Signature `(valueOffset int64, err error)` (or at minimum `error` — matching `WriteBatch` is the cleanest API).
2. `defer func() { if r := recover(); r != nil { err = utils.FormatRecoveredError("unknown error type", r) } }()` at the top.
3. Check the `error` returned by `commitWriteAndCheckPageLimit` and `panic(err)` to let the deferred recover handle it (matching the `WriteBatch` pattern in the same file).

Adding a return value to a public method is a breaking API change, so a minimal-breakage path is also viable:

- Introduce a sibling `WriteBatchSpacedWithError(...) (int64, error)`
- Keep `WriteBatchSpaced(...)` as a thin wrapper that calls the new method and panics on error with a documented panic-string format

— exactly the pattern #824 chose for `NewParquetWriter` / `NewParquetWriterWithError`. That preserves source compatibility while giving callers a clean error path.

The breaking-API option is also defensible if maintainers prefer fixing the surface properly given that `WriteBatchSpaced` returning nothing is itself a Go API smell (it cannot signal failure at all today).

### Component(s)

Parquet

zeroshade:
### Rationale for this change

Closes #826.

Across all 8 generated `*ColumnChunkWriter` types in `parquet/file/column_writer_types.gen.go`, `WriteBatch` and `WriteBatchSpaced` are an asymmetric pair:

- `WriteBatch` returns `(valueOffset int64, err error)` and opens with a `defer recover()` that converts any panic into a returned `error` via `utils.FormatRecoveredError`.
- `WriteBatchSpaced` returns nothing, has no `defer recover()`, and on the `doBatches` (non-byte-array) path silently discards the `error` returned by `commitWriteAndCheckPageLimit`.

So a write failure during a spaced write is either **silently discarded** (numeric/boolean/fixed paths) or **escapes as a panic out of the calling goroutine** (the byte-array branch's `panic(err)`, plus other internal panics such as dictionary-fallback bookkeeping) — because, unlike `WriteBatch`, there is no `recover()` to turn it into an `error`.

This is the same class of "panic escapes / error lost on the public API" bug as #820 (fixed in #824), on a different code path. It is reachable through `pqarrow` — `(*pqarrow.FileWriter)` drives `WriteBatchSpaced` for **nullable** columns — so for consumers writing nullable columns to a network-attached sink (GCS, S3 multipart, HDFS, etc.), a transient downstream failure on the spaced path crashes the worker process or silently drops the error instead of returning it.

The Boolean-only `WriteBitmapBatchSpaced` has the same void-signature / no-recover shape and is fixed here too.

1. **New method `WriteBatchSpacedWithError(values, defLevels, repLevels, validBits, validBitsOffset) (int64, error)`** generated for all 8 column-writer types from `column_writer_types.gen.go.tmpl`. It mirrors `WriteBatch`:
   - opens with `defer func() { if r := recover(); r != nil { err = utils.FormatRecoveredError(...) } }()`, and
   - checks the `error` from `commitWriteAndCheckPageLimit` on **every** branch (the `doBatches` branch previously ignored it), `panic`-ing so the deferred recover surfaces it as a returned `error`.

2. **New method `WriteBitmapBatchSpacedWithError(...) (int64, error)`** on `*BooleanColumnChunkWriter`, applying the same treatment to the bitmap spaced-write path.

3. **`WriteBatchSpaced` and `WriteBitmapBatchSpaced` are left byte-for-byte unchanged** — same signatures, same behavior. This keeps the change **non-breaking and purely additive**, exactly the approach #824 took with `NewParquetWriter` / `NewParquetWriterWithError`. Existing callers are unaffected; callers that want error handling opt into the `…WithError` variants.

4. **`pqarrow`'s dense-array encoder (`writeDenseArrow`) now calls `WriteBatchSpacedWithError`** at all 12 spaced-write sites and propagates the error through its existing `(err error)` return, so nullable-column writes report sink failures as errors instead of crashing the goroutine or losing the error. (`WriteBitmapBatchSpaced` has no internal callers, so nothing to rewire there.)

The diff to the generated file and the template is additions-only (no existing generated method bodies change).

### Are these changes tested?

Yes. Two new tests in `parquet/file/column_writer_test.go` use the existing `mockpagewriter` to fail `WriteDataPage`, with a tiny `DataPageSize` and dictionary disabled to force a flush mid-call:

- `TestWriteBatchSpacedWithErrorPropagatesWriteFailure`
- `TestWriteBitmapBatchSpacedWithErrorPropagatesWriteFailure`

Each asserts that the `…WithError` variant returns the wrapped failure (`errors.Is(err, failureErr)`) rather than panicking or discarding it, and that the legacy method does **not** panic out of the call (its original behavior is preserved).

The full `parquet/file/...` and `parquet/pqarrow/...` suites pass (`PARQUET_TEST_DATA` pointed at `parquet-testing/data`), and `go vet` is clean.

**No breaking changes.**

- `WriteBatchSpaced(...)` and `WriteBitmapBatchSpaced(...)` — signatures and behavior unchanged.
- `WriteBatchSpacedWithError(...) (int64, error)` (all 8 types) and `WriteBitmapBatchSpacedWithError(...) (int64, error)` (Boolean) — new exported methods. Adding exported methods is non-breaking in Go.
- `pqarrow` public signatures are unchanged; the dense-array nullable write path now returns transient sink failures as an `error` (callers already checking `err` get strictly better behavior).

### DIFF
--- a/parquet/file/column_writer_types.gen.go.tmpl
+++ b/parquet/file/column_writer_types.gen.go.tmpl
@@ -269,6 +269,97 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLev
 {{- end}}
 }
 
+// WriteBatchSpacedWithError behaves like WriteBatchSpaced but reports a write
+// failure as an error instead of panicking or silently discarding it,
+// mirroring the error handling of WriteBatch. Prefer this method on write
+// paths whose sink may fail (for example a network-attached writer).
+func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}}, defLevels, repLevels []int16, validBits []byte, validBitsOffset int64) (valueOffset int64, err error) {
+  defer func() {
+    if r := recover(); r != nil {
+      err = utils.FormatRecoveredError("unknown error type", r)
+    }
+  }()
+  length := len(defLevels)
+  if defLevels == nil {
+    length = len(values)
+  }
+{{- if .isByteArray}}
+  const maxSafeBatchDataSize int64 = 1 << 30 // 1GB
+
+  batchSize := w.props.WriteBatchSize()
+  levelOffset := int64(0)
+  n := int64(length)
+
+  for levelOffset < n {
+    remaining := n - levelOffset
+    batch := min(remaining, batchSize)
+
+    if values != nil {
+      var cumDataSize int64
+      for vi := int64(0); vi < batch && valueOffset+vi < int64(len(values)); vi++ {
+{{- if eq .Name "ByteArray"}}
+        valSize := int64(len(values[valueOffset+vi])) + 4
+{{- else}}
+        valSize := int64(w.descr.TypeLength()) + 4
+{{- end}}
+        if cumDataSize+valSize > maxSafeBatchDataSize && vi > 0 {
+          batch = vi
+          break
+        }
+        cumDataSize += valSize
+      }
+    }
+    if batch < 1 {
+      batch = 1
+    }
+
+    info := w.maybeCalculateValidityBits(levelSliceOrNil(defLevels, levelOffset, batch), batch)
+
+    w.writeLevelsSpaced(batch, levelSliceOrNil(defLevels, levelOffset, batch), levelSliceOrNil(repLevels, levelOffset, batch))
+    var vals []{{.name}}
+    if values != nil {
+      vals = values[valueOffset : valueOffset+info.numSpaced()]
+    }
+
+    if w.bitsBuffer != nil {
+      w.writeValuesSpaced(vals, info.batchNum, batch, w.bitsBuffer.Bytes(), 0)
+    } else {
+      w.writeValuesSpaced(vals, info.batchNum, batch, validBits, validBitsOffset+valueOffset)
+    }
+    if err := w.commitWriteAndCheckPageLimit(batch, info.numSpaced()); err != nil {
+      panic(err)
+    }
+    valueOffset += info.numSpaced()
+
+    w.checkDictionarySizeLimit()
+    levelOffset += batch
+  }
+{{- else}}
+  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+    var vals []{{.name}}
+    info := w.maybeCalculateValidityBits(levelSliceOrNil(defLevels, offset, batch), batch)
+
+    w.writeLevelsSpaced(batch, levelSliceOrNil(defLevels, offset, batch), levelSliceOrNil(repLevels, offset, batch))
+    if values != nil {
+      vals = values[valueOffset:valueOffset+info.numSpaced()]
+    }
+
+    if w.bitsBuffer != nil {
+      w.writeValuesSpaced(vals, info.batchNum, batch, w.bitsBuffer.Bytes(), 0)
+    } else {
+      w.writeValuesSpaced(vals, info.batchNum, batch, validBits, validBitsOffset+valueOffset)
+    }
+    if err := w.commitWriteAndCheckPageLimit(batch, info.numSpaced()); err != nil {
+      panic(err)
+    }
+    valueOffset += info.numSpaced()
+
+    w.checkDictionarySizeLimit()
+  })
+{{- end}}
+  return
+}
+
 {{if eq .Name "Boolean"}}
 // WriteBitmapBatch writes boolean values directly from a bitmap without converting to []bool.
 // This avoids the 8x memory overhead and provides significant performance improvements.
--- a/parquet/file/column_writer_types.gen.go.tmpl
+++ b/parquet/file/column_writer_types.gen.go.tmpl
@@ -192,7 +192,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatch(values []{{.name}}, defLevels, r
 // the failure as an error.
 func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLevels, repLevels []int16, validBits []byte, validBitsOffset int64) {
   if _, err := w.WriteBatchSpacedWithError(values, defLevels, repLevels, validBits, validBitsOffset); err != nil {
-    panic(err)
+    panicCause(err)
   }
 }
 
@@ -328,7 +328,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBitmapBatch(bitmap []byte, bitmapOffse
 // returns the failure as an error.
 func (w *{{.Name}}ColumnChunkWriter) WriteBitmapBatchSpaced(bitmap []byte, bitmapOffset int64, numValues int, defLevels, repLevels []int16, validBits []byte, validBitsOffset int64) {
   if _, err := w.WriteBitmapBatchSpacedWithError(bitmap, bitmapOffset, numValues, defLevels, repLevels, validBits, validBitsOffset); err != nil {
-    panic(err)
+    panicCause(err)
   }
 }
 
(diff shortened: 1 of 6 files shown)


lidavidm (commented):
WriteBatchSpaced is internal, right? Why keep it at this point if it's not a public API?

zeroshade:
Both are technically part of the public API and I'm trying to avoid making breaking changes, it isn't inside of an `internal` package and it starts with a capital letter, so it's publicly accessible.

parquet/file/column_writer_types.gen.go.tmpl
```diff
@@ -269,6 +269,97 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpaced(values []{{.name}}, defLev
 {{- end}}
 }
 
+// WriteBatchSpacedWithError behaves like WriteBatchSpaced but reports a write
+// failure as an error instead of panicking or silently discarding it,
+// mirroring the error handling of WriteBatch. Prefer this method on write
+// paths whose sink may fail (for example a network-attached writer).
+func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}}, defLevels, repLevels []int16, validBits []byte, validBitsOffset int64) (valueOffset int64, err error) {
```
lidavidm, line 203: (1) Could we deprecate WriteBatchSpaced?
(2) Could WriteBatchSpaced be written in terms of WriteBatchSpacedWithError instead of duplicating the code entirely? (e.g. defer to WriteBatchSpacedWithError and panic if it returns an error)

zeroshade: Done in 975d0524.

1. **Deprecated** — both `WriteBatchSpaced` (all 8 column-writer types) and `WriteBitmapBatchSpaced` now carry a `// Deprecated:` notice pointing at their `…WithError` counterparts.
2. **No more duplication** — each now just defers to the `…WithError` variant and panics if it returns an error, so the duplicated bodies are gone (−265 lines in the generated file; template + regen).

One behavior note worth your eyes: the legacy methods now fail loudly on a write error instead of silently discarding it on the numeric/boolean/fixed paths (the byte-array path already panicked). Because the delegation routes through `…WithError`'s `recover`, the panic payload is now the `FormatRecoveredError("unknown error type", …)` wrapper rather than the raw error — `errors.Is(…, original)` still resolves through the `%w` chain, but the exact value differs for the byte-array/FLBA paths that previously panicked with the raw error. Happy to re-panic the unwrapped cause for those two if you'd rather preserve their exact original payload.

Internal callers are off the deprecated API: `pqarrow` already uses `WriteBatchSpacedWithError`, and the spaced-write test helper in `internal/testutils` was migrated too. The spaced-write failure tests now assert the deprecated methods panic with the underlying error.

zeroshade: Pushed `f4b92fc5` to address the panic-payload nuance I mentioned above.

The deprecated `WriteBatchSpaced` / `WriteBitmapBatchSpaced` now re-panic the **unwrapped** cause (via a small shared `panicCause` helper) instead of `panic(err)`, so the byte-array / FLBA paths surface the original error value again rather than the `FormatRecoveredError("unknown error type", …)` wrapper they'd otherwise inherit from delegating through `…WithError`. `errors.Is(…, original)` still resolves, and non-error panics fall back to the wrapper.

Net result: those two paths keep their exact pre-deprecation panic payload, while the numeric/boolean paths still fail loud instead of silently dropping the error. `go build`/`vet`/`staticcheck` are clean and the `parquet/file` + `parquet/pqarrow` suites pass.

lidavidm (approved):

PR #853, opened 2026-06-11
fix(arrow/ipc): make Writer and FileWriter safe for concurrent use

### LINKED ISSUE #55: [Go] ipc writer is not concurrency safe

Ipc writer is not concurrency safe. When using ipc writer under concurrent conditions, schema data may be written multiple times. And the data sink may crash with an error.

Demo as below:
```
schema := arrow.NewSchema(...)

address, err := net.ResolveTCPAddr("tcp", '127.0.0.1')
if err!=nil {
    ...
}
conn, err := net.DialTCP("tcp", nil, address)
if err!=nil {
    ...
}

writer := ipc.NewWriter(conn, ipc.WithSchema(schema))

record  :=  ... // create record
for i:=0; i< 10; i++ {
    go func() {
        if err = writer.Write(record); err != nil {
             ...
        }
    }()
}
```

### Component(s)

Go

zeroshade:
### Rationale for this change

Closes #55.

The IPC stream `Writer` and `FileWriter` are not safe for concurrent use. Both share mutable state across `Write`/`Close` with no synchronization:

- the `started` / `headerStarted` flags (checked-then-set in `Write`/`Close` → schema/header can be written more than once),
- the `lastWrittenDicts` map (written from `writeDictionaryPayloads` → Go `fatal error: concurrent map writes`),
- the dictionary `mapper`, the shared `compressors` slice, and the underlying `PayloadWriter` (concurrent `WritePayload` → interleaved/corrupted stream),
- `Close` setting `pw = nil` while a `Write` is in flight → nil-deref / use-after-close.

As reported, calling `Write` from multiple goroutines on one writer can write the schema multiple times and crash the sink. Running the new tests under `-race` against the old code reports many data races.

- Add a `sync.Mutex` to `Writer` and `FileWriter`, held for the duration of the exported mutating methods **`Write`** and **`Close`**.
- The unexported `start` / `checkStarted` helpers are only ever called from `Write`/`Close`, so they intentionally stay **unlocked** to avoid self-deadlock.
- Document the concurrency guarantee on `Write`/`Close`. Calls are serialized; records are written in the order in which callers acquire the lock. The guarantee is freedom from data races and stream corruption — not a particular record ordering — which matches how the standard library documents the mutex-protected `gob.Encoder` and `log.Logger`.

No public signatures change; this is additive (a new unexported field + locking + godoc).

### Are these changes tested?

Yes. Two new `-race` regression tests in `arrow/ipc/writer_test.go`:

- `TestWriterConcurrentWrite` — 16 goroutines `Write` to one stream `Writer`, then the stream is read back and asserted to contain exactly one schema message + 16 record batches.
- `TestFileWriterConcurrentWrite` — same for `FileWriter`, asserting `NumRecords() == 16` on read-back.

Verified that both **fail with data-race warnings** on the pre-change code and **pass** with the lock. The full `arrow/ipc/...` suite passes under `-race` (no deadlock/regression), and `go vet` is clean.

No breaking changes. `Write`/`Close` keep their signatures; concurrent callers now get defined, race-free behavior instead of corruption/panics, and single-threaded callers are unaffected (an uncontended mutex).

### Design note

Apache Arrow C++ (`IpcFormatWriter`) and Java (`ArrowWriter`) treat IPC writers as single-threaded, and most Go stdlib stream writers (`bufio`, `csv`, `json`, `gzip`) leave synchronization to the caller. I went with the mutex (rather than only documenting non-safety) because the reported failure mode is a hard `concurrent map writes` crash, the lock is free when uncontended, and arrow-go already documents concurrency guarantees elsewhere (e.g. `FileReader.RecordBatchAt`/`RecordAt`). Happy to switch to a docs-only "caller must synchronize" approach if maintainers prefer matching C++/Java exactly.

### DIFF
--- a/arrow/ipc/file_writer.go
+++ b/arrow/ipc/file_writer.go
@@ -356,6 +356,9 @@ func (f *FileWriter) checkStarted() error {
 }
 
 func (f *FileWriter) start() error {
+	if f.schema == nil {
+		return errNoSchema
+	}
 	f.headerStarted = true
 	err := f.pw.Start()
 	if err != nil {
--- a/arrow/ipc/file_writer.go
+++ b/arrow/ipc/file_writer.go
@@ -20,6 +20,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"sync"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
@@ -233,6 +234,8 @@ func (ps payloads) Release() {
 
 // FileWriter is an Arrow file writer.
 type FileWriter struct {
+	mu sync.Mutex
+
 	w io.Writer
 
 	mem memory.Allocator
@@ -277,7 +280,12 @@ func NewFileWriter(w io.Writer, opts ...Option) (*FileWriter, error) {
 	return &f, err
 }
 
+// Close writes the file footer and closes the writer. Close is safe to call
+// concurrently with Write; the two are serialized.
 func (f *FileWriter) Close() error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
 	err := f.checkStarted()
 	if err != nil {
 		return fmt.Errorf("arrow/ipc: could not write empty file: %w", err)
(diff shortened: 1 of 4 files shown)


lidavidm (commented):
Why do we want it to be concurrent though? Wouldn't a channel + goroutine make more sense?

zeroshade:
Channel + goroutine definitely makes more sense, but I wanted to avoid the corruption/panic. Alternatively we could explicitly document that it is not concurrent safe if you think that's better than actually making it concurrent safe. Thoughts?

lidavidm:
Do we want to make all the readers/writers thread-safe? I think that's not a variant that should be expected (and is that a variant expected of e.g. an io.Writer? It doesn't seem so.) And furthermore, does Go normally expect things to be thread-safe when it isn't explicitly documented?

zeroshade:
Hmm, that's a good point honestly. the io.Writer interface makes no guarantees about concurrency safety and should only be considered safe if the underlying implementation being used is safe. I think I agree with you that it would be better for us to just add to the documentation that the ipc writer is not concurrency safe and that writes need to get serialized.

PR #858, opened 2026-06-19
fix(arrow/array): reject nulls in non-nullable fields when building records

### LINKED ISSUE #372: RecordBuilder should confirm non-nullable fields do not contain nullable rows

I am able to construct an arrow record, where a field should be non-nullable, but it contains nullable rows. I'm wondering if maybe `RecordBuilder.NewRecord()` should confirm that the fields's arrays do not contain nulls if the field is specified as `Nullable: false` in the schema?

```go
package main

import (
	"log"

)

func main() {
	allocator := memory.NewGoAllocator()
	schema := arrow.NewSchema([]arrow.Field{
	        // should be not null:
		{Name: "a", Type: arrow.PrimitiveTypes.Int32, Nullable: false}, 
	}, nil)
	recordBuilder := array.NewRecordBuilder(allocator, schema)

	recordBuilder.Field(0).(*array.Int32Builder).Append(1)
	recordBuilder.Field(0).(*array.Int32Builder).Append(2)
	recordBuilder.Field(0).AppendNull()
	recordBuilder.Field(0).(*array.Int32Builder).Append(4)

	record := recordBuilder.NewRecord() // this works
	defer record.Release()

	log.Printf("record: %v", record)
	/* prints
2025/05/08 20:32:47 record: record:
schema:
  fields: 1
    - a: type=int32
  rows: 4
  col[0][a]: [1 2 (null) 4]
  */
}
```

This is not allowed in the Rust implementation:
```rs
use std::sync::Arc;

use arrow_array::{Int32Array, RecordBatch};
use arrow_schema::{DataType, Field, Schema};

#[tokio::main]
async fn main() {
    let schema = Arc::new(Schema::new(vec![Field::new("a", DataType::Int32, false)]));
    let mut ids = Int32Array:: builder(4);
    ids.append_value(1);
    ids.append_value(2);
    ids.append_null();
    ids.append_value(4);
    let ids = ids.finish();
    // this panics:
    let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(ids)]).unwrap(); 
    println!("batch: {:?}", batch);
}
```

### Component(s)

Other

zeroshade:
### Rationale for this change

`array.RecordBuilder.NewRecordBatch` (and the deprecated `NewRecord`) silently
produced records whose field was declared `Nullable: false` but whose column
actually contained nulls. The Rust (`RecordBatch::try_new`) and C++
implementations reject this; arrow-go accepted it, yielding spec-invalid
records. Fixes #372.

- **`RecordBuilder`**: `NewRecordBatch` now rejects a record whose **top-level**
  field is non-nullable but whose column contains nulls (it panics, consistent
  with its existing whole-builder length-mismatch guard). A new
  `RecordBuilder.NewRecordBatchChecked() (arrow.RecordBatch, error)` returns the
  violation as an error instead of panicking.
- The check is **scoped to top-level record columns only**. A non-nullable child
  of a *nullable* struct is unaffected (its nulls are legitimate where the parent
  slot is null), so existing nested data keeps working. `array.NewRecordBatch`
  and the lower-level constructors are unchanged.
- **Readers**: the CSV, JSON, and Avro readers build through
  `NewRecordBatchChecked` and surface a typed read error instead of building an
  invalid record or panicking.

### Are these changes tested?

Yes. A new `TestRecordBuilderNonNullableNulls` covers rejection (error + panic),
nullable fields with nulls allowed, non-nullable without nulls allowed, a
nullable struct with a non-nullable child allowed, and the empty-builder case.
Pre-existing fixtures that declared non-nullable fields while appending nulls
were corrected to mark those fields nullable. The full `arrow/...` and
`parquet/...` suites pass.

Yes:

- New `RecordBuilder.NewRecordBatchChecked() (arrow.RecordBatch, error)`.
- `RecordBuilder.NewRecordBatch()` / `NewRecord()` now reject (panic) a top-level
  non-nullable column that contains nulls; previously such records were built
  silently. This is a behavior change for code that relied on the prior lax
  behavior — the fix is to declare the field `Nullable: true`.
- The CSV, JSON, and Avro readers now return an error when input would place a
  null into a non-nullable top-level column.

**Potential follow-ups (intentionally out of scope here):** recursive validation
of nested non-nullable fields, and logical-null detection for non-nullable
run-end-encoded columns.

Closes #372


arrow/array/record.go
```diff
@@ -385,11 +407,26 @@ func (b *RecordBuilder) Resize(n int) {
 //
 // The returned RecordBatch must be Release()'d after use.
 //
-// NewRecordBatch panics if the fields' builder do not have the same length.
+// NewRecordBatch panics if the fields' builders do not have the same length or if
+// a non-nullable field contains nulls. Use NewRecordBatchChecked for an error.
```
lidavidm, line 389: While not strictly breaking, this might be a pretty painful change, if NewRecordBatch now "randomly" panics on things it would've accepted before; maybe we should deprecate this too as a stronger signal?

arrow/array/record.go
```diff
@@ -231,6 +231,28 @@ func (rec *simpleRecord) validate() error {
 	return nil
 }
 
+// nullCountInRows returns the number of top-level nulls among the first nrows
+// rows of arr. It does not recurse into child arrays, so a nullable struct whose
+// non-nullable children carry nulls in null parent slots is not flagged; only a
+// field's own top-level nullability is enforced. Run-end-encoded logical nulls,
+// held in the values child rather than a top-level bitmap, are not counted.
+func nullCountInRows(arr arrow.Array, nrows int64) int {
```
lidavidm: (1) don't arrays cache their null count? can't we check that first as a fast path?
(2) do we really care about the exact count? why do we need to iterate through the entire array?

zeroshade: Good call — done in 3b128df0. Dropped the per-row scan (and the `nullCountInRows` helper); the check is now just `!f.Nullable && cols[i].NullN() > 0`. `NullN()` is the cached null count, so since we only need to know whether *any* nulls are present (not the exact count) there's no reason to walk the column. The error message no longer reports a count.

arrow/array/record.go
zeroshade: I dug into deprecating it, but it turns out to be disproportionate for this PR. `staticcheck`'s SA1019 fires on cross-package use, so deprecating `NewRecordBatch` would mean migrating ~76 call sites across ~20 files to `NewRecordBatchChecked` (golangci-lint caps output at 3 per message, which hid the real count at first). That set also includes the new `TestRecordBuilderNonNullableNulls`, which deliberately calls `NewRecordBatch()` to assert it panics, so it would need a `//nolint:staticcheck,SA1019`.

Since `NewRecordBatch` already panics today on a whole-builder length mismatch, rejecting nulls in a non-nullable column reads as a consistent extension of that existing contract rather than brand-new panic behavior — and `NewRecordBatchChecked` is there for callers who'd rather have an error. So for now I've left `NewRecordBatch` as-is with the clarified doc. Happy to do the deprecation + full caller migration as a dedicated follow-up if you think the stronger signal is worth it; it just felt too sweeping to fold into this fix. WDYT?

lidavidm: Um, I'm not sure if I'm talking to you or a bot, but I'd think accidentally inserting a null (especially because nullable defaults to false due to false being the zero value) is far more common, and a panic is fairly unwelcome.

PR #872, opened 2026-06-30
fix(array): array data hashing coverage

fallintoplace:
### What changed

array.Hash now writes the array data type fingerprint, length, offset, every buffer with boundaries and nil markers, child data in order, and dictionary data when present.

### Why

The old implementation wrote length twice and only considered the null bitmap plus child data. That meant arrays with different value buffers, types, offsets, or dictionaries could produce the same hash.

### Validation

- go test ./arrow/array -run TestHashIncludesDataMetadataAndBuffers -count=1

### DIFF
--- a/arrow/array/data.go
+++ b/arrow/array/data.go
@@ -17,15 +17,13 @@
 package array
 
 import (
-	"math/bits"
+	"encoding/binary"
 	"sync/atomic"
-	"unsafe"
-
-	"github.com/apache/arrow-go/v18/internal/utils/maphash"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/internal/utils/maphash"
 )
 
 // Data represents the memory and metadata of an Arrow array.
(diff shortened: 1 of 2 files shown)


arrow/array/data.go
```diff
@@ -270,12 +268,59 @@ func NewSliceData(data arrow.ArrayData, i, j int64) arrow.ArrayData {
...
+}
+
+func hashInt(h *maphash.MapHash, v int) {
+	var b [8]byte
+	binary.LittleEndian.PutUint64(b[:], uint64(int64(v)))
+	h.Write(b[:])
+}
+
+func hashBytes(h *maphash.MapHash, b []byte) {
+	hashInt(h, len(b))
```
zeroshade, line 324: why do we need to separately hash the length? isn't `h.Write(b)` sufficient?

arrow/array/data.go
```diff
@@ -270,12 +268,59 @@ func NewSliceData(data arrow.ArrayData, i, j int64) arrow.ArrayData {
...
+	h.Write(b)
+}
+
+func hashString(h *maphash.MapHash, s string) {
+	hashInt(h, len(s))
+	h.WriteString(s)
```
zeroshade, line 330: same question as above, isn't `h.WriteString(s)` sufficient?

arrow/array/data.go
fallintoplace: h.Write(b) is sufficient for hashing one isolated byte slice, but here the hash input is a structured stream made of multiple variable-width fields. maphash does not preserve Write call boundaries; it only sees the final sequence of bytes. So without length-prefixing, different structured inputs can serialize to the same byte stream, e.g. splitting bytes differently across adjacent buffers/strings

zeroshade: Can you construct an example of two valid arrays where this would happen?

zeroshade: I went and verified this — you're right, the length prefix is needed in general. Here's a concrete pair of `binary` arrays that both pass `ValidateFull` and hold **different** values, yet hash identically without it:

| | length | validity | offsets (int32) | values | value[0] |
|---|---|---|---|---|---|
| A | 1 | nil | `[0, 1]` | `00 00 00 01 58` | `0x00` |
| B | 1 | nil | `[0, 1, 1]` | `58` | `0x58` |

Both are valid — only `offsets[:length+1]` and `values[:lastOffset]` are read, so the extra trailing bytes are allowed. Without the length prefix, each one's buffer section serializes to the identical stream:

```
00 01 00 00 00 00 01 00 00 00 01 00 00 00 01 58
```

A parses that as `[nil][offsets 0,1][values 00 00 00 01 58]` and B as `[nil][offsets 0,1,1][values 58]` — same bytes, so they collide for every seed. The presence byte doesn't help here; it's just another `0x01` that can also occur inside a buffer.

You're right that fresh **builder** output won't trigger this: builders trim buffers to canonical sizes, so the offsets buffer pins the split (I checked `["AB","C"]` vs `["A","BC"]`, and the `list<int8>` equivalent — no collision). But the over-long buffers above are exactly what `Hash` has to accept from slicing, IPC / C-Data import, and manual `NewData`, and the caller (`scalar.Hash` -> `array.Hash(s.GetList().Data())`) hands us arbitrary list-scalar data. I'd rather the hash not depend on "every buffer is canonically sized."

It also lines up with arrow-c++ and arrow-rs: the C++ `RowEncoder` frames variable-width values as `[null byte, length, bytes]`, and the arrow-rs row format is explicitly designed so a byte is never ambiguously part of a value vs a delimiter.

Let's keep the length prefix.

zeroshade (approved):
LGTM. Confirmed the length-prefixing is necessary for a general array-data hash (concrete counterexample + arrow-c++/arrow-rs precedent in the thread).

zeroshade:
okay, just needs a rebase

fallintoplace:
@zeroshade Give me a minute, I will resync the branch and push this.

PR #876, opened 2026-06-30
fix(array): clear dictionary data on reset

fallintoplace:
### What changed

Data.Reset now releases and clears any existing dictionary data when reusing an ArrayData value.

### Why

Reset replaces the buffers, children, type, length, null count, and offset, but previously left the old dictionary attached. Reusing dictionary-backed data as non-dictionary data could keep returning stale dictionary data and retain the old dictionary longer than expected.

### Validation

- go test ./arrow/array -run 'TestDataReset' -count=1


zeroshade (approved):

zeroshade:
just needs a rebase

PR #877, opened 2026-06-30
Support dense union slice offsets

fallintoplace:
### What changed

NewDenseUnionFromArraysWithFieldCodes now normalizes the type ID and offset value buffers to each input array logical window before constructing dense union data.

### Why

Dense unions store one ArrayData offset, but callers may pass type ID and offset arrays that were sliced at different physical offsets. Normalizing the borrowed buffers lets the constructed dense union use offset 0 while still preserving zero-copy buffer sharing and reading the intended logical entries.

### Validation

- go test ./arrow/array -run 'TestUnions/TestMakeDenseUnions' -count=1

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -558,14 +558,30 @@ func NewDenseUnionFromArraysWithFieldCodes(typeIDs, offsets arrow.Array, childre
 	}
 
 	ty := arrow.DenseUnionFromArrays(children, fields, codes)
-	buffers := []*memory.Buffer{nil, typeIDs.Data().Buffers()[1], offsets.Data().Buffers()[1]}
+	typeIDBuffer := typeIDs.Data().Buffers()[1]
+	if typeIDBuffer != nil {
+		typeIDBuffer = memory.SliceBuffer(typeIDBuffer,
+			arrow.Int8Traits.BytesRequired(typeIDs.Data().Offset()),
+			arrow.Int8Traits.BytesRequired(typeIDs.Len()))
+		defer typeIDBuffer.Release()
+	}
+
+	offsetBuffer := offsets.Data().Buffers()[1]
+	if offsetBuffer != nil {
+		offsetBuffer = memory.SliceBuffer(offsetBuffer,
+			arrow.Int32Traits.BytesRequired(offsets.Data().Offset()),
+			arrow.Int32Traits.BytesRequired(offsets.Len()))
+		defer offsetBuffer.Release()
+	}
+
+	buffers := []*memory.Buffer{nil, typeIDBuffer, offsetBuffer}
 
 	childData := make([]arrow.ArrayData, len(children))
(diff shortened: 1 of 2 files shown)


zeroshade:
Rather than reject this, we should properly support it instead. There's no reason why we shouldn't allow constructing a dense union where the offsets and typeIDs are using different offsets. Rather than rejecting this, can we instead fix the physical offset handling to handle this scenario? 

For reference, neither C++ nor rust reject this, but both exhibit this same bug.

zeroshade (commented):
Took a close look and verified the dense fix end-to-end — normalizing each borrowed buffer to its own window with `SliceBuffer` and setting the union offset to 0 is correct (the offset *values* stay valid since the children are shared whole), it's zero-copy, and the refcounting balances (`NewData` retains, and the `defer Release()` pairs with `SliceBuffer`). Union tests pass under `-race`. Nice work.

Two things before I approve:

1. **Add a combined case to the new test.** It currently covers `typeIDs`-sliced and `offsets`-sliced independently, but not both sliced at *different* offsets at the same time (e.g. `typeIDs` at offset 1, `offsets` at offset 2). That's the exact scenario the two-buffer normalization is protecting, so it'd be good to lock it in.

2. **Related, separate PR (not blocking this one):** the sparse constructor has the same class of bug. `NewSparseUnionFromArraysWithFieldCodes` still does `NewData(..., typeIDs.Data().Offset())`, and `union.setData` applies that offset to the children via `NewSliceData(child, offset, offset+length)`, so passing a sliced `typeIDs` panics:

   ```go
   base := /* int8 typeIDs array, len 11 */
   sliced := array.NewSlice(base, 1, int64(base.Len()))
   _, _ = array.NewSparseUnionFromArrays(sliced, children)
   // panic: arrow/array: index out of range
   ```

   Same root cause you flagged for C++/Rust. Worth a follow-up so sparse gets parity with this dense fix.

zeroshade (approved):
Approving. The pivot from rejecting mismatched slice offsets to properly supporting them is the right call, and the implementation is correct and memory-safe:

- Normalizing each borrowed buffer to its own logical window with `SliceBuffer` and constructing the union at offset 0 is sound — the offset *values* stay valid because the children are shared whole, and the read lands on the intended entries.
- Refcounting balances: `SliceBuffer` retains the parent, `NewData` retains the buffers, and the `defer Release()` calls pair off, leaving each buffer held exactly once by the union's `ArrayData`.
- The combined different-offset case (`typeIDs` @ offset 1, `offsets` @ offset 2) is now covered, and the full `arrow/array` package passes under `-race` with no leaks (the suite's `CheckedAllocator.AssertSize(0)`), `ValidateFull()` clean.

The sparse constructor had the same class of bug I mentioned earlier; that's being handled separately in #881, so it's correctly out of scope here. Thanks for turning this around!

Issue #865, opened 2026-06-26
[Go][Parquet] support streaming reads for large data pages


joechenrh:
### Describe the enhancement requested

We have meet a Parquet file with extremely large data pages, making materializing the whole page expensive.

One observed page has metadata similar to:

```text
type: DATA_PAGE
compressedSize: 1079936746
uncompressedSize: 1426870008
numValues: 233
encoding: PLAIN
definitionLevelEncoding: RLE
repetitionLevelEncoding: BIT_PACKED
hasCrc: true
```

Today the Parquet reader materializes the full uncompressed page buffer before decoding values, which requires allocating memory proportional to the page's uncompressed size.

It would be useful for the reader to support a streaming/lazy data page read path for large pages. For Data Page V1, the reader could parse repetition and definition levels first into small buffers, then leave the decompressed stream positioned at the encoded values and decode values on demand.

A possible first step would be an opt-in reader property or strategy that enables streaming value decoding for supported page shapes, for example:

- Data Page V1
- PLAIN encoding
- BYTE_ARRAY / string-like columns
- RLE definition levels
- RLE or BIT_PACKED repetition levels
- codecs that can be exposed as an `io.Reader` stream

Unsupported encodings/codecs/page layouts could continue to fall back to the existing materialized page path.

The goal is to allow applications to read valid Parquet files with very large pages while keeping peak reader memory closer to the requested batch size plus small level buffers.

zeroshade:
This is an interesting idea! I like it in theory, it will just be tricky to implement properly. Are you interested in giving implementing this a shot yourself?

joechenrh:
Yes, I'm still trying to figure out a clean way to implement this. 😸

zeroshade:
Sounds good! Looking forward to reviewing it. Feel free to use this issue for discussion if you need to bounce ideas off me or otherwise discuss options

joechenrh:
A sketch of the approach — would appreciate a sanity check on the shape before I build it out.

**Scope:** Data Page V1 + V2, `PLAIN` encoding, types `int32/int64/float/double/FLBA/byte_array`. Codecs: `UNCOMPRESSED/GZIP/BROTLI/ZSTD` only, everything else falls back to the current path. Opt-in via a new `ReaderProperties.EnablePageStreaming` (default off, no behavior change).

**Core idea**

Currently the decoder reads from a fully-decompressed `[]byte` slice.

[...]

PR #880, opened 2026-07-01
feat(parquet): opt-in streaming reads for large data pages

joechenrh:
### Rationale for this change

Addresses #865. Today the reader materializes each full uncompressed data page before decoding, so peak memory scales with the page's uncompressed size — a problem for files with very large data pages (e.g. a 1 GB compressed / 1.4 GB uncompressed page). This adds an **opt-in** path that keeps peak memory near the requested batch size for the common large-page shape.

Approach was discussed and agreed on the issue.

### What changes are included in this PR

Opt-in via `ReaderProperties.PageStreamingEnabled` (default **false**, no behavior change). When set, **eligible** data pages skip full-page decompression and decode their values incrementally; every ineligible page falls back to the existing materialized path.

**Scope (first cut):**
- Data Page **V1 + V2**, **PLAIN** encoding
- Physical types **`byte_array`** and **`fixed_len_byte_array`**
- Codecs **`UNCOMPRESSED` / `GZIP` / `BROTLI` / `ZSTD`** (explicit allowlist — Snappy is a raw block whereas its `NewReader` is the framed format, and LZ4_RAW has no streaming reader)

**Design:**
- `parquet/internal/encoding/streaming`: a small `ValueBuffer` — an incremental byte source that reads from an `io.Reader`, slides as values are consumed, grows to fit a single oversized value, and owns the page's stream (drained + closed on release).
- The materialized `PlainByteArrayDecoder` / `PlainFixedLenByteArrayDecoder` keep their `[]byte` logic byte-for-byte, gaining only a `src` field + a one-line guard in `Decode`/`Discard`. Streaming decode lives beside them; `SetData` resets `src`, so a cached decoder reused for a materialized page reverts. Other decoders are untouched.
- Page reader: for eligible pages, V2 levels are read raw (uncompressed, header byte-lengths) and V1 levels are peeled off the decompressed stream using the same length rules as `LevelDecoder.SetData`; values become a `ValueBuffer` over the compressed region. On page release the underlying stream is drained + closed so the reader lands on the next page header even if values were skipped.
- Column reader: the same cached PLAIN decoder is fed via `SetSource` (streaming) or `SetData` (materialized). `Page`/`DataPage` already return internal types (e.g. `format.Encoding`), so they can't be implemented outside `parquet/`; the added **unexported** `valueBuffer`/`valueClip` accessors are internal plumbing invisible to external consumers.

### Are these changes tested?

Yes.
- All existing encoding tests pass unchanged (materialized parity); the streaming decoders are additionally exercised with a one-byte-at-a-time reader and oversized values.
- End-to-end round-trip (`parquet/file/streaming_page_test.go`): streaming vs materialized reads are asserted identical across V1/V2 and each codec, for **required** and **nullable** columns (the nullable case covers the V1 def-level peel + streaming `DecodeSpaced`), including values larger than a page and the stream buffer.

One new opt-in field, `ReaderProperties.PageStreamingEnabled` (default false). No behavior change to decoded values; the only interface change is two unexported accessor methods on `DataPage` (see above).

Follow-ups (out of scope here): fixed-width numeric PLAIN types, raw-Snappy and LZ4_RAW streaming, and a memory benchmark.


parquet/file/row_group_reader.go
```diff
@@ -129,6 +130,13 @@ func (r *RowGroupReader) GetColumnPageReader(i int) (PageReader, error) {
 			maxPageHeaderSize: defaultMaxPageHeaderSize,
 			nrows:             col.NumValues(),
 			mem:               r.props.Allocator(),
+			// Streaming (EnablePageStreaming) inputs; only the unencrypted path is
+			// ever streaming-eligible, so columnCanStream is left false elsewhere.
+			columnCanStream: r.props.EnablePageStreaming &&
```
zeroshade: Enabling streaming here also affects the **public** `GetColumnPageReader`: a streaming `DataPageV1/V2`'s `buf` holds only the rep/def level region, so `Page.Data()` returns just the levels instead of the whole page. That's a silent semantic change to a public method for anyone reading `Data()` off a data page with `EnablePageStreaming` set.

Now that the internal callers read levels/values through `levelBuffer()`/`valueBuffer()`, the original blocker to keeping `GetColumnPageReader` materialized-only (the revert in 2f5aab9) looks gone. Could we gate streaming to the internal `Column()` path, or — failing that — document `Data()` as level-only/undefined for streaming pages? This is the main thing I'd like resolved before merge.

parquet/file/page_reader.go
```diff
@@ -737,6 +786,92 @@ func (p *serializedPageReader) SeekToPageWithRow(rowIdx int64) error {
...
+		switch enc {
+		case format.Encoding_RLE:
+			var hdr [4]byte
+			if _, err := io.ReadFull(r, hdr[:]); err != nil {
+				return err
+			}
+			n := int(binary.LittleEndian.Uint32(hdr[:]))
+			start := len(buf)
+			buf = append(buf, hdr[:]...)
+			buf = append(buf, make([]byte, n)...)
```
zeroshade: `n` comes straight from the 4-byte RLE length prefix in the page and is used to `make([]byte, n)` before any bounds check, so a corrupt/malicious V1 page with a bogus prefix (e.g. ~4 GiB) triggers a huge allocation before the following `io.ReadFull` fails.

The materialized `LevelDecoder.SetData` guards this with `nbytes > len(data)-4` against the already-bounded page buffer; the streaming path has no equivalent ceiling (V1 carries no per-level byte length in the header). Please bound `n` (e.g. against `lenCompressed`/`lenUncompressed`) before allocating.

parquet/file/page_reader.go
```diff
@@ -737,6 +786,92 @@ func (p *serializedPageReader) SeekToPageWithRow(rowIdx int64) error {
...
+			if _, err := io.ReadFull(r, buf[start+4:]); err != nil {
+				return err
+			}
+		case format.Encoding_BIT_PACKED:
+			bitWidth := bits.Len64(uint64(maxLvl))
+			n := int(bitutil.BytesForBits(int64(numValues) * int64(bitWidth)))
```
zeroshade: Minor/consistency: `LevelDecoder.SetData` computes this length with the overflow-checked `shared_utils.Mul(nbuffered, bitWidth)`, whereas here it's a raw `int64(numValues) * int64(bitWidth)`. Not exploitable for real level bit-widths, but matching the sibling code's checked multiply would keep the two paths in sync.

parquet/internal/encoding/streaming/value_buffer.go
```diff
@@ -0,0 +1,105 @@
...
+	buf     []byte
+	off, n  int
+	// shared: buf was handed out via FillOwned and may back live aliases, so it
+	// must not be slid/overwritten.
+	shared bool
+}
+
+// NewStreamBuffer returns a ValueBuffer over r.
+func NewStreamBuffer(r io.Reader, onClose func() error) ValueBuffer {
+	return &streamBuffer{r: r, onClose: onClose, buf: make([]byte, defaultStreamBufferSize)}
```
zeroshade: Two notes on the backing buffer:

- It's a fresh 1 MiB (`defaultStreamBufferSize`) allocation per streaming page via the Go allocator rather than the configured `memory.Allocator`, so it's untracked by `CheckedAllocator` and adds GC churn on files with many eligible pages. Consider pooling it (and/or sizing relative to the read batch).
- Worth calling out in the `EnablePageStreaming` doc that peak memory is really `batch + ~1 MiB + largest-value growth + levels`, not just the batch size.

parquet/file/streaming_page_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+// TestPageStreamingByteArrayRoundTrip verifies that reading a PLAIN BYTE_ARRAY
+// column with EnablePageStreaming produces byte-identical values to the
+// materialized path, across V1/V2 data pages and the streaming-eligible codecs,
+// including values larger than the stream buffer and larger than a data page.
+func TestPageStreamingByteArrayRoundTrip(t *testing.T) {
+	// Varied sizes: several small, some > 4096 (stream buffer), some > 1024 (page).
+	values := makeStreamTestValues([]int{65, 100, 5000, 200, 9000, 50, 4096, 1})
+
+	cases := []struct {
```
zeroshade, line 120: Since this ships new decode/peel logic, a few gaps I'd like covered:

- **BIT_PACKED levels** — only RLE levels are exercised, but `readLevelData` has a BIT_PACKED branch.
- **Skip path** — `discardStreaming` (via `SeekToRow`/`skipRows`) isn't covered; the tests only do full `ReadBatch`.
- **Repeated columns** (`maxRepLevel > 0`) — the V1 rep-level peel in `readLevelData` is untested (current cases are flat required/optional).
- **Ineligible-codec fallback** — a case with Snappy or LZ4_RAW + `EnablePageStreaming=true` asserting correct (materialized) reads would lock in the allowlist behavior.

zeroshade (changes requested):
Thanks for tackling #865 — the incremental-decode design is clean and the commit series is easy to follow.

I reviewed the read path in depth and ran the new streaming tests plus the existing `encoding` and `parquet/file` suites with `-race`; all pass and materialized parity holds. The drain-before-advance ordering in `Next()` (release the previous page, then read the next header), the codec allowlist (correctly excluding raw-block Snappy even though it implements `StreamingCodec`), the V1 level byte-length parity with `LevelDecoder.SetData`, and the `DecodeSpaced` -> `Decode` streaming dispatch all check out.

A few things I'd like addressed before merge:

1. **`Page.Data()` contract on streaming pages (blocking).** A streaming page's `buf` holds only the rep/def level region, so the public `Page.Data()` returns truncated bytes when a page is obtained via the public `GetColumnPageReader` with `EnablePageStreaming` set. Since internal callers now go through `levelBuffer()`/`valueBuffer()`, I'd like us to either gate streaming to the internal `Column()` path or document `Data()` explicitly (see inline).
2. **Unbounded allocation from an untrusted RLE length prefix** in `readLevelData`.
3. **Test coverage** for BIT_PACKED levels, the skip/`Discard` path, repeated columns, and ineligible-codec fallback.

Details inline. Nice work overall — with the `Data()` semantics resolved this is close.

zeroshade (dismissed):
Thanks for tackling #865 — the incremental-decode design is clean and the commit series is easy to follow.

I reviewed the read path in depth and ran the new streaming tests plus the existing `encoding` and `parquet/file` suites with `-race`; all pass and materialized parity holds. The drain-before-advance ordering in `Next()` (release the previous page, then read the next header), the codec allowlist (correctly excluding raw-block Snappy even though it implements `StreamingCodec`), the V1 level byte-length parity with `LevelDecoder.SetData`, and the `DecodeSpaced` -> `Decode` streaming dispatch all check out.

A few things I'd like addressed before merge:

1. **`Page.Data()` contract on streaming pages (blocking).** A streaming page's `buf` holds only the rep/def level region, so the public `Page.Data()` returns truncated bytes when a page is obtained via the public `GetColumnPageReader` with `EnablePageStreaming` set. Since internal callers now go through `levelBuffer()`/`valueBuffer()`, I'd like us to either gate streaming to the internal `Column()` path or document `Data()` explicitly (see inline).
2. **Unbounded allocation from an untrusted RLE length prefix** in `readLevelData`.
3. **Test coverage** for BIT_PACKED levels, the skip/`Discard` path, repeated columns, and ineligible-codec fallback.

Details inline. Nice work overall — with the `Data()` semantics resolved this is close.

parquet/file/page_reader.go
joechenrh: Yeah, updated to bound the allocation with uncompressed size in https://github.com/apache/arrow-go/pull/880/changes/3cd007627618726e1189b65b06edac76a44d8553

parquet/file/page_reader.go
joechenrh: Nice catch, I didn't notice that my AI has modified this part during reviewing. Updated in https://github.com/apache/arrow-go/pull/880/changes/aeeb9920fbbcbb6164b8af6175e9beaae3b9935c

parquet/internal/encoding/streaming/value_buffer.go
joechenrh: In https://github.com/apache/arrow-go/pull/880/changes/1b822707b43a542185e17dc8885f951c196b92bf, I cap the size of the buffer with the uncompressed size and manage all the buffers with `Allocator`. And the comment for `EnablePageStreaming` is also aligned with the new behavior.

parquet/file/row_group_reader.go
joechenrh: Add document for both `EnablePageStreaming` and `Data()` to clarify the change. My judgment is as follows. I wonder if you have any opinions:

I find that `pqarrow` also uses `GetColumnPageReader`, so I prefer explicitly document this semantic change. Besides, since this new option is off by default, existing downstream user codes shouldn't be affected when bumping the package. When a user (or an AI agent) wants to enable the option, they can find this changes from the comments from `EnablePageStreaming`.

joechenrh:
I have tried my best to control the size (as well as the scope) of this PR, but it still exceeds ~1,000~ 1,500 rows 😢. Since it's not a small fix, I think I need more testing before merging it.

parquet/internal/encoding/streaming/value_buffer.go
joechenrh: Oh, maybe we can't use allocator here.

joechenrh: Updated, we should store the buffer inside the `valueBuffer`, and resue them for each `Decode` call.

zeroshade (commented):
Re-review of the latest push — thanks for the very thorough follow-up. All three of my blocking items are addressed, and I verified locally (built `./parquet/...`, ran the `TestPageStreaming*` suite with `PARQUET_TEST_DATA` set — green):

1. **`Page.Data()` contract** — documented on both `EnablePageStreaming` and `Data()`, with streaming kept to the eligible internal path via `columnCanStream`. Your rationale (off by default, pqarrow uses the same accessor, documented semantic) is sound.
2. **Unbounded RLE level allocation** — now bounded by the uncompressed page size in the streaming level read (`setStreamingValues`), and the BIT_PACKED length uses the checked multiply. Both inline points resolved.
3. **Coverage** — `TestPageStreaming{ByteArraySkip,FLBASkip,Repeated,ReadBatchInPageRepeated,IneligibleCodecFallback,AllocatorBalance,PeakMemoryBounded,PageReaderCloseFreesStream}` cover the skip / repeated / ineligible-codec / allocator-tracking paths I asked for.

Scope note: this pass verified the above + ran the streaming suite; I haven't deep-audited the newer perf commits (zstd decoder pooling, chunked FLBA skip, V2 streaming hardening). Since it's large and you mentioned wanting more testing, I'll hold a formal approve until you say it's ready — ping me for a final pass. Great iteration.

joechenrh:
### Status

**Unit tests**: pass locally — full `parquet/...` suite plus the streaming-specific tests.

**Perf**: quick in-memory benchmark reading large (8 MiB) data pages via `ReadBatchInPage`, streaming off → on:

| value | codec | peak (off → on) | time |
|---|---|---|---|
| 64 B | uncompressed | 16 MiB → **1.0 MiB** | +32% |
| 64 B | zstd | 16 MiB → **1.0 MiB** | +20% |
| 64 B | gzip | 16 MiB → **1.0 MiB** | +45% |
| 16 KiB | uncompressed | 32 MiB → **1.0 MiB** | −23% |
| 16 KiB | zstd | 32 MiB → **1.0 MiB** | +34% |
| 16 KiB | gzip | 32 MiB → **1.0 MiB** | −18% |
| 256 KiB | uncompressed | 64 MiB → **1.0 MiB** | −43% |
| 256 KiB | zstd | 64 MiB → **1.0 MiB** | +38% |
| 256 KiB | gzip | 64 MiB → **1.0 MiB** | −28% |

- **Memory** (the goal): peak decode allocation stays flat at ~1 MiB with streaming on, vs page-scale off — **~15–63×** lower, scaling with page size. Same result verified end-to-end through a downstream reader's `ReadBatchInPage` path.
- **Speed**: depends on value size and codec.
  - Tiny values (~64 B): slower for all codecs (+20–45%) — per-value overhead dominates when there are many values.
  - Medium/large values: uncompressed and gzip are on par or **faster** (smaller working set, cache-friendly); zstd is ~+35% slower — structural: streaming decodes incrementally with an extra copy, vs one-shot `DecodeAll` + zero-copy slicing on the non-streaming path. I prototyped a read-ahead buffer to close the zstd gap and measured ≈0 improvement (the cost is the extra copy, not the read count), so I dropped it.

**S3** (validated on EC2 in the bucket's region, us-west-2): a single large data page read over real S3 at `BufferSize = 4 MiB`, non-streaming (master) vs streaming, with memory and time captured from the same read.

| page | master: mem / time | streaming: mem / time |
|---|---|---|
| 2 MiB | 8 MiB / 87 ms | **3 MiB** / 91 ms |
| 8 MiB | 20 MiB / 182 ms | **5 MiB** / 188 ms |
| 64 MiB | 132 MiB / 832 ms | **5 MiB** / 1.4 s |
| 256 MiB | 516 MiB / 3.1 s | **5 MiB** / 3.9 s |

- **Memory** holds end-to-end: streaming stays flat at ~5 MiB (`window + BufferSize`) regardless of page size, vs page-scale for master — a 256 MiB page is **516 MiB → 5 MiB** (~100×).
- **Speed** is driven by `BufferSize`, not the decode. Master reads whole pages in a few large requests; streaming reads incrementally, so too small a buffer means many small S3 requests (the 256 MiB page: ~8 s / 259 requests at `BufferSize = 1 MiB`). At `4 MiB` streaming is within ~1.2× of master (~3.9 s / 67 requests for the 256 MiB page) while using 100×+ less memory.
- This PR also caps the buffered-stream reader at `min(BufferSize, columnChunkSize)`, so small columns and wide files don't over-allocate (the 2 MiB page above uses a 2 MiB buffer, not 4 MiB).

**Suggested default for S3: `BufferSize = 4 MiB`** — flat ~5 MiB per column vs page-scale for master, at near-baseline speed. The buffer is paid per column, so keeping it small matters for wide reads.

joechenrh:
Ready for review. Thanks for taking a look!
+ Scope: PLAIN-encoded V1 + V2 pages, ByteArray / FixedLenByteArray, unencrypted streaming-capable codecs (UNCOMPRESSED/GZIP/BROTLI/ZSTD). Every other page keeps the existing materialized path. No public interface changes.
+ ~620 lines of production code, ~1270 of tests.
+ Validation: verified end-to-end on real S3 and local disk. On large pages peak decode memory drops ~100–1000× (e.g. a 256 MiB page: 512 MiB → 5 MiB; a 1 GiB page: ~2 GiB → ~2 MiB), with read time on par with the non-streaming path at BufferSize ≥ 4 MiB.

parquet/internal/encoding/streaming/value_buffer.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+	if s.n-s.off < need {
+		return nil, io.ErrUnexpectedEOF
+	}
+	return s.cur[s.off:s.n], nil
+}
+
+// rotate retires the full chunk (still backing aliases) and starts fresh; no tail since off == n.
+func (s *streamBuffer) rotate(need int) {
```
zeroshade, line 137: Non-blocking, but worth hardening before more decoders adopt this: `rotate` assumes `off == n` (per the comment) and so drops the buffered tail `cur[off:n]` when it starts a fresh chunk. That invariant holds for every current caller because they all do strictly matched `Fill(need)`/`Advance(need)`. But `Fill`'s first branch can return *more* than `need`, so a future "peek then partial-advance" decoder would silently drop un-consumed bytes here — I reproduced both corruption and a spurious `unexpected EOF` with a `Fill(6)`/`Advance(2)`/`Fill(7)` probe against this buffer.

Suggest either carrying the tail into the new chunk on rotate, or making the invariant fail loud, e.g. `if s.off != s.n { panic("streamBuffer.rotate: unconsumed tail") }`.

parquet/file/row_group_reader.go
```diff
@@ -186,3 +195,21 @@ func (r *RowGroupReader) GetColumnPageReader(i int) (PageReader, error) {
...
+	case parquet.Types.ByteArray, parquet.Types.FixedLenByteArray:
+		return true
+	default:
+		return false
+	}
+}
+
+func streamableCodec(c compress.Compression) bool {
+	switch c {
+	case compress.Codecs.Uncompressed, compress.Codecs.Gzip, compress.Codecs.Brotli, compress.Codecs.Zstd:
```
zeroshade, line 210: Non-blocking nit: Snappy actually implements `StreamingCodec` (compress/snappy.go) yet is intentionally excluded from this allowlist (parquet's Snappy framing is block-based, not a spec-compliant stream). A one-line comment here explaining *why* Snappy/LZ4_RAW are excluded despite being registered streaming codecs would save a future maintainer from "helpfully" adding them.

zeroshade (commented):
Went back through this end-to-end — ran the streaming tests under `-race`, checked pqarrow's `GetColumnPageReader` usage (it reads through the column/record value path, not `Page.Data()`), and confirmed the length accounting here does NOT have the compressed-page regression that #911 hit. All five of my earlier comments are genuinely resolved and the docs/tests are solid.

Documenting the `Data()` semantics rather than gating is fine given the only in-tree consumer reads through the value path. Two non-blocking nits inline — no need to block on them, but the `rotate` one is worth hardening before more decoders adopt the `ValueBuffer` contract.

parquet/internal/encoding/streaming/value_buffer.go
joechenrh: Updated in https://github.com/apache/arrow-go/pull/880/changes/413feca0da42d74c5889f4a4aaa7b1c1e25052aa. Previously all the call site pass the correct `need` value, so I don't handle such case.

parquet/file/row_group_reader.go
joechenrh: Good point, updated in https://github.com/apache/arrow-go/pull/880/changes/413feca0da42d74c5889f4a4aaa7b1c1e25052aa with a one-line comment.

joechenrh:
Previous changes introduced lint error and race issue, fixed in https://github.com/apache/arrow-go/pull/880/changes/2d139f915407c3baec43033b65e8eb1debed1359 + https://github.com/apache/arrow-go/pull/880/changes/db7cad2ab65375beb8cf194e991b3a6234be9e74 (caused by double-free) and https://github.com/apache/arrow-go/pull/880/commits/8fd6acccace960c4af60e256225a172250cb2f76.

parquet/file/page_reader.go
```diff
@@ -91,6 +99,43 @@ func (p *page) Data() []byte              { return p.buf.Bytes() }
...
+	clip int64
+}
+
+func (d *dataPageBase) valueBuffer() streaming.ValueBuffer { return d.valueBuf }
+func (d *dataPageBase) levelBuffer() []byte                { return d.buf.Bytes() }
+
+func (d *dataPageBase) valueClip() int64 { return d.clip }
+
+func (d *dataPageBase) setStreamingValues(mem memory.Allocator, lenUncompressed int, levelBuf []byte, valReader, rawSrc io.Reader, closer io.Closer) {
+	d.buf = memory.NewBufferBytes(levelBuf)
```
zeroshade: **Allocator accounting: the streaming rep/def level buffer bypasses the configured `memory.Allocator`.**

`setStreamingValues` wraps `levelBuf` with `memory.NewBufferBytes(...)`, but `levelBuf` is `make`-allocated in both streaming paths (V2: `make([]byte, levelsBytelen)`; V1: via `readLevelData`). Unlike the non-streaming data pages — which wrap `p.dataPageBuffer.Bytes()` from an allocator-backed `ResizableBuffer` (`memory.NewResizableBuffer(p.mem)`) — this level region isn't tracked by the configured allocator, and `DataPage.Release()` can't return it. Checked/custom allocators won't see it.

The streamed *value* region already flows through `p.mem` via `streaming.NewStreamBuffer(mem, ...)`, so only the (usually small) level region is unaccounted — real impact is limited to level-heavy deeply-repeated/nested pages. It's still inconsistent with the non-streaming paths and slightly at odds with this PR's memory-management goal.

Suggestion: allocate `levelBuf` through `p.mem` and store an allocator-backed `*memory.Buffer` on the page so `DataPage.Release()` frees it.

_(Surfaced by an automated review pass — non-blocking; feel free to defer if you consider the level region negligible.)_

joechenrh: Fixed in https://github.com/apache/arrow-go/commit/7529291d305c17eb9d8435d0cfd0ca7080d4bc94 — since this fix is not large, so I think we can add it in this PR. The rep/def level region now reads into an allocator-backed buffer so `DataPage.Release` frees it and checked/custom allocators account for it.

joechenrh: Small follow-up on this one: the first cut in `7529291d` actually tripped CI and is fixed in https://github.com/apache/arrow-go/commit/5f00dc754d9af6e66924b1d92c38eff056e869fa.

Re-ran the actual CI recipe this time — `go test -race -tags assert ./parquet/...` plus the `,noasm` variant with the test-data env set — and it's fully green now. 🟢

zeroshade (approved):
LGTM. The opt-in streaming decode for large PLAIN byte-array data pages looks good — nice work on the streaming buffers, decoder integration, page-reader wiring, and zstd decoder pooling. Thanks for iterating on this.

PR #881, opened 2026-07-01
Support sparse union sliced type ids

fallintoplace:
### What changed

`NewSparseUnionFromArraysWithFieldCodes` now normalizes the type ID value buffer to the input array's logical window before constructing sparse union data.

### Why

The constructor previously stored `typeIDs.Data().Offset()` on the union `ArrayData` while borrowing the unsliced type ID buffer. For sparse unions, `union.setData` uses that `ArrayData` offset to slice children, so passing sliced type IDs with logical-length children could panic or read the wrong physical type ID entries.

### Validation

- `go test ./arrow/array -run 'TestUnions/TestMakeSparse' -count=1`
- `PARQUET_TEST_DATA=/Users/hoangvu/Code/OSS/arrow-go/parquet-testing/data ARROW_TEST_DATA=/Users/hoangvu/Code/OSS/arrow-go/arrow-testing/data go test ./... -count=1`

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -289,7 +289,15 @@ func NewSparseUnionFromArraysWithFieldCodes(typeIDs arrow.Array, children []arro
 		return nil, errors.New("arrow/array: type codes must have same length as children")
 	}
 
-	buffers := []*memory.Buffer{nil, typeIDs.Data().Buffers()[1]}
+	typeIDBuffer := typeIDs.Data().Buffers()[1]
+	if typeIDBuffer != nil {
+		typeIDBuffer = memory.SliceBuffer(typeIDBuffer,
+			arrow.Int8Traits.BytesRequired(typeIDs.Data().Offset()),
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Approving. This is the correct sparse counterpart to the dense fix in #877 — it normalizes the type-id buffer to its own logical window with `SliceBuffer` and builds the union at offset 0, so `union.setData` no longer slices the children by a stale `ArrayData` offset (which previously panicked when sliced type ids were passed with full-length children).

Verified end-to-end: the `defer Release()` is set up before the child-length check, so the error path doesn't leak; refcounts balance (`SliceBuffer` retains the parent, `NewData` retains the buffers); the new `sliced type ids` case asserts `array.Equal` + `ValidateFull()` + offset 0; and `arrow/array` passes under `-race` with no leaks. Since sparse has only the one buffer to normalize, this single case is complete coverage. Thanks for the quick follow-up!

PR #883, opened 2026-07-02
fix(parquet): align DataPageV2 pages to row boundaries in spaced writes

### LINKED ISSUE #882: parquet/pqarrow: DataPageV2 writer can split repeated byte-array/list values across page boundaries
We hit this in production with Parquet files written by a Go writer and read by Rust/DataFusion. The original bad file is large and contains PII, so it cannot be shared. The repro below is synthetic and minimal.

Repro repository: [https://github.com/punkeel/parquet-repro-1](https://github.com/punkeel/parquet-repro-1)

The repro uses Arrow Go's public writer path:

```go
pqarrow.NewFileWriter(...)
writer.WriteBuffered(record)
```

Minimal shape:

```text
c1: required list<nullable bool>

row 0: [true, null]
row 1: [true]
```

The actual null element is important: it makes `pqarrow` use the `WriteBatchSpaced` path. If the element is nullable/optional but the data has no nulls, this tiny repro does not hit the same path.

With DataPageV2 enabled, Arrow Go writes page 2 starting with repetition level `1`, i.e. inside row 0's list:

```text
OK  c1 page 01: num_rows=1 num_values=1 first_rep=0 reps=[0]
BAD c1 page 02: num_rows=1 num_values=2 first_rep=1 reps=[1 0]
```

That appears to violate the Parquet DataPageV2 rule. The spec says in [`parquet.thrift`](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift):

> Every page must begin at a row boundary (repetition_level = 0): rows must **not** be split across page boundaries when using V2 data pages.

# To Reproduce

```sh
git clone https://github.com/punkeel/parquet-repro-1
cd parquet-repro-1

go run . -out repro.parquet
go run ./cmd/inspect -in repro.parquet
```

Observed inspector output:

```text
inspecting repro.parquet
column 0: c1.list.element (max repetition level 1)
DataPageV2 pages must start at row boundaries, i.e. first_rep must be 0.
OK  c1 page 01: num_rows=1 num_values=1 first_rep=0 reps=[0]
BAD c1 page 02: num_rows=1 num_values=2 first_rep=1 reps=[1 0]

FOUND: this file contains DataPageV2 pages for repeated c1 that start with rep_level != 0.
That means those pages begin inside an existing list, not at a row boundary.
```

The checked-in `repro.parquet` is 544 bytes:

```text
sha256: 623e19d44b23c1855c225e60aa8027e0c8d1456cd51210f475643b5277b57579
```

This also reproduces with a larger batch boundary:

```sh
go run . -out /tmp/repro-bool-long.parquet -list-len 20001 -batch-size 20000
go run ./cmd/inspect -in /tmp/repro-bool-long.parquet
```

That prints:

```text
OK  c1 page 01: num_rows=1 num_values=20000 first_rep=0 reps=[0 1 1 1 1 1 1 1 1 1 1 1]
BAD c1 page 02: num_rows=1 num_values=2 first_rep=1 reps=[1 0]
```

# Expected behavior

For repeated columns written as DataPageV2, Arrow Go should only start pages at `repetition_level = 0`. If the writer cannot satisfy that for `WriteBatchSpaced`, it should either align the split to a row boundary or return an error instead of writing a non-compliant file.

# Version information

```text
github.com/apache/arrow-go/v18 v18.6.0
Go module directive: go 1.25.5
Local Go tool used for the repro: go1.26.4 darwin/arm64
```

The production writer also used:

```text
github.com/apache/iceberg-go v0.6.0
```

The repro omits Iceberg because the invalid bytes are emitted by the Arrow Go Parquet writer after the Arrow record/schema already exists.

# Related

This was originally noticed via an Arrow Rust reader failure: [apache/arrow-rs#10243](https://github.com/apache/arrow-rs/issues/10243). That issue is expected to be closed or treated as reader-side clarification because the original file is non-compliant. The remaining bug is on the Arrow Go writer side.

# Suspected source

In `github.com/apache/arrow-go/v18 v18.6.0`, the regular batch path has DataPageV2 row-boundary handling for repeated columns. The analogous `WriteBatchSpaced` path appears not to apply the same alignment. `pqarrow` uses `WriteBatchSpaced` when nullable values or nullable-parent state must be preserved, which is why the actual null in this repro is the key trigger.

[cc @truffle-dev @etseidl for visibility]

truffle-dev:
### Rationale for this change

Closes #882.

DataPageV2 requires every page to begin at a row boundary (first repetition level == 0) so that the offset index can be used for page pruning. `WriteBatch` already guarantees this: it batches through the rep-level-aware `columnWriter.doBatches`, which shrinks each batch back to the previous `rep_level == 0`.

The spaced and dictionary-index write paths did not. `WriteBatchSpaced`/`WriteBatchSpacedWithError` (non-byteArray), the boolean bitmap spaced path, and `WriteDictIndices` all split by a fixed `WriteBatchSize` using the free `doBatches` with no rep-level awareness. Because a page flush is only ever checked at a batch boundary, a repeated column written through those paths under DataPageV2 could end a page mid-row, so the next page started at `rep_level != 0`. The byteArray/fixedLenByteArray spaced paths use an adaptive size scan but had no row-boundary alignment either.

Edited in `column_writer_types.gen.go.tmpl` and regenerated `column_writer_types.gen.go`:

- Route the non-byteArray spaced, boolean bitmap spaced, and `WriteDictIndices` paths through `w.doBatches(int64(length), repLevels, ...)`. The method falls back to the identical free `doBatches` behavior for V1, `repLevels == nil`, or `MaxRepetitionLevel() == 0`, so this is a no-op except for the repeated-column DataPageV2 case. The spaced/dict actions keep their own value cursor (`valueOffset += info.numSpaced()`), so variable batch sizes are safe.
- Add the same `rep_level == 0` shrink to the byteArray and fixedLenByteArray spaced manual loops, mirroring the alignment `WriteBatch` already performs for those types.

### Are these changes tested?

Yes. New test `TestWriteBatchSpacedV2RowBoundary` writes a repeated `int32` column through `WriteBatchSpacedWithError` under DataPageV2 with a `WriteBatchSize` that is not a multiple of the row width and a small data page size, then reads every page back and asserts each DataPageV2 starts at `rep_level == 0`.

Verified red-then-green by stashing the fix:

```
--- FAIL: TestWriteBatchSpacedV2RowBoundary
    DataPageV2 #2 starts mid-row (first repetition level = 1, want 0)
```

With the fix, `./parquet/file/...` and `./parquet/pqarrow/...` pass (`PARQUET_TEST_DATA` set).

No API changes. Repeated columns written via the spaced and dictionary-index paths under DataPageV2 now produce spec-compliant pages that start on row boundaries.

---

Disclosure: I am Truffle, an AI software agent. I wrote and verified this change and understand every line.

### DIFF
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -684,25 +684,8 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
 		batchStart, batch int64
 	)
 	for batchStart = 0; batchStart+batchSize < int64(len(repLevels)); batchStart += batch {
-		// check one past the last value of the batch for if it's a new row
-		// if it's not, shrink the batch back to the beginning of a previous
-		// row boundary to end on.
-		batch = batchSize
-		for batch > 0 && repLevels[batchStart+batch] != 0 {
-			batch--
-		}
-		// if the batch shrank all the way to zero there is no row boundary
-		// at or before the requested split (a single row is wider than the
-		// batch size). growing forward to the next row boundary keeps the
-		// whole row in one batch rather than looping forever on a zero-length
-		// batch.
-		if batch == 0 {
-			batch = batchSize
-			for batchStart+batch < int64(len(repLevels)) && repLevels[batchStart+batch] != 0 {
-				batch++
-			}
-		}
 		// batchStart <--> batch now begins and ends on a row boundary!
+		batch = alignBatchToRowBoundary(repLevels, batchStart, batchSize)
 		action(batchStart, batch)
 	}
 	action(batchStart, int64(len(repLevels))-batchStart)
@@ -716,6 +699,30 @@ func doBatches(total, batchSize int64, action func(offset, batch int64)) {
 	if total%batchSize > 0 {
 		action(numBatches*batchSize, total%batchSize)
 	}
+}
+
+// alignBatchToRowBoundary adjusts batch so that repLevels[offset+batch] lands on
+// a row boundary (repetition level 0) or the end of the level slice. A repeated
+// row must never span a DataPageV2 page boundary, so it first shrinks toward the
+// previous boundary. If there is no boundary at or before the requested split -
+// the current row is wider than batch - it grows forward to the next one so the
+// whole row stays in a single batch and the caller keeps making progress rather
+// than looping on a zero-length batch. offset must already sit on a boundary.
+func alignBatchToRowBoundary(repLevels []int16, offset, batch int64) int64 {
+	n := int64(len(repLevels))
+	if offset+batch >= n {
+		return batch
+	}
+	b := batch
+	for b > 0 && repLevels[offset+b] != 0 {
+		b--
+	}
+	if b > 0 {
+		return b
+	}
+	for b = batch; offset+b < n && repLevels[offset+b] != 0; b++ {
+	}
+	return b
 }
 
 func levelSliceOrNil(rep []int16, offset, batch int64) []int16 {
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -685,10 +685,22 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
 	)
 	for batchStart = 0; batchStart+batchSize < int64(len(repLevels)); batchStart += batch {
 		// check one past the last value of the batch for if it's a new row
-		// if it's not, shrink the batch and feel back to the beginning of a
-		// previous row boundary to end on
+		// if it's not, shrink the batch back to the beginning of a previous
+		// row boundary to end on.
 		batch = batchSize
-		for ; repLevels[batchStart+batch] != 0; batch-- {
+		for batch > 0 && repLevels[batchStart+batch] != 0 {
+			batch--
+		}
+		// if the batch shrank all the way to zero there is no row boundary
+		// at or before the requested split (a single row is wider than the
+		// batch size). growing forward to the next row boundary keeps the
+		// whole row in one batch rather than looping forever on a zero-length
+		// batch.
+		if batch == 0 {
+			batch = batchSize
+			for batchStart+batch < int64(len(repLevels)) && repLevels[batchStart+batch] != 0 {
+				batch++
+			}
 		}
 		// batchStart <--> batch now begins and ends on a row boundary!
 		action(batchStart, batch)
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -657,7 +657,12 @@ func (w *columnWriter) Close() (err error) {
 func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(offset, batch int64)) {
 	batchSize := w.props.WriteBatchSize()
 	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
-	// use the regular doBatches function
+	// use the regular doBatches function.
+	//
+	// NOTE: the spec also requires row-aligned pages for V1 when an OffsetIndex
+	// is present (PageIndexEnabled). That gap predates this change and also
+	// affects the WriteBatch path, so it is tracked separately rather than
+	// widened into this DataPageV2 fix.
 	if w.props.DataPageVersion() == parquet.DataPageV1 || repLevels == nil || w.descr.MaxRepetitionLevel() == 0 {
 		doBatches(total, batchSize, action)
 		return
@@ -675,6 +680,17 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
 		panic("columnwriter: not enough repetition levels for batch to write")
 	}
 
+	// nothing to write: a non-nil but empty repLevels slice would index
+	// repLevels[0] below and panic (recovered as an opaque error).
+	if total == 0 {
+		return
+	}
+
+	// drive the loop over exactly the number of levels we were asked to write.
+	// a caller passing a repLevels slice longer than total must not spill the
+	// extra levels into the column.
+	repLevels = repLevels[:total]
+
 	if repLevels[0] != 0 {
 		panic("columnwriter: batch writing for V2 data pages must start at a row boundary")
 	}
@@ -688,7 +704,11 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
 		batch = alignBatchToRowBoundary(repLevels, batchStart, batchSize)
 		action(batchStart, batch)
 	}
-	action(batchStart, int64(len(repLevels))-batchStart)
+	// a grow-to-end alignment on an oversized final row can leave batchStart
+	// already at len(repLevels); skip the trailing zero-length action.
+	if batchStart < int64(len(repLevels)) {
+		action(batchStart, int64(len(repLevels))-batchStart)
+	}
 }
 
 func doBatches(total, batchSize int64, action func(offset, batch int64)) {
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -659,10 +659,10 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
 	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
 	// use the regular doBatches function.
 	//
-	// NOTE: the spec also requires row-aligned pages for V1 when an OffsetIndex
-	// is present (PageIndexEnabled). That gap predates this change and also
-	// affects the WriteBatch path, so it is tracked separately rather than
-	// widened into this DataPageV2 fix.
+	// The spec also requires row-aligned pages for V1 when an OffsetIndex is
+	// present (PageIndexEnabled). That gap is on the WriteBatch path and is
+	// out of scope for the DataPageV2 row alignment here; it is tracked in
+	// https://github.com/apache/arrow-go/issues/887.
 	if w.props.DataPageVersion() == parquet.DataPageV1 || repLevels == nil || w.descr.MaxRepetitionLevel() == 0 {
 		doBatches(total, batchSize, action)
 		return
--- a/parquet/file/column_writer_types.gen.go.tmpl
+++ b/parquet/file/column_writer_types.gen.go.tmpl
@@ -214,6 +214,8 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
   const maxSafeBatchDataSize int64 = 1 << 30 // 1GB
 
   batchSize := w.props.WriteBatchSize()
+  isV2WithRep := w.props.DataPageVersion() != parquet.DataPageV1 &&
+    repLevels != nil && w.descr.MaxRepetitionLevel() > 0
   levelOffset := int64(0)
   n := int64(length)
 
@@ -234,6 +236,14 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
           break
         }
         cumDataSize += valSize
+      }
+    }
+
+    // V2 row-boundary alignment: a repeated row must not span a page
+    // boundary, so shrink the batch back to the previous rep_level == 0.
+    if isV2WithRep && levelOffset+batch < n {
+      for batch > 1 && repLevels[levelOffset+batch] != 0 {
+        batch--
       }
     }
     if batch < 1 {
@@ -262,7 +272,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
     levelOffset += batch
   }
 {{- else}}
-  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+  w.doBatches(int64(length), repLevels, func(offset, batch int64) {
     var vals []{{.name}}
     info := w.maybeCalculateValidityBits(levelSliceOrNil(defLevels, offset, batch), batch)
 
@@ -347,7 +357,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBitmapBatchSpacedWithError(bitmap []by
     length = numValues
   }
 
-  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+  w.doBatches(int64(length), repLevels, func(offset, batch int64) {
     info := w.maybeCalculateValidityBits(levelSliceOrNil(defLevels, offset, batch), batch)
 
     w.writeLevelsSpaced(batch, levelSliceOrNil(defLevels, offset, batch), levelSliceOrNil(repLevels, offset, batch))
@@ -384,7 +394,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteDictIndices(indices arrow.Array, defLe
 
   dictEncoder := w.currentEncoder.(encoding.DictEncoder)
 
-  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+  w.doBatches(int64(length), repLevels, func(offset, batch int64) {
     info := w.maybeCalculateValidityBits(levelSliceOrNil(defLevels, offset, batch), batch)
     w.writeLevelsSpaced(batch, levelSliceOrNil(defLevels, offset, batch), levelSliceOrNil(repLevels, offset, batch))
 
--- a/parquet/file/column_writer_types.gen.go.tmpl
+++ b/parquet/file/column_writer_types.gen.go.tmpl
@@ -125,10 +125,20 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatch(values []{{.name}}, defLevels, r
       }
     }
 
-    // V2 row-boundary alignment
+    // V2 row-boundary alignment: a repeated row must not span a page
+    // boundary. Shrink the batch back to the previous rep_level == 0; if
+    // there is no row boundary at or before the requested split (a single
+    // row is wider than the batch), grow forward to the next one so the
+    // whole row lands in one batch instead of looping without progress.
     if isV2WithRep && levelOffset+batch < n {
-      for batch > 1 && repLevels[levelOffset+batch] != 0 {
+      for batch > 0 && repLevels[levelOffset+batch] != 0 {
         batch--
+      }
+      if batch == 0 {
+        batch = batchSize
+        for levelOffset+batch < n && repLevels[levelOffset+batch] != 0 {
+          batch++
+        }
       }
     }
     if batch < 1 {
@@ -240,10 +250,19 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
     }
 
     // V2 row-boundary alignment: a repeated row must not span a page
-    // boundary, so shrink the batch back to the previous rep_level == 0.
+    // boundary. Shrink the batch back to the previous rep_level == 0; if
+    // there is no row boundary at or before the requested split (a single
+    // row is wider than the batch), grow forward to the next one so the
+    // whole row lands in one batch instead of looping without progress.
     if isV2WithRep && levelOffset+batch < n {
-      for batch > 1 && repLevels[levelOffset+batch] != 0 {
+      for batch > 0 && repLevels[levelOffset+batch] != 0 {
         batch--
+      }
+      if batch == 0 {
+        batch = batchSize
+        for levelOffset+batch < n && repLevels[levelOffset+batch] != 0 {
+          batch++
+        }
       }
     }
     if batch < 1 {
--- a/parquet/file/column_writer_types.gen.go.tmpl
+++ b/parquet/file/column_writer_types.gen.go.tmpl
@@ -125,21 +125,11 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatch(values []{{.name}}, defLevels, r
       }
     }
 
-    // V2 row-boundary alignment: a repeated row must not span a page
-    // boundary. Shrink the batch back to the previous rep_level == 0; if
-    // there is no row boundary at or before the requested split (a single
-    // row is wider than the batch), grow forward to the next one so the
-    // whole row lands in one batch instead of looping without progress.
-    if isV2WithRep && levelOffset+batch < n {
-      for batch > 0 && repLevels[levelOffset+batch] != 0 {
-        batch--
-      }
-      if batch == 0 {
-        batch = batchSize
-        for levelOffset+batch < n && repLevels[levelOffset+batch] != 0 {
-          batch++
-        }
-      }
+    // V2 row-boundary alignment: a repeated row must not span a page boundary,
+    // so snap the batch onto the nearest row boundary (or keep a single wide row
+    // whole). See alignBatchToRowBoundary in column_writer.go.
+    if isV2WithRep {
+      batch = alignBatchToRowBoundary(repLevels, levelOffset, batch)
     }
     if batch < 1 {
       batch = 1
@@ -249,21 +239,11 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
       }
     }
 
-    // V2 row-boundary alignment: a repeated row must not span a page
-    // boundary. Shrink the batch back to the previous rep_level == 0; if
-    // there is no row boundary at or before the requested split (a single
-    // row is wider than the batch), grow forward to the next one so the
-    // whole row lands in one batch instead of looping without progress.
-    if isV2WithRep && levelOffset+batch < n {
-      for batch > 0 && repLevels[levelOffset+batch] != 0 {
-        batch--
-      }
-      if batch == 0 {
-        batch = batchSize
-        for levelOffset+batch < n && repLevels[levelOffset+batch] != 0 {
-          batch++
-        }
-      }
+    // V2 row-boundary alignment: a repeated row must not span a page boundary,
+    // so snap the batch onto the nearest row boundary (or keep a single wide row
+    // whole). See alignBatchToRowBoundary in column_writer.go.
+    if isV2WithRep {
+      batch = alignBatchToRowBoundary(repLevels, levelOffset, batch)
     }
     if batch < 1 {
       batch = 1
--- a/parquet/file/column_writer_v2_row_boundary_test.go
+++ b/parquet/file/column_writer_v2_row_boundary_test.go
@@ -18,12 +18,18 @@ package file_test
 
 import (
 	"bytes"
+	"fmt"
 	"testing"
+	"time"
 
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/parquet"
 	"github.com/apache/arrow-go/v18/parquet/compress"
 	"github.com/apache/arrow-go/v18/parquet/file"
 	"github.com/apache/arrow-go/v18/parquet/internal/encoding"
+	"github.com/apache/arrow-go/v18/parquet/pqarrow"
 	"github.com/apache/arrow-go/v18/parquet/schema"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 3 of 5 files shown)


punkeel:
Thanks for working on this. I tested the PR locally against the minimized `pqarrow` repro in [punkeel/parquet-repro-1](https://github.com/punkeel/parquet-repro-1).

I tested PR head `31b0bf60` with the repro module using:

```go
replace github.com/apache/arrow-go/v18 => /tmp/arrow-go-pr883
```

I do not think this fully fixes the issue yet. I see two remaining failure modes.

## 1. Variable-width / byte-array-like nullable list elements still write invalid DataPageV2 pages

These cases still produce a `DataPageV2` page that starts with `repetition_level = 1`, i.e. the page starts inside an existing list instead of at a row boundary:

```text
c1: required list<nullable T>

row 0: [value, null]
row 1: [value]
```

Writer settings:

```text
DataPageV2
data page size = 1
batch size = 1
dictionary disabled
uncompressed
```

Types I tested that still fail:

- `string`
- `binary`
- `fixed_size_binary(1)`

For `list<nullable string>`, the generated pages are:

```text
OK  page 1: rows=1 values=1 first_rep=0 reps=[0]
BAD page 2: rows=1 values=2 first_rep=1 reps=[1 0]
```

So page 2 is still not row-aligned. It begins by continuing row 0's list, then reaches row 1.

DuckDB can read the generated file successfully:

```text
[x, NULL]
[tail]
```

I also tested a less tiny long-list variant to make sure this is not only a `batch-size=1` artifact:

```sh
go run . -out /tmp/repro-pr883-string-long.parquet -list-len 20001 -batch-size 20000
go run ./cmd/inspect -in /tmp/repro-pr883-string-long.parquet
```

That still produces a non-row-aligned V2 page:

```text
OK  page 1: rows=1 values=1 first_rep=0 reps=[0]
BAD page 2: rows=0 values=20000 first_rep=1 reps=[1 1 1 1 ...]
OK  page 3: rows=1 values=1 first_rep=0 reps=[0]
```

## 2. Fixed-width nullable list elements appear to hang / make no progress

For the same shape:

```text
c1: required list<nullable T>

row 0: [value, null]
row 1: [value]
```

the writer did not return within 2 seconds for:

- `bool`
- `int32`
- `float64`

This looks like a no-progress loop in the fixed-width `WriteBatchSpaced` path when the requested split would be inside a row/list rather than at a row boundary. My guess is that the batching code can shrink the current batch to zero when there is no row boundary before the requested split, so the callback keeps retrying without consuming any values. I have not fully traced that part, but the externally visible behavior is that these small repros do not complete.

## Tests I think are worth adding to this PR

I think the PR should cover both "still writes invalid pages" and "hangs / no progress".

Suggested small tests:

1. `list<nullable string>` with rows `[x, null]`, `[tail]`
   - Use DataPageV2, small data page size, `WithBatchSize(1)`, dictionary disabled.
   - Assert every DataPageV2 page starts with `rep_level == 0`.
   - This currently still fails because page 2 starts with `rep_level == 1`.

2. `list<nullable binary>` with rows `[[]byte("x"), null]`, `[[]byte("y")]`
   - Same assertion.
   - This checks the byte-array path without UTF8/string.

3. `list<nullable fixed_size_binary(1)>` with rows `[0x01, null]`, `[0x02]`
   - Same assertion.
   - This checks the fixed-size byte-array path.

4. `list<nullable float64>` with rows `[1.5, null]`, `[2.5]`
   - Use DataPageV2, small data page size, `WithBatchSize(1)`, dictionary disabled.
   - Assert the writer returns.
   - Then assert every DataPageV2 page starts with `rep_level == 0`.

5. Optional but useful: a longer `list<nullable string>` case, e.g. row 0 has `20001` elements and the writer uses `WithBatchSize(20000)`.
   - This is less minimal, but it shows the issue is not only caused by the artificial `batch-size=1` setup.

One important debugging detail: the trigger is not merely that the list element type is nullable/optional. The data needs an actual null element. With no null values, `pqarrow` can take a different path and the issue may not reproduce. The actual null sends the leaf writer through the `WriteBatchSpaced` path, which is where these remaining cases seem to be.

truffle-dev:
Thanks for the thorough repro, @punkeel — you nailed the root cause. Both failure modes share one gap: when a single repeated row is wider than the write batch size, the shrink-to-row-boundary step has no earlier boundary to back off to. Fixed in d421e980.

1. **No-progress hang (fixed-width elements).** In `columnWriter.doBatches` the shrink loop hit `batch == 0` (as you guessed), and `action(batchStart, 0)` left `batchStart` unchanged, so the callback kept retrying without consuming any values. It now grows the batch forward to the next `rep_level == 0` when the shrink can't reach one, keeping the whole oversized row in a single batch.

2. **Mid-row DataPageV2 (byte-array elements).** The ByteArray/FixedLenByteArray spaced manual loops clamped at `batch = 1` instead of hanging, so they still emitted a page starting at `rep_level = 1`. Same shrink-then-grow fix applied there (in the template + regenerated `.gen.go`).

Tests in `column_writer_v2_row_boundary_test.go` now cover nullable list elements of `string`, `binary`, `fixed_size_binary(1)` and `float64` with rows wider than the batch size, plus a single list of length 20001 with batch size 20000. Every row ends with an actual null element so the leaf routes through `WriteBatchSpaced`, per your note that the null is the real trigger. The write runs under a 30s timeout, so a regression to the no-progress loop fails the test rather than hanging the suite. Red-green verified by stashing the source fix: the three byte-array cases assert a mid-row page and the two fixed-width cases time out; all pass with the fix.

`parquet/file` and `parquet/pqarrow` suites are green with `PARQUET_TEST_DATA` set.

Analysis and fix by an autonomous AI agent; traced against source, verified red-green, and I can own follow-ups.

punkeel:
Can the for loops computing batch size and co be simplified further? Is there already a helper for that in the codebase? Or in the Java parquet library, maybe?

truffle-dev:
Good call - the three copies (the `doBatches` method plus the ByteArray/FixedLenByteArray manual loops) were duplicating the same alignment, so I pulled them into one `alignBatchToRowBoundary(repLevels, offset, batch)` in `column_writer.go`. There wasn't an existing helper; the loops grew up independently.

Collapsing them also fixed a small inconsistency: the generated loops restarted the forward grow from `WriteBatchSize` instead of the value-scanned batch, so they could overshoot a nearer row boundary. The helper grows from the requested batch in all callers now.

I didn't lift anything from parquet-mr - Java doesn't split a write into batches like this. `ColumnWriteStoreBase` only size-checks (and flushes a page) at record boundaries via `endRecord()`, so the row-boundary guarantee falls out of the structure there rather than a reusable helper. This helper is the equivalent within the Go writer's batch model.

Suites still green with `PARQUET_TEST_DATA` set, and the null-element regression tests stay red when the helper is neutered to a no-op. 037b4513.

punkeel (approved):

punkeel:
Code LGTM, with zero prior experience in this codebase.

My local tests/reproductions seem to be working fine now.


----

@truffle-dev please write a few more unit tests for `alignBatchToRowBoundary`, with edge cases in mind (0, 1, large value).

truffle-dev:
Added `TestAlignBatchToRowBoundary` in `column_writer_align_test.go` - table-driven, twelve cases. Covers the edges you called out: batch 0 (stays 0, callers clamp to 1), batch 1 both on- and mid-row, and a large batch both landing on a boundary and shrinking back across several rows. Plus already-on-boundary, shrink to the previous boundary, grow forward when a single row is wider than the batch (including growing to the end of the slice), the tail cases where offset+batch reaches or passes the end, and a non-zero offset. Each case also asserts the returned batch ends on a row boundary.

Confirmed load-bearing: neuter the helper to `return batch` and the six shrink/grow cases go red while the pass-through cases stay green. 7c3c7879.

parquet/file/column_writer.go
punkeel, line 694: I'm surprised this didn't catch the bug. @truffle-dev is this assertion in the wrong spot? do we need this assertion in more places?

parquet/file/column_writer.go
```diff
@@ -684,13 +684,8 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
...
 	)
 	for batchStart = 0; batchStart+batchSize < int64(len(repLevels)); batchStart += batch {
-		// check one past the last value of the batch for if it's a new row
-		// if it's not, shrink the batch and feel back to the beginning of a
-		// previous row boundary to end on
-		batch = batchSize
-		for ; repLevels[batchStart+batch] != 0; batch-- {
-		}
 		// batchStart <--> batch now begins and ends on a row boundary!
+		batch = alignBatchToRowBoundary(repLevels, batchStart, batchSize)
```
zeroshade, line 704: Three hardening items around this loop:

1. **`total` vs `len(repLevels)`** — this branch iterates on `len(repLevels)` and only uses `total` in the guard above. If a caller passes a `repLevels` slice longer than `total`, it will write more levels than requested. Consider `repLevels = repLevels[:total]` right after the `len(repLevels) < total` check (or require exact equality).
2. **`total == 0`** with a non-nil but empty `repLevels` hits the earlier `repLevels[0]` guard and panics (recovered as an opaque error). An early `if total == 0 { return }` avoids it.
3. **Zero-length tail** — when a single oversized row makes `alignBatchToRowBoundary` grow to the end, the trailing `action(batchStart, int64(len(repLevels))-batchStart)` can fire with a 0-length batch. Guard with `if batchStart < int64(len(repLevels))`.

parquet/file/column_writer_types.gen.go.tmpl
```diff
@@ -384,7 +393,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteDictIndices(indices arrow.Array, defLe
 
   dictEncoder := w.currentEncoder.(encoding.DictEncoder)
 
-  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+  w.doBatches(int64(length), repLevels, func(offset, batch int64) {
```
zeroshade, line 396: `WriteDictIndices` is now correctly routed through the rep-aware `w.doBatches`, but I don't see a test exercising this path under DataPageV2 with a repeated column — all three new tests set `WithDictionaryDefault(false)`, so the dictionary path is never hit. Could we add one?

Note from testing: dictionary columns rarely produce more than one data page (page sizing uses `DictEncoder.ObservedRawSize()`), so forcing a multi-page split likely needs many distinct values and/or a high `WithDictionaryPageSizeLimit`.

parquet/file/column_writer_types.gen.go.tmpl
```diff
@@ -347,7 +356,7 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBitmapBatchSpacedWithError(bitmap []by
     length = numValues
   }
 
-  doBatches(int64(length), w.props.WriteBatchSize(), func(offset, batch int64) {
+  w.doBatches(int64(length), repLevels, func(offset, batch int64) {
```
zeroshade, line 359: Same here — the boolean bitmap spaced path is rerouted through `w.doBatches`, but there's no DataPageV2 + repeated boolean test covering it. Worth a small regression test.

parquet/file/column_writer_types.gen.go.tmpl
```diff
@@ -236,6 +238,13 @@ func (w *{{.Name}}ColumnChunkWriter) WriteBatchSpacedWithError(values []{{.name}
         cumDataSize += valSize
       }
     }
+
+    // V2 row-boundary alignment: a repeated row must not span a page boundary,
+    // so snap the batch onto the nearest row boundary (or keep a single wide row
+    // whole). See alignBatchToRowBoundary in column_writer.go.
+    if isV2WithRep {
+      batch = alignBatchToRowBoundary(repLevels, levelOffset, batch)
```
zeroshade, line 246: Non-blocking: after the 1GB `maxSafeBatchDataSize` scan above, `alignBatchToRowBoundary` can grow the batch *forward* past that cap when a single row is wider than it. That's unavoidable under V2 (a row can't span pages), and `FlushCurrentPage` does guard `uncompressed > MaxInt32` — but the failure comes late (after encoding a huge buffer), and per-value `BYTE_ARRAY` lengths are cast to `uint32`. Consider an early, explicit error when an aligned single row can't fit Parquet's int32 page/value-size limits.

parquet/file/column_writer_v2_row_boundary_test.go
```diff
@@ -0,0 +1,324 @@
...
+	require.NoError(t, rgw.Close())
+	require.NoError(t, w.Close())
+
+	assertPagesStartOnRowBoundary(t, buf.Bytes())
+}
+
+// assertPagesStartOnRowBoundary reads the first column chunk back and requires
+// that every DataPageV2 begins at a row boundary (its first repetition level is
+// 0). The test forces multiple pages, so a mid-row split will surface here.
+func assertPagesStartOnRowBoundary(t *testing.T, raw []byte) {
```
zeroshade, line 110: This helper (and the tests using it) only decode repetition levels — they never compare read-back values against the input. Given the fix makes batch sizes variable (a value-cursor hazard), a value round-trip check on at least one repeated case would materially strengthen these tests. I ran one for `list<list<int32>>` and it round-trips correctly, so this is about locking it in.

parquet/file/column_writer_v2_row_boundary_test.go
```diff
@@ -0,0 +1,324 @@
...
+		v2Count++
+		var dec encoding.LevelDecoder
+		require.NoError(t, dec.SetDataV2(v2.RepetitionLevelByteLen(), maxRep, int(v2.NumValues()), v2.Data()))
+		levels := make([]int16, v2.NumValues())
+		n, _ := dec.Decode(levels)
+		require.Greater(t, n, 0)
+		require.Zero(t, levels[0], "wide row page starts mid-row")
+	}
+	require.NoError(t, pr.Err())
+	require.Equal(t, 1, v2Count, "a single wide row must occupy exactly one DataPageV2")
```
zeroshade, line 233: Heads-up: this test passes even with the fix reverted. A single ~160KB row fits in one page regardless of alignment, and a single row can't produce a mid-row page — so the `v2Count == 1` / `first_rep == 0` assertions here are effectively decorative. The real regression guard is the 30s timeout against the no-progress hang. Consider documenting it as a hang guard, or shaping the data so a mid-row split is actually possible.

parquet/file/column_writer_v2_row_boundary_test.go
```diff
@@ -0,0 +1,324 @@
...
+// fails the test instead of hanging the whole suite.
+func writeNullableListV2WithinTimeout(t *testing.T, elemType arrow.DataType, listNullable bool, rowWidths []int, batchSize, pageSize int64, timeout time.Duration) []byte {
+	t.Helper()
+
+	type result struct {
+		b   []byte
+		err error
+	}
+	done := make(chan result, 1)
+	go func() {
```
zeroshade, line 253: Minor: if the batching ever regresses to the no-progress loop, this goroutine leaks and spins a CPU forever after `t.Fatalf` returns (the test binary keeps running the rest of the suite). Not critical since the timeout still fails the test, but passing a cancellation signal into the write would avoid the leak.

zeroshade (changes requested):
Thanks for this — the core DataPageV2 fix is sound, and I verified it locally: `go generate ./parquet/file/...` reproduces the committed `.gen.go` exactly, the new regression tests are genuinely red-green (neutering `alignBatchToRowBoundary` to `return batch` makes `TestWriteBatchSpacedV2RowBoundary` and all four `TestWriteSpacedV2NullableListRowBoundary` subtests fail with "starts mid-row"), and `./parquet/file/...` + `./parquet/pqarrow/...` pass. Nice catch that the fix also repairs per-page `num_rows` and the OffsetIndex `first_row_index`, not just the first repetition level.

Requesting changes for a few items before merge — mostly hardening and coverage for paths this PR touches, plus a decision on a closely-related V1 case. Specifics are inline; summary here.

### 1. Correctness hardening in `columnWriter.doBatches`
- **`total` vs `len(repLevels)`**: the V2 branch drives its loop off `len(repLevels)` and uses `total` only in the guard. If a caller ever passes a `repLevels` slice longer than `total`, it writes more levels than requested. Suggest `repLevels = repLevels[:total]` after the existing `< total` check (or require exact equality).
- **`total == 0`** with a non-nil empty `repLevels` indexes `repLevels[0]` and panics (recovered as an opaque error). An early `if total == 0 { return }` is cleaner.
- **Zero-length tail**: after a grow-to-end on an oversized row, the trailing `action(batchStart, len-batchStart)` can fire with a 0-length batch; guard with `if batchStart < int64(len(repLevels))`.

### 2. Test coverage for modified-but-untested paths
- `WriteDictIndices` and the boolean `WriteBitmapBatchSpacedWithError` are both rerouted through `w.doBatches` here, but no test exercises them under V2 + a repeated column (all three new tests set `WithDictionaryDefault(false)`; none use bool). Please add coverage.
- The boundary tests only decode repetition levels — none compare read-back values to the input. Since this change makes batch sizes variable (a value-cursor hazard), please add a value round-trip assertion for at least one repeated case. I ran one for `list<list<int32>>` (maxRep=2) and it round-trips correctly, so this is about locking it in.

### 3. Related spec gap: V1 + OffsetIndex (please decide)
The spec requires row-aligned pages for V1 too *when an OffsetIndex is present* ("If an OffsetIndex is present, a page must begin at a row boundary"; and "Page indexes require pages to start and end at row boundaries, regardless of which page header is used"). `doBatches` falls back to the unaligned free `doBatches` for **all** V1, regardless of `PageIndexEnabled`. I confirmed empirically: a repeated column written as V1 + `WithPageIndexEnabled(true)` + small pages produced **12 of 16** data pages beginning mid-row — i.e. an invalid OffsetIndex. This predates this PR and affects `WriteBatch` too, so it may belong in a separate issue — but since this PR is about exactly this invariant, it would be ideal to either handle it here (align when `DataPageVersion() == V2 || PageIndexEnabledFor(col)`) or file a linked follow-up.

### Non-blocking nits (inline)
- Oversized single row can grow past the 1GB byte-array cap (unavoidable under V2, but an early explicit error beats a late large-allocation failure).
- The wide-row test's assertions are effectively decorative (it passes even with the fix reverted); the real guard is the timeout, and the helper leaks a spinning goroutine on a hang regression.

The helper extraction is clean and the template/generated pair stays in sync — thanks again.

truffle-dev:
Thanks both. All in `c60d0313`.

**Hardening (`doBatches`)**
1. `total == 0` returns early before the `repLevels[0]` index — `c60d0313`
2. `repLevels = repLevels[:total]` after the length check, so an over-long slice can't spill extra levels — `c60d0313`
3. trailing `action` guarded by `batchStart < len(repLevels)`, skipping the zero-length tail after a grow-to-end — `c60d0313`

**Coverage**
- `TestWriteBitmapBatchSpacedV2RowBoundary` — boolean spaced path, V2 + repeated. Your instinct was right that the first cut was weak: `pageSize=256` coincidentally aligned the flushes, so it passed even reverted. Dropped it to 64; now 12/16 pages start mid-row without the fix — `c60d0313`
- `TestWriteDictIndicesV2RowBoundary` — dict path via pqarrow, low-cardinality `list<int32>` under page pressure so the indices span multiple data pages. Red-green — `c60d0313`
- `TestWriteSpacedV2NestedListRoundTrip` — `list<list<int32>>` (maxRep=2) value round-trip through pqarrow read-back, locking in the variable batch sizes — `c60d0313`

**Wide-row test** — reworded as a hang guard: one row can't split, so the boundary asserts are decorative and the 30s timeout is the real guard. Noted the goroutine leak is bounded to the failing run (the writer exposes no cancellation hook) — `c60d0313`

**V1 + OffsetIndex** — reproduced it (16 pages, 12 mid-row) and filed #887. It also hits `WriteBatch`, so I kept it out of this PR to keep the V2 fix landable; glad to fold it in here instead if you'd prefer.

**1GB cap** — left as-is for now; `FlushCurrentPage`'s `MaxInt32` guard still catches it, just late. Can add the early explicit error here or in the #887 follow-up, your call.

@punkeel on the `:678` assertion — it's an entry precondition. The first `doBatches` call always starts at level 0, so it only checks that the whole call begins on a boundary. The bug was sub-batches *ending* mid-row inside the loop (a fixed split), which this assertion can't see. That per-batch boundary is now guaranteed by construction in `alignBatchToRowBoundary` rather than asserted, and the red-green tests lock it in — so I kept the one entry check instead of scattering more.

parquet/file/column_writer.go
```diff
@@ -657,7 +657,12 @@ func (w *columnWriter) Close() (err error) {
 func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(offset, batch int64)) {
 	batchSize := w.props.WriteBatchSize()
 	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
-	// use the regular doBatches function
+	// use the regular doBatches function.
+	//
+	// NOTE: the spec also requires row-aligned pages for V1 when an OffsetIndex
+	// is present (PageIndexEnabled). That gap predates this change and also
+	// affects the WriteBatch path, so it is tracked separately rather than
```
punkeel: "separately" -> link where it's tracked, or issue number.

parquet/file/column_writer.go
```diff
@@ -657,7 +657,12 @@ func (w *columnWriter) Close() (err error) {
 func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(offset, batch int64)) {
 	batchSize := w.props.WriteBatchSize()
 	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
-	// use the regular doBatches function
+	// use the regular doBatches function.
+	//
+	// NOTE: the spec also requires row-aligned pages for V1 when an OffsetIndex
+	// is present (PageIndexEnabled). That gap predates this change and also
```
punkeel: "this change" is not a good comment in code

parquet/file/column_writer.go
```diff
@@ -684,16 +700,15 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
...
-		}
 		// batchStart <--> batch now begins and ends on a row boundary!
+		batch = alignBatchToRowBoundary(repLevels, batchStart, batchSize)
 		action(batchStart, batch)
 	}
-	action(batchStart, int64(len(repLevels))-batchStart)
+	// a grow-to-end alignment on an oversized final row can leave batchStart
+	// already at len(repLevels); skip the trailing zero-length action.
+	if batchStart < int64(len(repLevels)) {
+		action(batchStart, int64(len(repLevels))-batchStart)
```
punkeel, line 710: is that line covered by a test?

truffle-dev:
Thanks @punkeel.

1. Dropped "this change" from the note — contentless once it's in history — `ab6fa342`
2. The V1/OffsetIndex gap is tracked in #887; the note links it now instead of hand-waving "separately" — `ab6fa342`
3. Yes, covered. The trailing `action` (the final partial batch) runs in `TestWriteBatchSpacedV2RowBoundary`. The guard on line 709 that *skips* a zero-length trailing action is exercised by `TestWriteSpacedV2NullableListWideRow`: a single row wider than the batch (20001 vals, batch 20000) grows to the row end, leaving `batchStart == len(repLevels)`, so the trailing call is skipped and it asserts exactly one DataPageV2. `go test -covermode=set` shows both the branch condition and the trailing block execute.

parquet/file/column_writer.go
truffle-dev: Not the wrong spot, just narrower than it looks. It's a precondition on the batch handed to `doBatches`, and `WriteBatchSpaced` always enters on a fresh row, so `repLevels[0]` is always 0 here — it can't trip on this bug. The mid-row split happened a layer down: `commitWriteAndCheckPageLimit` flushes once buffered size crosses `DataPageSize`, and pre-fix the batches weren't row-aligned, so a page could close mid-row. This entry check never sees the per-page split.

The fix makes that invariant constructive rather than asserted — `alignBatchToRowBoundary` hands every `action()` a batch that begins and ends on a boundary, so each flush lands on one by construction. No broader runtime assertion is needed for correctness.

If you'd want defense-in-depth against a future regression in the helper, the spot is `FlushCurrentPage` just before `buildDataPageV2`: assert the flushed page's first repetition level is 0 and `numBufferedRows` matches the row-start count. That guards the boundary where the split actually happens instead of relying on the row-boundary tests. Glad to add it if you'd like it in the tree.

punkeel:
I’ve been running 7c3c7879dd4e91a328d69afdc8a5fcbcff4bf414 since it got added here, seems to be running smoothly. Not noticing data corruption / panics so far

truffle-dev:
Appreciate the prod datapoint, @punkeel. The spaced-write path this touches is exactly where a misaligned DataPageV2 would surface as silent corruption rather than a loud failure, so a clean run against real data is a good signal. Thanks for keeping it on the branch.

zeroshade (approved):
Re-review of `ab6fa342`. All items from my earlier review are addressed, and I re-verified locally (go1.26.4).

**Verified locally**
- `go generate ./parquet/file/...` reproduces the committed `column_writer_types.gen.go` with **no diff** — template and generated stay in sync.
- `./parquet/file/...` builds clean; the targeted suite is green: `TestAlignBatchToRowBoundary` (12 cases), `TestWriteBatchSpacedV2RowBoundary`, `TestWriteSpacedV2NullableList{RowBoundary,WideRow}`, plus the three new ones below.
- Red-green still holds: neutering `alignBatchToRowBoundary` to `return batch` fails every boundary test with `starts mid-row` (and the shrink/grow unit cases), so the new coverage is genuinely load-bearing — including the bool and dict paths that were previously untested.

**Prior asks → resolved**
1. **`doBatches` hardening** — all three in place: `total == 0` early return before the `repLevels[0]` index; `repLevels = repLevels[:total]` so an over-long slice can't spill extra levels; and the `batchStart < len(repLevels)` guard so a grow-to-end on an oversized final row doesn't fire a zero-length trailing action.
2. **Coverage for the rerouted paths** — `TestWriteBitmapBatchSpacedV2RowBoundary` (bool; `pageSize=64` so flushes don't coincidentally align), `TestWriteDictIndicesV2RowBoundary` (dict indices spanning multiple data pages), and value round-trips via that dict test + `TestWriteSpacedV2NestedListRoundTrip` (`list<list<int32>>`, maxRep=2). All confirmed red-green, which also locks in the variable batch sizes the fix introduces.
3. **V1 + OffsetIndex** — filed as #887, and the code comment now cites it instead of hand-waving "separately". Keeping it out of this PR is the right call: it also affects `WriteBatch`, and this PR cleanly lands the DataPageV2 invariant. (Also picks up @punkeel's "this change" / "separately" comment fixes.)

**Non-blocking, fine as follow-ups**
- An oversized single row can still grow past the 1 GB `maxSafeBatchDataSize` cap (a row can't span a V2 page, so this is inherent); the late `FlushCurrentPage` `MaxInt32` guard remains the backstop. An early explicit error would be nicer — reasonable to fold into #887.
- The wide-row test is now documented as a hang guard (its page asserts can't catch a single-row mid-row split; the 30s timeout is the real guard).

The helper extraction is clean and the template/generated pair is consistent. Thanks for the thorough iteration and the red-green discipline throughout. LGTM — approving.

truffle-dev:
Thanks @zeroshade. Both non-blocking notes are tracked in #887: it already carries the V1 + OffsetIndex path, and I've added the oversized-single-row cap follow-up there so it isn't lost. Appreciate the careful re-review.

parquet/file/column_writer.go
```diff
@@ -675,6 +680,17 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
...
+	// nothing to write: a non-nil but empty repLevels slice would index
+	// repLevels[0] below and panic (recovered as an opaque error).
+	if total == 0 {
+		return
+	}
+
+	// drive the loop over exactly the number of levels we were asked to write.
+	// a caller passing a repLevels slice longer than total must not spill the
+	// extra levels into the column.
+	repLevels = repLevels[:total]
```
zeroshade, line 692: Hardening items 1 & 2 ✅ — `total == 0` returns before the `repLevels[0]` index below, and `repLevels = repLevels[:total]` stops a caller's over-long slice from spilling extra levels into the column. Both correctly sit after the `len(repLevels) < total` guard.

parquet/file/column_writer.go
```diff
@@ -684,16 +700,15 @@ func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(off
...
-		for ; repLevels[batchStart+batch] != 0; batch-- {
-		}
 		// batchStart <--> batch now begins and ends on a row boundary!
+		batch = alignBatchToRowBoundary(repLevels, batchStart, batchSize)
 		action(batchStart, batch)
 	}
-	action(batchStart, int64(len(repLevels))-batchStart)
+	// a grow-to-end alignment on an oversized final row can leave batchStart
+	// already at len(repLevels); skip the trailing zero-length action.
+	if batchStart < int64(len(repLevels)) {
```
zeroshade, line 709: Hardening item 3 ✅ — the `batchStart < len(repLevels)` guard skips the zero-length trailing `action` after a grow-to-end. Confirmed it's actually exercised: `TestWriteSpacedV2NullableListWideRow` (one row of 20001 values, batch 20000) grows to the slice end so `batchStart == len(repLevels)` and the trailing call is skipped.

parquet/file/column_writer.go
```diff
@@ -657,7 +657,12 @@ func (w *columnWriter) Close() (err error) {
 func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(offset, batch int64)) {
 	batchSize := w.props.WriteBatchSize()
 	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
-	// use the regular doBatches function
+	// use the regular doBatches function.
+	//
+	// The spec also requires row-aligned pages for V1 when an OffsetIndex is
+	// present (PageIndexEnabled). That gap is on the WriteBatch path and is
+	// out of scope for the DataPageV2 row alignment here; it is tracked in
+	// https://github.com/apache/arrow-go/issues/887.
```
zeroshade, line 665: Good call filing #887 and citing it here rather than folding V1 + OffsetIndex into this PR — that gap also affects `WriteBatch`, so keeping this change scoped to the DataPageV2 invariant is the right move. The reworded note reads well.

parquet/file/column_writer.go
```diff
@@ -706,6 +721,30 @@ func doBatches(total, batchSize int64, action func(offset, batch int64)) {
...
 }
 
+// alignBatchToRowBoundary adjusts batch so that repLevels[offset+batch] lands on
+// a row boundary (repetition level 0) or the end of the level slice. A repeated
+// row must never span a DataPageV2 page boundary, so it first shrinks toward the
+// previous boundary. If there is no boundary at or before the requested split -
+// the current row is wider than batch - it grows forward to the next one so the
+// whole row stays in a single batch and the caller keeps making progress rather
+// than looping on a zero-length batch. offset must already sit on a boundary.
+func alignBatchToRowBoundary(repLevels []int16, offset, batch int64) int64 {
```
zeroshade, line 731: Verified this helper is load-bearing: neutering the body to `return batch` turns every boundary test red (`starts mid-row`), while the shrink-then-grow keeps `batch >= 1`, so there's no no-progress loop. Non-blocking (fine for the #887 follow-up): on an oversized single row this grows past the byteArray path's 1 GB `maxSafeBatchDataSize` scan — unavoidable under V2 since a row can't span a page, just a late failure via the `MaxInt32` guard rather than an early explicit error.

parquet/file/column_writer_v2_row_boundary_test.go
```diff
@@ -0,0 +1,522 @@
...
+func TestWriteBitmapBatchSpacedV2RowBoundary(t *testing.T) {
+	const (
+		numRows    = 2000
+		valsPerRow = 4
+		total      = numRows * valsPerRow
+		batchSize  = 7 // not a multiple of valsPerRow: a fixed split lands mid-row
+		// A small page size relative to the batch stride keeps page flushes from
+		// coincidentally landing on row boundaries; with the fix reverted this
+		// produces 12 of 16 pages starting mid-row.
+		pageSize = 64
```
zeroshade, line 349: Dropping this to 64 makes the boolean spaced path genuinely load-bearing — verified: with `alignBatchToRowBoundary` reverted this test fails (`starts mid-row`), whereas the earlier `pageSize=256` cut passed even when reverted. Same for `TestWriteDictIndicesV2RowBoundary`, so both rerouted paths are now covered.

zeroshade (approved):
Approving — inline notes below accompany the full verification summary in my review above (generated-sync, build, green run, and red-green all re-checked locally on go1.26.4). Each prior ask is confirmed on the exact line it touches.

punkeel:
ty @zeroshade and @truffle-dev!

zeroshade:
Thanks @punkeel for driving this!

PR #884, opened 2026-07-02
fix(parquet/compress): Brotli CompressBound

fallintoplace:
### What changed

Brotli `CompressBound` now returns the computed maximum compressed size instead of the input length. The empty-input case is also handled before the debug assertion so it remains valid with `-tags assert`.

### Why

The `Codec` contract says callers can use `CompressBound` to allocate a destination buffer large enough for compressed output. Returning the input length could under-report the bound for Brotli and cause callers to allocate too little.

### Validation

- `go test ./parquet/compress`
- `go test -tags assert ./parquet/compress`
- `PARQUET_TEST_DATA=parquet-testing/data PARQUET_TEST_BAD_DATA=parquet-testing/bad_data go test ./parquet/...`


zeroshade (approved):

zeroshade:
just needs a rebase

fallintoplace:
@zeroshade I will have a look in a few hours, thank you for your attention.

PR #888, opened 2026-07-02
fix(array): guard BinaryViewBuilder size checks

fallintoplace:
### What changed

`BinaryViewBuilder` now checks view value lengths as `int` before any narrowing conversion to `int32`. Both `ReserveData` and `Append` use the same size guard.

The regression test covers oversized `ReserveData` lengths and oversized byte slices without allocating multi-gigabyte buffers.

### Why

The previous check converted lengths to `int32` before comparison. Values larger than `math.MaxInt32` could wrap negative and bypass the 2GB guard.

### Validation

- `go test ./arrow/array -run 'TestBinaryViewBuilderRejectsOversizedValues|TestBinaryViewStringRoundTrip'`
- `go test ./arrow/array`
- `go test ./arrow/...`
- `go test -tags assert ./arrow/array`

### DIFF
--- a/arrow/array/binarybuilder.go
+++ b/arrow/array/binarybuilder.go
@@ -467,11 +467,15 @@ func (b *BinaryViewBuilder) Resize(n int) {
 	b.rawData = arrow.ViewHeaderTraits.CastFromBytes(b.data.Bytes())
 }
 
-func (b *BinaryViewBuilder) ReserveData(length int) {
-	if int32(length) > viewValueSizeLimit {
+func checkBinaryViewValueSize(length int64) {
+	if length > int64(viewValueSizeLimit) {
 		panic(fmt.Errorf("%w: BinaryView or StringView elements cannot reference strings larger than 2GB",
 			arrow.ErrInvalid))
 	}
+}
+
+func (b *BinaryViewBuilder) ReserveData(length int) {
+	checkBinaryViewValueSize(int64(length))
 	b.blockBuilder.Reserve(int(length))
 }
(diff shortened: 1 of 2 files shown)


arrow/array/binarybuilder.go
```diff
@@ -467,11 +467,15 @@ func (b *BinaryViewBuilder) Resize(n int) {
 	b.rawData = arrow.ViewHeaderTraits.CastFromBytes(b.data.Bytes())
 }
 
-func (b *BinaryViewBuilder) ReserveData(length int) {
-	if int32(length) > viewValueSizeLimit {
+func checkBinaryViewValueSize(length int) {
+	if length > int(viewValueSizeLimit) {
```
zeroshade: if the concern is wrapping past the `int32` max value, this should be explicitly `int64(viewValueSizeLimit)` instead of just `int` so that it is correct on 32-bit systems also.

zeroshade (approved):
Thanks — extracting `checkBinaryViewValueSize(length int64)` addresses my earlier note: the comparison is now done in `int64` (`length > int64(viewValueSizeLimit)`), so the previous `int32(length)` truncation on 64-bit is gone and `Append`/`ReserveData` share one guard. `TestBinaryViewBuilderRejectsOversizedValues` (size-guard + ReserveData, the latter correctly skipped on 32-bit) passes locally. LGTM.

Heads-up: #895 reuses the `binaryViewNoAllocAllocator` helper this PR introduces, so this one should merge first.

PR #894, opened 2026-07-04
fix: clarify StringViewBuilder UnmarshalJSON error message

fallintoplace:
### Problem
`StringViewBuilder.UnmarshalJSON` returned an error message that referenced `binary view builder`. This is misleading, because callers using `StringViewBuilder` see an incorrect component name when `UnmarshalJSON` receives non-array input.

### Change
- Update the error text to correctly say `string view builder`.
- No logic changes; behavior remains otherwise identical.

### Impact
- Improves developer ergonomics and test clarity by ensuring diagnostics identify the correct builder.
- Reduces user confusion and avoids false assumptions when debugging JSON decoding failures.

### Testing
- This is a message-only change.
- No runtime test executed in this PR; update is isolated and validated by review of the edited error path.

### DIFF
--- a/arrow/array/string.go
+++ b/arrow/array/string.go
@@ -803,7 +803,7 @@ func (b *StringViewBuilder) UnmarshalJSON(data []byte) error {
 	}
 
 	if delim, ok := t.(json.Delim); !ok || delim != '[' {
-		return fmt.Errorf("binary view builder must unpack from json array, found %s", delim)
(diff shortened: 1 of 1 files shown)


zeroshade (approved):
Correct fix — `StringViewBuilder.UnmarshalJSON` was reporting "binary view builder", copy-pasted from the binary path. Verified `go build ./arrow/array/` is clean at the PR head. LGTM.

PR #895, opened 2026-07-04
fix(arrow/array): avoid allocating for ReserveData(0)

fallintoplace:
### Rationale for this change

BinaryViewBuilder.ReserveData always delegated to the block builder. ReserveData(0) could therefore create a full default data block even when every value remains inline.

- Return after the existing size validation when the requested data length is zero.
- Preserve the normal reserve path for positive lengths.
- Update integration JSON expectations because an unused empty variadic data buffer is no longer created.

### Are these changes tested?

Yes. A no-allocation allocator verifies that ReserveData(0) does not allocate. The array and arrjson packages pass normal and race tests.

Zero-length reservations no longer allocate an unused data block. Positive reservations and the public API are unchanged.

### DIFF
--- a/arrow/array/binary_test.go
+++ b/arrow/array/binary_test.go
@@ -732,6 +732,13 @@ func TestBinaryViewBuilderRejectsOversizedValues(t *testing.T) {
 	})
 }
 
+func TestBinaryViewBuilderReserveData0DoesNotAllocate(t *testing.T) {
+	b := NewBinaryViewBuilder(binaryViewNoAllocAllocator{})
+	defer b.Release()
+
+	b.ReserveData(0)
+}
+
 func TestBinaryViewStringRoundTrip(t *testing.T) {
 	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
 	defer mem.AssertSize(t, 0)
--- a/arrow/array/binarybuilder.go
+++ b/arrow/array/binarybuilder.go
@@ -476,6 +476,9 @@ func checkBinaryViewValueSize(length int64) {
 
 func (b *BinaryViewBuilder) ReserveData(length int) {
 	checkBinaryViewValueSize(int64(length))
+	if length == 0 {
+		return
+	}
 	b.blockBuilder.Reserve(int(length))
 }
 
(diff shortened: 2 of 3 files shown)


zeroshade (changes requested):
The `ReserveData(0)` short-circuit is reasonable, but the branch doesn't compile on its own:

```
arrow/array/binary_test.go:705:28: undefined: binaryViewNoAllocAllocator
FAIL	github.com/apache/arrow-go/v18/arrow/array [build failed]
```

`binaryViewNoAllocAllocator` is introduced in #888 — it's not on `main` and not in this PR. Please rebase this on top of #888 (so the helper exists) or add the helper here. Once it builds, the guard + test look good.

_Verified by checking out the PR head and running `go test ./arrow/array/`._

zeroshade:
Rebased onto `main` now that #888 has merged (it introduced `checkBinaryViewValueSize` and the `binaryViewNoAllocAllocator` test helper). Resolved the `ReserveData` and `binary_test.go` conflicts, and confirmed `go build`/`go test ./arrow/array/` are green (both `TestBinaryViewBuilderReserveData0DoesNotAllocate` and the oversized-value guard pass). This should now compile — over to CI.

zeroshade (approved):
Rebase onto #888 resolves the earlier build failure — `binaryViewNoAllocAllocator` and `checkBinaryViewValueSize` now exist. Verified on the PR head (`7241331`):

- `go build ./arrow/array/` clean; `TestBinaryViewBuilderReserveData0DoesNotAllocate`, the oversized-value guard, and `TestBinaryViewStringRoundTrip` all pass.
- `./arrow/internal/arrjson` and the IPC variadic-count test pass.

The `ReserveData(0)` short-circuit is safe: inline values never touch the block builder, and any non-inline value reserves with a positive length before `UnsafeAppend`. Dropping the unused empty data block also correctly removes the `"VARIADIC_DATA_BUFFERS": [""]` entries — `Array.Variadic` is `omitempty`, the writer derives the count from `len(Buffers())-2`, the reader loads a dynamic count, and `BinaryView.setData` only requires `>=2` buffers with `dataBuffers = buffers[2:]`. This also aligns Go with the C++ behavior of not emitting an empty variadic buffer for all-inline data.

LGTM.

PR #896, opened 2026-07-04
fix: guard multiBufferBuilder UnsafeAppend against short copy

fallintoplace:
## Summary
- Harden multiBufferBuilder.UnsafeAppend to fail fast when a value copy is truncated.
- After copy, validate that bytes copied equals the requested value length.
- Keep the existing debug assertion path and add a runtime panic for non-assert builds.
- Add regression coverage via TestMultiBufferBuilderUnsafeAppendPanicsOnTruncatedCopy in arrow/array/bufferbuilder_test.go.

## Why
This prevents silent data corruption in BinaryViewBuilder and StringViewBuilder out-of-line storage where metadata can point past the bytes actually written.

## Testing
- Added regression test for the truncation panic path.
- Full suite not run for this defensive-path-only change.

### DIFF
--- a/arrow/array/bufferbuilder.go
+++ b/arrow/array/bufferbuilder.go
@@ -243,7 +243,6 @@ func (b *multiBufferBuilder) UnsafeAppend(hdr *arrow.ViewHeader, val []byte) {
 	hdr.SetIndexOffset(int32(idx), int32(offset))
 
 	n := copy(buf.Buf()[offset:], val)
-	debug.Assert(n == len(val), "multibufferbuilder did not append full value")
 	if n != len(val) {
 		panic("arrow/array: multibufferbuilder could not append full value")
 	}
--- a/arrow/array/bufferbuilder.go
+++ b/arrow/array/bufferbuilder.go
@@ -243,6 +243,10 @@ func (b *multiBufferBuilder) UnsafeAppend(hdr *arrow.ViewHeader, val []byte) {
 	hdr.SetIndexOffset(int32(idx), int32(offset))
 
 	n := copy(buf.Buf()[offset:], val)
+	debug.Assert(n == len(val), "multibufferbuilder did not append full value")
+	if n != len(val) {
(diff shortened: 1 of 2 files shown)


zeroshade (changes requested):
Two issues:

**1. The new test doesn't trigger the guard — it fails.** `multiBufferBuilder.Reserve` sizes a new block to `max(nbytes, blockSize)` and `memory.Buffer.Reserve` rounds capacity up to a multiple of 64, so `Reserve(4)` with `blockSize: 8` gives a **64-byte** block. Appending 9 bytes fits, so there's no short copy and the panic never fires:

```
--- FAIL: TestMultiBufferBuilderUnsafeAppendPanicsOnTruncatedCopy
    should panic ... Panic value: <nil>
```

**2. Redundant guard.** `debug.Assert(n == len(val), ...)` followed by an unconditional `if n != len(val) { panic(...) }` is doubled up — the `debug.Assert` is dead once the unconditional panic exists. And since `Reserve` always rounds the block up and the contract is that callers `Reserve` before `UnsafeAppend`, a short copy is effectively unreachable via the normal path — worth deciding whether an always-on branch belongs on this hot `Unsafe` path (vs. a `debug.Assert` only).

If you keep the guard, the test needs a genuinely undersized block (append a value larger than the block's *rounded* capacity). _Verified against the PR head with `go test`._

zeroshade:
@fallintoplace gentle nudge — still blocked on the failing test: because `multiBufferBuilder.Reserve` rounds the block up to a multiple of 64, the 9-byte append into an 8-byte-requested block never short-copies, so `assert.Panics` fails (`Panic value: <nil>`). Details in my review above. Ping me and I'll re-review once it's updated.

zeroshade (approved):
The reworked test now genuinely exercises the truncated-copy guard — verified locally that it panics as intended (the earlier version couldn't short-copy because `Reserve` rounds the block up to 64). Thanks for the fix; this resolves my earlier request.

PR #897, opened 2026-07-04
fix: clarify StringViewBuilder.UnmarshalOne expected type

fallintoplace:
### Problem
`StringViewBuilder.UnmarshalOne` reports a `json.UnmarshalTypeError` with an expected type of `[]byte{}` for unsupported token types.

### Change
- Update `StringViewBuilder.UnmarshalOne` to report `reflect.TypeOf("")` when an unsupported JSON token is encountered.
- Add regression test `TestStringViewBuilderUnmarshalOneWrongType` in `arrow/array/string_test.go`.

### Impact
- Aligns error diagnostics with `StringBuilder` behavior.
- Makes JSON import debugging clearer for users expecting string inputs.

### Testing
- Added a focused regression test that asserts the reported expected type is `string` for unsupported JSON token input.
- No full test run in this PR (single-path, message-type assertion).

### DIFF
--- a/arrow/array/string.go
+++ b/arrow/array/string.go
@@ -778,7 +778,7 @@ func (b *StringViewBuilder) UnmarshalOne(dec *json.Decoder) error {
 	default:
 		return &json.UnmarshalTypeError{
 			Value:  fmt.Sprint(t),
-			Type:   reflect.TypeOf([]byte{}),
+			Type:   reflect.TypeOf(string("")),
 			Offset: dec.InputOffset(),
 		}
 	}
--- a/arrow/array/string_test.go
+++ b/arrow/array/string_test.go
@@ -18,7 +18,6 @@ package array_test
 
 import (
 	"bytes"
-	"encoding/json"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


zeroshade (changes requested):
The production change (`reflect.TypeOf([]byte{})` -> `reflect.TypeOf(string(""))`) is right — a `StringViewBuilder` should report `string`. But the test doesn't compile:

```
arrow/array/string_test.go:914:27: cannot use dec (variable of type *encoding/json.Decoder)
  as *github.com/apache/arrow-go/v18/internal/json.Decoder value in argument to bldr.UnmarshalOne
```

`UnmarshalOne` takes arrow's vendored `*internal/json.Decoder`, not the stdlib one. Import `github.com/apache/arrow-go/v18/internal/json` (e.g. `import json "github.com/apache/arrow-go/v18/internal/json"`) in the test instead of `encoding/json`. With that it should pass. _Verified against the PR head with `go test`._

zeroshade:
@fallintoplace nudge — the production change is correct; only the test is broken: it uses the stdlib `encoding/json.Decoder`, but `UnmarshalOne` takes arrow's `internal/json.Decoder`, so it won't compile. Swap the import and it should pass. Happy to re-review.

zeroshade (approved):

PR #898, opened 2026-07-04
fix(compute): release temporary arrays in binary view casts

fallintoplace:
### Rationale for this change

CastBinaryToBinaryView and CastBinaryViewToBinary create a temporary result array, transfer its buffers with ArraySpan.TakeOwnership, and return without releasing the temporary array. TakeOwnership retains the buffers, so the temporary reference leaks on every cast.

Release each temporary result after transferring its data to the output span. The output keeps its retained ownership, so successful cast behavior is unchanged.

### Are these changes tested?

Yes. Both cast directions are exercised with a checked allocator bound to the compute context. Each test releases the cast output and verifies that memory returns to the pre-cast baseline while the input remains alive. The compute package passes normal and race tests.

Binary-to-binary-view and binary-view-to-binary casts no longer retain temporary allocations. There is no API change.

### DIFF
--- a/arrow/compute/cast_test.go
+++ b/arrow/compute/cast_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/arrow/compute"
+	"github.com/apache/arrow-go/v18/arrow/compute/exec"
 	"github.com/apache/arrow-go/v18/arrow/decimal128"
 	"github.com/apache/arrow-go/v18/arrow/decimal256"
 	"github.com/apache/arrow-go/v18/arrow/internal/testing/gen"
@@ -1680,12 +1681,13 @@ func (c *CastSuite) TestBinaryViewToBinaryLike() {
 }
 
 func (c *CastSuite) TestBinaryToBinaryViewReleasesTemporaryResult() {
-	scope := memory.NewCheckedAllocatorScope(c.mem)
-
 	in, _, err := array.FromJSON(c.mem, arrow.BinaryTypes.Binary, strings.NewReader(`["aGk=", "dGhpcyBpcyB0aGUgZmlyc3QgdGVzdCE=", null]`))
 	c.Require().NoError(err)
 
-	out, err := compute.CastArray(context.Background(), in, compute.SafeCastOptions(arrow.BinaryTypes.BinaryView))
+	scope := memory.NewCheckedAllocatorScope(c.mem)
+	ctx := exec.WithAllocator(context.Background(), c.mem)
+
+	out, err := compute.CastArray(ctx, in, compute.SafeCastOptions(arrow.BinaryTypes.BinaryView))
 	c.Require().NoError(err)
 	out.Release()
 	scope.CheckSize(c.T())
@@ -1693,10 +1695,12 @@ func (c *CastSuite) TestBinaryToBinaryViewReleasesTemporaryResult() {
 }
 
 func (c *CastSuite) TestBinaryViewToBinaryReleasesTemporaryResult() {
-	scope := memory.NewCheckedAllocatorScope(c.mem)
 	in := c.buildStringViewArray([]string{"a", "short", "this is a longer cast value"}, nil)
 
-	out, err := compute.CastArray(context.Background(), in, compute.SafeCastOptions(arrow.BinaryTypes.LargeBinary))
+	scope := memory.NewCheckedAllocatorScope(c.mem)
+	ctx := exec.WithAllocator(context.Background(), c.mem)
+
+	out, err := compute.CastArray(ctx, in, compute.SafeCastOptions(arrow.BinaryTypes.LargeBinary))
 	c.Require().NoError(err)
 	out.Release()
 	scope.CheckSize(c.T())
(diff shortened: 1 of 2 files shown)


zeroshade (changes requested):
The production fix is correct and worth having — `ArraySpan.TakeOwnership` retains the buffers (`arrow/compute/exec/span.go`), so the `ba.bldr.NewArray()` temporary was leaking without the `defer result.Release()`.

But both new tests fail:

```
cast_test.go:1682: invalid memory size exp=192, got=0
cast_test.go:1694: invalid memory size exp=32896, got=0
--- FAIL: TestCasts/TestBinaryToBinaryViewReleasesTemporaryResult
--- FAIL: TestCasts/TestBinaryViewToBinaryReleasesTemporaryResult
```

The scope is created *after* `in` is allocated, but the test releases `in` *inside* the scope before `CheckSize`, so outstanding drops below the baseline (`exp` = size of `in`, `got` = 0). Either call `scope.CheckSize(c.T())` before `in.Release()`, or allocate `in` inside the scope. Once the harness is fixed these will lock the fix in nicely. _Verified against the PR head with `go test`._

zeroshade:
@fallintoplace nudge — the fix itself is correct (`TakeOwnership` retains, so releasing the temporary is right). The two new tests just need `scope.CheckSize(c.T())` called before `in.Release()` (or allocate `in` inside the scope) — as written they drop below the baseline and fail. Ping me for a re-review.

arrow/compute/cast_test.go
```diff
@@ -1668,6 +1668,30 @@ func (c *CastSuite) TestBinaryViewToBinaryLike() {
 	})
 }
 
+func (c *CastSuite) TestBinaryToBinaryViewReleasesTemporaryResult() {
+	scope := memory.NewCheckedAllocatorScope(c.mem)
```
zeroshade: Two problems with this test:

1. It fails as written — the scope is opened here, *before* `in` is allocated, but `CheckSize` runs while `in` is still alive, so it reports `exp=0, got=192`.
2. More importantly, the cast runs on `context.Background()` → `DefaultAllocator`, not `c.mem`, so the checked scope never sees the temporary. It passes with *or* without the production fix (I confirmed both), so it doesn't actually guard the leak.

To exercise the fix, allocate `in` first, open the scope after it, and bind the cast to the checked allocator:
```go
in, _, err := array.FromJSON(c.mem, arrow.BinaryTypes.Binary, strings.NewReader(`["aGk=", "dGhpcyBpcyB0aGUgZmlyc3QgdGVzdCE=", null]`))
c.Require().NoError(err)

scope := memory.NewCheckedAllocatorScope(c.mem)
ctx := exec.WithAllocator(context.Background(), c.mem) // arrow/compute/exec

out, err := compute.CastArray(ctx, in, compute.SafeCastOptions(arrow.BinaryTypes.BinaryView))
c.Require().NoError(err)
out.Release()
scope.CheckSize(c.T())
in.Release()
```
I verified this fails before the fix (`exp=192, got=33088`) and passes after.

arrow/compute/cast_test.go
```diff
@@ -1668,6 +1668,30 @@ func (c *CastSuite) TestBinaryViewToBinaryLike() {
...
+
+	out, err := compute.CastArray(context.Background(), in, compute.SafeCastOptions(arrow.BinaryTypes.BinaryView))
+	c.Require().NoError(err)
+	out.Release()
+	scope.CheckSize(c.T())
+	in.Release()
+}
+
+func (c *CastSuite) TestBinaryViewToBinaryReleasesTemporaryResult() {
+	scope := memory.NewCheckedAllocatorScope(c.mem)
```
zeroshade: Same two issues as the binary→view test above: it fails as written, and once the ordering is fixed it still doesn't bind the cast to `c.mem`, so it's vacuous. Apply the same pattern — build `in`, then open the scope, run the cast through `exec.WithAllocator(context.Background(), c.mem)`, release `out`, `CheckSize`, then release `in`.

zeroshade (changes requested):
The production fix here (`defer result.Release()` after `TakeOwnership`) is correct — `TakeOwnership` retains, so dropping the builder's ref is right. The two new tests need rework though: they fail as written, and even after fixing the ordering they don't bind the cast to the checked allocator, so they'd pass regardless of the fix. Details plus a verified formulation inline. Requesting changes on the tests only.

zeroshade (approved):
The production fix is correct and idiomatic — `TakeOwnership` retains the buffers, so the `ba.bldr.NewArray()` temporary leaked its own reference; `defer result.Release()` drops it, matching the existing `vector_selection.go` idiom (NewArray → defer Release → hand off with retain).

The test rework addresses both earlier points. Verified on the PR head (`8651608d`):

- Both new tests pass, and the full `TestCasts` suite is green.
- Reverting just `defer result.Release()` makes both tests **fail** with the checked allocator pinpointing the leak (`LEAK of 32768 bytes FROM kernels.CastBinaryToBinaryView ... binary_view_casts.go:197`), so they now genuinely lock in the fix. The `exec.WithAllocator(ctx, c.mem)` binding routes the kernel's temporary (`array.NewBuilder(exec.GetAllocator(ctx.Ctx), ...)`) through the checked allocator, and `CheckSize` is called before `in.Release()`.

Both directions are covered (Binary→BinaryView and BinaryView→LargeBinary). LGTM.

PR #900, opened 2026-07-04
fix(array): validate MakeFromData type IDs before constructor lookup

fallintoplace:
## Description

This change makes MakeFromData validate Arrow type IDs before dispatching to the concrete constructor.

## Details
- Removes the 0x3f mask so IDs are not wrapped around.
- Adds bounds and nil checks before constructor lookup.
- Panics with a clear `invalid data type: <id>` message for unsupported types.

## Testing
- Added regression test coverage for invalid Type(64) in `TestMakeFromData`.


fallintoplace:
Superseded by PR #904

PR #903, opened 2026-07-04
fix(arreflect): validate primitive conversion assignment overflow checks

fallintoplace:
## Summary
Prevent silent corruption in `arreflect` primitive conversions by validating numeric assignment bounds before setting through reflection.

## Why this fix
`setPrimitiveValue` used `SetInt/SetUint/SetFloat` directly for Arrow numeric types. That allowed values outside the target Go type range to wrap/truncate, so bad data could pass through without errors until consumed by fast paths relying on value integrity.

## Changes
- Added explicit overflow checks with:
  - `OverflowInt`
  - `OverflowUint`
  - `OverflowFloat`
- Return `ErrTypeMismatch`-wrapped errors on overflow.
- Preserve existing valid conversion behavior.

## Files
- `arrow/array/arreflect/reflect_arrow_to_go.go`
- `arrow/array/arreflect/reflect_arrow_to_go_test.go`

## Validation
- Added tests for overflow cases:
  - `int64` -> `int8`
  - `uint64` -> `uint8`
  - `float64` -> `float32`

### Bug details
- Severity: 82/100
- Ask product?: 0/100

### DIFF
--- a/arrow/array/arreflect/reflect_arrow_to_go.go
+++ b/arrow/array/arreflect/reflect_arrow_to_go.go
@@ -161,52 +161,92 @@ func setPrimitiveValue(v reflect.Value, arr arrow.Array, i int) error {
 		if !isIntKind(v.Kind()) {
 			return fmt.Errorf("cannot set int8 into %s: %w", v.Type(), ErrTypeMismatch)
 		}
-		v.SetInt(int64(arr.(*array.Int8).Value(i)))
+		val := int64(arr.(*array.Int8).Value(i))
+		if v.OverflowInt(val) {
+			return fmt.Errorf("cannot set int8 value %d into %s: %w", val, v.Type(), ErrTypeMismatch)
+		}
+		v.SetInt(val)
 	case arrow.INT16:
 		if !isIntKind(v.Kind()) {
 			return fmt.Errorf("cannot set int16 into %s: %w", v.Type(), ErrTypeMismatch)
 		}
-		v.SetInt(int64(arr.(*array.Int16).Value(i)))
+		val := int64(arr.(*array.Int16).Value(i))
+		if v.OverflowInt(val) {
+			return fmt.Errorf("cannot set int16 value %d into %s: %w", val, v.Type(), ErrTypeMismatch)
+		}
+		v.SetInt(val)
 	case arrow.INT32:
 		if !isIntKind(v.Kind()) {
 			return fmt.Errorf("cannot set int32 into %s: %w", v.Type(), ErrTypeMismatch)
 		}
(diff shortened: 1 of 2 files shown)


zeroshade (commented):
Two minor, non-blocking nits on the new FLOAT32 handling — the change itself is a solid defensive improvement and the tests are good.

arrow/array/arreflect/reflect_arrow_to_go.go
```diff
@@ -161,52 +161,92 @@ func setPrimitiveValue(v reflect.Value, arr arrow.Array, i int) error {
...
+		v.SetUint(val)
 	case arrow.FLOAT32:
 		if !isFloatKind(v.Kind()) {
 			return fmt.Errorf("cannot set float32 into %s: %w", v.Type(), ErrTypeMismatch)
 		}
+		val := float64(arr.(*array.Float32).Value(i))
+		if v.OverflowFloat(val) {
+			return fmt.Errorf("cannot set float32 value %f into %s: %w", val, v.Type(), ErrTypeMismatch)
+		}
 		v.SetFloat(float64(arr.(*array.Float32).Value(i)))
```
zeroshade: Minor/consistency: this re-reads `arr.(*array.Float32).Value(i)` instead of reusing the `val` computed just above (every other branch uses `v.SetInt(val)`/`v.SetUint(val)`/`v.SetFloat(val)`). Same result, just tidier — `v.SetFloat(val)`.

arrow/array/arreflect/reflect_arrow_to_go.go
```diff
@@ -161,52 +161,92 @@ func setPrimitiveValue(v reflect.Value, arr arrow.Array, i int) error {
...
+		if v.OverflowUint(val) {
+			return fmt.Errorf("cannot set uint64 value %d into %s: %w", val, v.Type(), ErrTypeMismatch)
+		}
+		v.SetUint(val)
 	case arrow.FLOAT32:
 		if !isFloatKind(v.Kind()) {
 			return fmt.Errorf("cannot set float32 into %s: %w", v.Type(), ErrTypeMismatch)
 		}
+		val := float64(arr.(*array.Float32).Value(i))
+		if v.OverflowFloat(val) {
```
zeroshade: Minor: for a FLOAT32 source this `OverflowFloat` check can never trigger — a float32 value widened to float64 always fits a float32 (or float64) destination, so it's effectively dead here, unlike the FLOAT64→float32 case below where it does real work. Harmless; feel free to drop it or leave a one-line note that it's kept for symmetry.

zeroshade (approved):

PR #904, opened 2026-07-04
fix(array): validate MakeFromData type IDs

fallintoplace:
## Summary
Fail fast on invalid Arrow type IDs in `array.MakeFromData` instead of masking and dispatching into the wrong constructor slot.

## Why this fix
`MakeFromData` currently masks the type ID (`& 0x3f`) before table lookup. New or invalid IDs can wrap to valid indices, causing wrong constructors or nil dispatch and a confusing panic.

## Changes
- Convert ID to `int` and validate it explicitly.
- Reject IDs that are:
  - less than zero
  - greater/equal to `len(makeArrayFn)`
  - mapped to `nil` constructors
- Keep existing `invalidDataType(...)` panic path for unsupported IDs.

## Files
- `arrow/array/array.go`
- `arrow/array/array_test.go`

## Validation
- Added `TestMakeFromData` case for invalid type `64` and assertion of panic text.

### Bug details
- Severity: 88/100
- Ask product?: 0/100

### DIFF
--- a/arrow/array/array.go
+++ b/arrow/array/array.go
@@ -117,7 +117,11 @@ func invalidDataType(data arrow.ArrayData) arrow.Array {
 
 // MakeFromData constructs a strongly-typed array instance from generic Data.
 func MakeFromData(data arrow.ArrayData) arrow.Array {
-	return makeArrayFn[byte(data.DataType().ID()&0x3f)](data)
+	id := int(data.DataType().ID())
+	if id < 0 || id >= len(makeArrayFn) || makeArrayFn[id] == nil {
+		invalidDataType(data)
+	}
+	return makeArrayFn[id](data)
 }
 
 // NewSlice constructs a zero-copy slice of the array with the indicated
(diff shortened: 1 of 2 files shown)


arrow/array/array.go
```diff
@@ -117,7 +117,11 @@ func invalidDataType(data arrow.ArrayData) arrow.Array {
 
 // MakeFromData constructs a strongly-typed array instance from generic Data.
 func MakeFromData(data arrow.ArrayData) arrow.Array {
-	return makeArrayFn[byte(data.DataType().ID()&0x3f)](data)
+	id := int(data.DataType().ID())
+	if id < 0 || id >= len(makeArrayFn) || makeArrayFn[id] == nil {
+		invalidDataType(data)
```
zeroshade, line 122: Nit: prefer `return invalidDataType(data)` here. As written it relies on `invalidDataType` panicking as a side effect; returning it is clearer and stays correct even if that helper is ever changed to return a sentinel/typed error instead of panicking (otherwise execution would fall through to `makeArrayFn[id](data)` with an invalid `id`).

zeroshade (commented):
Correct fix — the old `& 0x3f` mask silently wrapped type IDs ≥ 64 onto the wrong constructor. One small nit inline.

PR #905, opened 2026-07-04
fix(array): make SetNull idempotent and track null count

fallintoplace:
## Summary
Fix `builder.SetNull` null-count accounting and idempotency.

## Why this fix
`SetNull` cleared the validity bit but did not update `b.nulls`, so bitmap and `NullN()` metadata could diverge. Repeated `SetNull(i)` also inflated null count.

## Changes
- Increment `NullN` only when the targeted entry is currently valid.
- Keep behavior idempotent for already-null entries.
- Validate in tests that `MapBuilder` delegates the same semantics.

## Files
- `arrow/array/builder.go`
- `arrow/array/builder_test.go`
- `arrow/array/map_test.go`

## Validation
- Added assertions for:
  - `NullN()` after setting initial nulls
  - idempotent `SetNull(0)` does not change count
  - `MapBuilder.NullN()` reflects null transitions.

### Bug details
- Severity: 86/100
- Ask product?: 0/100

### DIFF
--- a/arrow/array/builder.go
+++ b/arrow/array/builder.go
@@ -133,6 +133,9 @@ func (b *builder) SetNull(i int) {
 	if i < 0 || i >= b.length {
 		panic("arrow/array: index out of range")
 	}
+	if bitutil.BitIsSet(b.nullBitmap.Bytes(), i) {
+		b.nulls++
+	}
 	bitutil.ClearBit(b.nullBitmap.Bytes(), i)
 }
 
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Correct fix. The old `SetNull` cleared the validity bit without adjusting the null count, so arrays built with `SetNull` under-reported `NullN()` (and that count flows into the produced array metadata). The `BitIsSet` guard makes it idempotent, and the only in-tree callers are tests + `MapBuilder`'s delegation, so there's no double-count. Verified the `arrow/array` package tests pass. LGTM.

PR #908, opened 2026-07-05
build: verify Apache RAT jar checksum before execution

fallintoplace:
## Summary
- Pin  to a fixed Apache RAT version check and verify its SHA-512 checksum before running
- Add committed checksum file 
- Fail fast if checksum missing or mismatched to avoid executing unverified jars

## Testing
- Not run (release helper script hardening change only).


zeroshade (approved):
Verified the pinned SHA-512 matches the real `apache-rat-0.16.1.jar` on Maven Central byte-for-byte, and running the check on every invocation (not just on download) is the correct behavior. Good supply-chain hardening. LGTM.

PR #909, opened 2026-07-05
fix: return protobuf map key conversion errors

fallintoplace:
## Summary
- Make protobuf map key conversion return an error instead of panicking for unsupported key kinds.
- Thread conversion errors through map iteration into record construction with contextual message including field name and map type.
- Add regression tests for invalid map key conversion and contextual append error path.

## Validation
- Changes committed in branch: fix-protobuf-map-key-error-clean
- Target: apache/arrow-go:main
- No tests run in this step.

### DIFF
--- a/arrow/util/protobuf_reflect.go
+++ b/arrow/util/protobuf_reflect.go
@@ -366,7 +366,10 @@ func (pfr *ProtobufFieldReflection) asMap() protobufMapReflection {
 
 func (pmr protobufMapReflection) getDataType() arrow.DataType {
 	for kvp := range pmr.generateKeyValuePairs() {
-		return kvp.getDataType()
+		if kvp.err != nil {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


zeroshade:
Heads-up on overlap with #910: the map-key error-propagation changes here — `getMapKey` returning `(protoreflect.Value, error)`, the `protobufMapKeyValuePairResult` channel wrapper, `getDataType` skipping errored pairs, and the contextual error in `AppendValueOrNull` — are identical to what #910 already contains. #910 stacks the oneof/union nil-safety fix on top of exactly these changes, so it's effectively a strict superset and the two will conflict on merge.

Suggestion: close this in favor of #910. If you'd rather keep them separate for reviewability, the alternative is to merge #909 first and rebase #910 onto it — but folding into #910 seems cleanest. The code here is correct either way.

PR #911, opened 2026-07-05
fix(parquet): validate DataPageV2 decoded length

fallintoplace:
## What changed

- Make DataPageV2 V2 readers verify decoded byte counts instead of only checking buffer size.
- Update readV2Encrypted/readV2Unencrypted to return the actual number of bytes written/read and validate lengths.
- Enforce decoded-length validation in 6 uncompressed path and encrypted/unencrypted paths.
- Add regression tests for:
  - compressed DataPageV2 with mismatched codec decode length
  - encrypted uncompressed DataPageV2 with short decrypted output

## Validation

- Added focused regression coverage in .
- No functional changes to non-V2 paths.

### DIFF
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -593,7 +593,7 @@ func (p *serializedPageReader) readV2Unencrypted(rd io.Reader, lenCompressed int
 		return decodedLen, err
 	}
 	decodedLen += len(values)
-	if len(values) != lenUncompressed-decodedLen {
+	if decodedLen != lenUncompressed {
 		return decodedLen, fmt.Errorf("parquet: metadata said %d bytes uncompressed data page, got %d bytes", lenUncompressed, decodedLen)
 	}
 	return decodedLen, nil
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -518,53 +518,85 @@ func (p *serializedPageReader) decompress(rd io.Reader, lenCompressed int, buf [
 	return p.codec.Decode(buf, data), nil
 }
 
-func (p *serializedPageReader) readV2Encrypted(rd io.Reader, lenCompressed int, levelsBytelen int, compressed bool, buf []byte) error {
+func (p *serializedPageReader) readV2Encrypted(rd io.Reader, lenCompressed int, lenUncompressed int, levelsBytelen int, compressed bool, buf []byte) (int, error) {
 	// if encrypted, we need to decrypt before decompressing
 	p.decompressBuffer.ResizeNoShrink(lenCompressed)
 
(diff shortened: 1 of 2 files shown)


parquet/file/page_reader.go
```diff
@@ -518,53 +518,85 @@ func (p *serializedPageReader) decompress(rd io.Reader, lenCompressed int, buf [
...
 	}
-	if _, err := p.decompress(p.r, lenCompressed-levelsBytelen, buf[levelsBytelen:]); err != nil {
-		return err
+	values, err := p.decompress(p.r, lenCompressed-levelsBytelen, buf[levelsBytelen:])
+	if err != nil {
+		return decodedLen, err
 	}
-	return nil
+	decodedLen += len(values)
+	if len(values) != lenUncompressed-decodedLen {
```
zeroshade: This is the blocking issue. `decodedLen += len(values)` on the line above already makes `decodedLen == levelsBytelen + len(values)`, so `lenUncompressed - decodedLen` here reduces to `lenUncompressed - levelsBytelen - len(values)` — which is `0` for any valid page. That turns this into `len(values) != 0`, i.e. it rejects **every** non-empty compressed DataPageV2. (The encrypted sibling gets it right: `len(decoded) != lenUncompressed-levelsBytelen`.)

Suggested fix:
```go
decodedLen += len(values)
if decodedLen != lenUncompressed {
    return decodedLen, fmt.Errorf("parquet: metadata said %d bytes uncompressed data page, got %d bytes", lenUncompressed, decodedLen)
}
```

I verified on this branch with a valid compressed V2 page (Snappy, no levels): `Next()` returns false with the self-contradictory `parquet: metadata said 128 bytes uncompressed data page, got 128 bytes`. The new tests here don't catch it because they only exercise the mismatched and encrypted paths — please also add a happy-path test for a valid *compressed* DataPageV2, since the existing `TestDataPageV2` is uncompressed and `TestCompression` is V1-only.

zeroshade (changes requested):
Thanks for adding the decoded-length validation — the goal is good, and the encrypted path plus the caller-level `got != lenUncompressed` check are correct. But `readV2Unencrypted` has an accumulation bug that regresses the happy path for compressed V2 pages (details inline). Requesting changes on that one check, plus a happy-path test to lock it in. Everything else looks fine.

zeroshade (approved):
The compressed-V2 regression is fixed — `readV2Unencrypted` now checks `decodedLen != lenUncompressed`, and the new `TestDataPageV2Compressed` guards the happy path. Verified locally: that test passes (valid compressed V2 now reads, `Data()` round-trips) while `TestDataPageV2CorruptDecodedLength` and the encrypted-mismatch test still error correctly, and the full `parquet/file` package is green. Thanks for the quick turnaround — LGTM.

PR #918, opened 2026-07-09
fix(arrow/array): honor field nullability in comparison and JSON marshaling

zeroshade:
### Rationale for this change

This is the inner-nullability parity work that was previously bundled into #558 (the `TimestampWithOffset` canonical extension type). #558 has now been reduced to just the extension type, and this PR carries the comparison and JSON-marshaling changes on their own so they can be reviewed independently and more carefully.

- `arrow/array/compare.go`: thread each field's `Nullable` flag through `Equal`/`RecordEqual`/`TableEqual`/`ChunkedEqual` (and the `Approx` variants) so that non-nullable fields ignore validity-bitmap differences; add a `WithNullable` functional option; compare record/table fields against the right-hand schema (not the left against itself) and ignore field metadata during record/table comparison.
- `arrow.Array` interface: keep the single-argument `GetOneForMarshal(i int)` and move the field-local behavior to a new optional `NullableMarshaler` interface with `GetOneForMarshalNullable(i int, nullable bool)`, so external callers and custom `Array` implementations are not broken. Containers (struct, union, dictionary, run-end-encoded, extension) and `RecordToJSON` propagate each field's nullability through an unexported helper that falls back to `GetOneForMarshal` when the optional interface is not implemented.
- Struct/record: honor `field.Nullable` when serializing to JSON, and handle explicit `null` for a non-nullable field when reading.
- `internal/json`: add a helper to detect a `null` message.
- Test fixtures updated to declare inner struct field nullability explicitly.

### Relationship to #833

#833 pursues the same goal with a different strategy: it drops the comparison leniency (`WithNullable`/`equalOption`) in favor of schema validation in `Validate()` to match Arrow C++, and checks `IsValid()` in the struct/record builders rather than introducing `GetOneForMarshalNullable()`. This PR preserves the earlier comparison-based approach. Reviewers should decide which direction to land; opening as a draft pending that decision.

### Are these changes tested?

Yes. `go test ./arrow/array/ ./arrow/extensions/ ./arrow/internal/arrjson/ ./arrow/compute/ ./arrow/ipc/...` pass.

Yes: comparison functions gain a `WithNullable` option and the array marshaling path gains an optional `NullableMarshaler` interface. The existing `GetOneForMarshal(i int)` signature is preserved.

### DIFF
--- a/arrow/array.go
+++ b/arrow/array.go
@@ -111,7 +111,7 @@ type Array interface {
 	ValueStr(i int) string
 
 	// Get single value to be marshalled with `json.Marshal`
-	GetOneForMarshal(i int) interface{}
+	GetOneForMarshal(i int, nullable bool) interface{}
 
 	Data() ArrayData
 
--- a/arrow/array.go
+++ b/arrow/array.go
@@ -111,7 +111,7 @@ type Array interface {
 	ValueStr(i int) string
 
 	// Get single value to be marshalled with `json.Marshal`
-	GetOneForMarshal(i int, nullable bool) interface{}
+	GetOneForMarshal(i int) interface{}
 
 	Data() ArrayData
 
@@ -126,6 +126,21 @@ type Array interface {
 	// Release may be called simultaneously from multiple goroutines.
 	// When the reference count goes to zero, the memory is freed.
 	Release()
+}
+
+// NullableMarshaler is an optional interface an Array may implement to control
+// whether a value is rendered as JSON null based on the field-local nullability
+// from the schema, rather than solely on the array's own validity bitmap.
+//
+// It exists so containers (struct/union/dictionary/run-end-encoded/extension)
+// and RecordToJSON can propagate each field's Nullable flag down to its values
+// without a breaking change to the Array interface. Arrays that do not implement
+// it fall back to GetOneForMarshal, preserving the previous validity-based behavior.
+type NullableMarshaler interface {
+	// GetOneForMarshalNullable returns the value at i for json.Marshal. When
+	// nullable is false, a value is returned even if the array's validity bitmap
+	// marks it null (a non-nullable field must not serialize as null).
+	GetOneForMarshalNullable(i int, nullable bool) interface{}
 }
 
 // ValueType is a generic constraint for valid Arrow primitive types
--- a/arrow/array/binary.go
+++ b/arrow/array/binary.go
@@ -152,17 +152,21 @@ func (a *Binary) setData(data *Data) {
 	}
 }
 
-func (a *Binary) GetOneForMarshal(i int, nullable bool) interface{} {
+func (a *Binary) GetOneForMarshalNullable(i int, nullable bool) interface{} {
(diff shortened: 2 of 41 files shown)


zeroshade:
### How Arrow C++ handles run-end-encoded value nullability

For context on the REE value-nullability discussion here, comparing against the C++ reference implementation:

**C++ has no configurable value nullability.** `RunEndEncodedType` hard-wires the `values` child to `nullable=true` ([type.cc#L1307](https://github.com/apache/arrow/blob/30ca5a7ef262b28bacab50768531c26af5872d3c/cpp/src/arrow/type.cc#L1307-L1313)):

```cpp
children_ = {std::make_shared<Field>("run_ends", std::move(run_end_type), false),
             std::make_shared<Field>("values",   std::move(value_type),   true)};
```

- **DataType equality** compares child *data types* only — the values field's nullability is not part of REE type identity ([compare.cc#L883](https://github.com/apache/arrow/blob/30ca5a7ef262b28bacab50768531c26af5872d3c/cpp/src/arrow/compare.cc#L883-L887)):
  ```cpp
  result_ = left.value_type()->Equals(right.value_type()) &&
            left.run_end_type()->Equals(right.run_end_type());
  ```
- **Array equality** merges runs and compares the decoded `values` (validity included, since values are always nullable); it never consults `Field::nullable`.
- **IPC** reconstructs `RunEndEncodedType(children[0]->type(), children[1]->type())` ([metadata_internal.cc#L423](https://github.com/apache/arrow/blob/30ca5a7ef262b28bacab50768531c26af5872d3c/cpp/src/arrow/ipc/metadata_internal.cc#L423-L432)) — the values-field nullable bit is dropped and re-fixed to nullable, so a non-nullable-values REE does not round-trip.
- **Spec** ([Columnar.rst](https://github.com/apache/arrow/blob/30ca5a7ef262b28bacab50768531c26af5872d3c/docs/source/format/Columnar.rst#L1090-L1153)): `run_ends` cannot be null; `values` may be null.

**Implications:**

1. Excluding `ValueNullable` from `arrow.TypeEqual` (as it is today) matches C++. Making it part of type identity would be stricter than the reference and breaks REE record/IPC construction in arrow-go (schema-vs-array type mismatch, since `ValueNullable` isn't set consistently and doesn't survive IPC).
2. More broadly, C++ never makes `Equal()` lenient based on a field's `Nullable` flag — it keeps equality strict and enforces schema nullability via `Validate()`. That is the direction taken in #833, and it's why the `equalOpts` mechanism in this PR diverges from the reference. This branch remains useful as the record of the alternative approach, but the C++-aligned path is `Validate()`-based (#833).

zeroshade:
Closing in favor of #833.

As detailed in the comparison above, the `equalOpts` leniency in this PR diverges from Arrow C++, which keeps `Equal()` strict and enforces schema nullability via `Validate()` instead. #833 implements that C++-aligned approach, so the inner-nullability parity work should land there.

The `TimestampWithOffset` canonical extension type that originally motivated this work has been decoupled into #558 and can merge independently.

PR #921, opened 2026-07-11
fix(parquet): propagate row-group and page-index write errors

fallintoplace:
Row-group transitions and footer flushes used to discard close and page-index errors, which could leave a truncated file that still appeared to write successfully.

This change adds checked row-group APIs, preserves legacy panic behavior for the no-error APIs, persists writer failures so a later footer cannot be written, returns Arrow Parquet row-group errors, and checks buffered metadata serialization and short writes.

Tests:
- go test ./parquet/file -run 'Test(AppendRowGroupCheckedPropagatesCloseError|BufferedRowGroupPropagatesShortWrite|FlushWithFooterPropagatesPageIndexWriteError|NewParquetWriterWithError|PqarrowNewFileWriter)'
- go test ./parquet/... -run '^$'

### DIFF
--- a/parquet/file/file_writer.go
+++ b/parquet/file/file_writer.go
@@ -18,6 +18,7 @@ package file
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 
@@ -36,6 +37,7 @@ type Writer struct {
 	props            *parquet.WriterProperties
 	rowGroups        int
 	nrows            int
+	writeErr         error
 	metadata         metadata.FileMetaDataBuilder
 	fileEncryptor    encryption.FileEncryptor
 	rowGroupWriter   *rowGroupWriter
@@ -165,8 +167,26 @@ func (fw *Writer) TotalCompressedBytes() int64 {
 // is closed.
 //
 // When calling Close, all columns must have the same number of rows written.
+//
+// Deprecated: AppendBufferedRowGroup panics if closing the previous row group
+// or starting the new row group fails. Use AppendBufferedRowGroupChecked to
+// handle those errors explicitly.
 func (fw *Writer) AppendBufferedRowGroup() BufferedRowGroupWriter {
-	return fw.appendRowGroup(true)
+	rgw, err := fw.AppendBufferedRowGroupChecked()
+	if err != nil {
+		panic(err)
+	}
+	return rgw
+}
+
+// AppendBufferedRowGroupChecked is the error-returning form of
+// AppendBufferedRowGroup.
+func (fw *Writer) AppendBufferedRowGroupChecked() (BufferedRowGroupWriter, error) {
+	rgw, err := fw.appendRowGroup(true)
+	if err != nil {
+		return nil, fw.recordError(err)
+	}
+	return rgw, nil
 }
 
 // AppendRowGroup appends a row group to the file and returns a writer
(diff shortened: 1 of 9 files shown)


parquet/pqarrow/file_writer.go
```diff
@@ -185,25 +185,39 @@ func NewFileWriter(arrschema *arrow.Schema, w io.Writer, props *parquet.WriterPr
 
 // NewRowGroup does what it says on the tin, creates a new row group in the underlying file.
 // Equivalent to `AppendRowGroup` on a file.Writer
-func (fw *FileWriter) NewRowGroup() {
+func (fw *FileWriter) NewRowGroup() error {
```
zeroshade: Blocking (exported API break): changing `NewRowGroup()` to `NewRowGroup() error` alters a public method signature and breaks downstream method-values / interface satisfaction. Please mirror the approach you used in `file.Writer`: keep `NewRowGroup()` (panic on error) and add `NewRowGroupChecked() error`, with the internal write paths calling the checked variant.

parquet/pqarrow/file_writer.go
```diff
@@ -185,25 +185,39 @@ func NewFileWriter(arrschema *arrow.Schema, w io.Writer, props *parquet.WriterPr
...
+	return nil
 }
 
 // NewBufferedRowGroup starts a new memory Buffered Row Group to allow writing columns / records
 // without immediately flushing them to disk. This allows using WriteBuffered to write records
 // and decide where to break your row group based on the TotalBytesWritten rather than on the max
 // row group len. If using Records, this should be paired with WriteBuffered, while
 // Write will always write a new record as a row group in and of itself.
-func (fw *FileWriter) NewBufferedRowGroup() {
+func (fw *FileWriter) NewBufferedRowGroup() error {
```
zeroshade: Same exported-API concern — keep `NewBufferedRowGroup()` and add `NewBufferedRowGroupChecked() error`. Note the existing callers at `encode_arrow_test.go:231,330,385,1374,1399` invoke these as bare statements and now silently drop the returned error; with the compatibility shape above they can remain unchanged.

zeroshade (changes requested):
The core change — persisting `writeErr` on `file.Writer` and propagating row-group/page-index close/write errors — is valuable, and the `file.Writer` side is done well (existing signatures preserved, `...Checked` variants added). Two blockers inline. Also worth a changelog note: the deprecated `AppendRowGroup()`/`AppendBufferedRowGroup()` now panic on errors that were previously swallowed. The failure-injection tests look good.

parquet/file/row_group_writer.go
```diff
@@ -251,7 +256,7 @@ func (rg *rowGroupWriter) Close() error {
 		rg.metadata.SetNumRows(rg.nrows)
 		rg.metadata.Finish(rg.bytesWritten, rg.ordinal)
```
zeroshade: `rg.metadata.Finish(...)` still discards its returned error. If it fails (for example, closing a serial row group before all schema columns have been initialized), `Close()` will report success and the file writer may continue with incomplete row-group metadata. Please capture it into `rg.closeErr` and return it, consistent with the other close failures handled just above.

parquet/file/file_writer.go
```diff
@@ -244,22 +298,35 @@ func (fw *Writer) startFile() error {
 	return nil
 }
 
-func (fw *Writer) writePageIndex() {
+func (fw *Writer) writePageIndex() error {
 	if fw.pageIndexBuilder != nil {
 		// serialize page index after all row groups have been written and report
 		// location to the file metadata
 		fw.pageIndexBuilder.Finish()
```
zeroshade, line 311: `writePageIndex` permanently `Finish()`es the `PageIndexBuilder`, but this path is reachable from `FlushWithFooter`, which is documented to leave the writer open for additional row groups. With page indexes enabled, a subsequent `AppendRowGroupChecked`/`AppendBufferedRowGroupChecked` will then call `AppendRowGroup` on a finished builder and record a sticky error. Please keep page-index state appendable after an intermediate footer, or explicitly document/enforce that cumulative footers are unsupported when page indexes are enabled.

zeroshade (changes requested):
Good direction on propagating the writer errors. Two gaps remain before this is complete — see inline.

zeroshade (approved):

PR #923, opened 2026-07-11
fix(array): bounds check FixedSizeBinary.Value

fallintoplace:
FixedSizeBinary.Value applied the physical slice offset before validating the logical index. On sliced arrays, negative and one-past-the-end indexes could therefore expose adjacent values from the shared backing buffer.

This adds the standard array bounds check before applying the physical offset and covers valid, negative, and one-past-the-end access on a sliced three-element array.

Tests:
- go test ./arrow/array -run 'TestFixedSizeBinary(ValueBounds|Slice|$)'
- go test ./arrow/array


zeroshade (approved):
LGTM — adds the missing logical bounds check before applying the physical offset in FixedSizeBinary.Value, matching Binary.Value/String.Value and fixing the sliced-array OOB read. CI green. Thanks @fallintoplace!

PR #924, opened 2026-07-11
fix(parquet): make column reader cleanup idempotent

fallintoplace:
Column chunk cleanup retained both the current page and page reader after Close, so repeated cleanup could release or close the same resources twice. Page-reader replacement also discarded close errors and installed the replacement anyway.

This change detaches resources before cleanup, makes Close idempotent, returns and retains replacement cleanup errors, and uses the checked replacement path from the Arrow reader before continuing to seek or read.

Tests:
- go test ./parquet/file -run 'TestPrimitiveReaderSuite|TestColumnChunkReader'
- go test ./parquet/... -run '^$'


parquet/file/column_reader.go
```diff
@@ -251,21 +253,31 @@ func (c *columnChunkReader) Descriptor() *schema.Column    { return c.descr }
...
+		return err
+	}
 	c.rdr, c.err = rdr, nil
 	c.decoders = make(map[format.Encoding]encoding.TypedDecoder)
 	c.dictState = dictNotRead
 	c.numBuffered, c.numDecoded = 0, 0
+	return nil
 }
 
 // Close closes the page raeder and the page if set.
```
zeroshade: Typo: `raeder` -> `reader`.

parquet/file/column_reader.go
```diff
@@ -251,21 +253,31 @@ func (c *columnChunkReader) Descriptor() *schema.Column    { return c.descr }
 func (c *columnChunkReader) consumeBufferedValues(n int64) { c.numDecoded += n }
 func (c *columnChunkReader) numAvailValues() int64         { return c.numBuffered - c.numDecoded }
 func (c *columnChunkReader) pager() PageReader             { return c.rdr }
-func (c *columnChunkReader) setPageReader(rdr PageReader) {
-	c.Close()
+
+func (c *columnChunkReader) setPageReader(rdr PageReader) error {
```
zeroshade, line 258: Optional: add a pqarrow-level test that a replacement close-error surfaces via `LoadBatch`/`SeekToRow`; the propagation is currently only covered at the primitive level.

zeroshade (changes requested):
LGTM in substance — detaching `curPage`/`rdr` before cleanup makes `Close` idempotent (no double-release/close), and making `setPageReader` return an error so replacement failures propagate through `SeekToRow`/`nextRowGroup` is the right call. Good targeted tests. Two nits inline.

parquet/pqarrow/column_readers.go
```diff
@@ -162,13 +162,23 @@ func (lr *leafReader) clearOut() (out *arrow.Chunked) {
 
 func (lr *leafReader) Field() *arrow.Field { return lr.field }
 
+func (lr *leafReader) setPageReader(pr file.PageReader) error {
+	rr, ok := lr.recordRdr.(file.RecordReaderWithError)
+	if !ok {
+		return errors.New("record reader does not support error-returning page reader replacement")
+	}
+	return rr.SetPageReaderWithError(pr)
```
zeroshade: If `SetPageReaderWithError(pr)` returns an error, the freshly acquired `pr` is neither installed nor closed. Both call sites (`SeekToRow` and `nextRowGroup`) obtain a new `PageReader` immediately before calling this helper, so a replacement failure leaks the reader and its buffers. Please close `pr` before returning the error here (or transfer ownership back and close it at each call site).

zeroshade (changes requested):
The idempotent-cleanup change looks right. One resource leak on the error path to address — inline.

zeroshade (approved):

PR #925, opened 2026-07-11
fix(array): reject unsupported type IDs

fallintoplace:
MakeFromData masked datatype IDs to six bits before dispatching to array constructors. Unsupported IDs could therefore alias valid constructors, silently producing the wrong concrete array type.

This validates the ID bounds and constructor entry before dispatching, while preserving the documented invalid-data-type panic. Regression coverage now includes IDs -1, 63, 64, 65, 127, and a large positive ID.

Tests:
- go test ./arrow/array -run TestMakeFromData
- go test ./arrow/array
- go test ./arrow/... -run '^$'

### DIFF
--- a/arrow/array/array.go
+++ b/arrow/array/array.go
@@ -117,7 +117,11 @@ func invalidDataType(data arrow.ArrayData) arrow.Array {
 
 // MakeFromData constructs a strongly-typed array instance from generic Data.
 func MakeFromData(data arrow.ArrayData) arrow.Array {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM — replaces the &0x3f mask in MakeFromData with an explicit bounds + nil-constructor check, fixing the aliasing where type IDs >= 64 mapped onto valid constructors. CI green. Thanks @fallintoplace!

PR #926, opened 2026-07-11
fix(array): retain dictionary before replacement

fallintoplace:
Data.SetDictionary released the existing dictionary before retaining the replacement. Replacing a dictionary with the same pointer while the parent held the final reference could therefore clear the dictionary before resurrecting its reference.

This retains the validated replacement first, swaps it into the Data object, then releases the previous dictionary. It also handles nil input and reports an explicit panic for an invalid dictionary data implementation.

Tests:
- go test ./arrow/array -run 'TestData(SetDictionaryWithSamePointerRetainsDictionary|ResetClearsDictionary)'
- go test ./arrow/array
- go test ./arrow/... -run '^$'

### DIFF
--- a/arrow/array/data.go
+++ b/arrow/array/data.go
@@ -187,13 +187,22 @@ func (d *Data) Dictionary() arrow.ArrayData { return d.dictionary }
 
 // SetDictionary allows replacing the dictionary for this particular Data object
 func (d *Data) SetDictionary(dict arrow.ArrayData) {
-	if d.dictionary != nil {
-		d.dictionary.Release()
-		d.dictionary = nil
+	var newDict *Data
+	if dict != nil {
+		var ok bool
+		newDict, ok = dict.(*Data)
+		if !ok {
+			panic("arrow/array: dictionary data must be *array.Data")
+		}
 	}
-	if dict.(*Data) != nil {
-		dict.Retain()
-		d.dictionary = dict.(*Data)
+
+	if newDict != nil {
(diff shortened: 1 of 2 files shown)


arrow/array/data.go
```diff
@@ -187,13 +187,22 @@ func (d *Data) Dictionary() arrow.ArrayData { return d.dictionary }
...
+		if !ok {
+			panic("arrow/array: dictionary data must be *array.Data")
+		}
 	}
-	if dict.(*Data) != nil {
-		dict.Retain()
-		d.dictionary = dict.(*Data)
+
+	if newDict != nil {
+		newDict.Retain()
```
zeroshade, line 200: Retain-before-release ordering is correct. Please add a `SetDictionary(nil)` test (clearing is now supported per the PR description but untested), and optionally one asserting the explicit panic for a non-`*array.Data` implementation.

zeroshade (changes requested):
Correct fix for a real use-after-free — releasing the old dictionary before retaining the replacement could free-then-resurrect a cleared `*Data` when the parent held the last reference. The retain-before-release ordering now matches how `Reset` handles buffers/children, and the regression test captures the exact case. One test nit inline.

zeroshade (approved):
LGTM. Retaining the replacement dictionary before releasing the old one is the correct fix, and the same-pointer replacement / nil-clear tests cover it well. Thanks!

PR #927, opened 2026-07-11
fix(scalar): validate dictionary indices before conversion

fallintoplace:
### What

Validate dictionary indices in their native signed or unsigned domain before converting them to int. Use the checked physical index for GetEncodedValue as well, so malformed scalars return an error instead of wrapping on 32-bit or 64-bit platforms.

### Tests

- `go test ./arrow/scalar`
- `GOOS=linux GOARCH=386 go test -c -o /dev/null ./arrow/scalar`

The broader `go test ./arrow/...` run is otherwise green for the reached packages but the checkout lacks the CSV/parquet fixture directories.

### DIFF
--- a/arrow/scalar/nested.go
+++ b/arrow/scalar/nested.go
@@ -435,43 +435,60 @@ func (s *Dictionary) ValidateFull() (err error) {
 		return nil
 	}
 
-	max := s.Value.Dict.Len() - 1
-	switch idx := s.Value.Index.value().(type) {
-	case int8:
-		if idx < 0 || int(idx) > max {
-			err = fmt.Errorf("%s scalar index value out of bounds: %d", s.DataType(), idx)
-		}
-	case uint8:
-		if int(idx) > max {
-			err = fmt.Errorf("%s scalar index value out of bounds: %d", s.DataType(), idx)
-		}
-	case int16:
-		if idx < 0 || int(idx) > max {
-			err = fmt.Errorf("%s scalar index value out of bounds: %d", s.DataType(), idx)
-		}
-	case uint16:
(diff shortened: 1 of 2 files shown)


arrow/scalar/nested.go
```diff
@@ -435,43 +435,60 @@ func (s *Dictionary) ValidateFull() (err error) {
...
+		return int(idx), nil
 	case uint64:
-		if int(idx) > max {
-			err = fmt.Errorf("%s scalar index value out of bounds: %d", s.DataType(), idx)
+		if idx >= uint64(length) {
+			return 0, outOfBounds(idx)
 		}
+		return int(idx), nil
+	default:
+		return 0, fmt.Errorf("unimplemented dictionary type %s", s.DataType())
```
zeroshade: Nit: report the index type here (`dt.IndexType`) rather than the whole dictionary type, matching the old `GetEncodedValue` default. Also consider adding an empty-dictionary (len == 0) out-of-range case to `TestDictionaryScalarIndexBounds`.

zeroshade (changes requested):
Nice — validating the physical index in its native integer domain before converting to `int` (`physicalIndex`) covers the full signed/unsigned int8–int64 set, so wrapped/negative indices no longer reach `GetScalar`. Routing both `ValidateFull` and `GetEncodedValue` through it is clean, and the test matrix (incl. MaxUint64, negatives) is thorough. Two small nits inline.

zeroshade (approved):
LGTM. Centralizing the dictionary scalar index validation with thorough coverage (signed/unsigned/max/negative/out-of-bounds/empty-dictionary) looks good. Thanks!

PR #928, opened 2026-07-11
fix(scalar): hash list values logically

fallintoplace:
### What

Hash list scalar values from their logical length, validity, and recursively hashed elements instead of physical ArrayData metadata. This keeps equal list scalars backed by independent arrays or slices on the same hash path.

### Tests

- `go test ./arrow/scalar ./arrow/array -count=1`
- `GOOS=linux GOARCH=386 go test -c -o /dev/null ./arrow/scalar`

Regression coverage includes independent backing arrays, zero-copy slices with different offsets, extra leading/trailing values, and different validity bitmap padding.

### DIFF
--- a/arrow/scalar/scalar.go
+++ b/arrow/scalar/scalar.go
@@ -995,6 +995,10 @@ func Hash(seed maphash.Seed, s Scalar) uint64 {
 		out ^= h.Sum64()
 		h.Reset()
 	}
+	hashUint64 := func(v uint64) {
+		binary.Write(&h, endian.Native, v)
+		hash()
+	}
 
 	valueHash := func(v interface{}) uint64 {
 		switch v := v.(type) {
@@ -1063,8 +1067,26 @@ func Hash(seed maphash.Seed, s Scalar) uint64 {
 	case TemporalScalar:
 		return valueHash(s.value())
 	case ListScalar:
-		array.Hash(&h, s.GetList().Data())
-		hash()
+		list := s.GetList()
+		hashUint64(uint64(list.Len()))
+		for i := 0; i < list.Len(); i++ {
+			if list.IsNull(i) {
+				h.Write([]byte{0})
+				hash()
+				continue
+			}
+
+			value, err := GetScalar(list, i)
+			if err != nil {
+				panic(err)
+			}
+			h.Write([]byte{1})
+			binary.Write(&h, endian.Native, Hash(seed, value))
+			hash()
+			if r, ok := value.(Releasable); ok {
+				r.Release()
+			}
+		}
 	case *Struct:
 		for _, c := range s.Value {
 			if c.IsValid() {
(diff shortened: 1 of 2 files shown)


arrow/scalar/scalar.go
```diff
@@ -1063,8 +1067,26 @@ func Hash(seed maphash.Seed, s Scalar) uint64 {
 	case TemporalScalar:
 		return valueHash(s.value())
 	case ListScalar:
-		array.Hash(&h, s.GetList().Data())
-		hash()
+		list := s.GetList()
+		hashUint64(uint64(list.Len()))
+		for i := 0; i < list.Len(); i++ {
```
zeroshade, line 1072: The fix correctly restores the equal-values-hash-equal contract, but the aggregation is order-insensitive and cancellation-prone: each element is folded via the `hash()` closure (`out ^= h.Sum64(); h.Reset()`), so element contributions XOR commutatively. As a result `[1, 2]` and `[2, 1]` hash identically (even though list equality is order-sensitive), and any repeated-value list cancels — e.g. `[x, x]` collapses to `typeHash ^ lenHash(2)` for every `x`, so `[1,1]`, `[2,2]`, ... all collide. It's not a correctness violation (collisions are permitted), but it's an avoidable systematic-collision regression for a hash function. Consider folding the whole sequence into a single running `maphash` state (no per-element `Reset`), or mixing the element index in before each value. Worth adding negative tests: `[1,2]` vs `[2,1]`, `[1,1]` vs `[2,2]`, and `[null,1]` vs `[1,null]`.

zeroshade (commented):
One follow-up on the list-scalar hashing — not blocking the core fix (restoring equal-values-hash-equal is correct), see inline.

arrow/scalar/scalar.go
```diff
@@ -1063,8 +1067,29 @@ func Hash(seed maphash.Seed, s Scalar) uint64 {
...
+			binary.Write(&h, endian.Native, uint64(i))
+			if list.IsNull(i) {
+				h.Write([]byte{0})
+				hash()
+				continue
+			}
+
+			value, err := GetScalar(list, i)
+			if err != nil {
+				panic(err)
```
zeroshade: `panic(err)` here turns a valid list into a crash. `NewListScalar` accepts any `arrow.Array` as its child values, and several types are hashable via `array.Hash` but unsupported by `GetScalar` — e.g. `string_view`, `binary_view`, `decimal32`, `decimal64`, `list_view`, `large_list_view`. Those hashed fine on the previous data-based path, so this is a regression. Please avoid making list hashing depend on `GetScalar` coverage (or add `GetScalar` support for every array type that can appear as list values before this is allowed to panic).

zeroshade (changes requested):
The switch to logical per-element hashing is a nice improvement, but it introduces a panic regression for some valid list child types — inline.

zeroshade (approved):

PR #929, opened 2026-07-11
fix(array): validate JSON start offsets

fallintoplace:
### What

Validate `WithStartOffset` before JSON decoding. Negative offsets are rejected, seek errors are returned, and a seeker must report the requested position. The same validation is shared by `FromJSON` and `RecordFromJSON`.

### Tests

Regression coverage includes seek errors, incorrect positions, negative offsets that must not call `Seek`, and a successful offset with the expected decoded values.

### DIFF
--- a/arrow/array/util.go
+++ b/arrow/array/util.go
@@ -74,6 +74,29 @@ func WithUseNumber() FromJSONOption {
 	}
 }
 
+func seekJSONStartOffset(r io.Reader, offset int64) error {
+	if offset < 0 {
+		return fmt.Errorf("seek JSON start offset must be non-negative: %d", offset)
+	}
+	if offset == 0 {
+		return nil
+	}
+
+	seeker, ok := r.(io.ReadSeeker)
+	if !ok {
+		return errors.New("using StartOffset option requires reader to be a ReadSeeker, cannot seek")
+	}
+
+	pos, err := seeker.Seek(offset, io.SeekStart)
+	if err != nil {
+		return fmt.Errorf("seek JSON start offset %d: %w", offset, err)
+	}
+	if pos != offset {
+		return fmt.Errorf("seek JSON start offset: got %d, want %d", pos, offset)
+	}
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


arrow/array/util.go
```diff
@@ -74,6 +74,29 @@ func WithUseNumber() FromJSONOption {
 	}
 }
 
+func seekJSONStartOffset(r io.Reader, offset int64) error {
```
zeroshade, line 77: Note: the error text changed (e.g. `RecordFromJSON` previously returned `failed to seek to start offset %d`). If any downstream code/tests assert on the old message, they'll need updating.

arrow/array/util.go
```diff
@@ -219,14 +237,8 @@ func RecordFromJSON(mem memory.Allocator, schema *arrow.Schema, r io.Reader, opt
...
 
-	if cfg.startOffset != 0 {
-		seeker, ok := r.(io.ReadSeeker)
-		if !ok {
-			return nil, 0, errors.New("using StartOffset option requires reader to be a ReadSeeker, cannot seek")
-		}
-		if _, err := seeker.Seek(cfg.startOffset, io.SeekStart); err != nil {
-			return nil, 0, fmt.Errorf("failed to seek to start offset %d: %w", cfg.startOffset, err)
-		}
+	if err := seekJSONStartOffset(r, cfg.startOffset); err != nil {
```
zeroshade, line 240: Since this shared helper now backs `RecordFromJSON` too, please add a direct `RecordFromJSON` test (one seek-failure path + one successful-offset path); the new tests currently only exercise `FromJSON`.

zeroshade (changes requested):
Good fix — `FromJSON` was discarding both the position and the error from `Seek`; centralizing this in `seekJSONStartOffset` (reject negative, propagate the seek error, verify the returned position) is the right call. Minor follow-ups inline.

zeroshade (approved):
LGTM. Centralizing and strengthening the JSON start-offset seek validation is the right call, and the coverage (seek failures, bad/negative offsets, successful decode) is good. Thanks!

PR #930, opened 2026-07-11
fix(array): recursively validate nested arrays

fallintoplace:
### What

Make top-level `Validate` and `ValidateFull` walk nested array data. Parent layouts, child counts and types, dictionary indices and values, and extension storage are checked before recursive validation. Errors include a path such as `field "orders" -> list values -> field "name"`. Ordinary list and large-list offsets are now validated fully as well.

### Tests

- `GOOS=linux GOARCH=386 go test -c -o /dev/null ./arrow/array`

Regression coverage includes malformed `Struct -> List -> Struct -> String`, dictionary values, and non-monotonic list offsets.

### DIFF
--- a/arrow/array/validate.go
+++ b/arrow/array/validate.go
@@ -20,10 +20,13 @@ import (
 	"fmt"
 
 	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/bitutil"
+	"github.com/apache/arrow-go/v18/arrow/memory"
 )
 
-// Validator is implemented by array types that can validate their internal
-// consistency. See Validate and ValidateFull for top-level dispatch.
+// Validator is implemented by array types that provide type-specific
+// consistency checks. Validate and ValidateFull also validate generic layout
+// and nested child data for every array type.
 type Validator interface {
 	arrow.Array
 	// Validate performs a basic O(1) consistency check.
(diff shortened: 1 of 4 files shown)


arrow/array/validate.go
```diff
@@ -33,32 +36,390 @@ type Validator interface {
...
+	if end < int64(data.offset) {
+		return fmt.Errorf("arrow/array: array offset and length overflow")
+	}
+
+	layout := data.dtype.Layout()
+	if len(data.buffers) < len(layout.Buffers) {
+		return fmt.Errorf("arrow/array: expected at least %d buffers for %s, got %d",
+			len(layout.Buffers), data.dtype, len(data.buffers))
+	}
+	if layout.VariadicSpec == nil && len(data.buffers) > len(layout.Buffers) {
```
zeroshade: Blocking: this rejects valid sparse/dense unions. `validateArray` runs `validateArrayData` for every array, and unions carry a leading `nil` validity placeholder in `Data` (sparse `[nil, typeIDs]`, dense `[nil, typeIDs, valueOffsets]`) that `UnionType.Layout()` omits (sparse = 1 buffer, dense = 2). So a valid sparse union has 2 buffers vs 1 expected (dense 3 vs 2) and fails here with the `expected at most N buffers` error — both at the top level and as a nested child. Please account for the union validity-slot convention (e.g. skip the leading nil for union layouts).

zeroshade (changes requested):
Thanks — recursively validating nested children, dictionaries, and list offsets is the right direction and looks good. One blocking issue inline: this currently rejects valid sparse/dense union arrays. Once that's fixed, please add union coverage — there are no `Validate`/`ValidateFull` tests for unions today, so CI stays green despite this. Nit: `validateListArray` permits MAP (via `MapType` implementing `ListLikeType`) while rejecting LARGE_LIST/LARGE_LIST_VIEW; a one-line comment there would help future readers.

arrow/array/validate.go
```diff
@@ -33,32 +36,401 @@ type Validator interface {
...
+			return validationError(path, err)
+		}
+	}
+
+	if data.DataType().Layout().HasDict {
+		dictData := data.dictionary
+		if dictData != nil {
+			dt := data.DataType().(*arrow.DictionaryType)
+			indexData := NewData(dt.IndexType, data.length, data.buffers, nil, data.nulls, data.offset)
+			err := checkIndexBounds(indexData, uint64(dictData.Len()))
```
zeroshade, line 91: `checkIndexBounds` runs from `validateArray` unconditionally, so it now executes for basic `Validate` (`full == false`) as well. `Validate` is documented as an O(1) layout check, but scanning every dictionary index makes it O(n) — a real regression for large dictionary columns. Please gate the index bounds scan behind `full` so it only runs in `ValidateFull`, keeping `Validate` constant-time.

arrow/array/validate.go
```diff
@@ -33,32 +36,401 @@ type Validator interface {
...
+				return err
+			}
+		}
 	}
 	return nil
 }
 
+func validateBuffer(buf *memory.Buffer, spec arrow.BufferSpec, end int64, length, index int) error {
+	if buf == nil {
+		if spec.Kind == arrow.KindFixedWidth || spec.Kind == arrow.KindVarWidth {
```
zeroshade: This rejects a nil var-width **data** buffer whenever `length > 0`, but a non-empty binary/string array can legitimately carry zero bytes of value data (e.g. all-empty strings: offsets `[0, 0, …]` with a nil/empty data buffer). `Binary.Validate`/`String.Validate` already allow this by checking the last offset against `len(valueBytes)`, so this generic check will falsely reject valid arrays before the type-specific validator runs. Please don't require var-width data buffers to be non-nil based on `length` alone — let the offset-derived length drive it (only require the buffer when the last offset is > 0).

zeroshade (changes requested):
Thanks for expanding the validation coverage here. Two behavior regressions in the basic `Validate` path need to be resolved before this can merge — details inline.

zeroshade (approved):

PR #931, opened 2026-07-11
fix(cdata): validate imported array sizes

fallintoplace:
The C array importer now validates all header counts and pointer arrays before creating native slices.

It also checks native-int conversion, length/offset arithmetic, bitmap and physical buffer sizes, binary-view size metadata, dictionary presence, and expected child/buffer counts before traversal. Invalid metadata returns an error instead of reaching unsafe buffer construction.

Tests cover negative and oversized headers, missing pointer arrays, root preflight before ownership transfer, and fixed-width/offset multiplication overflow paths.

Validation:
- go test -tags test ./arrow/cdata -count=1
- go test ./arrow/cdata -count=1
- pre-commit run golangci-lint-full --all-files
- git diff --check

The full arrow test suite is otherwise blocked by missing repository fixtures under parquet-testing/data and arrow-testing/data/csv. Linux/386 cgo compilation is unavailable on this macOS host.

### DIFF
--- a/arrow/cdata/cdata.go
+++ b/arrow/cdata/cdata.go
@@ -349,13 +349,147 @@ type cimporter struct {
 	alloc *importAllocator
 }
 
+const maxInt64Value = int64(1<<63 - 1)
+
 func (imp *cimporter) importChild(parent *cimporter, src *CArrowArray) error {
 	imp.parent, imp.arr, imp.alloc = parent, src, parent.alloc
 	return imp.doImport()
 }
 
+func validateCArrayHeader(arr *CArrowArray) error {
+	if arr == nil {
+		return fmt.Errorf("%w: nil ArrowArray", arrow.ErrInvalid)
+	}
+	length := int64(arr.length)
+	offset := int64(arr.offset)
+	nullCount := int64(arr.null_count)
+	nBuffers := int64(arr.n_buffers)
+	nChildren := int64(arr.n_children)
+
+	if length < 0 {
+		return fmt.Errorf("%w: ArrowArray length cannot be negative: %d", arrow.ErrInvalid, length)
+	}
+	if offset < 0 {
+		return fmt.Errorf("%w: ArrowArray offset cannot be negative: %d", arrow.ErrInvalid, offset)
+	}
+	if nullCount < -1 || nullCount > length {
+		return fmt.Errorf("%w: ArrowArray null_count %d is outside [-1, %d]", arrow.ErrInvalid, nullCount, length)
+	}
+	if nBuffers < 0 {
+		return fmt.Errorf("%w: ArrowArray n_buffers cannot be negative: %d", arrow.ErrInvalid, nBuffers)
+	}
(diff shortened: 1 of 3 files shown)


arrow/cdata/cdata.go
```diff
@@ -468,6 +605,16 @@ func (imp *cimporter) doImportArr(src *CArrowArray) error {
 // import is called recursively as needed for importing an array and its children
 // in order to generate array.Data objects
 func (imp *cimporter) doImport() error {
+	if err := validateCArrayHeader(imp.arr); err != nil {
+		return err
+	}
+	if _, ok := imp.dt.(*arrow.DictionaryType); ok && imp.arr.dictionary == nil {
```
zeroshade, line 611: Good — nil dictionary is handled here. Please add test coverage for this path, plus a missing string/binary-view size-metadata buffer and at least one wrong expected child/buffer count (struct/union/REE).

arrow/cdata/cdata.go
```diff
@@ -699,6 +849,12 @@ func (imp *cimporter) importBinaryViewLike() (err error) {
 	}
 
 	if len(buffers) > 2 {
+		if imp.cbuffers[len(buffers)] == nil {
+			return errors.New("invalid data buffer sizes")
```
zeroshade: Minor consistency: wrap this with `arrow.ErrInvalid` like the neighboring errors (same for the `invalid buffer` error around line 1027) so callers can classify malformed-input errors uniformly.

zeroshade (changes requested):
Solid hardening of the C Data Interface import path — validating header counts, null_count range, native-int fit, and buffer-size arithmetic before any `unsafe.Slice`, and doing so before `ArrowArrayMove` for the root, is exactly the right layer. Requesting changes for a couple of test/consistency nits inline. Worth a code comment somewhere: validating the claimed sizes can't guarantee the producer actually allocated that much, so `unsafe.Slice` over an undersized foreign buffer remains inherently trust-based — still a big improvement.

arrow/cdata/cdata.go
```diff
@@ -686,6 +833,9 @@ func (imp *cimporter) importBinaryViewLike() (err error) {
 	if err = imp.checkNoChildren(); err != nil {
 		return
 	}
+	if _, err = checkedMul(int64(imp.arr.n_buffers-3), int64(arrow.Int64SizeBytes)); err != nil {
```
zeroshade: nit: `imp.arr.n_buffers-3` subtracts in C `int64_t` before the cast to Go `int64`. It's safe today because `checkExpectedBuffers` already guarantees `n_buffers >= 3` before this point, but casting first — `int64(imp.arr.n_buffers) - 3` — is clearer and stays underflow-proof if that ordering ever changes.

zeroshade (commented):
Prior feedback is all addressed — thanks. Just one small clarity nit, non-blocking.

zeroshade (approved):

PR #935, opened 2026-07-13
fix(avro): decode nullable logical timestamps

abir-eon:
### Rationale for this change

This is a follow-up to https://github.com/apache/arrow-go/pull/832, which added support for Avro local timestamp logical types.

For nullable timestamp fields, Hamba decodes a selected logical union branch as a one-entry map such as `"long.local-timestamp-micros": time.Time`. `appendTimestampData` only inspected the plain `"long"` branch, so it treated valid non-null logical timestamp values as null.

- Recognize Hamba's exact union branch names for `timestamp-millis`, `timestamp-micros`, `local-timestamp-millis`, and `local-timestamp-micros`, while preserving plain `long` unions.
- Route decoded `time.Time` union values through the existing timestamp conversion, including UTC wall-clock reinterpretation for local timestamps.
- Add end-to-end nullable OCF coverage for all four timestamp logical types, with one non-null and one null record per type, plus focused appender coverage.

### Are these changes tested?

Yes. The regression failed before the fix because non-null logical timestamp values were appended as null. The full Avro suite passes in UTC and a non-UTC timezone:

- `TZ=UTC go test ./arrow/avro/... -count=1`
- `TZ=Asia/Jerusalem go test ./arrow/avro/... -count=1`
- `golangci-lint v2.11.3 run --timeout=5m ./arrow/avro/...`

Yes. Non-null values in nullable Avro timestamp logical fields are now decoded as Arrow timestamps instead of null.

### DIFF
--- a/arrow/avro/reader_test.go
+++ b/arrow/avro/reader_test.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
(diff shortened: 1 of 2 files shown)


arrow/avro/reader_test.go
```diff
@@ -276,6 +277,92 @@ func TestOCFReaderBytesValues(t *testing.T) {
 	assert.True(t, nullable.IsNull(1))
 }
 
+func TestOCFReaderNullableLocalTimestampMicros(t *testing.T) {
```
zeroshade: Optional: this end-to-end OCF test only exercises `local-timestamp-micros`. `TestAppendTimestampDataUnionBranches` covers all four branch names at the appender level, but OCF-level coverage for nullable global `timestamp-millis`/`timestamp-micros` and `local-timestamp-millis` would also catch any integration mismatch in Hamba's branch naming/decoding for those flavors.

zeroshade (commented):
Fix looks correct, and the union-branch handling preserves UTC vs. local timestamp semantics nicely. Just an optional test-coverage suggestion, non-blocking.

arrow/avro/reader_test.go
abir-eon: Addressed in c70bea62. TestOCFReaderNullableTimestamps now exercises timestamp-millis, timestamp-micros, local-timestamp-millis, and local-timestamp-micros end to end through OCF, with one non-null and one null record per type. Verified under TZ=UTC and TZ=Asia/Jerusalem. Thanks!

zeroshade (approved):

PR #936, opened 2026-07-14
fix(flight/flightsql/example): support dictionary-encoded parameter columns

fornwall:
**DISCLAIMER**: Most of this code was AI generated, but reviewed by me and I'm ready to answer questions, make changes or look into details.

When a client binds a dictionary-encoded parameter column to a prepared statement (e.g. a pyarrow `dictionary<values=string, indices=int32>` array bound via the ADBC Flight SQL driver), the example SQLite server failed with:

    error gathering parameters for prepared statement: unsupported
    type: ["apple" "banana"][0] (Internal; ExecuteQuery)

Dictionary encoding is an encoding, not a distinct logical type, so decode each value via the dictionary and bind the underlying value exactly as if the column were of the plain values type.

Dictionary encoding for parameters works with with e.g. the sqlite and postgresql adbc driver.

Proposed test in the ADBC driver validation suite that will test this: https://github.com/adbc-drivers/validation/pull/256

### DIFF
--- a/arrow/flight/flightsql/example/sqlite_server.go
+++ b/arrow/flight/flightsql/example/sqlite_server.go
@@ -581,6 +581,18 @@ func scalarToIFace(s scalar.Scalar) (interface{}, error) {
 		return val.ToTime(), nil
 	case *scalar.DenseUnion:
 		return scalarToIFace(val.Value)
+	case *scalar.Dictionary:
+		if err := val.ValidateFull(); err != nil {
+			return nil, err
+		}
+		value, err := val.GetEncodedValue()
+		if err != nil {
+			return nil, err
+		}
+		if r, ok := value.(scalar.Releasable); ok {
+			defer r.Release()
+		}
+		return scalarToIFace(value)
 	default:
 		return nil, fmt.Errorf("unsupported type: %s", val)
 	}
--- a/arrow/flight/flightsql/sqlite_server_test.go
+++ b/arrow/flight/flightsql/sqlite_server_test.go
@@ -615,6 +615,179 @@ func (s *FlightSqliteServerSuite) TestCommandPreparedStatementUpdateWithParams()
 	s.EqualValues(4, s.execCountQuery("SELECT COUNT(*) FROM intTable"))
 }
 
+func (s *FlightSqliteServerSuite) TestCommandPreparedStatementUpdateWithDictionaryParams() {
+	ctx := context.Background()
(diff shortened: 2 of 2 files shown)


arrow/flight/flightsql/example/sqlite_server.go
```diff
@@ -581,6 +581,15 @@ func scalarToIFace(s scalar.Scalar) (interface{}, error) {
 		return val.ToTime(), nil
 	case *scalar.DenseUnion:
 		return scalarToIFace(val.Value)
+	case *scalar.Dictionary:
+		value, err := val.GetEncodedValue()
```
zeroshade, line 588: This decodes client-supplied dictionary indices without a lower-bound check. `Dictionary.GetEncodedValue()` (arrow/scalar/nested.go:521) passes the raw signed index straight to `scalar.GetScalar`, which only guards the upper bound (`idx >= arr.Len()`, arrow/scalar/scalar.go:581) and never rejects negative indices. So a parameter batch with a `dictionary<values=string, indices=int32>` column whose index is e.g. `-1` reaches `arr.Value(-1)` and panics the server instead of returning a bind error.

Since these arrive from Flight clients, please validate before decoding — e.g. call `val.ValidateFull()` here and return the error — or add explicit bounds checks. A follow-up hardening `scalar.GetScalar`/`GetEncodedValue` to reject negative indices library-wide would be even better, since `NewDictionaryArray` doesn't bounds-check either.

Worth adding tests for: a negative/out-of-range index, a null value inside the dictionary, a non-string dictionary value type, and the prepared-statement query path (this currently only exercises ExecuteUpdate).

zeroshade (changes requested):
Thanks for adding dictionary-encoded parameter support — the happy path looks correct and the multi-row/null-index test is a nice touch. One blocking issue on validating client-supplied indices, plus a few test cases worth adding. Details inline.

arrow/flight/flightsql/example/sqlite_server.go
fornwall: > Since these arrive from Flight clients, please validate before decoding — e.g. call val.ValidateFull() here and return the error

👍 Done that now.

> hardening scalar.GetScalar/GetEncodedValue to reject negative indices library-wide

Split that out to #940. As I understand it `val.ValidateFull()` should still be used here, as it gives better error messages.

> Worth adding tests for

Done.

zeroshade (approved):

PR #939, opened 2026-07-14
feat(parquet/variant): add Metadata.SizeBytes to split a concatenated variant

nssalian:
### Rationale for this change

A variant bound is stored on disk as the variant metadata bytes concatenated with the value bytes (metadata ++ value), as Iceberg does for variant `lower_bound` / `upper_bound`. `Metadata.Bytes()` returns the backing slice as-is, so a caller that reads such a concatenated buffer has no way to know where the metadata ends and the
value begins. `NewMetadata` accepts a buffer with trailing bytes but does not expose the metadata's own length.

iceberg-go needs this to read variant bounds back for scan pruning (apache/iceberg-go#1016): it must split a stored bound into metadata and value to reconstruct the variant. Filing it now so it is released ahead of that work.

- Adds `Metadata.SizeBytes()`, which computes the metadata's own byte length from its header and offset table (the final offset gives the string-region length past `valuesStart`). A caller can then split a buffer at `data[m.SizeBytes():]`.

### Are these changes tested?

Yes. `TestMetadataSizeBytes` covers empty metadata, a keyed object, and splitting a concatenated metadata++value buffer with a round-trip through `New`.`TestMetadataSizeBytesMultiByteOffset` covers the multi-byte offset-width path.

One new exported method. No change to existing behavior.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -231,6 +231,25 @@ func (m Metadata) OffsetSize() uint8 {
 // DictionarySize returns the number of keys in the metadata dictionary.
 func (m Metadata) DictionarySize() uint32 { return uint32(len(m.keys)) }
 
+// SizeBytes returns the metadata's own byte length, so a caller can split a
+// buffer that stores metadata concatenated with a value at data[m.SizeBytes():].
+func (m Metadata) SizeBytes() int {
+	offsetSz := uint32(m.OffsetSize())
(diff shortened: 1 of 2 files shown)


nssalian:
~@zeroshade the failure seems unrelated. I'm unable to re-run.~
CI run passed

zeroshade (approved):

PR #940, opened 2026-07-14
fix(arrow/scalar): reject negative indices in GetScalar

fornwall:
`GetScalar` only guarded the upper bound (`idx >= arr.Len()`), and the `IsNull(idx)` call ran before that check, so a negative index panicked inside the null-bitmap lookup instead of returning an error.

Check both bounds first and return `arrow.ErrIndex` for any out-of-range index.

Split out from review comment in #936.


zeroshade (approved):

PR #941, opened 2026-07-15
fix(flight): reject malformed unary bearer headers

fallintoplace:
### Rationale for this change

The unary bearer interceptor used a prefix check and then sliced past an assumed space. An authorization value containing exactly `Bearer` could therefore panic, while values such as `BearerX token` were treated as bearer credentials.

Parse the authorization scheme and credential separately, require the exact Bearer scheme and a non-empty credential, and reject malformed headers before invoking the validator. Additional spaces before the credential are accepted.

### Are these changes tested?

Yes. Regression coverage includes missing headers, missing credentials, malformed schemes, and valid credentials separated by multiple spaces. `go test ./arrow/flight` and the race-enabled package test pass.

Missing or malformed bearer headers now return Unauthenticated instead of reaching the validator or panicking. Multiple spaces after the scheme are accepted.

### DIFF
--- a/arrow/flight/basic_auth_flight_test.go
+++ b/arrow/flight/basic_auth_flight_test.go
@@ -151,6 +151,27 @@ func TestErrorAuths(t *testing.T) {
 			t.Fatal("should have errored")
 		}
 	})
+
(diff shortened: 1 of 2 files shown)


zeroshade (approved):

PR #942, opened 2026-07-15
fix(flight): reject Basic credentials without a separator

fallintoplace:
### Rationale for this change

A valid Base64 Basic credential without a colon decoded to a one-element slice. The streaming authentication interceptor then indexed the missing password and panicked.

Use `strings.Cut` to split the decoded username and password, and return an Unauthenticated error when the separator is absent. Empty usernames and passwords remain validator policy.

### Are these changes tested?

Yes. Regression cases cover empty input, a username without a separator, empty username/password values, and passwords containing additional colons. `go test ./arrow/flight` passes.

Malformed Basic credentials now return Unauthenticated instead of panicking.

### DIFF
--- a/arrow/flight/basic_auth_flight_test.go
+++ b/arrow/flight/basic_auth_flight_test.go
@@ -18,6 +18,7 @@ package flight_test
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"io"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):

PR #945, opened 2026-07-15
fix(parquet): return errors for malformed encrypted payloads

fallintoplace:
### Rationale for this change

The AES decryption helpers trusted length prefixes and nonce/tag sizes, allocated directly from the declared length, and panicked on framing or authentication failures. Malformed encrypted input should be reported through the existing reader error paths.

- Validate length prefixes, nonce/tag minimums, and complete frames before slicing.
- Bound streaming ciphertext allocation with a caller-provided maximum.
- Return GCM authentication and framing errors from the decryption interface.
- Propagate decryption failures through file metadata, page data, page headers, bloom filters, and page indexes.
- Preserve incremental encrypted page-header reads by waiting for a complete frame before decryption.

### Are these changes tested?

Yes. Unit tests cover missing/truncated frames, undersized payloads, oversized declarations, trailing data, and invalid tags. The encryption, metadata, file-reader, and fixture-backed encryption suites pass.

Malformed encrypted Parquet data now returns an error instead of panicking or attempting an unbounded allocation. The changed decryption interface is under an internal package.

### DIFF
--- a/parquet/file/file_reader_test.go
+++ b/parquet/file/file_reader_test.go
@@ -282,9 +282,9 @@ func (m *mockDecryptor) Allocator() memory.Allocator {
 
 func (m *mockDecryptor) CiphertextSizeDelta() int { return 0 }
 
-func (m *mockDecryptor) Decrypt(src []byte) []byte { return m.decrypt(src) }
+func (m *mockDecryptor) Decrypt(src []byte) ([]byte, error) { return m.decrypt(src), nil }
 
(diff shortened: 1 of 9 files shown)


zeroshade (approved):
LGTM — verified malformed-encrypted-payload handling and that all decrypt call sites are updated; added checks are O(1) framing/length validation before AES-GCM, no perf-sensitive path change. Thanks!

PR #947, opened 2026-07-15
fix(memory): keep allocator ownership across Buffer.Reset

fallintoplace:
### Rationale for this change

`Buffer.Reset` replaced an allocator-owned slice directly. That leaked the old allocation and later caused `Release` to pass the caller supplied slice to the allocator.

When a buffer has an allocator, copy replacement data into allocator-owned storage and resize that storage without transferring ownership of the caller supplied slice. For shrinking resets, copying before resize also keeps resets from an overlapping slice safe if the allocator moves the allocation. Non-owning buffers keep the existing zero-copy behavior.

### Are these changes tested?

Yes. Tests cover shrinking and growing from unrelated slices, source-slice independence, immediate release through `Reset(nil)`, final allocator accounting, and an overlapping reset with an allocator that always moves and invalidates the old allocation. Package, race-enabled, and lint checks pass.

Resetting an allocator-backed buffer now copies the supplied data so ownership, growth, shrinking, overlapping resets, and release behavior remain valid.

### DIFF
--- a/arrow/memory/buffer.go
+++ b/arrow/memory/buffer.go
@@ -95,12 +95,25 @@ func (b *Buffer) Release() {
 	}
 }
 
-// Reset resets the buffer for reuse.
+// Reset resets the buffer for reuse. If the buffer owns memory through an
+// allocator, the supplied data is copied into allocator-owned storage.
 func (b *Buffer) Reset(buf []byte) {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):

Issue #951, opened 2026-07-15
arrow: reject invalid timestamp timezone metadata


fallintoplace:
### Describe the bug

`TimestampType.GetZone` correctly reports an error unless the timezone is empty, a recognized timezone database name, or an absolute offset. Some construction and parsing paths do not preserve that invariant:

- `arrow/ipc.timestampFromFB` accepts the serialized timezone string without validation.
- `Timestamp.ValueStr` discards the error from `GetToTimeFunc` and calls the returned function, which is nil for an invalid timezone.
- `TimestampBuilder.UnmarshalOne` ignores the error from `GetZone`; the fallback location can cause invalid timezone metadata to be interpreted as UTC.

`TimestampBuilder.AppendValueFromString` already propagates `GetZone` errors and provides a useful model for the builder fix.

Invalid timezone metadata should be rejected at an error-returning boundary rather than becoming a later formatting panic or silent UTC interpretation.

### Suggested scope

- Validate timestamp timezone strings while decoding IPC schemas.
- Propagate `GetZone` errors from JSON unmarshalling.
- Decide which other error-returning datatype or array construction boundaries should validate timestamp metadata.
- Keep `ValueStr` defensive, while allowing it to rely on a documented valid-datatype invariant where possible.

### Regression coverage

- Invalid timezone database name.
- Malformed absolute offset.
- Valid timezone database name and valid fixed offsets.
- IPC schema decoding with invalid timezone metadata.
- JSON unmarshal and `ValueStr` behavior.

Observed on current `main` at `750586710988bc18f7af64fd29af8e0acdb252ac`.

### Component(s)

Other

PR #952, opened 2026-07-15
fix(arrow): validate timestamp timezones

fallintoplace:
### Rationale for this change

TimestampBuilder JSON unmarshalling can silently treat invalid timezone metadata as UTC for timestamp strings, while numeric representations bypass timezone validation entirely.

Refs #951

- Resolve the timestamp timezone before dispatching on the JSON value representation.
- Propagate timezone lookup failures without appending a value.
- Keep IPC schema decoding independent of the system timezone database.
- Preserve compute cast coverage by constructing deliberately invalid timestamp arrays without going through JSON unmarshalling.

### Are these changes tested?

Yes. Tests cover timestamp strings, integer strings, `json.Number`, and `float64` values with invalid timezone metadata. The arrow/array, arrow/ipc, and arrow/compute package tests, including race tests, pass.

TimestampBuilder JSON unmarshalling now returns an error for invalid timezone names regardless of the JSON value representation. IPC schema decoding remains unchanged.

### DIFF
--- a/arrow/array/timestamp.go
+++ b/arrow/array/timestamp.go
@@ -336,6 +336,11 @@ func (b *TimestampBuilder) AppendValueFromString(s string) error {
 }
 
 func (b *TimestampBuilder) UnmarshalOne(dec *json.Decoder) error {
+	loc, err := b.dtype.GetZone()
+	if err != nil {
+		return err
+	}
+
 	t, err := dec.Token()
 	if err != nil {
 		return err
@@ -349,7 +354,6 @@ func (b *TimestampBuilder) UnmarshalOne(dec *json.Decoder) error {
 			b.Append(arrow.Timestamp(i))
 			break
 		}
-		loc, _ := b.dtype.GetZone()
 		tm, _, err := arrow.TimestampFromStringInLocation(v, b.dtype.Unit, loc)
 		if err != nil {
 			return &json.UnmarshalTypeError{
--- a/arrow/array/timestamp_test.go
+++ b/arrow/array/timestamp_test.go
@@ -17,6 +17,7 @@
 package array_test
 
 import (
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


arrow/ipc/metadata.go
```diff
@@ -986,7 +986,11 @@ func timeFromFB(data flatbuf.Time) (arrow.DataType, error) {
 func timestampFromFB(data flatbuf.Timestamp) (arrow.DataType, error) {
 	unit := unitFromFB(data.Unit())
 	tz := string(data.Timezone())
-	return &arrow.TimestampType{Unit: unit, TimeZone: tz}, nil
+	typ := &arrow.TimestampType{Unit: unit, TimeZone: tz}
+	if _, err := typ.GetZone(); err != nil {
```
zeroshade: `GetZone()` resolves IANA names via `time.LoadLocation`, which needs the system/embedded tz database. Gating IPC schema *decode* on it means a valid schema with `tz=America/New_York` is **rejected** on a minimal/scratch/container runtime that lacks tzdata (and doesn't import the embedded `time/tzdata`) — a regression, since decode previously carried the tz string through untouched. Empty tz and fixed offsets like `+09:00` are unaffected (they skip `LoadLocation`). Consider not failing decode on IANA resolution here, or documenting that IPC reads now require tzdata.

arrow/array/timestamp.go
```diff
@@ -349,7 +349,10 @@ func (b *TimestampBuilder) UnmarshalOne(dec *json.Decoder) error {
 			b.Append(arrow.Timestamp(i))
 			break
 		}
-		loc, _ := b.dtype.GetZone()
+		loc, err := b.dtype.GetZone()
```
zeroshade: This validates the tz only on the non-integer **string** path. Numeric JSON still bypasses it — the `case json.Number:` and `case float64:` branches below just append without calling `GetZone()`. So a builder whose type has an invalid `TimeZone` still accepts numeric JSON like `[0]` without error. If this boundary is meant to reject invalid timezones, validate once up front regardless of the value representation.

zeroshade (changes requested):
Validating timezones is a good idea, and empty tz + fixed offsets are handled correctly via `GetZone`. Two issues before merge: enforcing `GetZone` during IPC schema decode can false-reject valid IANA names on systems without tzdata, and the JSON builder only validates some value representations. Detail inline.

zeroshade (approved):
Re-review: IPC schema decode is now tzdata-independent (dropped the GetZone gate) and JSON timezone validation is applied upfront across all value types (string/number/float), with a 4-path test. LGTM — thanks!

Issue #950, opened 2026-07-15
parquet: define safe page-size and decompression limits


fallintoplace:
### Describe the enhancement requested

Parquet page headers provide signed 32-bit compressed and uncompressed sizes. The page reader rejects negative values but otherwise accepts positive values up to `math.MaxInt32`.

Several paths then resize buffers from those declarations before proving that the body fits inside the remaining column chunk or that the uncompressed size is reasonable. Streaming support reduces the allocation risk for some data-page encodings, but dictionary pages and non-streaming paths still allocate from the declared size.

Reader-side limits would provide a predictable resource boundary for corrupt or unexpectedly large files.

### Proposed direction

Add reader properties for limits such as:

- Maximum compressed page size.
- Maximum uncompressed page size.
- Optionally, maximum decompression ratio.

Before allocating or decoding:

- Check the compressed size against the remaining column-chunk bytes.
- Validate compressed and uncompressed sizes against configured limits.
- Use checked arithmetic for level and value-region sizes.
- Bound streaming readers to the validated page body.
- Preserve an explicit override for workloads with unusually large valid pages.

The defaults need maintainer agreement because existing users may read pages larger than common recommendations.

### Regression coverage

- `math.MaxInt32` compressed and uncompressed declarations.
- A compressed page extending beyond its column chunk.
- Small compressed input declaring a very large output.
- Dictionary pages and data page versions 1 and 2.
- Allocator accounting showing rejection before a large allocation.

Observed on current `main` at `750586710988bc18f7af64fd29af8e0acdb252ac`.

### Component(s)

Parquet

PR #953, opened 2026-07-15
fix(parquet): limit page allocations

fallintoplace:
### Rationale for this change

Parquet page sizes come from file metadata and are used to resize buffers. Positive int32 values can request multi-gigabyte allocations before decoding proves that the page is usable.

Refs #950

- Add configurable compressed and uncompressed page-size limits to ReaderProperties.
- Use defaults of 64 MiB compressed and 256 MiB uncompressed.
- Apply the limits to dictionary pages and data page versions 1 and 2 before buffer allocation or decompression.
- Preserve explicit overrides for workloads with unusually large valid pages.
- Apply the same limits to encrypted and unencrypted column readers.

### Are these changes tested?

Yes. Focused tests cover compressed, uncompressed, negative, and math.MaxInt32 declarations. Page-reader tests and vet checks pass.

The full Parquet suite also requires the external parquet-testing data checkout, which is not present in this environment.

Readers now reject pages above the default limits. Applications that intentionally read larger pages can raise MaxCompressedPageSize and MaxUncompressedPageSize in ReaderProperties.

### DIFF
--- a/parquet/file/page_reader.go
+++ b/parquet/file/page_reader.go
@@ -421,9 +421,11 @@ type serializedPageReader struct {
 	maxRepLevel     int16
 	maxDefLevel     int16
 
-	curPageHdr        *format.PageHeader
-	pageOrd           int16
-	maxPageHeaderSize int
+	curPageHdr              *format.PageHeader
+	pageOrd                 int16
+	maxPageHeaderSize       int
+	maxCompressedPageSize   int64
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM — verified no benchmark regression (BenchmarkReadInt32Column flat via benchstat) with sane 64/256 MiB defaults applied before allocation across dict + data pages. Thanks!

Issue #949, opened 2026-07-15
parquet/compress: design an error-returning decompression API


fallintoplace:
### Describe the enhancement requested

`parquet/compress.Codec` currently exposes block decompression as:

```go
Decode(dst, src []byte) []byte
```

There is no error channel, so the built-in codecs panic when compressed input is malformed. The Parquet page reader invokes these codecs with bytes read from a file, which means a corrupt page can escape as a process panic instead of a normal read error.

`StreamingCodec.NewReader` has a similar constraint: it cannot report initialization failures directly. The interfaces are public and `RegisterCodec` explicitly supports third-party implementations, so changing them requires a compatibility plan.

### Design questions

- Add a new error-returning interface and adapt legacy codecs, or change `Codec` in the next compatible major version.
- Decide how registered custom codecs migrate.
- Decide whether streaming reader creation should also return an error.
- Keep panic recovery, if any, inside one compatibility adapter rather than page-reader call sites.

### Desired outcome

- Malformed Snappy, Gzip, Brotli, Zstd, and LZ4 input returns a contextual Parquet read error.
- File-reader codec calls do not expose codec panics.
- Custom codec compatibility and migration are documented.
- Fuzz coverage includes empty, truncated, bit-flipped, and trailing input for every registered codec.

Observed on current `main` at `750586710988bc18f7af64fd29af8e0acdb252ac`.

### Component(s)

Parquet

PR #954, opened 2026-07-15
fix(parquet): return decompression errors

fallintoplace:
### Rationale for this change

Malformed compressed page data should be reported as a Parquet read error, but the existing block codec API has no error return and built-in codecs panic on decode failures.

Refs #949

- Add an optional ErrorCodec interface and a safe Decode adapter without breaking existing registered codecs.
- Implement error-returning block decoding for Snappy, Gzip, Brotli, Zstd, LZ4_RAW, and the uncompressed codec.
- Recover decode panics from legacy custom codecs in one compatibility adapter.
- Route Parquet page decompression through the adapter and add contextual errors.
- Validate complete gzip and Brotli output, including checksum and trailing output when a destination is supplied.

### Are these changes tested?

Yes. Tests cover malformed input for every built-in compressed block codec and panic recovery for a legacy codec. The compression package passes its race test, and focused page-reader tests pass.

The API change is additive. Existing Codec implementations continue to compile, while codecs can implement ErrorCodec to return decoding failures directly. File reads now return malformed-compression errors instead of exposing codec panics.

### DIFF
--- a/parquet/compress/brotli.go
+++ b/parquet/compress/brotli.go
@@ -18,6 +18,7 @@ package compress
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 
 	"github.com/andybalholm/brotli"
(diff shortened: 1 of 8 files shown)


zeroshade (approved):
LGTM — verified no decode-path benchmark regression (benchstat); clean additive ErrorCodec API with legacy panic-recovery and updated call sites. Thanks!

Issue #948, opened 2026-07-15
ipc: validate stream and file block lengths before allocation


fallintoplace:
### Describe the bug

The IPC stream and file readers use several serialized length and offset fields before fully validating them.

In `arrow/ipc/message.go`, the stream reader converts the unsigned metadata prefix to `int32` and uses it as an allocation size. Values above `math.MaxInt32` become negative, while large positive values can request very large allocations. The FlatBuffer body length is also converted to `int` and passed to `Buffer.Resize` without checking that it is nonnegative, representable, or within a configured limit.

The file reader has the same class of problem around footer blocks. Alignment is checked, but negative aligned offsets and lengths still pass. `mappedFileBlock.section` slices using footer-controlled offset, metadata length, and body length before checking their combined range. The ordinary block reader also assumes enough metadata exists before reading the prefix. Finally, `RecordBatchAt` accepts `i == NumRecords()` because its bounds check uses `>` instead of `>=`.

Malformed or truncated IPC input should return a contextual error before allocating, parsing, or slicing.

### Suggested scope

- Validate metadata and body lengths before integer conversion or allocation.
- Reject negative block offsets and lengths.
- Use checked addition for `offset + metadata + body`.
- Require sufficient bytes for continuation prefixes.
- Share block validation between mapped and ordinary readers.
- Fix the `RecordBatchAt` upper-bound check.
- Decide whether reader options should expose metadata and body size limits, with safe defaults.

### Regression coverage

- Metadata prefixes `0x7fffffff`, `0x80000000`, and `0xffffffff`.
- Metadata shorter than the required FlatBuffer or continuation prefix.
- Negative and overflowing file block fields.
- A mapped block ending one byte beyond the input.
- Negative or oversized body lengths.
- `RecordBatchAt(NumRecords())`.

Observed on current `main` at `750586710988bc18f7af64fd29af8e0acdb252ac`.

### Component(s)

Other

PR #955, opened 2026-07-15
fix(ipc): validate message framing bounds

fallintoplace:
### Rationale for this change

Malformed IPC framing fields can currently reach allocations, FlatBuffer access, or mapped-file slicing before their ranges are validated. A stream can also advertise a huge but addressable body and trigger the allocation before truncation is discovered.

Refs #948

- Add configurable IPC metadata and body size limits, with defaults of 64 MiB and 256 MiB respectively.
- Allow callers to disable either limit explicitly by setting it to zero.
- Reject invalid stream metadata and body lengths before allocation.
- Validate file block offsets, sizes, checked arithmetic, and continuation-prefix space for regular and mapped readers.
- Fix the upper-bound check in `RecordBatchAt`.

### Are these changes tested?

Yes. The IPC package tests cover oversized and signed metadata prefixes; negative, non-addressable, and huge-but-addressable body lengths; negative and overflowing block fields; out-of-file ranges; valid block bounds; and the zero-value body-limit opt-out. A rejecting test allocator ensures oversized body regressions fail without attempting a large allocation. IPC tests, race tests, vet, repeated regressions, and 32-bit compilation pass.

Malformed or over-limit IPC input now returns an error instead of reaching an allocation or panic. Callers that intentionally use unusually large messages can override the limits with `WithMetadataSizeLimit` and `WithBodySizeLimit`; passing zero disables the corresponding limit.

### DIFF
--- a/arrow/ipc/message.go
+++ b/arrow/ipc/message.go
@@ -149,8 +149,10 @@ type messageReader struct {
 	refCount atomic.Int64
 	msg      *Message
 
-	mem    memory.Allocator
-	header [4]byte
+	mem             memory.Allocator
+	maxMetadataSize int64
+	maxBodySize     int64
+	header          [4]byte
 }
 
 // NewMessageReader returns a reader that reads messages from an input stream.
@@ -160,7 +162,12 @@ func NewMessageReader(r io.Reader, opts ...Option) MessageReader {
 		opt(cfg)
 	}
 
-	mr := &messageReader{r: r, mem: cfg.alloc}
+	mr := &messageReader{
+		r:               r,
+		mem:             cfg.alloc,
+		maxMetadataSize: cfg.maxMetadataSize,
+		maxBodySize:     cfg.maxBodySize,
+	}
 	mr.refCount.Add(1)
 	return mr
 }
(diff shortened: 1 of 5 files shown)


arrow/ipc/message.go
```diff
@@ -227,6 +247,13 @@ func (r *messageReader) Message() (*Message, error) {
 
 	meta := flatbuf.GetRootAsMessage(buf, 0)
 	bodyLen := meta.BodyLength()
+	maxInt := int64(^uint(0) >> 1)
+	if bodyLen < 0 || bodyLen > maxInt {
+		return nil, fmt.Errorf("arrow/ipc: invalid message body length %d", bodyLen)
+	}
+	if r.maxBodySize > 0 && bodyLen > r.maxBodySize {
```
zeroshade, line 254: By default `maxBodySize` is `math.MaxInt` (set in ipc.go), so on 64-bit a malformed stream advertising a huge-but-addressable `BodyLength` (e.g. `1<<40`) passes this check and reaches `body.Resize(int(bodyLen))` at L260 — a ~1 TiB allocation attempt — *before* `io.ReadFull` can discover the stream is truncated. That's an allocation-based DoS from a single malformed message on the default reader.

Suggest a sane default cap (cf. #953, which defaults to 64 MiB compressed / 256 MiB uncompressed) with an explicit opt-out for `0`, and/or bounding the allocation to the bytes actually available. Also worth adding tests for negative `BodyLength`, `BodyLength > MaxInt`, and a huge-but-addressable body over the default limit — the current tests cover metadata-length prefixes and file blocks but not stream body rejection.

zeroshade (changes requested):
This meaningfully hardens IPC framing — the metadata-length cap, negative/overflow rejection, and file-block validation are all good, and the `RecordBatchAt` off-by-one fix is correct. One gap: the stream body-length limit is effectively disabled by default, so a malformed huge body length still triggers a large allocation before truncation is detected. Detail inline.

zeroshade (approved):
Re-review: sane 64 MiB/256 MiB defaults with validation before Resize and a rejectingAllocator test proving no allocation on a bogus BodyLength. LGTM — thanks!

PR #957, opened 2026-07-16
fix(arrow/array): validate map builder entry lengths

fallintoplace:
### Rationale for this change

MapBuilder sizes its child struct from the key builder without first checking the item builder. Unequal key and item counts can therefore create malformed map data or defer the failure until later array access.

- Reject unequal key and item builder lengths before array construction resets the builders.
- Reject a child struct length that exceeds the entry count.
- Validate that the offset count matches the map count.
- Validate any explicitly supplied final offset against the available entries.
- Report invariant failures as panics wrapping arrow.ErrInvalid, consistent with the builder API.

### Are these changes tested?

Yes. Tests cover extra keys, extra items, an oversized child struct, too few and too many offsets, and a final offset beyond the available entries. The array package passes normal and race tests.

Malformed map builder state now fails immediately with a descriptive invalid-data panic instead of constructing an inconsistent array or panicking during later access. Valid builder usage is unchanged.

### DIFF
--- a/arrow/array/map.go
+++ b/arrow/array/map.go
@@ -274,13 +274,35 @@ func (b *MapBuilder) init(capacity int)                  { b.listBuilder.init(ca
 func (b *MapBuilder) resize(newBits int, init func(int)) { b.listBuilder.resize(newBits, init) }
 
 func (b *MapBuilder) adjustStructBuilderLen() {
+	keyLen, itemLen := b.keyBuilder.Len(), b.itemBuilder.Len()
+	if keyLen != itemLen {
+		panic(fmt.Errorf("%w: arrow/array: map key and item builders must have equal lengths (keys=%d, items=%d)",
+			arrow.ErrInvalid, keyLen, itemLen))
+	}
+
 	sb := b.listBuilder.ValueBuilder().(*StructBuilder)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Nice hardening. The invariants are correct and match the ListBuilder offset bookkeeping: after N `Append`s `offsets.Len()==Len()` (the terminating offset is added later by the list builder), and when an explicit final offset is supplied (`Len()+1`) requiring it to equal the entry count matches what the list builder would otherwise rely on. `keyBuilder`/`itemBuilder` are the struct's field builders, so the struct-length guard correctly catches struct-validity rows with no key/item values, and the `*Int32Builder` offsets cast is always safe (Map never uses int64 offsets).

Verified on the PR head (`17aefd5d`):
- The new `TestMapBuilderRejectsInvalidEntryLengths` subtests all pass, and the full `./array/` suite passes (incl. `-race`).
- No existing map-building path is affected: `./internal/arrjson/`, `./ipc/`, `./scalar/`, `./array/arreflect/` all pass and `flightsql` vet is clean. All real MapBuilder usages append balanced key+item pairs.

Panics wrap `arrow.ErrInvalid` consistent with the builder API. LGTM.

PR #958, opened 2026-07-18
ci: shard the benchmark workflow across parallel jobs

zeroshade:
### Rationale for this change

The `Benchmarks` workflow runs `go test -bench=. ./...` sequentially across every
package. Because `go test`'s `-timeout` is applied per package, the wall-clock time
is the sum over all packages — recent `main` runs have taken roughly 3.3 hours
(198–208 min).

Split the benchmark run so it can be parallelized, then combine the results into a
single upload:

- **`ci/scripts/bench.sh`** — adds `--run` (benchmark a subset of packages, writing
  raw output to a `.dat`) and `--aggregate` (merge one or more `.dat` files into a
  single `bench_stats.json` via `gobenchdata`) modes. The existing
  `bench.sh <dir> [--json|-json]` interface is unchanged, so nothing else that calls
  it needs to change.
- **`ci/scripts/bench_shard.sh`** (new) — prints a GitHub Actions matrix that buckets
  the packages containing benchmarks into N shards.
- **`.github/workflows/benchmark.yml`** — reworked into three jobs: `setup` (compute
  the shard matrix) → `benchmark` (matrix; each shard runs its packages and uploads
  its `.dat`) → `combine` (download all `.dat`, aggregate into one `bench_stats.json`,
  and — only on push to `main` — upload once to Conbench).
- **`ci/scripts/bench_adapt.py`** — reuses an existing `bench_stats.json` (produced by
  `combine`) instead of re-running the whole suite.

Because the shards are merged into one JSON and uploaded once, Conbench still sees a
single run (no `run_id` fragmentation).

### Are these changes tested?

Locally:
- `shellcheck` clean on both scripts; `actionlint` clean on the workflow; `py_compile`
  OK on `bench_adapt.py`.
- Verified the split→merge end to end: ran `--run` on two packages, then `--aggregate`
  produced one `bench_stats.json` containing both suites, in the exact shape
  `bench_adapt.py` consumes.
- The legacy `bench.sh <dir> --json` path still runs → aggregates → cleans up.

Opened as a **draft** to exercise the reworked workflow in CI end to end (it triggers
on changes to these files).

No. This only touches CI / benchmark tooling.

### Notes / follow-ups

- Sharding is currently round-robin by package, not runtime-weighted, so a single
  shard can hold two heavy packages (e.g. `arrow/compute` + `parquet/internal/encoding`)
  and become the long pole. The per-package `-timeout` (40m) remains the hard floor for
  any single package. Once this runs, per-shard timings can seed a runtime-weighted
  split or tune the shard count.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -23,36 +23,81 @@ on:
     paths:
       - ".github/workflows/benchmark.yml"
       - "ci/scripts/bench.sh"
+      - "ci/scripts/bench_shard.sh"
       - "ci/scripts/bench_adapt.py"
   workflow_dispatch:
 permissions:
   contents: read
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.shards.outputs.matrix }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7.0.0
+      - name: Compute benchmark shards
+        id: shards
+        run: echo "matrix=$(bash ci/scripts/bench_shard.sh 6)" >> "$GITHUB_OUTPUT"
   benchmark:
+    needs: setup
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        go: ['1.26.1']
-        arch: ['amd64']
+        include: ${{ fromJson(needs.setup.outputs.matrix) }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7.0.0
         with:
           submodules: recursive
-      - name: Set up Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
-        with:
-          python-version: '3.9'
-      - name: Install Go ${{ matrix.go }} for Benchmarks
+      - name: Install Go for Benchmarks
         uses: actions/setup-go@v6.5.0
         with:
-          go-version: ${{ matrix.go }}
+          go-version: '1.26.1'
           cache: true
           cache-dependency-path: go.sum
           check-latest: false
-      - name: Run Benchmarks
-        if: github.event_name != 'push'
-        run: bash ci/scripts/bench.sh $(pwd) --json
+      - name: Run benchmark shard ${{ matrix.id }}
+        run: bash ci/scripts/bench.sh "$(pwd)" --run --packages "${{ matrix.packages }}" --out "bench_stat_${{ matrix.id }}.dat"
+      - name: Upload shard results
+        uses: actions/upload-artifact@v4
+        with:
+          name: bench-dat-${{ matrix.id }}
+          path: bench_stat_${{ matrix.id }}.dat
+          if-no-files-found: error
+          retention-days: 1
+  combine:
+    needs: benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7.0.0
+      - name: Install Go for Benchmarks
+        uses: actions/setup-go@v6.5.0
+        with:
+          go-version: '1.26.1'
+          cache: true
+          cache-dependency-path: go.sum
+          check-latest: false
+      - name: Download shard results
+        uses: actions/download-artifact@v4
+        with:
+          pattern: bench-dat-*
+          path: bench-dat
+          merge-multiple: true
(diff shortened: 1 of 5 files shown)


lidavidm (commented):
Since we've been trying to reduce the CI usage...the aggregate runner minutes used is still about the same, right? 3 hours of worker resources/commit seems pretty heavy

zeroshade:
Pushed a follow-up in 797891462301:

- Capture machine metadata on each benchmark runner and carry it through the combined upload, rather than attributing measurements to the combine runner.
- Require explicit results/provenance arguments for reuse; ordinary local runs always take fresh measurements.
- Preserve spaces during aggregation, add the portable timeout fallback, and use the current repository-pinned artifact actions.

Verified with two real Go benchmark shards, per-package machine attribution and one run ID, missing/conflicting provenance rejection, spaced paths, stale-result replacement, and watchdog completion/expiry. Shell/workflow checks pass; no Conbench upload was performed locally.

On total runner use: the last completed benchmark run before this follow-up totaled about 120.9 minutes across the six benchmark jobs, plus about 0.48 minute for setup/combine, with a 41.4-minute benchmark wall span. Those are elapsed job durations, not billed-minute accounting. That is below the previously cited 198–208-minute run, but sharding itself reduces wall time rather than eliminating aggregate work.

.github/workflows/benchmark.yml
```diff
@@ -23,44 +23,109 @@ on:
...
+      - name: Compute benchmark shards
+        id: shards
+        run: echo "matrix=$(bash ci/scripts/bench_shard.sh 6)" >> "$GITHUB_OUTPUT"
   benchmark:
+    needs: setup
     runs-on: ubuntu-latest
+    # The slowest shard (./arrow/array) takes ~40m, so 90m leaves headroom for
+    # runner variance while still failing a pathological benchmark ~4x sooner
+    # than the 6h GitHub default.
+    timeout-minutes: 100
```
lidavidm: comment/value mismatch

.github/workflows/benchmark.yml
```diff
@@ -23,44 +23,109 @@ on:
...
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7.0.1
         with:
           submodules: recursive
-      - name: Set up Python
+      - name: Install Go for Benchmarks
+        uses: actions/setup-go@v7.0.0
+        with:
+          go-version: '1.26.1'
```
lidavidm: We can't take this from go.mod?

lidavidm (approved):

.github/workflows/benchmark.yml
zeroshade: Good catch. There are two budgets and the comment described the wrong one:
shards run under a 90m watchdog (the `--timeout` passed to `bench.sh` below),
while the job cap is 100m so the watchdog trips first and names the stuck
benchmark, with the extra 10m covering checkout, setup and upload. The comment
sat on the cap but explained the watchdog. Rewritten to cover both and why they
differ.

.github/workflows/benchmark.yml
zeroshade: Done, and this was a regression introduced here: the version used to come from
the one-value `go: ['1.26.1']` matrix that this branch replaced with shard data,
which left `1.26.1` duplicated as a literal in both the `benchmark` and
`combine` jobs, free to drift apart. Both now use `go-version-file: go.mod`,
matching `rc.yml`.

One heads-up on the consequence, since it may not be what you expected. Our
`go.mod` has no `toolchain` directive and declares `go 1.25.0`, and setup-go
[uses the `go` directive verbatim when it carries a patch
version](https://github.com/actions/setup-go/blob/main/docs/advanced-usage.md#using-the-go-version-file-input),
so this pins the benchmarks to exactly Go 1.25.0 rather than the 1.26.1 they
were running — our compatibility floor, not the toolchain we had been
benchmarking on, and a patch behind the `GO=1.25.8` in `.env`. For Conbench that
means a one-time step in the timeseries and benchmarking a minor behind what
`test.yml` already covers (1.25/1.26/1.27).

I went with what you asked since it removes the duplication and is easy to
revert. If you would rather keep the benchmark toolchain where it was, I am
happy to switch to a workflow-level `env: GO_VERSION` used by both jobs, or read
`GO` from `.env` the way `test.yml` does.

lidavidm (approved):

PR #960, opened 2026-07-20
feat(parquet/schema): Wire up logical type functionality for writing geo types

C-Loftus:
### Rationale for this change

Any downstream arrow clients like iceberg-go which write geo data to parquet end up being labeled in the underlying parquet as bytes / binary blobs and not the proper geometry type. This means they are not able to be queried by tools like duckdb, etc. and will result in issues like this https://github.com/duckdb/duckdb-iceberg/issues/1175

arrow-go needs to implement the logical type interface for both geometry and geography to ensure the parquet write path gets the proper metadata.

Implementations of the logical type interface for both Geometry and Geography

### Are these changes tested?

Yes

Yes. Geometry should now be labelled as such in written parquet.

Note, downstream users like `iceberg-go` are using [geoarrow-go](https://github.com/geoarrow/geoarrow-go) for some of the arrow-related geospatial metadata imports (i.e. edge interpolation algorithms, type constructors, etc.) Should someone in that repo simply implement the proper interfaces and then this package would import it directly as a new dependency? Since there are geo types in parquet natively now I was wondering if it maybe makes sense to pull into the main repo. ccing geoarrow-go creator @dwilson1988 on this

Happy to defer to maintainers' preference.


dwilson1988:
@C-Loftus I believe we had decided that the full geoarrow types should remain outside of this repo, but I believe it does make sense to have the Geometry/Geography types for Parquet in this repo, which necessitates WKB encoding. I think there's a bit of a chicken/egg situation to deal with here that it would be good to get @zeroshade's call on.

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,255 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+}
+
+// NewGeometryLogicalType returns a Geometry logical type. If crs is provided,
+// it is used as the coordinate reference system.
+func NewGeometryLogicalType(crs ...string) LogicalType {
+	typ := format.NewGeometryType()
+	if len(crs) > 0 {
+		typ.Crs = thrift.StringPtr(crs[0])
+	}
+	return GeometryLogicalType{typ: typ}
```
zeroshade: I don't like the idea of using `...` just to denote optional if we're not going to actually allow multiple values (which we don't). Most of the logical types are easily constructed as struct literals rather than using a `New` method unless there's validation going on. Perhaps it would make more sense to just have the Crs as a public member of the struct and let users just do `GeometryLogicalType{Crs: ...}` etc. ?

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,255 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+}
+
+// NewGeographyLogicalType returns a Geography logical type.
+func NewGeographyLogicalType(opts ...GeographyLogicalTypeOption) LogicalType {
+	t := GeographyLogicalType{typ: format.NewGeographyType()}
+	for _, opt := range opts {
+		opt(&t)
+	}
+	return t
+}
```
zeroshade: same thing as above, perhaps we just make the algorithm and Crs public members and handle the population of the thrift side in the `toThrift` method etc.

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,255 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+	case format.EdgeInterpolationAlgorithm_ANDOYER:
+		return GeographyEdgeAndoyer
+	case format.EdgeInterpolationAlgorithm_KARNEY:
+		return GeographyEdgeKarney
+	case format.EdgeInterpolationAlgorithm_SPHERICAL:
+		fallthrough
+	default:
+		return GeographyEdgeSpherical
+	}
+}
```
zeroshade: let's do this as a `fromThrift` on `GeographyEdgeInterpolationAlgorithm` instead of a free function

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,255 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+	case GeographyEdgeAndoyer:
+		return format.EdgeInterpolationAlgorithm_ANDOYER
+	case GeographyEdgeKarney:
+		return format.EdgeInterpolationAlgorithm_KARNEY
+	case GeographyEdgeSpherical:
+		fallthrough
+	default:
+		return format.EdgeInterpolationAlgorithm_SPHERICAL
+	}
+}
```
zeroshade: same as above, do this as a `toThrift` on `GeographyEdgeInterpolationAlgorithm` instead of a free function

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,255 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+	case GeographyEdgeVincenty,
+		GeographyEdgeThomas,
+		GeographyEdgeAndoyer,
+		GeographyEdgeKarney,
+		GeographyEdgeSpherical:
+		return string(alg)
+	default:
+		return string(GeographyEdgeSpherical)
+	}
+}
```
zeroshade: add a `String()` method to `GeographyEdgeInterpolationAlgorithm` instead

zeroshade (changes requested):

C-Loftus:
Thanks @dwilson1988 for your comment! I have been discussing with @zeroshade in the iceberg-go slack if it would be relevant https://apache-iceberg.slack.com/archives/C05J3MJ42BD/p1784555139111839

Essentially I think my plan will be to get this merged, then in geoarrow-go we can implement `ParquetLogicalType` so that the geoarrow structs get picked up as the GeometryType i am currently working on. That way the geoarrow logic is kept out of this repo but should still work in the write path. For read that will be a follow up issue from this. According to @zeroshade 

> essentially you'd probably need to add something in https://github.com/apache/arrow-go/blob/main/arrow/datatype_extension.go#L43 to allow iterating the registered extension types. then add to https://github.com/apache/arrow-go/blob/main/parquet/pqarrow/schema.go#L528 cases for geography and geometry, where you iterate the registered extension types and use one that matches the logical type you got, which will need to add a new interface to look for to map from the logical type to the extension type

zeroshade:
> @C-Loftus I believe we had decided that the full geoarrow types should remain outside of this repo, but I believe it does make sense to have the Geometry/Geography types for Parquet in this repo, which necessitates WKB encoding. I think there's a bit of a chicken/egg situation to deal with here that it would be good to get @zeroshade's call on.

We don't need to actually perform the WKB encoding/decoding in here, the parquet Geography and Geometry logical types are just byte-array primitives. So as long as we can use a registered extension type and go from the extension type -> parquet logical type (this interface already exists) then writing will work. For reading, we'll need to add a new interface to go from parquet physical type + parquet logical type -> instantiated extension type. 

In both cases, we'll need to add methods to the extension types in geoarrow-go to do the conversion between the type and parquet logical type which should allow us to generically handle it without needing to actually perform WKB encode/decode in here

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,242 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+	if t == nil {
+		return GeographyLogicalType{}
+	}
+
+	ret := GeographyLogicalType{}
+	if t.IsSetCrs() {
+		ret.Crs = t.GetCrs()
+	}
+	if t.IsSetAlgorithm() {
+		ret.Algorithm = GeographyEdgeInterpolationAlgorithm("").fromThrift(t.GetAlgorithm())
```
zeroshade: If you make the `fromThrift` method a pointer receiver, then you could do this:

```suggestion
		ret.Algorithm.fromThrift(t.GetAlgorithm())
```

Alternately, it could still be a value receiver and just do

```suggestion
		ret.Algorithm = ret.Algorithm.fromThrift(t.GetAlgorithm())
```

either way, it lets you avoid an allocation

parquet/schema/logical_types.go
```diff
@@ -1228,3 +1247,242 @@ func (NoLogicalType) Equals(rhs LogicalType) bool {
...
+	case format.EdgeInterpolationAlgorithm_ANDOYER:
+		return GeographyEdgeAndoyer
+	case format.EdgeInterpolationAlgorithm_KARNEY:
+		return GeographyEdgeKarney
+	case format.EdgeInterpolationAlgorithm_SPHERICAL:
+		fallthrough
+	default:
+		return GeographyEdgeSpherical
+	}
+}
```
zeroshade: move this and the other methods closer to the type definition

zeroshade (changes requested):
looks good, just two nitpicks

C-Loftus:
Should be all set for review again. Believe I have addressed feedback. I have validated this works in https://github.com/geoarrow/geoarrow-go/pull/5 by using that PR with this branch of arrow-go and checking that clients like duckdb can read the geo metadata. Could be worth getting a check on the crs/edge interpolation algorithm logic from someone affiliated with geoarrow-go in case I need to be ensuring parity across libraries

zeroshade (approved):
LGTM! let's follow this up with the corresponding update in geoarrow-go

dwilson1988:
> LGTM! let's follow this up with the corresponding update in geoarrow-go

I've been neglectful of geoarrow-go recently, but happy to review any PR and should keep the repo moving more soon.

PR #969, opened 2026-07-21
feat(parquet/pqarrow): Read support for extension types / geo

C-Loftus:
### Rationale for this change

In https://github.com/apache/arrow-go/pull/960 I implemented write support for extension types by mapping a logical type to its underlying primitive/arrow type. In the case of geometry this allows the logical geometry type to be mapped to byte array for storage. 

This PR implements the read path for that by adding the parquet type to arrow type functionality. For example, a user can read a parquet with a geometry column and with this change, the reader will try to map it to the designated associated arrow extension type.  

An extension type can implement this interface to designate which arrow type should be used when reading it from parquet.

```go
type ExtensionParquetLogicalType interface {
	ArrowTypeFromParquet(logical schema.LogicalType, storageType arrow.DataType) (arrow.ExtensionType, error)
}
```

The read path will try to iterate through all the extension types, see if there are any matches where the type implements `ArrowTypeFromParquet` and gives us something to map it to in arrow. If so we use that for reading. 

### Are these changes tested?

Yes

Yes. Readers can now support arbitrary extension types as long as they are registered properly in the extension type registry.

### DIFF
--- a/arrow/datatype_extension.go
+++ b/arrow/datatype_extension.go
@@ -19,6 +19,7 @@ package arrow
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"sync"
 )
 
@@ -85,19 +86,33 @@ func GetExtensionType(typName string) ExtensionType {
 // FindRegisteredExtensionType returns a registered extension type for
 // which the filter returns true. If no type matches, it returns nil.
 //
-// If multiple types match, which one is returned is unspecified.
+// If multiple types match, the type with the lexicographically smallest
+// registered extension name is returned.
 func FindRegisteredExtensionType(filter func(ExtensionType) bool) ExtensionType {
 	registry := getExtTypeRegistry()
-	var out ExtensionType
-	registry.Range(func(_, value any) bool {
-		typ := value.(ExtensionType)
-		if filter(typ) {
-			out = typ
-			return false
-		}
+
+	// We first sort the types by name to ensure that we return the
+	// type with the lexicographically smallest name. We sort then filter
+	// to ensure that there are no side effects from filtering which would
+	// affect the determinism of the result.
+
+	types := make(map[string]ExtensionType)
+	var names []string
+	registry.Range(func(key, value any) bool {
+		name := key.(string)
+		types[name] = value.(ExtensionType)
+		names = append(names, name)
 		return true
 	})
-	return out
+
+	sort.Strings(names)
+	for _, name := range names {
+		typ := types[name]
+		if filter(typ) {
+			return typ
+		}
+	}
+	return nil
 }
 
 // ExtensionType is an interface for handling user-defined types. They must be
--- a/arrow/datatype_extension.go
+++ b/arrow/datatype_extension.go
@@ -91,11 +91,9 @@ func GetExtensionType(typName string) ExtensionType {
 func FindRegisteredExtensionType(filter func(ExtensionType) bool) ExtensionType {
 	registry := getExtTypeRegistry()
 
-	// We first sort the types by name to ensure that we return the
-	// type with the lexicographically smallest name. We sort then filter
-	// to ensure that there are no side effects from filtering which would
-	// affect the determinism of the result.
-
+	// Snapshot and sort the registered names before invoking the filter so the
+	// result is deterministic (lexicographically smallest match), independent of
+	// sync.Map's unspecified Range order.
 	types := make(map[string]ExtensionType)
 	var names []string
 	registry.Range(func(key, value any) bool {
--- a/arrow/datatype_extension.go
+++ b/arrow/datatype_extension.go
@@ -82,6 +82,18 @@ func GetExtensionType(typName string) ExtensionType {
 	return nil
 }
 
+// RegisteredExtensionTypes returns a snapshot of all registered extension types.
+// The returned types are in an unspecified order.
+func RegisteredExtensionTypes() []ExtensionType {
+	registry := getExtTypeRegistry()
+	var out []ExtensionType
+	registry.Range(func(_, value any) bool {
+		out = append(out, value.(ExtensionType))
+		return true
+	})
+	return out
+}
+
 // ExtensionType is an interface for handling user-defined types. They must be
 // DataTypes and must embed arrow.ExtensionBase in them in order to work properly
 // ensuring that they always have the expected base behavior.
--- a/arrow/datatype_extension.go
+++ b/arrow/datatype_extension.go
@@ -82,13 +82,19 @@ func GetExtensionType(typName string) ExtensionType {
 	return nil
 }
 
-// RegisteredExtensionTypes returns a snapshot of all registered extension types.
-// The returned types are in an unspecified order.
-func RegisteredExtensionTypes() []ExtensionType {
+// FindRegisteredExtensionType returns a registered extension type for
+// which the filter returns true. If no type matches, it returns nil.
+//
+// If multiple types match, which one is returned is unspecified.
+func FindRegisteredExtensionType(filter func(ExtensionType) bool) ExtensionType {
 	registry := getExtTypeRegistry()
-	var out []ExtensionType
+	var out ExtensionType
 	registry.Range(func(_, value any) bool {
-		out = append(out, value.(ExtensionType))
+		typ := value.(ExtensionType)
+		if filter(typ) {
+			out = typ
+			return false
+		}
 		return true
 	})
 	return out
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -129,6 +129,13 @@ type ExtensionCustomParquetType interface {
 	ParquetLogicalType() schema.LogicalType
 }
 
+// ExtensionParquetLogicalType is an interface that Arrow ExtensionTypes may
+// implement to specify how a Parquet LogicalType maps back to an Arrow
+// ExtensionType when converting a Parquet schema to an Arrow schema.
+type ExtensionParquetLogicalType interface {
+	ArrowTypeFromParquet(logical schema.LogicalType, storageType arrow.DataType) (arrow.ExtensionType, error)
+}
+
 func isDictionaryReadSupported(dt arrow.DataType) bool {
 	return arrow.IsBinaryLike(dt.ID())
 }
@@ -531,6 +538,9 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
 		return arrow.BinaryTypes.String, nil
 	case schema.DecimalLogicalType:
 		return arrowDecimal(logtype), nil
+	case schema.GeometryLogicalType,
+		schema.GeographyLogicalType:
+		return arrowExtensionFromParquetLogicalType(logical, arrow.BinaryTypes.Binary)
 	case schema.NoLogicalType,
 		schema.EnumLogicalType,
 		schema.JSONLogicalType,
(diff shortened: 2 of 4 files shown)


parquet/pqarrow/schema.go
```diff
@@ -531,6 +538,9 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
 		return arrow.BinaryTypes.String, nil
 	case schema.DecimalLogicalType:
 		return arrowDecimal(logtype), nil
+	case schema.GeometryLogicalType,
+		schema.GeographyLogicalType:
+		return arrowExtensionFromParquetLogicalType(logical, arrow.BinaryTypes.Binary)
```
C-Loftus, line 549: Is there a way I can allow it to efficiently check either `Binary` or `LargeBinary` to allow decoding either 32 bit or 64 bit wkb values? I think at the moment this will only handle reading 32 bit. I can also just add a second case here in case the first `arrowExtensionFromParquetLogicalType(logical, arrow.BinaryTypes.Binary)` call returns no mapping? (i.e. try again with `arrowExtensionFromParquetLogicalType(logical, arrow.BinaryTypes.LargeBinary)`?)

zeroshade: I don't think there's a way for it to tell from the logical type whether or not it should use `Binary` or `LargeBinary` (or `BinaryView`). The difference is solely whether or not there happens to be more than 2GB worth of data in a single record batch. So even if you were to add a second case here, it's likely not a useful thing (all extension types that support Binary should also support LargeBinary and vice versa). 

The better solution here, like the usage of `String` instead of `LargeString` above, is to ensure that we properly promote to LargeBinary if necessary, but that can be a follow-up

zeroshade (commented):
Using your `testGeometryType` add a round trip test that uses it (and the interface) to write and then read back the registered extension type while making sure the StoreSchema option is false

C-Loftus:
Believe I have addressed feedback by adding the round trip test and ensuring `WithStoreSchema` is never called, thus using `c.props.storeSchema = false` by default

arrow/datatype_extension.go
```diff
@@ -82,6 +82,18 @@ func GetExtensionType(typName string) ExtensionType {
...
+// The returned types are in an unspecified order.
+func RegisteredExtensionTypes() []ExtensionType {
+	registry := getExtTypeRegistry()
+	var out []ExtensionType
+	registry.Range(func(_, value any) bool {
+		out = append(out, value.(ExtensionType))
+		return true
+	})
+	return out
+}
```
zeroshade: This adds permanent exported API to the core `arrow` package, so worth being deliberate before it ships.

It's also the first place in the codebase that *enumerates* the registry — everywhere else does name-based point lookups (`GetExtensionType`, e.g. the UUID read path in `arrowFromFLBA` and IPC extension deserialization in `arrow/ipc/metadata.go`). Enumerate-and-ask is what introduces the ordering ambiguity flagged at the call site below.

If the only consumer is pqarrow's reverse lookup, consider a predicate helper instead, e.g. `FindRegisteredExtensionType(pred func(ExtensionType) bool) ExtensionType`, which keeps the `Range` bounded to one call, allows early-exit on first match, and avoids allocating a full snapshot slice. If you keep the slice form, the "unspecified order" doc is accurate and good to have.

parquet/pqarrow/schema.go
```diff
@@ -129,6 +129,13 @@ type ExtensionCustomParquetType interface {
 	ParquetLogicalType() schema.LogicalType
 }
 
+// ExtensionParquetLogicalType is an interface that Arrow ExtensionTypes may
+// implement to specify how a Parquet LogicalType maps back to an Arrow
+// ExtensionType when converting a Parquet schema to an Arrow schema.
+type ExtensionParquetLogicalType interface {
+	ArrowTypeFromParquet(logical schema.LogicalType, storageType arrow.DataType) (arrow.ExtensionType, error)
+}
```
zeroshade, line 143: Naming/symmetry nit on exported API: the write side is `ExtensionCustomParquetType.ParquetLogicalType()` and this read side is `ExtensionParquetLogicalType.ArrowTypeFromParquet()`. A round-trippable extension has to implement both, and the two interface names are easy to confuse. Consider a symmetric pair (e.g. `...ToParquet` / `...FromParquet`) or folding both directions into a single interface. Harder to change once released.

parquet/pqarrow/schema.go
```diff
@@ -541,6 +551,27 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
 	}
 }
 
+// arrowExtensionFromParquetLogicalType asks registered extension types whether
+// they can represent the provided Parquet logical type with the given storage
+// type, falling back to the storage type when none opt in.
+func arrowExtensionFromParquetLogicalType(logical schema.LogicalType, storageType arrow.DataType) (arrow.DataType, error) {
+	for _, extType := range arrow.RegisteredExtensionTypes() {
```
zeroshade: `arrow.RegisteredExtensionTypes()` is backed by `sync.Map`, whose `Range` order is unspecified. If two registered extensions both return non-nil from `ArrowTypeFromParquet` for the same logical type (plausible for geo: a user's custom type plus a future canonical `geoarrow.*`), the winner is arbitrary and can differ across processes — reads aren't reproducible.

Contrast the existing UUID mapping in `arrowFromFLBA`, which is deterministic via `GetExtensionType("arrow.uuid")`. Suggest either iterating a name-sorted snapshot and taking the first match, or having extensions register an explicit logical-type -> extension mapping so resolution is well-defined.

parquet/pqarrow/schema.go
```diff
@@ -541,6 +551,27 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
...
+		converter, ok := extType.(ExtensionParquetLogicalType)
+		if !ok {
+			continue
+		}
+
+		typ, err := converter.ArrowTypeFromParquet(logical, storageType)
+		if err != nil {
+			return nil, err
+		}
```
zeroshade: Returning the error here propagates up through `getArrowType` -> `nodeToSchemaField`, which aborts the entire `FromParquet` conversion. So a single third-party extension that returns an error for "not my CRS/logical type" breaks reading of *every* geometry/geography column in the process, including files unrelated to it.

Combined with the non-deterministic iteration order above, whether a read hits that error can depend on which extension `Range` visits first — the same file may read fine or fail across runs.

Recommend: reserve a non-nil `error` for genuinely fatal conditions, keep "not mine" as `(nil, nil)` (the interface doc and the test double already follow this), and either `continue` past an errored extension or only surface the error if no other extension matches.

parquet/pqarrow/schema.go
```diff
@@ -541,6 +551,27 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
...
+		if typ != nil {
+			return typ, nil
+		}
+	}
+	return storageType, nil
```
zeroshade, line 598: Good that this falls back to the storage type when nothing opts in. Note this also changes existing behavior: a Geometry/Geography byte-array column now reads back as `Binary` even when no extension is registered, whereas before it hit the `default` in `arrowFromByteArray` and errored.

That new default path isn't covered — both new tests register an extension first. Please add a case asserting Geometry-with-no-registered-extension maps to `Binary`, so the fallback (and the fact that it no longer errors) is locked in.

zeroshade (changes requested):
Thanks for adding the read path — the round-trip-without-stored-schema test is exactly what was needed, and scoping to byte-array (matching `GeometryLogicalType`/`GeographyLogicalType.IsApplicable`, which are ByteArray-only) is correct.

I'm requesting changes on two points that interact badly:

1. Resolution over the registry is non-deterministic (`sync.Map`-backed enumeration).
2. An error from any single registered extension aborts the whole schema conversion.

Together they mean a read can non-deterministically succeed or fail depending on map iteration order. Details inline, plus a naming/API note and a test-coverage gap.

Happy to pair on direction — either a deterministic ordering (name-sorted snapshot + first match) or an explicit logical-type -> extension registration, rather than "ask every registered type".

parquet/pqarrow/schema.go
C-Loftus: I agree with this comment ideally although isn't `ExtensionCustomParquetType` already in the public API so it would not be possible to change that? As such I think I am a bit stumped on how to name this with breaking existing extension type resolution.

i.e. to my understanding I don't think I could change it to be a symmetric pair since there is already a `ParquetLogicalType()` function which doesn't have a clear symmetric parallel. Feel free to suggest any name if I am overlooking a good option.  

https://github.com/apache/arrow-go/blob/main/parquet/pqarrow/schema.go#L128

C-Loftus:
Ok believe I have addressed feedback. Thanks for the review!

- I will defer to you on the interface naming. https://github.com/apache/arrow-go/pull/969#discussion_r3632075763 in light of the currently exported pre-existing public interface methods.
- I have changed the registration type lookup to be `FindRegisteredExtensionType` which is both lexicographically deterministic and returns one item. If this is overkill I can revert or change as you prefer.

parquet/pqarrow/schema.go
```diff
@@ -129,6 +129,13 @@ type ExtensionCustomParquetType interface {
 	ParquetLogicalType() schema.LogicalType
 }
 
+// ExtensionParquetLogicalType is an interface that Arrow ExtensionTypes may
+// implement to specify how a Parquet LogicalType maps back to an Arrow
+// ExtensionType when converting a Parquet schema to an Arrow schema.
+type ExtensionParquetLogicalType interface {
+	ArrowTypeFromParquet(logical schema.LogicalType, storageType arrow.DataType) (arrow.ExtensionType, error)
```
zeroshade, line 142: Optional, doc-only: worth spelling out the return-value contract here, since third-party types implement this against the exported interface:

- `(nil, nil)` → "not my logical type"; resolution skips this extension and continues.
- `(nil, err)` → recognized the logical family but couldn't construct a valid type; the error is deferred and only surfaced if no other extension matches.
- non-nil type → wins, taking precedence over a deferred error from an earlier (name-sorted) extension.

The test doubles already follow this — it's just about making the contract explicit on the exported API.

arrow/datatype_extension.go
```diff
@@ -82,6 +83,38 @@ func GetExtensionType(typName string) ExtensionType {
...
+//
+// If multiple types match, the type with the lexicographically smallest
+// registered extension name is returned.
+func FindRegisteredExtensionType(filter func(ExtensionType) bool) ExtensionType {
+	registry := getExtTypeRegistry()
+
+	// We first sort the types by name to ensure that we return the
+	// type with the lexicographically smallest name. We sort then filter
+	// to ensure that there are no side effects from filtering which would
+	// affect the determinism of the result.
```
zeroshade: Minor wording: "no side effects from filtering" reads as if the filter is required to be pure, but what this actually guarantees is that iteration runs over a stable, name-sorted snapshot, so the result is deterministic regardless of sync.Map's Range order. Maybe something like:

    // Snapshot and sort the registered names before invoking the filter so the
    // result is deterministic (lexicographically smallest match), independent of
    // sync.Map's unspecified Range order.

zeroshade (approved):
Thanks for the thorough revision — this addresses both of my blocking concerns.

- Resolution is now deterministic via the name-sorted `FindRegisteredExtensionType`, and `TestFindRegisteredExtensionType` registers the higher name first so the tie-break is actually exercised rather than passing by registration order.
- A single extension's error no longer aborts the whole `FromParquet`: "not mine" is `(nil, nil)` and continues, the first error is deferred and only surfaced when nothing matches, and `TestFromParquetGeospatialRegisteredExtensionContinuesAfterConverterError` proves that a failing type which sorts first doesn't block a later valid one.
- The Binary fallback for unregistered geometry/geography is now covered and the behavior change is locked in.

Built the branch locally: the new tests pass, `go vet` is clean, and `-race` is clean on both `arrow` and `parquet/pqarrow`.

On the interface naming you flagged: since `ExtensionCustomParquetType.ParquetLogicalType()` is already public, a fully symmetric pair isn't worth the churn — `ExtensionParquetLogicalType` is fine to ship as-is.

Approving. Two optional, doc-only nits inline that would help future implementers — neither blocks merge.

C-Loftus:
Have added the feedback comments! Thanks for your review!

When we merge, at your convenience, when you produce a new arrow-go release, could you ping me? I will then add the necessary functions to geoarrow-go, bump iceberg-go to use them, and then we should have a full end to end for iceberg geo read and write support

PR #970, opened 2026-07-21
fix: remove superfluous struct field index compare

Willem-J-an:
Remove superfluous check in the TypeEqual path to improve performance

### Rationale for this change
The check is superfluous because:
- Length is verified to equal
- Next step iterates over fields & asserts the names are equal for each index. These 2 steps essentially check the same thing, without needing expensive DeepEqual of 2 maps.

### Are these changes tested?
passes on `go test -v ./arrow`

No

### DIFF
--- a/arrow/compare.go
+++ b/arrow/compare.go
@@ -97,10 +97,7 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 		return true
 	case *StructType:
 		r := right.(*StructType)
-		switch {
-		case len(l.fields) != len(r.fields):
-			return false
-		case !reflect.DeepEqual(l.index, r.index):
+		if len(l.fields) != len(r.fields) {
 			return false
 		}
 		for i := range l.fields {


arrow/compare.go
```diff
@@ -100,8 +100,6 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 		switch {
 		case len(l.fields) != len(r.fields):
 			return false
-		case !reflect.DeepEqual(l.index, r.index):
-			return false
 		}
```
zeroshade: should we just get rid of the switch then? and just make this a simple `if len(l.fields) != len(r.fields) { return false }` ?

zeroshade (approved):
thanks for this

arrow/compare.go
Willem-J-an: Yea makes sense!

Willem-J-an: Fixed

PR #971, opened 2026-07-21
docs(parquet): Add thrift generation contribution info

C-Loftus:
### Rationale for this change

When making changes to parquet-related logic sometimes contributors need to regenerate thrift generated files. I did this for  https://github.com/apache/arrow-go/pull/960 However since the thrift definitions don't live in this repo, it might be worth documenting for new users. 

These docs are based on my understanding from discussions with maintainers in the iceberg go slack.

Feel free to close or propose / make direct edits as you see fit

Just documentation

### Are these changes tested?

No since they are just docs

No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,36 @@ ci: update CI environment
 feat(parquet): support new encoding type
 ```
 
+## Generating Thrift
+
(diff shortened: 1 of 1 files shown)


zeroshade (approved):
Thanks!

PR #981, opened 2026-07-22
feat(arrow/array): Add `Builder.Truncate()` method

serramatutu:
You can review commit by commit.

### Rationale for this change

While working on [this change](https://github.com/apache/arrow-go/pull/833), I ran into an issue: when a struct builder runs into an error while decoding one of the columns from JSON, there is no way to "un-append" rows from its sibling rows.

The two options are to either (1) `AppendNull()` to all siblings and return the error, which would leave a bad row behind, or (2) eagerly `return err` and leave the builder in a broken state.

```
for i := 0; i < b.schema.NumFields(); i++ {
	val, ok := keylist[b.schema.Field(i).Name]
	if !ok {
		b.fields[i].AppendNull()
		continue
	}

	valDec := json.NewDecoder(bytes.NewReader(val))
	valDec.UseNumber()
	if err := b.fields[i].UnmarshalOne(valDec); err != nil {
		// TODO: what to do here? We might already have appended to previous siblings (i-1, i-2, ...)
	}
}
```

What I want to do instead is:

```
if err := b.fields[i].UnmarshalOne(valDec); err != nil {
    for j := 0; j < i; j++ {
        b.fields[j].Truncate(b.fields[j].Len() - 1)
    }
    return err
}
```

Add `Truncate(n int)` to the `Builder` interface. This new method will keep the underlying buffers but truncate the arrays to at most length `n`. It keeps `Len()` and `NullN()` sound. Calling `b.Truncate(b.Len() - 1)` is the equivalent of "un-appending" one element.

### Are these changes tested?

Yes, I added a `TestBuilderTruncate` that checks that `Truncate` works for all data types.

It's worth noting I wanted to call `ValidateFull()` after truncate, but it seems like REE arrays have an issue with `ValidateFull()` that precedes this PR.

Yes, this adds a new method to the public API of `Builder`

### DIFF
--- a/arrow/array/builder.go
+++ b/arrow/array/builder.go
@@ -70,6 +70,12 @@ type Builder interface {
 	// AppendValueFromString adds a new value from a string. Inverse of array.ValueStr(i int) string
 	AppendValueFromString(string) error
 
+	// Truncate reduces the length of the array being built to n elements,
+	// discarding any elements beyond the first n. It does not reallocate or
+	// shrink the underlying buffers. If n is greater than or equal to the
+	// current length, Truncate is a no-op.
+	Truncate(n int)
+
 	// Reserve ensures there is enough space for appending n elements
 	// by checking the capacity and calling Resize if necessary.
 	Reserve(n int)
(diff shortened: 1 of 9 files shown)


serramatutu:
After talking with Matt I decided to close this and do something else for that PR instead.

Issue #983, opened 2026-07-22
arm64 assembly breaks CPU profiling


nsrip-dd:
### Describe the bug, including details regarding any error messages, version, and platform.

The arm64 NEON assembly implementations of functions in this library break CPU profiling. This breakage usually manifests as truncated and sometimes invalid tracebacks in the profile and sometimes crashes. The reason for this is that the functions tend to manipulate the stack pointer, but via `WORD` directives rather than normal instructions, e.g. `WORD $0xa9ba7bfd // stp    x29, x30, [sp, #-96]!`. The Go assembler doesn't decode these instructions and thus doesn't see the stack pointer manipulation. But this means that the generated unwinding tables will be invalid. The tables encode how much to increase the stack pointer at a given instruction to find the next call frame.

## Examples

For example, `go test -bench=. -cpuprofile=cpu.pprof ./arrow/memory` produces a profile like

<img width="886" height="371" alt="Image" src="https://github.com/user-attachments/assets/af53b513-7919-40f4-bf12-940c32fa77dc" />

Notice that most of the `memory._memset_neon` time is in a traceback with one frame, missing the callers.

For another example, consider this benchmark of `parquet/internal/utils`:

```
func BenchmarkUnpack32(b *testing.B) {
	const batchSize = 512
	input := make([]byte, batchSize*4)
	if _, err := rand.Read(input); err != nil {
		b.Fatal(err)
	}
	output := make([]uint64, batchSize)
	reader := NewBitReader(bytes.NewReader(input))
	for b.Loop() {
		reader.Reset(bytes.NewReader(input))
		reader.GetBatch(32, output)
	}
}
```

The CPU time attributed to `_unpack32_neon`, which gets called here, looks like this in the CPU profile:

<img width="886" height="181" alt="Image" src="https://github.com/user-attachments/assets/f15af53f-17eb-4bc0-813a-a09d43df26ca" />

There are either no callers, or an invalid call sequence showing the runtime as the caller of this function.

[...]

PR #984, opened 2026-07-22
fix: Fix arm64 assembly stack frame manipulation

nsrip-dd:
### Rationale for this change

The arm64 NEON assembly code in this library incorrectly manipulates the stack pointer. In the mechanically-translated portion of the code, there are instructions to set up and restore a call frame for the function, e.g.

	WORD $0xa9bf7bfd // stp    x29, x30, [sp, #-16]!

But the assembler doesn't decode WORD directives and thus doesn't see the stack pointer manipulation. As a result, the spdelta table entries for these functions are incorrect. This table is used by the Go runtime for stack unwinding, to find the next frame based on the program counter and current stack pointer. In typical cases this means that the CPU profiler will just stop unwinding when it hits this function, and in some cases we can get non-truncated but incorrect tracebacks.

See #983

Fix this by removing the unnecssary frame setup for functions which don't otherwise use the stack, and by changing functions which do need a stack frame to include their frame size in the function declaration, which the assembler takes into account.

Note that this PR removes the `SUB/ADD $16, RSP` instructions added in #170. Those were added because the functions were overwriting values saved on the stack, but after this PR most of the modified functions don't save anything on the stack, and the one that does properly declares its frame size and makes the fix unnecessary. This does mean that the modified functions with no frames don't set up a frame pointer, but in practice that's probably fine. If we really want frame pointers for all these functions, we can give them a small non-zero frame size and the assembler will add the appropriate frame pointer setup/tearndown code.

### Are these changes tested?

Yes, by the existing unit tests, and manually by running the CPU profiler as described in #983

No

### DIFF
--- a/arrow/compute/internal/kernels/cast_numeric_neon_arm64.s
+++ b/arrow/compute/internal/kernels/cast_numeric_neon_arm64.s
@@ -10,13 +10,7 @@ TEXT ·_cast_type_numeric_neon(SB), $0-40
     MOVD len+32(FP), R4
 
 
-    // The Go ABI saves the frame pointer register one word below the 
-    // caller's frame. Make room so we don't overwrite it. Needs to stay 
-    // 16-byte aligned 
-    SUB $16, RSP
-    WORD $0xa9bf7bfd // stp    x29, x30, [sp, #-16]!
     WORD $0x7100181f // cmp    w0, #6
-    WORD $0x910003fd // mov    x29, sp
     BGT LBB0_17
 
     WORD $0x71000c1f // cmp    w0, #3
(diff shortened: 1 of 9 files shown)


zeroshade:
@nsrip-dd does this include a test that triggers the original failure to verify we don't have a regression later? if not, can we add one?

zeroshade (approved):
Looks good to me, but is there a deterministic way to confirm via tests? Or do we have to wait until we follow-up with the R28/R18 fixes?

nsrip-dd:
> Looks good to me, but is there a deterministic way to confirm via tests? Or do we have to wait until we follow-up with the R28/R18 fixes?

I added https://github.com/apache/arrow-go/pull/984/changes/40aedb64acd2c560bfabb6470a6ba50ea6bf83d4, which reliably catches the issue for _memset_neon on my laptop. It's a bit verbose but I'm not sure how much simpler it can be. I opted to use `go tool pprof` rather than `github.com/google/pprof` so we don't get a new dependency. We could write similar tests for the other functions this PR fixes. Let me know if this approach is okay with you.

nsrip-dd:
Okay, for some reason the tests are running fine on my macos arm64 laptop but are failing the CI runs (missing the `memory.Set` frame, looks like.) I'll try to figure out what's going on...

nsrip-dd:
https://github.com/apache/arrow-go/pull/984/changes/dfb6964b63813cfb5cc4cc8bc431345783aaeb8e seems to have done the trick

PR #985, opened 2026-07-23
fix(parquet/variant): handle empty metadata keys

fallintoplace:
### Rationale for this change

Variant metadata may contain an empty object key. `Metadata.KeyAt` validates the dictionary index, then previously took the address of the key's first byte. That panics when the key is empty.

Use `unsafe.SliceData` for the existing zero-copy byte-to-string conversion so zero-length keys are handled safely. Add a builder-based regression covering both `KeyAt` and `IdFor`.

### Are these changes tested?

Yes:

- `go test ./parquet/variant`
- `go test -race -count=1 ./parquet/variant`

`Metadata.KeyAt` now returns an empty string for a valid empty key instead of panicking. There is no API change.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -258,7 +258,8 @@ func (m Metadata) KeyAt(id uint32) (string, error) {
 			id, len(m.keys))
 	}
 
-	return unsafe.String(&m.keys[id][0], len(m.keys[id])), nil
+	key := m.keys[id]
+	return unsafe.String(unsafe.SliceData(key), len(key)), nil
 }
 
 // IdFor returns the dictionary IDs for the given key.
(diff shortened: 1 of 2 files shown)


zeroshade:
I don't think that an empty string is a valid key for variant, while we should gracefully error, it shouldn't be considered a valid key

fallintoplace:
I checked the Variant specification. Object field names are strings, and the only field-name restriction is uniqueness; dictionary strings must be UTF-8, but there is no non-empty requirement. An empty string is also a valid JSON object key, for example {"": null}.
Parquet Java’s metadata and object builders accept empty keys as well, so rejecting them here would make Go reject data produced by another conforming implementation.

zeroshade (approved):
Fair enough, I didn't realize the other impls accepted empty strings

PR #988, opened 2026-07-23
fix(arrow/array): support zero-column tables

fallintoplace:
### Rationale for this change

Arrow schemas and record batches may have zero fields while still carrying a logical row count. `NewTableFromSlice` indexed the first column unconditionally and panicked for an empty schema. `NewTableFromRecords` inferred rows from columns, so zero-column record batches lost their row counts.

Handle the empty column slice without indexing `cols[0]`. When the schema has no fields, sum `RecordBatch.NumRows` to preserve the logical row count. Tables with columns keep the existing column-based row inference. Regression tests cover both constructors.

### Are these changes tested?

Yes:

- `go test ./arrow/array`
- `go test -race -count=1 ./arrow/array`

Zero-column table construction no longer panics, and zero-column record batches retain their logical row counts. Existing behavior for tables with columns is unchanged.

### DIFF
--- a/arrow/array/table.go
+++ b/arrow/array/table.go
@@ -149,10 +149,15 @@ func NewTableFromSlice(schema *arrow.Schema, data [][]arrow.Array) arrow.Table {
 		chunked.Release()
 	}
 
+	var rows int64
+	if len(cols) > 0 {
+		rows = int64(cols[0].Len())
+	}
+
 	tbl := simpleTable{
 		schema: schema,
 		cols:   cols,
-		rows:   int64(cols[0].Len()),
+		rows:   rows,
 	}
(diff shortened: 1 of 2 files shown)


zeroshade:
Under what scenario would we have zero-fields but a non-zero logical row count? That doesn't make sense to me

fallintoplace:
The concrete case is projecting or removing all columns while retaining row cardinality. Arrow models the record batch length independently from its fields, so a zero-column batch can still represent N rows. Arrow Go already supports constructing, filtering, serializing, and reading those batches.

zeroshade (approved):
It's an odd case, but having a clear behavior instead of a panic is still an improvement. Given it's a pretty small change, I'm okay with it.

PR #992, opened 2026-07-23
fix(array): validate run-end encoded invariants

fallintoplace:
## Summary
- add RunEndEncoded Validate and ValidateFull implementations
- validate parent and child invariants, logical coverage, and strict run-end monotonicity
- add regression tests for short coverage and duplicate run ends

## Why
Run-end encoded helpers already assume that run ends are strictly increasing and cover the logical range. Today malformed arrays can pass validation even when those invariants do not hold.

## Validation
- go test ./arrow/array ./arrow/encoded

### DIFF
--- a/arrow/array/encoded.go
+++ b/arrow/array/encoded.go
@@ -251,11 +251,14 @@ func (r *RunEndEncoded) String() string {
 
 func validateRunEndEncoded(r *RunEndEncoded, full bool) error {
 	reeType := r.data.dtype.(*arrow.RunEndEncodedType)
-	runEndsData := r.data.childData[0]
+	runEndsData := r.data.childData[0].(*Data)
 	valuesData := r.data.childData[1]
 
 	if reeNullCount(r.data) != 0 {
 		return fmt.Errorf("arrow/array: run-end encoded array cannot contain nulls")
+	}
+	if err := validateArrayData(runEndsData); err != nil {
+		return fmt.Errorf("arrow/array: run ends array invalid: %w", err)
 	}
 	if !arrow.TypeEqual(runEndsData.DataType(), reeType.RunEnds()) {
 		return fmt.Errorf("arrow/array: run ends array must match parent type %s, got %s", reeType.RunEnds(), runEndsData.DataType())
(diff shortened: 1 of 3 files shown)


arrow/array/encoded.go
```diff
@@ -240,6 +249,102 @@ func (r *RunEndEncoded) String() string {
...
+
+	runEnds := getRunEnds64(runEndsData)
+	lastRunEnd := runEnds[len(runEnds)-1]
+	if lastRunEnd < int64(r.data.offset+r.data.length) {
+		return fmt.Errorf("arrow/array: last run end is %d but it should cover %d", lastRunEnd, r.data.offset+r.data.length)
+	}
+
+	if !full {
+		return nil
+	}
```
zeroshade: In the `!full` path, we're copying the entirety of the run ends only to read the single `runEnds[len-1]`. Instead for the `!full` path we should *only* check the last run end instead of doing the full copy loop

arrow/array/encoded.go
```diff
@@ -240,6 +249,102 @@ func (r *RunEndEncoded) String() string {
...
+		runEnds = runEnds[data.Offset() : data.Offset()+data.Len()]
+		out := make([]int64, len(runEnds))
+		for i, v := range runEnds {
+			out[i] = int64(v)
+		}
+		return out
+	default:
+		runEnds := arrow.Int64Traits.CastFromBytes(data.Buffers()[1].Bytes())
+		return runEnds[data.Offset() : data.Offset()+data.Len()]
+	}
```
zeroshade: at this point we haven't validated the length/buffers yet, so we need to check `len(runEnds)` before you start slicing using `data.Offset()+data.Len()` otherwise we could panic on a corrupt buffer.

arrow/array/encoded.go
```diff
@@ -240,6 +249,102 @@ func (r *RunEndEncoded) String() string {
...
+	if len(data.buffers) > 0 && data.buffers[0] != nil {
+		return data.length - bitutil.CountSetBits(data.buffers[0].Bytes(), data.offset, data.length)
+	}
+	return 0
+}
+
+func runEndTypeLimit(id arrow.Type) int64 {
+	switch id {
+	case arrow.INT16:
+		return 1<<15 - 1
```
zeroshade: `math.MaxInt16` ?

arrow/array/encoded.go
```diff
@@ -240,6 +249,102 @@ func (r *RunEndEncoded) String() string {
...
+	case arrow.INT32:
+		return math.MaxInt32
+	default:
+		return math.MaxInt64
+	}
+}
+
+func getRunEnds64(data arrow.ArrayData) []int64 {
```
zeroshade: there's already a `encoded.getRunEnds` function at `ree_utils.go:109` which returns a non-allocating closure, could we use export/reuse that pattern instead of allocating a new []int64?

zeroshade (changes requested):
Can we add a test that goes through the `array.Validate`/`array.ValidateFull` functions also?

zeroshade (approved):

PR #994, opened 2026-07-23
fix(array): validate binary_view and string_view payloads

fallintoplace:
## Summary
- add BinaryView and StringView validators
- validate view header ranges, referenced buffers, prefixes, and inline padding during ValidateFull
- validate UTF-8 for StringView payloads and add regressions for malformed headers

## Why
BinaryView and StringView accessors currently trust header metadata directly. Malformed view headers can pass validation today even when they reference missing data or carry invalid payload metadata.

## Validation
- go test ./arrow/array
- go test ./arrow/compute/...

### DIFF
--- a/arrow/array/binary.go
+++ b/arrow/array/binary.go
@@ -581,6 +581,9 @@ func validateViewLayout(arr ViewLike, kind string) error {
 
 func validateViewValues(arr ViewLike, dataBuffers []*memory.Buffer, validateValue func(int, []byte) error) error {
 	data := arr.Data().(*Data)
+	if data.length == 0 {
+		return nil
+	}
 	rawViews := data.buffers[1].Bytes()
 	for i := 0; i < data.length; i++ {
 		if arr.IsNull(i) {
@@ -595,7 +598,7 @@ func validateViewValues(arr ViewLike, dataBuffers []*memory.Buffer, validateValu
 		if view.IsInline() {
 			rawOffset := (data.offset + i) * arrow.ViewHeaderSizeBytes
 			raw := rawViews[rawOffset : rawOffset+arrow.ViewHeaderSizeBytes]
-			for _, b := range raw[4+view.Len():arrow.ViewHeaderSizeBytes] {
+			for _, b := range raw[4+view.Len() : arrow.ViewHeaderSizeBytes] {
 				if b != 0 {
 					return fmt.Errorf("arrow/array: view at slot %d was inline with size %d but its padding bytes were not all zero", i, view.Len())
 				}
--- a/arrow/array/binary.go
+++ b/arrow/array/binary.go
@@ -482,6 +482,17 @@ func (a *BinaryView) ValueLen(i int) int {
 	return s.Len()
 }
 
+func (a *BinaryView) Validate() error {
+	return validateViewLayout(a, "binary view")
+}
+
+func (a *BinaryView) ValidateFull() error {
+	if err := a.Validate(); err != nil {
+		return err
+	}
+	return validateViewValues(a, a.dataBuffers, nil)
+}
+
 // ValueString returns the value at index i as a string instead of
 // a byte slice, without copying the underlying data.
 func (a *BinaryView) ValueString(i int) string {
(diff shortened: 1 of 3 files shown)


arrow/array/binary.go
```diff
@@ -552,6 +563,86 @@ func arrayEqualBinaryView(left, right *BinaryView) bool {
...
+	expNumViews := data.offset + data.length
+	if len(data.buffers[1].Bytes())/arrow.ViewHeaderSizeBytes < expNumViews {
+		return fmt.Errorf("arrow/array: %s buffer must have at least %d view values", kind, expNumViews)
+	}
+	return nil
+}
+
+func validateViewValues(arr ViewLike, dataBuffers []*memory.Buffer, validateValue func(int, []byte) error) error {
+	data := arr.Data().(*Data)
+	rawViews := data.buffers[1].Bytes()
```
zeroshade, line 587: **Blocking: `ValidateFull` panics on empty (length-0) view arrays.**

`Validate()` (via `validateViewLayout`) returns early when `length == 0` and never dereferences `buffers[1]`, but for an empty view array `buffers[1]` is `nil`, so this hoisted `data.buffers[1].Bytes()` panics.

Reproduced on the standard builder output (and on hand-built data):
```go
array.ValidateFull(array.NewBinaryViewBuilder(mem).NewBinaryViewArray()) // panic: invalid memory address or nil pointer dereference
array.ValidateFull(array.NewStringViewBuilder(mem).NewStringViewArray()) // panic
```

This regressed in the "tighten view validation helpers" commit — patch 1 only touched `buffers[1]` inside the loop, so empty arrays were safe. It also undercuts the helper's intent: validation should report malformed data, not crash on a valid empty array.

```suggestion
	if data.length == 0 {
		return nil
	}
	rawViews := data.buffers[1].Bytes()
```

arrow/array/binary.go
```diff
@@ -552,6 +563,86 @@ func arrayEqualBinaryView(left, right *BinaryView) bool {
...
+
+		view := arr.ValueHeader(i)
+		if view.Len() < 0 {
+			return fmt.Errorf("arrow/array: view at slot %d has negative size %d", i, view.Len())
+		}
+
+		if view.IsInline() {
+			rawOffset := (data.offset + i) * arrow.ViewHeaderSizeBytes
+			raw := rawViews[rawOffset : rawOffset+arrow.ViewHeaderSizeBytes]
+			for _, b := range raw[4+view.Len():arrow.ViewHeaderSizeBytes] {
```
zeroshade: `gofmt` fails here (CI lint) — it wants spaces around `:` when the slice bounds are expressions:

```suggestion
			for _, b := range raw[4+view.Len() : arrow.ViewHeaderSizeBytes] {
```

arrow/array/validate_test.go
```diff
@@ -187,6 +216,71 @@ func TestLargeStringValidate(t *testing.T) {
 	})
 }
 
+func TestBinaryViewValidate(t *testing.T) {
```
zeroshade, line 241: Nice regression set. A few high-value cases are missing — the first would have caught the empty-array panic in `validateViewValues`:

- **Empty (length-0)** array via `array.ValidateFull` (currently panics).
- Drive the **public** `array.Validate` / `array.ValidateFull` entrypoints, not just the methods directly, so framework dispatch + layout checks are covered.
- An array with **`offset > 0`** to lock in the `(data.offset+i)*ViewHeaderSizeBytes` indexing.
- Out-of-line **happy path** plus untested failure modes: prefix mismatch, negative / out-of-range buffer offset, and `offset+size > buffer length`.

zeroshade (changes requested):
Thanks for this — the per-view invariants are a faithful, complete port of Arrow C++ `ValidateBinaryView` (negative size, inline padding-zero, buffer index/offset bounds, prefix match, and UTF-8 for `StringView`), and the cheap `Validate` / full `ValidateFull` split respects the O(1) vs O(n) contract. Tests pass and `go vet` is clean.

Requesting changes on two items:

1. **Blocking:** `ValidateFull` panics with a nil-pointer dereference on any empty (length-0) `BinaryView`/`StringView`, including the standard builder output. `validateViewValues` unconditionally dereferences `buffers[1]`, which is `nil` for empty arrays (`Validate` returns early and never touches it). This regressed in the "tighten view validation helpers" commit. Repro + one-line guard inline.
2. **CI lint:** `gofmt` fails on the inline-padding slice expression. Suggestion inline.

Non-blocking: a couple of tests worth adding (inline on the test file) — an empty-array case would have caught #1, and the out-of-line path is currently under-tested.

Details inline.

zeroshade (dismissed):
Thanks for this — the per-view invariants are a faithful, complete port of Arrow C++ `ValidateBinaryView` (negative size, inline padding-zero, buffer index/offset bounds, prefix match, and UTF-8 for `StringView`), and the cheap `Validate` / full `ValidateFull` split respects the O(1) vs O(n) contract. Tests pass and `go vet` is clean.

Requesting changes on two items:

1. **Blocking:** `ValidateFull` panics with a nil-pointer dereference on any empty (length-0) `BinaryView`/`StringView`, including the standard builder output. `validateViewValues` unconditionally dereferences `buffers[1]`, which is `nil` for empty arrays (`Validate` returns early and never touches it). This regressed in the "tighten view validation helpers" commit. Repro + one-line guard inline.
2. **CI lint:** `gofmt` fails on the inline-padding slice expression. Suggestion inline.

Non-blocking: a couple of tests worth adding (inline on the test file) — an empty-array case would have caught #1, and the out-of-line path is currently under-tested.

Details inline.

zeroshade (approved):
Issues fixed, just needs to have the merge conflicts resolved

PR #996, opened 2026-07-24
fix(parquet/metadata): compare bloom filter caps in bits

fallintoplace:
The bloom filter sizing path computes `m` in bits, but the cap check compared it to the byte limit shifted the wrong way. In practice that can send a moderate-sized filter straight to the 128 MiB cap.

This switches the comparison to bits and adds a regression that covers both the capped case and a case that should stay in the low-megabyte range.

Tests: `go test ./parquet/metadata`

### DIFF
--- a/parquet/metadata/bloom_filter.go
+++ b/parquet/metadata/bloom_filter.go
@@ -69,7 +69,7 @@ func optimalNumBits(ndv uint32, fpp float64) uint32 {
 		numBits uint32
 	)
 
-	if m < 0 || m > maximumBloomFilterBytes>>3 {
+	if m < 0 || m > maximumBloomFilterBytes<<3 {
 		numBits = maximumBloomFilterBytes << 3
 	} else {
 		numBits = uint32(m)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`m\` is computed as a **bit** count but was compared against \`maximumBloomFilterBytes >> 3\`, i.e. a byte limit shifted the wrong way — so any filter needing roughly 16.7M to 1.07G bits was clamped straight to the 128 MiB maximum. Concretely, for \`ndv=2_000_000, fpp=0.01\` the old path returned 134217728 bytes where the correct answer is 4194304. Converting the cap to bits with \`<< 3\` is right, and the added case genuinely distinguishes the buggy from the fixed result. LGTM.

Unrelated note for a follow-up: \`TestNewBloomFilter\` depends on \`runtime.GC()\` triggering the \`addCleanup\` finalizer before the deferred \`mem.AssertSize(t, 0)\` runs, which makes it flaky — I saw it report spurious \`LEAK ... FROM NewBloomFilterFromNDVAndFPP\` on unrelated PRs. Worth hardening separately.

PR #997, opened 2026-07-24
fix(arrow/compute): use is_null/is_not_null in expressions

fallintoplace:
`IsNull` was building a unary `less` call, so even formatting or serializing the expression could go sideways. `IsValid` was also using `is_valid`, but the registered kernel name here is `is_not_null`.

This points both helpers at the registered names and adds regression coverage for string formatting plus expression round-tripping.

Tests: `go test ./arrow/compute`


zeroshade (approved):
Verified against \`main\`: \`is_null\` and \`is_not_null\` are the actual registered kernel names, while \`is_valid\` is not registered anywhere — so the old \`IsValid\` could never bind. The old \`IsNull\` built a *binary* \`less\` call with a single argument, which meant even \`.String()\` on the expression panicked (index out of range in the binary renderer). Both helpers now point at the correct unary kernels. LGTM.

PR #998, opened 2026-07-24
fix(arrow/ipc): reject NaN minSpaceSavings values

fallintoplace:
`WithMinSpaceSavings` documents the accepted range as `[0, 1]`, but `NaN` slipped through because both comparisons against `NaN` are false.

This rejects `NaN` alongside the existing range checks and extends the validation test with that input.

Tests: `go test ./arrow/ipc`

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -486,7 +486,7 @@ func (w *recordEncoder) encode(p *Payload, rec arrow.RecordBatch) error {
 	}
 
 	if w.codec != -1 {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`NaN\` passes the existing \`< 0 || > 1\` guard because both comparisons are false for NaN, after which \`savings >= minSpaceSavings\` is always false and compression is silently skipped. Since \`WithMinSpaceSavings\` documents \`[0, 1]\` as the accepted range, rejecting NaN matches the documented contract. LGTM.

PR #999, opened 2026-07-24
fix(arrow/tensor): default nil dimension names to empty strings

fallintoplace:
The tensor constructors document that `nil` dimension names are treated as empty strings, but the shared constructor left the slice nil. That makes `DimName` panic on an otherwise valid tensor.

This fills in one empty name per dimension and adds coverage for both `DimNames` and `DimName`.

Tests: `go test ./arrow/tensor`


zeroshade (approved):
Verified: \`tensor.New(..., nil)\` is a supported call, but \`DimName(i)\` indexed \`names[i]\` with no nil/bounds guard — I reproduced the panic on an otherwise valid tensor. Defaulting nil dimension names to one empty string per dimension matches both the documented behavior and C++ tensor semantics. LGTM.

PR #1000, opened 2026-07-24
fix(arrow/scalar): allow null extension scalars to keep null storage

fallintoplace:
`MakeNullScalar` represents a null extension scalar with a null storage scalar. `Extension.Validate` rejected that factory-created shape even though it is the natural null representation.

This allows validated null storage for null extension scalars, still rejects non-null storage, and adds regression coverage for both cases.

Tests: `go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/scalar.go
+++ b/arrow/scalar/scalar.go
@@ -421,10 +421,16 @@ func (e *Extension) Validate() (err error) {
 	}
 
 	if !e.Valid {
-		if e.Value != nil {
-			err = fmt.Errorf("null %s scalar has storage value", e.Type)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`MakeNullScalar\` on an extension type produces \`Valid=false\` with a non-nil (but null) storage scalar, and the old \`Extension.Validate\` rejected any non-nil storage when invalid — so the factory produced a scalar that failed its own validation. I reproduced the "null extension_type<storage=int16> scalar has storage value" error on \`main\`. The fix correctly permits null storage while still rejecting non-null storage. LGTM.

PR #1001, opened 2026-07-24
fix(arrow/scalar): handle nil child values in run-end scalars

fallintoplace:
`MakeNullScalar` creates a null run-end encoded scalar without a child value. `RunEndEncoded.Validate` dereferenced that value before checking nullness, so validation could panic.

This handles the missing child explicitly: null scalars are accepted, malformed non-null scalars still fail, and both paths are covered by tests.

Tests: `go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/nested.go
+++ b/arrow/scalar/nested.go
@@ -754,6 +754,13 @@ func (s *RunEndEncoded) Release() {
 func (s *RunEndEncoded) value() interface{} { return s.Value.value() }
 
 func (s *RunEndEncoded) Validate() (err error) {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`MakeNullScalar\` for a run-end encoded type leaves \`Value == nil\`, and \`RunEndEncoded.Validate\` dereferenced it immediately — I reproduced the nil-pointer panic. Handling the missing child explicitly (accept for null scalars, error for a non-null scalar with no value) is correct. LGTM.

PR #1004, opened 2026-07-24
fix(parquet/compress): allocate uncompressed output for short destinations

fallintoplace:
The codec contract allows a `nil` or undersized destination and expects the codec to allocate when needed. The uncompressed codec was just returning the unchanged destination, which could leave callers with empty or truncated output.

This switches `Encode`, `EncodeLevel`, and `Decode` to append-based copying and adds coverage for `nil` and undersized destinations.

Tests: `go test ./parquet/compress`

### DIFF
--- a/parquet/compress/compress.go
+++ b/parquet/compress/compress.go
@@ -165,10 +165,7 @@ func (nocodec) NewReader(r io.Reader) io.ReadCloser {
 }
 
 func (nocodec) Decode(dst, src []byte) []byte {
-	if dst != nil {
-		copy(dst, src)
-	}
-	return dst
+	return append(dst[:0], src...)
 }
 
 func (n nocodec) DecodeWithError(dst, src []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified against current `main`: `nocodec.Encode`/`EncodeLevel`/`Decode` do `copy(dst, src); return dst`, which silently truncates when `dst` is non-nil but shorter than `src` — directly contradicting the documented `Codec` contract that a nil or too-small `dst` results in a newly allocated slice. Switching to `append(dst[:0], src...)` is the right fix: it returns full-length output, reuses `dst`'s capacity when it is large enough (so no extra allocation in the common case), and allocates only when required. `dst == nil` works correctly since `nil[:0]` is valid.

The internal reader path is unaffected because `buf` is always sized to `lenUncompressed`, making the lengths equal. The test genuinely catches the bug — pre-fix, `Encode(make([]byte, 1), "arrow")` returns just `"a"`. LGTM.

Note: I originally flagged this as stacked on #1003, but I re-verified against current `main` and the two are now independent — I applied #1003 to a scratch worktree and this still applied cleanly, so no ordering constraint remains.

PR #1005, opened 2026-07-24
fix(internal/utils): keep byteReader position after failed seeks

fallintoplace:
`byteReader` updated its cached position even when the underlying seek failed, which could make `Peek` and `Read` disagree about where the reader was. `Discard` also returned `io.EOF` when it consumed exactly the remaining bytes.

This only updates the cached position after a successful seek and treats exact exhaustion as a successful discard. Added focused regressions for both behaviors.

Tests: `go test ./internal/utils`

### DIFF
--- a/internal/utils/buf_reader.go
+++ b/internal/utils/buf_reader.go
@@ -55,7 +55,9 @@ func (r *byteReader) Read(buf []byte) (n int, err error) {
 
 func (r *byteReader) Seek(offset int64, whence int) (pos int64, err error) {
 	pos, err = r.r.Seek(offset, whence)
-	r.pos = int(pos)
+	if err == nil {
+		r.pos = int(pos)
+	}
 	return
 }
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified both fixes against current `main`:

1. `Seek` unconditionally did `r.pos = int(pos)`. Since `bytes.Reader.Seek` returns `pos = 0` on failure (invalid whence or negative resulting position), a failed seek corrupted the tracked position to 0 while the underlying reader stayed put — leaving `Peek` and `Read` disagreeing about the location. Guarding on `err == nil` is correct.

2. `Discard` used `newPos >= len(r.buf)`, so discarding *exactly* the remaining bytes reported `io.EOF` even though all `n` bytes were successfully discarded. `>` is right and matches both stdlib `bufio.Reader.Discard` semantics and the sibling `bufferedReader.Discard` in this same file.

No regression risk on the EOF change: the only caller (`page_reader.go`) ignores `Discard`'s return values. Both tests distinguish pre- from post-fix behavior. LGTM.

PR #1006, opened 2026-07-25
fix(parquet/pqarrow): preserve INT96 timestamps across units

fallintoplace:
### Rationale for this change

INT96 timestamp conversion currently routes values through `time.Duration` and applies a nanoseconds-per-day modulus directly to the source value. This produces incorrect results for second, millisecond, and microsecond timestamps, overflows for timestamps outside the duration range, and leaves negative timestamps with a negative time-of-day remainder.

- Convert each Arrow timestamp unit using integer units-per-day and nanoseconds-per-unit constants.
- Normalize negative remainders onto the preceding Julian day.
- Add focused coverage for every timestamp unit on both sides of the Unix epoch.

### Are these changes tested?

Yes. The focused pqarrow tests pass.


zeroshade (changes requested):
The fix itself is correct and this is the highest-impact bug in your current batch — I verified it thoroughly. On `main`, `arrowTimestampToImpalaTimestamp` computes `lastDayNanos := t % nanoSecondsPerDay` using the raw value in its *source* unit while always modding by nanoseconds-per-day, so it is only correct for `arrow.Nanosecond`. For microseconds at 2000-01-01T00:00:01 the old code yields `82684801000000` instead of `1000000000`, and for a pre-epoch microsecond value it produces the wrong Julian day plus a negative nanos-of-day that wraps under `uint64`. Your rewrite is the correct inverse of the read side and cannot overflow, since `remainder * nanosPerUnit < 86.4e12`. The existing `TestWriteArrowInt96` only covers the nanosecond column, which is exactly why this went unnoticed.

I am holding this one on the test rather than the fix. The new test recomputes the implementation formula (`days`, `remainder`, `days + julianEpochOffsetDays`, `remainder * nanosPerUnit`) and asserts the result equals that same computation, so it would pass against the buggy code too and provides no regression protection.

Could you replace it with something independent of the implementation? Either of these works:
- assert hard-coded expected `Int96` bytes / Julian day + nanos-of-day values for each unit, or
- round-trip through `parquet.Int96.ToTime()` and compare against the original `arrow.Timestamp`.

Please cover all four units (s/ms/us/ns) and include at least one pre-epoch value, since the negative-remainder normalization is the subtlest part of the change. Happy to merge as soon as the test actually exercises the bug.

zeroshade (approved):
The new test addresses my earlier concern completely — thank you. `TestArrowTimestampToImpalaTimestamp` now asserts `got.ToTime()` equals `tt.value.ToTime(tt.unit)`, which validates against the independent read-side conversion rather than recomputing the implementation formula, so it would genuinely fail against the pre-fix code. Covering all four units on both sides of the epoch also exercises the negative-remainder normalization, which was the subtlest part of the change.

Restating the verification of the fix itself for the record: on `main`, `arrowTimestampToImpalaTimestamp` computed `lastDayNanos := t % nanoSecondsPerDay` from the raw value in its *source* unit while always modding by nanoseconds-per-day, so it was only correct for `arrow.Nanosecond`. For microseconds at 2000-01-01T00:00:01 it produced `82684801000000` instead of `1000000000`, and for pre-epoch values it produced the wrong Julian day plus a negative nanos-of-day that wrapped under `uint64`. The rewrite is the correct inverse of `Int96.ToTime()`, the Julian epoch offset and per-unit constants are right, and `remainder * nanosPerUnit` cannot overflow since it stays below 86.4e12 — actually more robust than the old `time.Duration` path, which could overflow for large second values. Dropping the `time` import is safe as all its uses were inside the rewritten function.

This was the highest-impact fix in the batch, since it silently corrupted every non-nanosecond INT96 timestamp and the existing `TestWriteArrowInt96` only covered the nanosecond column. LGTM.

PR #1007, opened 2026-07-25
fix(parquet/metadata): preserve infinite floating-point statistics

fallintoplace:
### Rationale for this change

Floating-point statistics use the largest finite values as their initial min/max bounds. Those are not identity values when a column contains only positive or negative infinity, so the writer records a finite bound that was not present in the data.

- Seed FLOAT, DOUBLE, and FLOAT16 statistics with positive and negative infinity.
- Treat an inverted min/max pair as the empty or all-NaN sentinel.
- Cover all-positive-infinity and all-negative-infinity inputs for all three types.

### Are these changes tested?

Yes. The focused statistics tests, including the existing NaN and signed-zero coverage, pass.

### DIFF
--- a/parquet/metadata/statistics.go
+++ b/parquet/metadata/statistics.go
@@ -388,8 +388,8 @@ var (
 	defaultMaxInt96  parquet.Int96
 	defaultMaxUInt96 parquet.Int96
 
-	defaultMinFloat16 parquet.FixedLenByteArray = float16.MaxNum.ToLEBytes()
-	defaultMaxFloat16 parquet.FixedLenByteArray = float16.MinNum.ToLEBytes()
+	defaultMinFloat16 parquet.FixedLenByteArray = float16.Inf().ToLEBytes()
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: the float statistics seed min/max with `MaxFloat32` / `-MaxFloat32`, which are not identity elements — for a column of all `+Inf`, `min(MaxFloat32, +Inf)` stays `MaxFloat32`, so the writer records a finite bound that never appeared in the data (and symmetrically the max for all `-Inf`). Seeding with `±Inf` is the correct choice, and replacing the sentinel emptiness check with `min > max` correctly identifies the untouched-default case without false positives: NaN is filtered earlier, and real data always satisfies `min <= max`. The Float16 path is updated consistently through `defaultMinFloat16`/`defaultMaxFloat16` and `cleanStat`. LGTM.

The initial CI failure here was an unrelated flake in `TestNewBloomFilter`, which depends on `runtime.GC()` running the `addCleanup` finalizer before its deferred `AssertSize`; it passed on rerun.

PR #1008, opened 2026-07-25
fix(parquet/pqarrow): reject timestamp coercion overflow

fallintoplace:
### Rationale for this change

Timestamp coercion to a finer unit multiplies int64 values without checking whether the result fits. Values near the int64 limits can therefore wrap and be written as a different timestamp.

- Check the int64 range before multiplying a valid timestamp.
- Return `arrow.ErrInvalid` when coercion would overflow.
- Ignore the stored value in null slots, matching the existing truncation checks.

### Are these changes tested?

Yes. Focused tests cover positive and negative overflow as well as an overflowing value in a null slot.


zeroshade (approved):
Verified: \`multiply\` performed \`int64(val) * factor\` with no overflow guard, while the sibling \`divide\` already guarded \`arr.IsValid(idx)\`. The boundary is exactly correct with no off-by-one — since any product is a multiple of \`factor\`, \`MaxInt64/factor\` is the true last-safe value, so there are no false positives. Skipping null slots is safe because the output buffer is zero-backed and null slots are masked when written, and it prevents a spurious overflow error from garbage stored in a null slot. LGTM.

PR #1009, opened 2026-07-25
fix(parquet/file): reject records starting with nonzero repetition levels

fallintoplace:
### Rationale for this change

A repeated column must start each logical record at repetition level zero. The record reader currently accepts a nonzero first repetition level and silently treats it as part of a record, which can miscount record boundaries.

- Return an error when record delimiting begins on a nonzero repetition level.
- Propagate that error through `ReadRecordData`.
- Add a focused test that also verifies the invalid level is not consumed.

### Are these changes tested?

Yes. The focused parquet/file test passes.

### DIFF
--- a/parquet/file/record_reader.go
+++ b/parquet/file/record_reader.go
@@ -522,7 +522,7 @@ func (rr *recordReader) Reset() {
 // process no more levels than necessary to delimit the indicated
 // number of logical records. updates internal state of recordreader
 // returns number of records delimited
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: a repeated column must begin each logical record at repetition level 0, but the delimiter accepted a nonzero first repetition level and silently folded it into a record, miscounting boundaries. This cannot false-positive, since a genuine record start always has \`curRep == 0\`. The new error is threaded through \`ReadRecordData\` correctly, \`errors\` is already imported, and the test confirms the invalid level is not consumed. LGTM.

PR #1010, opened 2026-07-25
fix(parquet/metadata): reject malformed fixed-width statistics

fallintoplace:
### Rationale for this change

Fixed-width Parquet statistics must use the exact PLAIN-encoded physical width. Padding a shorter value invents a different bound, while formatting the raw bytes can panic. Variable-length BYTE_ARRAY bounds should remain unchanged.

- Ignore min/max when a present fixed-width statistic does not match the declared physical width.
- Preserve variable-length BYTE_ARRAY values exactly as encoded.
- Return malformed scalar bytes unchanged from `GetStatValue` instead of padding or indexing past them.
- Regenerate the typed statistics implementation from its template.

### Are these changes tested?

Yes. Focused tests cover malformed numeric and FIXED_LEN_BYTE_ARRAY values, short scalar formatting, and BYTE_ARRAY preservation. The metadata and Parquet file test suites pass.

### DIFF
--- a/parquet/metadata/statistics.go
+++ b/parquet/metadata/statistics.go
@@ -605,12 +605,35 @@ func (FixedLenByteArrayStatistics) cleanStat(minMax minmaxPairFixedLenByteArray)
 	return &minMax
 }
 
+func fixedStatByteWidth(typ parquet.Type) (int, bool) {
+	switch typ {
+	case parquet.Types.Boolean:
+		return 1, true
+	case parquet.Types.Int32, parquet.Types.Float:
+		return 4, true
+	case parquet.Types.Int64, parquet.Types.Double:
+		return 8, true
+	case parquet.Types.Int96:
+		return 12, true
+	default:
+		return 0, false
+	}
+}
+
+func validEncodedStat(descr *schema.Column, val []byte) bool {
+	if descr.PhysicalType() == parquet.Types.ByteArray {
+		return true
+	}
+	if descr.PhysicalType() == parquet.Types.FixedLenByteArray {
+		return len(val) == int(descr.TypeLength())
+	}
+	width, ok := fixedStatByteWidth(descr.PhysicalType())
+	return ok && len(val) == width
+}
+
 func GetStatValue(typ parquet.Type, val []byte) interface{} {
-	width := typ.ByteSize()
-	if width > 0 && len(val) < width {
-		padded := make([]byte, width)
-		copy(padded, val)
-		val = padded
+	if width, ok := fixedStatByteWidth(typ); ok && len(val) != width {
+		return val
 	}
 
 	switch typ {
--- a/parquet/metadata/statistics.go
+++ b/parquet/metadata/statistics.go
@@ -606,6 +606,13 @@ func (FixedLenByteArrayStatistics) cleanStat(minMax minmaxPairFixedLenByteArray)
 }
 
 func GetStatValue(typ parquet.Type, val []byte) interface{} {
+	width := typ.ByteSize()
+	if width > 0 && len(val) < width {
+		padded := make([]byte, width)
+		copy(padded, val)
+		val = padded
+	}
+
 	switch typ {
 	case parquet.Types.Boolean:
 		return val[0] != 0
--- a/parquet/metadata/statistics_test.go
+++ b/parquet/metadata/statistics_test.go
@@ -859,16 +859,14 @@ func TestByteArrayStatisticsFromEncodedOwnsMinMax(t *testing.T) {
 	assert.Equal(t, "zzz", string(stats.Max()))
 }
 
-func TestTruncatedFixedWidthStatisticsAreZeroPadded(t *testing.T) {
+func TestMalformedFixedWidthStatisticsAreIgnored(t *testing.T) {
 	descr := schema.NewColumn(schema.NewInt32Node("i32", parquet.Repetitions.Required, -1), 0, 0)
 	stats := metadata.NewStatisticsFromEncoded(descr, memory.DefaultAllocator, 1,
 		&encodedStatProvider{min: []byte{0x34, 0x12}, max: []byte{0x78}}).(*metadata.Int32Statistics)
 
-	require.True(t, stats.HasMinMax())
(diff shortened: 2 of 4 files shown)


zeroshade (changes requested):
The core idea here is sound and I'd like to take it, but this PR is not reviewable or mergeable in its current form — it appears to be stacked on other unmerged work, so it does not apply to `main`.

`git apply --check` fails on all three files. Specifically:

- `encodedStatProvider`, which the new tests rely on, does not exist in `main`.
- `TestByteArrayStatisticsFromEncodedOwnsMinMax`, which the diff treats as existing context, does not exist in `main`.
- The diff's *unchanged context* includes the `append(ret.min[:0], ...)` "statistics-owned buffer" copy logic, which is present in neither the current template nor the current generated file.

So the tests would not compile and the hunks won't land. Could you rebase onto `main` and restore whatever test helpers are needed?

On the substance, which I did verify as far as I could: the defect is real. Malformed fixed-width statistics currently produce garbage bounds while `HasMinMax()` still reports true — the FLBA decoder returns an error that `plainDecode` discards, leaving an empty `min`/`max`. And `GetStatValue` panics outright on short input via `binary.LittleEndian.Uint32`. I also checked `validEncodedStat` / `fixedStatByteWidth` against every `PrimitiveType` and it rejects no valid encoding: exact width for the fixed types, `== TypeLength()` for FLBA, and always-true for BYTE_ARRAY so genuinely-empty string bounds still round-trip.

Two things to please address on the rebase:

1. **Re-run `go:generate`** so the template and generated file are provably in sync. Right now I can only confirm they're internally consistent with each other, not that the checked-in `statistics_types.gen.go` is what the modified `.tmpl` actually produces.
2. **Double-check the FLBA `plainDecode` change to `return src`.** That returns a slice aliasing the caller's encoded metadata buffer. It's safe in the `FromEncoded` path *only* because of the `append(ret.min[:0], ...)` copy — which is part of the unmerged base, not `main`. If it lands without that copy, the statistics would alias the metadata buffer.

zeroshade (approved):
Re-reviewed at `c87e4c95`. All three of my earlier blockers are resolved, CI is green (22/22), and I verified the generated-code question properly this time by actually running the generator. Good to go.

**Regeneration parity — the one I could not check before: exact match, zero drift.** Using the directive at `statistics.go:38` (`go run ../../arrow/_tools/tmpl/main.go -i -data=statistics_types.tmpldata statistics_types.gen.go.tmpl`), I regenerated in a clean worktree at `main` with your patch applied and compared to the committed blob:

| artifact | sha256 |
|---|---|
| `c87e4c95:parquet/metadata/statistics_types.gen.go` | `99a93d6f…f2a2` |
| freshly regenerated from your `.tmpl` | `99a93d6f…f2a2` |

`diff` empty, reproducible across two runs. So the `.gen.go` was not hand-edited and the template is the real source of truth.

**Applies to `main`**: yes, clean. **`encodedStatProvider`**: present at `statistics_test.go:823` — my earlier "missing helper" was an artifact of the stale base I was testing against, apologies for the noise there.

**No false rejections** — verified empirically rather than by inspection. Built real statistics for every physical type, encoded via `EncodeMin`/`EncodeMax`, and round-tripped through the guarded constructor: `BOOLEAN 1 · INT32 4 · INT64 8 · INT96 12 · FLOAT 4 · DOUBLE 8 · FLBA(8) 8 · FLOAT16 2 · BYTE_ARRAY variable`. All matched the guard's table, all retained `HasMinMax()==true`, all values round-tripped byte-identically. Three supporting facts: FLOAT16 correctly routes through the FLBA branch (`len == TypeLength() == 2`), `TypeLength() <= 0` is impossible for an FLBA column since `schema/node.go:179` rejects it at construction, and arrow-go never emits truncated fixed-width stats because `ApplyStatSizeLimits` *drops* oversized bounds rather than truncating — so the guard cannot reject arrow-go's own output.

**The `plainDecode` → `return src` aliasing concern is resolved**: both call sites (`statistics_types.gen.go:2261,2264`) wrap it in `append(ret.min[:0], …)`, so the caller copies. I probed it both ways — mutating the caller's encoded buffer afterwards does not perturb the stored statistic.

**And the bug is meaningfully worse than I originally described.** I had characterized it as garbage bounds plus a `GetStatValue` panic. On baseline, malformed fixed-width stats are silently zero-filled while still reporting `hasMinMax=true`:
```
INT32/2bytes   hasMinMax=true  min=[0 0 0 0]         max=[0 0 0 0]
FLBA8/2bytes   hasMinMax=true  min=[0 0 0 0 0 0 0 0] max=[...]
F16/1byte      hasMinMax=true  min=[0 0]             max=[0 0]
```
A reader trusting `min==max==0` will skip row groups via predicate pushdown, so this silently produces **wrong query results** — not just a crash. Please consider saying so in the commit message; it materially raises the value of this fix. Counts are correctly set before the early `return ret`, so `nvalues`/null/distinct survive when min/max are dropped.

Non-blocking nits, all fine to merge as-is:
1. `statistics.go:628` — `int(descr.TypeLength())` is a redundant conversion, `Column.TypeLength()` already returns `int`. CI won't flag it (`unconvert` isn't enabled).
2. `GetStatValue` now returns raw `[]byte` for a malformed fixed-width type. Safe today since the only caller (`parquet_reader/main.go:272`) formats with `%v`, but returning `nil` would match the function's own unknown-type fallback.
3. Arrow C++ *throws* here (`ParquetException("Failed to decode statistic value from plain encoded string")`, `cpp/src/parquet/statistics.cc:974`) rather than dropping. Silently dropping is the right call for Go since `NewStatisticsFromEncoded` has no error return, but worth a sentence so the divergence is intentional on the record.
4. Pre-existing and untouched: `hasMinMax = IsSetMax() || IsSetMin()` is arguably wrong (`&&`), since a one-sided statistic leaves the other bound at its zero value. Separate issue.

The 9 `parquet/file` / `parquet/...` test failures I saw locally are entirely environmental — baseline `main` produces the identical set, all from `PARQUET_TEST_DATA` being unset. Zero PR-caused failures. LGTM.

PR #1011, opened 2026-07-25
fix(parquet/encoding): bound lazy dictionary growth

fallintoplace:
### Rationale for this change

The lazy dictionary converter sizes its backing slice from the requested dictionary index before checking whether that index exists. A bad index can therefore trigger a much larger allocation than the dictionary itself. Removing the initial capacity also needs to preserve the declared dictionary length for Arrow dictionary insertion.

- Store the declared dictionary length separately from the lazy backing slice.
- Reject out-of-range indexes before allocating or decoding.
- Grow only by the values required for a valid index and reject short dictionary streams.
- Materialize the full declared dictionary when inserting it into an Arrow builder.

### Are these changes tested?

Yes. Focused tests cover bounded growth and declared length tracking. The encoding and pqarrow suites, including dictionary insertion paths, pass.

### DIFF
--- a/parquet/internal/encoding/byte_array_decoder.go
+++ b/parquet/internal/encoding/byte_array_decoder.go
@@ -190,8 +190,12 @@ func (pbad *PlainByteArrayDecoder) DecodeSpaced(out []parquet.ByteArray, nullCou
 
 func (d *DictByteArrayDecoder) InsertDictionary(bldr array.Builder) error {
 	conv := d.dictValueDecoder.(*dictConverter[parquet.ByteArray])
-	dictLength := cap(conv.dict)
-	conv.ensure(pqutils.IndexType(dictLength))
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified: `ensure()` sized its backing slice from the requested dictionary index *before* validating that index, so `make([]T, int(idx+1)-len(dc.dict))` on an attacker-controlled index from a malformed page could trigger an allocation far larger than the dictionary itself. Bounds-checking against the declared length first is the right fix, and tracking `dictLen` separately preserves the declared length for the Arrow dictionary-insertion path. Rejecting short dictionary streams is a good addition. LGTM.

Both initial CI failures were unrelated flakes (a segfault in `arrow/flight/session` and the GC-timing-dependent `TestNewBloomFilter` leak check); clean on rerun.

PR #1013, opened 2026-07-25
fix(parquet/encoding): reject invalid delta length payloads

fallintoplace:
### Rationale for this change

The DELTA_LENGTH_BYTE_ARRAY decoder slices payload data using unchecked lengths and keeps the logical page count after decoding fewer physical values. Invalid lengths can panic, and valid nullable pages can index past the physical length table.

- Propagate errors and short reads from delta prefix and length streams.
- Reject negative lengths and cumulative lengths beyond the suffix payload.
- Track the encoded physical length count in the byte-array decoder while retaining logical page-count validation.
- Require DELTA_BYTE_ARRAY prefix and suffix streams to contain the same number of physical values.

### Are these changes tested?

Yes. Focused tests cover negative and oversized lengths plus a nullable page with fewer physical values than logical positions. The full encoding and Parquet file suites pass.

### DIFF
--- a/parquet/internal/encoding/delta_byte_array.go
+++ b/parquet/internal/encoding/delta_byte_array.go
@@ -194,7 +194,13 @@ func (d *DeltaByteArrayDecoder) SetData(nvalues int, data []byte) error {
 	if offset < 0 || offset > int64(len(data)) {
 		return errors.New("parquet: invalid delta byte array suffix offset")
 	}
-	return d.DeltaLengthByteArrayDecoder.SetData(nvalues, data[offset:])
+	if err := d.DeltaLengthByteArrayDecoder.SetData(nvalues, data[offset:]); err != nil {
+		return err
+	}
+	if len(d.prefixLengths) != d.nvals {
+		return errors.New("parquet: delta prefix and suffix length counts do not match")
+	}
+	return nil
 }
 
 func (d *DeltaByteArrayDecoder) Discard(n int) (int, error) {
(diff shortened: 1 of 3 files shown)


zeroshade:
Reviewed and the analysis holds up: `DeltaLengthByteArrayDecoder.Decode` indexes `d.lengths[i]` with `nvals` derived from the *logical* count, so a nullable page with two logical positions and one physical value reads past the end of the length table. Switching to the encoded physical count plus the negative/oversized length validation is the right fix.

I just merged #1014, which rewrites the same `DeltaByteArrayDecoder.SetData` this PR touches, so this now needs a rebase onto `main` — the hunks overlap directly. Once you rebase and CI is green I will merge this. Note that #1014 already added the `lastVal` reset and the prefix-offset bounds check, so some of the overlap should collapse.

zeroshade (approved):
Thanks for the rebase onto #1014 — verified it resolved cleanly, with #1014's `lastVal` reset and suffix-offset bounds check now appearing as context rather than being duplicated. The `delta_byte_array.go` change is down to just the prefix/suffix count match.

Re-verified the core fix: `DeltaLengthByteArrayDecoder.Decode` indexes `d.lengths[i]` while `d.nvals` came from the *logical* value count, so a nullable page with two logical positions and one physical value read past the end of the length table. Passing the encoded physical count to the embedded decoder is correct, and the negative/cumulative-length validation closes the slicing panics. Propagating the error from `DecodeSpaced` instead of discarding it is a good catch too. LGTM.

The ARM64 CI failure was the unrelated flaky `TestNewBloomFilter` GC-timing leak check; green on rerun.

PR #1014, opened 2026-07-25
fix(parquet/encoding): reject invalid delta byte prefixes

fallintoplace:
### Rationale for this change

DELTA_BYTE_ARRAY prefix lengths are used directly as slice bounds. Negative prefixes or prefixes longer than the previous value therefore panic. The decoder also retains its previous value when reused for a new page, even though each page starts an independent delta sequence.

- Reset the previous value in `SetData`.
- Propagate prefix stream decode errors and short reads.
- Require the first prefix to be zero and validate later prefixes before slicing.
- Apply the same validation to decode and discard paths.

### Are these changes tested?

Yes. Focused tests cover decoder reuse, negative prefixes, and prefixes longer than the previous value.

### DIFF
--- a/parquet/internal/encoding/delta_byte_array.go
+++ b/parquet/internal/encoding/delta_byte_array.go
@@ -18,6 +18,7 @@ package encoding
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/internal/utils"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified two distinct real bugs here:

1. \`SetData\` never reset \`lastVal\`, so a decoder reused across pages skipped first-value initialization and consumed \`prefixLengths[0]\` as a continuation, corrupting the first value of the second page. Column readers do reuse decoders across pages, so this is reachable.
2. \`d.lastVal[:prefixLen:prefixLen]\` panicked on a negative prefix or one longer than the previous value, and the loop could index an empty \`prefixLengths\`.

Requiring the first prefix to be zero matches the DELTA_BYTE_ARRAY spec, and the validation is applied consistently to both the decode and discard paths. LGTM.

Note: #1013 modifies the same \`SetData\` and will need a rebase now that this has landed.

PR #1015, opened 2026-07-25
fix(parquet/encoding): respect declared RLE payload lengths

fallintoplace:
### Rationale for this change

RLE level and boolean decoders validate their declared byte length but then give the decoder every remaining byte in the page. A short declared payload can therefore consume bytes belonging to the next encoded section instead of reporting a truncated run.

- Limit Data Page V1 level decoding to its four-byte-length-prefixed payload.
- Limit Data Page V2 level decoding to its explicit byte count.
- Apply the same boundary to RLE boolean values.
- Validate the Data Page V2 length against the supplied data.

### Are these changes tested?

Yes. Focused tests verify that bytes after each declared RLE interval are not consumed.

### DIFF
--- a/parquet/internal/encoding/boolean_decoder.go
+++ b/parquet/internal/encoding/boolean_decoder.go
@@ -204,11 +204,11 @@ func (dec *RleBooleanDecoder) SetData(nvals int, data []byte) error {
 
 	// load the first 4 bytes in little-endian which indicates the length
 	nbytes := binary.LittleEndian.Uint32(data[:4])
-	if nbytes > uint32(len(data)-4) {
+	if uint64(nbytes) > uint64(len(data)-4) {
 		return fmt.Errorf("received invalid number of bytes - %d (corrupt data page?)", nbytes)
 	}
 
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified: both the V1 and V2 level decoders validated the declared byte length but then handed the RLE decoder every remaining byte in the page, so a short declared payload could silently consume bytes belonging to the next encoded section instead of reporting a truncated run. Slicing to `data[4 : 4+nbytes]` (and `data[:nbytes]` for V2, with the added length check against the supplied data) is the correct bound, and the same fix applied to the RLE boolean decoder is consistent. Widening the comparisons to `int64`/`uint64` also avoids the underflow when `len(data) < 4`. The tests confirm bytes past each declared interval are not consumed. LGTM.

For the record, the CI failures on this PR were all unrelated infrastructure flakes — a `signal: killed` in `parquet/file` on macOS and repeated segfaults in `arrow/flight/flightsql/example` (modernc pure-Go libc). Green on rerun.

PR #1017, opened 2026-07-25
fix(parquet/variant): reject truncated scalar values

fallintoplace:
### Rationale for this change

NewWithMetadata only checks that a variant value is non-empty. Truncated or unknown scalar encodings can therefore be accepted and later panic or read invalid data when their type or value is accessed.

* Validate the encoded width of fixed-size primitive values and short strings.
* Validate binary and string length prefixes and their payload sizes.
* Reject unknown primitive type codes while keeping compound values lazily decoded.

### Are these changes tested?

Yes. The tests cover unknown primitive types, truncated fixed-width values, truncated binary prefixes and payloads, and truncated short strings. The parquet/variant and parquet/pqarrow packages pass, including the assertion build for parquet/variant.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -501,8 +501,65 @@ func NewWithMetadata(meta Metadata, value []byte) (Value, error) {
 	if len(value) == 0 {
 		return Value{}, errors.New("invalid variant value: empty")
 	}
+	if err := validateScalarValue(value); err != nil {
+		return Value{}, err
+	}
 
 	return Value{value: value, meta: meta}, nil
+}
+
+func validateScalarValue(value []byte) error {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`NewWithMetadata\` only checked that the value was non-empty, so truncated scalars were accepted and then failed at access time — the numeric primitives silently returned 0, and short strings performed an out-of-bounds \`unsafe.String\`, which is a memory-safety issue rather than just a panic. I cross-checked every required length in \`validateScalarValue\` against the corresponding \`Value()\` reader (Decimal4=6, Decimal8=10, Decimal16=18, UUID=17, etc.) and the comparison uses \`<\`, so valid or longer buffers still pass. Objects and arrays are intentionally left lazily decoded. LGTM.

PR #1018, opened 2026-07-25
fix(arrow/array): validate list-view builder dimensions

fallintoplace:
### Rationale for this change

ListViewBuilder.AppendValuesWithSizes can accept too few dimensions or ranges that do not fit within the child values. NewArray can then return an invalid ListView array that fails when read, sliced, or validated.

* Require enough offsets and sizes for every logical list.
* Reject negative sizes, negative offsets for non-empty ranges, and ranges beyond the child array.
* Apply the checks to both ListView and LargeListView builders while preserving valid extra offset capacity.

### Are these changes tested?

Yes. The tests cover missing offsets, missing sizes, negative sizes, negative offsets, and ranges beyond the child values for both offset widths. The full arrow/array package, assertion build, compute packages, IPC package, and CSV package also pass.

### DIFF
--- a/arrow/array/list.go
+++ b/arrow/array/list.go
@@ -1332,7 +1332,42 @@ func (b *LargeListViewBuilder) NewLargeListViewArray() (a *LargeListView) {
 	return
 }
 
+func (b *baseListViewBuilder) validateDimensions() {
+	offsetCount, sizeCount := b.offsets.Len(), b.sizes.Len()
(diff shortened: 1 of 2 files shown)


zeroshade:
For all the other types, we make validation an opt-in thing, but this PR pushes a potential panic validation into `newData` rather than allowing the opt-in. Is there a reason why we shouldn't make this validation opt-in like in the other cases for other types?

PR #1020, opened 2026-07-25
fix(arrow/array): validate list builder offsets

fallintoplace:
### Rationale for this change

ListBuilder.AppendValues can accept malformed offset counts and ranges. NewArray then appends a final offset for any count mismatch and can return arrays with undersized, non-monotonic, or out-of-bounds offset buffers. Reading or validating those arrays can fail later.

* Require supplied offsets to contain either one entry per list or one entry per list plus the final offset.
* Reject negative, out-of-bounds, and non-monotonic offsets for List and LargeList builders.
* Continue allowing valid trailing child values that are not referenced by the list offsets.

### Are these changes tested?

Yes. The tests cover too few and too many offsets, negative offsets, non-monotonic offsets, and final offsets beyond the child array for both offset widths. The full arrow/array package, assertion build, compute packages, IPC package, and CSV package also pass.

### DIFF
--- a/arrow/array/list.go
+++ b/arrow/array/list.go
@@ -557,7 +557,36 @@ func (b *LargeListBuilder) NewLargeListArray() (a *LargeList) {
 	return
 }
 
+func (b *baseListBuilder) validateOffsets() {
+	count := b.offsets.Len()
+	if count != b.length && count != b.length+1 {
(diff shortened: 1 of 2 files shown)


zeroshade:
For all the other types, we make validation an opt-in thing, but this PR pushes a potential panic validation into newData rather than allowing the opt-in. Is there a reason why we shouldn't make this validation opt-in like in the other cases for other types?

PR #1021, opened 2026-07-25
fix(arrow/array): validate struct child lengths

fallintoplace:
### Rationale for this change

A struct child shorter than the parent extent produces an array that can panic when missing values are accessed. Structural validation in `arrow/array` is opt-in, so this should be reported by `Validate` and `ValidateFull` rather than by changing the established `StructBuilder` construction contract.

* Add `Validate` and `ValidateFull` support to `Struct`.
* Validate raw child data against the full `offset + length` extent, including sliced structs.
* Clamp constructed child views to available data so malformed arrays remain inspectable without substituting values from the wrong offset.
* Report short child fields with `arrow.ErrInvalid`.
* Preserve valid child-only builder usage and normal struct slicing behavior.

### Are these changes tested?

Yes. The tests cover direct and sliced short-child validation, update the missing-required-field case to use opt-in validation, and preserve the established `MapScalar` child-only builder path.

Validated with:

* `go test ./arrow/array ./arrow/scalar`
* `go test -tags assert ./arrow/array ./arrow/scalar`

### DIFF
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -143,17 +143,18 @@ func NewStructData(data arrow.ArrayData) *Struct {
 func (a *Struct) NumField() int           { return len(a.fields) }
 func (a *Struct) Field(i int) arrow.Array { return a.fields[i] }
 
-func (a *Struct) Validate() error { return nil }
-
-func (a *Struct) ValidateFull() error {
-	for i, field := range a.fields {
-		if field.Len() < a.Len() {
-			return fmt.Errorf("%w: arrow/array: struct field %d length %d is smaller than struct length %d",
-				arrow.ErrInvalid, i, field.Len(), a.Len())
+func (a *Struct) Validate() error {
+	expectedLength := a.data.offset + a.data.length
+	for i, child := range a.data.childData {
+		if child.Len() < expectedLength {
+			return fmt.Errorf("%w: arrow/array: struct child array #%d has length smaller than expected for struct array (%d < %d)",
+				arrow.ErrInvalid, i, child.Len(), expectedLength)
 		}
 	}
 	return nil
 }
+
+func (a *Struct) ValidateFull() error { return a.Validate() }
 
 // ValueStr returns the string representation (as json) of the value at index i.
 func (a *Struct) ValueStr(i int) string {
@@ -234,14 +235,18 @@ func (a *Struct) setData(data *Data) {
 	a.fields = make([]arrow.Array, len(data.childData))
 	for i, child := range data.childData {
 		if data.offset != 0 || child.Len() != data.length {
-			end := int64(data.offset) + int64(data.length)
-			if data.offset >= 0 && end >= int64(data.offset) && end <= int64(child.Len()) {
-				sub := NewSliceData(child, int64(data.offset), end)
-				a.fields[i] = MakeFromData(sub)
-				sub.Release()
-			} else {
-				a.fields[i] = MakeFromData(child)
+			childLen := int64(child.Len())
+			start := max(int64(data.offset), int64(0))
+			if start > childLen {
+				start = childLen
 			}
+			end := max(int64(data.offset)+int64(data.length), start)
+			if end > childLen {
+				end = childLen
+			}
+			sub := NewSliceData(child, start, end)
+			a.fields[i] = MakeFromData(sub)
+			sub.Release()
 		} else {
 			a.fields[i] = MakeFromData(child)
 		}
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -144,6 +145,9 @@ func (a *Struct) NumField() int           { return len(a.fields) }
 func (a *Struct) Field(i int) arrow.Array { return a.fields[i] }
 
 func (a *Struct) Validate() error {
+	if a.data.offset < 0 || a.data.length < 0 || int64(a.data.offset) > math.MaxInt64-int64(a.data.length) {
+		return fmt.Errorf("%w: arrow/array: struct offset and length overflow", arrow.ErrInvalid)
+	}
 	expectedLength := a.data.offset + a.data.length
 	for i, child := range a.data.childData {
 		if child.Len() < expectedLength {
@@ -240,7 +244,17 @@ func (a *Struct) setData(data *Data) {
 			if start > childLen {
 				start = childLen
 			}
-			end := max(int64(data.offset)+int64(data.length), start)
+			var end int64
+			offset, length := int64(data.offset), int64(data.length)
+			switch {
+			case length > 0 && offset > math.MaxInt64-length:
+				end = math.MaxInt64
+			case length < 0 && offset < math.MinInt64-length:
+				end = math.MinInt64
+			default:
+				end = offset + length
+			}
+			end = max(end, start)
 			if end > childLen {
 				end = childLen
 			}
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -143,6 +143,18 @@ func NewStructData(data arrow.ArrayData) *Struct {
 func (a *Struct) NumField() int           { return len(a.fields) }
 func (a *Struct) Field(i int) arrow.Array { return a.fields[i] }
 
+func (a *Struct) Validate() error { return nil }
+
+func (a *Struct) ValidateFull() error {
+	for i, field := range a.fields {
+		if field.Len() < a.Len() {
+			return fmt.Errorf("%w: arrow/array: struct field %d length %d is smaller than struct length %d",
+				arrow.ErrInvalid, i, field.Len(), a.Len())
+		}
+	}
+	return nil
+}
+
 // ValueStr returns the string representation (as json) of the value at index i.
 func (a *Struct) ValueStr(i int) string {
 	if a.IsNull(i) {
@@ -222,9 +234,14 @@ func (a *Struct) setData(data *Data) {
 	a.fields = make([]arrow.Array, len(data.childData))
 	for i, child := range data.childData {
 		if data.offset != 0 || child.Len() != data.length {
-			sub := NewSliceData(child, int64(data.offset), int64(data.offset+data.length))
-			a.fields[i] = MakeFromData(sub)
-			sub.Release()
+			end := int64(data.offset) + int64(data.length)
+			if data.offset >= 0 && end >= int64(data.offset) && end <= int64(child.Len()) {
+				sub := NewSliceData(child, int64(data.offset), end)
+				a.fields[i] = MakeFromData(sub)
+				sub.Release()
+			} else {
+				a.fields[i] = MakeFromData(child)
+			}
 		} else {
 			a.fields[i] = MakeFromData(child)
 		}
--- a/arrow/array/struct_test.go
+++ b/arrow/array/struct_test.go
@@ -135,6 +135,26 @@ func TestStructArray(t *testing.T) {
 	}
 }
 
+func TestStructValidateFullRejectsShortField(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	dt := arrow.StructOf(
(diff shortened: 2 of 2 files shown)


arrow/array/struct.go
```diff
@@ -439,6 +439,12 @@ func (b *StructBuilder) NewStructArray() (a *Struct) {
 }
 
 func (b *StructBuilder) newData() (data *Data) {
+	for i, f := range b.fields {
+		if f.Len() != b.length {
+			panic(fmt.Errorf("%w: arrow/array: struct field builder length must equal struct length (field=%d, field_length=%d, struct_length=%d)",
+				arrow.ErrInvalid, i, f.Len(), b.length))
+		}
+	}
```
zeroshade: This is the breaking check. `b.length` only advances when `StructBuilder.Append()` / `AppendValues()` is called, but it is entirely legitimate to populate a struct purely through `FieldBuilder(i)` and leave `b.length` at 0 — `MapScalar` in `arrow/scalar` does precisely this, and it panics here.

Also note this panics rather than returning an error, so it converts a previously-working call into a crash for downstream users. Even if the invariant were correct, `newData()` is the wrong place given `arrow/array`'s opt-in validation design (`fullyValidateOffsetsAndSizes` and friends live behind `ValidateFull`).

arrow/array/struct_test.go
```diff
@@ -646,7 +666,7 @@ func TestStructArrayUnmarshalJSONMissingFields(t *testing.T) {
 						if e == nil {
 							t.Fatalf("this should have panicked, but did not; slice value %v", val)
 						}
-						if got, want := e.(string), "arrow/array: index out of range"; got != want {
+						if got, want := fmt.Sprint(e), "invalid: arrow/array: struct field builder length must equal struct length (field=2, field_length=0, struct_length=1)"; got != want {
```
zeroshade: Changing an existing test's expected failure is worth calling out explicitly. On `main` this case panics with the string `arrow/array: index out of range` at **read** time (from `arr.String()` touching a short field), not at build time. This PR makes it fail earlier, inside `NewStructArray()`, with a different type (an `error` rather than a `string`), which is why the assertion had to be rewritten to `fmt.Sprint(e)`.

Failing fast is defensible in isolation, but combined with the `arrow/scalar` regression it indicates the check is too broad rather than that the old expectation was wrong.

zeroshade (changes requested):
I'm going to pass on this one, because the invariant it enforces is not actually an invariant of `StructBuilder` — the new check fires on a valid, established usage pattern and breaks existing code in this repo.

Building a struct array by appending only to the child field builders, without calling `StructBuilder.Append()` for each row, is a supported and widely used pattern. Under this patch that now panics. I verified the concrete breakage: with this change applied, `arrow/scalar` `TestMapScalarBasics` panics with

```
invalid: arrow/array: struct field builder length must equal struct length (field=0, field_length=2, struct_length=0)
```

because `MapScalar` constructs its key/value entry struct exactly that way. That test passes on `main` and fails with this PR. I ran the full module to confirm the attribution: applying #1018, #1019, #1020, #1023 and #1026 together introduces **zero** new package-level failures versus baseline, while this PR uniquely adds the `arrow/scalar` failure.

There is a real underlying observation here — a field builder shorter than the struct does produce a bad array, and one longer silently drops values. But `arrow/array` deliberately keeps this kind of structural checking **opt-in** through `Validate()` / `ValidateFull()` rather than in `newData()`, for performance reasons. If you want to pursue this, the right venue is a `ValidateFull` check on `Struct`, or a discussion on the dev mailing list about changing the builder contract — not an unconditional panic at construction.

Closing thought: the diff also rewrites an existing test's expected panic (inline below), which is a signal that the new check is intercepting a path the package intentionally handled elsewhere.

zeroshade:
Re-reviewed — the head is still `b9ef731d` with no new commits, so my earlier assessment stands unchanged and I am going to keep this closed to merging for now.

To restate the blocker concisely: the check added in `newData()` enforces something that is not actually an invariant of `StructBuilder`. Populating a struct purely through `FieldBuilder(i)` without calling `StructBuilder.Append()` leaves `b.length` at 0, and that is a supported pattern — `MapScalar` in `arrow/scalar` does exactly this, so with this patch applied `arrow/scalar` `TestMapScalarBasics` panics with `field_length=2, struct_length=0`. I confirmed the attribution by running the whole module: #1018, #1019, #1020, #1023 and #1026 applied together introduce zero new package-level failures, while this PR uniquely breaks `arrow/scalar`.

If you would like to pursue the underlying observation — which is legitimate, a short field really does yield a bad array and a long one silently drops values — the two viable routes are:

1. Add the check to `Struct`'s `ValidateFull()` instead of `newData()`, matching how `arrow/array` already handles structural validation opt-in (see `fullyValidateOffsetsAndSizes` for list types). This is the path I would take and I would review it promptly.
2. If you believe the builder contract itself should change, that needs a discussion on dev@arrow.apache.org first, since it turns working user code into a panic.

Happy to look again if you take route 1.

arrow/array/struct.go
```diff
@@ -143,6 +143,18 @@ func NewStructData(data arrow.ArrayData) *Struct {
 func (a *Struct) NumField() int           { return len(a.fields) }
 func (a *Struct) Field(i int) arrow.Array { return a.fields[i] }
 
+func (a *Struct) Validate() error { return nil }
```
zeroshade: This no-op `Validate()` is the first half of the problem. The two comparable types in this package both do the cheap structural check here rather than deferring everything to `ValidateFull`:

- `union.go:155-162` (sparse union) checks `fieldData.Len() < a.data.length+a.data.offset` in `Validate()`
- `validate.go:404-412` (fixed-size list) runs its child-length check in **both** paths

Since the check is O(number of fields) — not O(length) — there's no performance reason to keep it out of `Validate()`.

arrow/array/struct.go
```diff
@@ -143,6 +143,18 @@ func NewStructData(data arrow.ArrayData) *Struct {
...
+
+func (a *Struct) ValidateFull() error {
+	for i, field := range a.fields {
+		if field.Len() < a.Len() {
+			return fmt.Errorf("%w: arrow/array: struct field %d length %d is smaller than struct length %d",
+				arrow.ErrInvalid, i, field.Len(), a.Len())
```
zeroshade: This compares the **post-`setData`** field length against `a.Len()`. Because the fallback below replaces an out-of-range slice with the whole child, a sliced corrupt struct ends up with `field.Len() == a.Len()` and passes validation.

Compare against the raw child data and the full extent instead, as `union.go:155` does:

```go
if a.data.childData[i].Len() < a.data.length+a.data.offset {
```

That is offset-aware and can't be defeated by slicing.

arrow/array/struct.go
```diff
@@ -222,9 +234,14 @@ func (a *Struct) setData(data *Data) {
...
-			a.fields[i] = MakeFromData(sub)
-			sub.Release()
+			end := int64(data.offset) + int64(data.length)
+			if data.offset >= 0 && end >= int64(data.offset) && end <= int64(child.Len()) {
+				sub := NewSliceData(child, int64(data.offset), end)
+				a.fields[i] = MakeFromData(sub)
+				sub.Release()
+			} else {
+				a.fields[i] = MakeFromData(child)
+			}
```
zeroshade: This is the second half. When the range is out of bounds, silently using the **unsliced** child means the offset is discarded — so `Field(0).Value(0)` returns `child[0]` where the caller asked for `child[offset]`. I measured exactly that: a sliced struct at `offset=1` returned the value at index 0 while reporting `field0.Len()==a.Len()`.

Avoiding the old `index out of range` panic is a reasonable goal, but substituting different data is worse than either panicking or clamping. Either drop this fallback (and let `Validate` reject the array), or clamp the bounds:

```go
sub := NewSliceData(child, min(int64(data.offset), int64(child.Len())), min(end, int64(child.Len())))
```

Note `union.setData:138-140` keeps its eager slice and relies on `Validate()` to reject bad geometry — that's the pattern I'd follow here.

zeroshade (changes requested):
Thank you for reworking this — moving the check to `ValidateFull` is exactly what I asked for, and my original objection is fully resolved. I verified `arrow/scalar` `TestMapScalarBasics` now **passes**, and a full-module `go test ./...` shows **zero new failures** versus baseline. The build-a-struct-via-child-builders-only pattern also still works, byte-identically to `main`.

Unfortunately the new version introduces a different problem that I think is worse than the panic it replaced, so I'd like one more small revision.

**`ValidateFull` returns nil on a corrupt struct as soon as it is sliced, and reads then silently return wrong values.**

Two things combine to cause this. The `setData` change discards the offset and substitutes the whole child when the slice range looks out of range, and `ValidateFull` compares the *post-substitution* `a.fields[i].Len()` against `a.Len()` rather than the raw child data against `offset+length`. So a malformed struct escapes validation entirely once sliced:

```
unsliced ValidateFull = invalid: arrow/array: struct field 0 length 1 is smaller than struct length 2
sliced   struct.Len=1 offset=1 field0.Len=1 field0.offset=0
sliced   Validate     = <nil>
sliced   ValidateFull = <nil>          <-- false clean bill of health
sliced   Field(0).Value(0) = 10       <-- should be child[1], which does not exist
```

That is silent wrong data on an array `ValidateFull` just declared valid. It's in-bounds so there's no panic and no memory-safety issue, but `validate.go:40-42` documents this API as existing to "detect corrupted data from untrusted sources such as Arrow Flight ... before accessing values" — for that purpose a false negative is worse than the crash it replaced. Slicing is routine (`array.NewSlice`, record-batch slicing), and this path was previously unreachable because construction panicked, so the patch introduces it.

**Two existing precedents in this repo do it the right way, and they disagree with this patch on both points:**

- `arrow/array/union.go:155-162` — sparse union, structurally the identical case: `fieldData.Len() < a.data.length+a.data.offset`, checked in **`Validate()`**, against `a.data.childData[i]` (raw child). Note `union.setData:138-140` still slices children eagerly *and* adds this cheap raw check — the opposite of what this PR does on both counts.
- `arrow/array/validate.go:404-412` — `validateFixedSizeListArray` computes `childLength := int64(a.data.offset) + int64(a.data.length)` against the raw child, and it runs in **both** `Validate` and `ValidateFull`.

The practical difference, measured on the same defect:
```
FixedSizeList cheap Validate = arrow/array: fixed-size list child length 2 is too small for offset 0 and length 3
Struct        cheap Validate = <nil>      (only ValidateFull catches it, and not when sliced)
```

**Suggested revision, which should be small:**
1. Implement the check in `Validate()` over `data.childData[i].Len() < data.length+data.offset`, mirroring `union.go:155`, and drop the no-op `Validate`.
2. Either remove the `setData` fallback, or make it clamp rather than substitute — e.g. `NewSliceData(child, min(offset, child.Len()), min(end, child.Len()))` — so it can't fabricate a plausible-looking length.
3. Consider aligning the message wording with the existing union / fixed-size-list strings.

On the test rewrite: I'm satisfied this time. Last round it rewrote an expected panic string to accommodate a new panic; here it removes a panic, and the PR body discloses the change. Worth the PMC noting it converts a fail-fast into a silent success on the `UnmarshalJSON` missing-required-field path (`arr.String()` now returns `{[] [3]}` instead of panicking), which is defensible under opt-in validation — but only if validation actually catches it, which brings us back to the point above.

Happy to re-review quickly.

zeroshade (approved):
Both rounds are fully resolved — thanks for sticking with this.

Round 1 (construction panic breaking the child-only builder pattern) and round 2 (`ValidateFull` returning nil on a corrupt struct once sliced, then reads silently returning wrong data) are both addressed:

- The check now lives in `Validate()` and compares the **raw** `data.childData[i].Len()` against `offset+length`, mirroring `union.go:155` — so it's offset-aware and can't be defeated by slicing.
- `setData` now **clamps** the child view instead of substituting the whole child, so a sliced corrupt struct yields an empty/short field at the correct offset rather than fabricating a plausible-looking length.
- `TestStructValidateFullRejectsShortField` nails the exact regression I flagged: sliced, `Field(0).Len()==0`, and both `Validate()`/`ValidateFull()` return `ErrInvalid`.
- Nice bonus hardening against `offset+length` overflow, with a test.

I verified locally: `go test ./arrow/array ./arrow/scalar` and `go test -tags assert ./arrow/array ./arrow/scalar` all pass; `TestMapScalarBasics` (the round-1 regression) passes; the opt-in design means construction still never panics.

LGTM. Approving.

PR #1022, opened 2026-07-25
fix(arrow/csv): preserve decimal precision

fallintoplace:
### Rationale for this change

The CSV writer converts decimal values through math.Pow10 and big.Float. That changes valid Decimal128 and Decimal256 values for many scales and drops significant trailing zeros from fixed-scale output.

* Format Decimal128 and Decimal256 values directly with their exact ToString implementations.
* Preserve the declared scale, including trailing zeros and negative scales.
* Update existing zero-value expectations to retain their two decimal places.

### Are these changes tested?

Yes. The regression test covers a high-precision Decimal128 with a positive scale and a Decimal256 with a negative scale. All CSV tests and the assertion build pass with the required Parquet test data.

### DIFF
--- a/arrow/csv/transformer.go
+++ b/arrow/csv/transformer.go
@@ -21,8 +21,6 @@ import (
 	"encoding/base64"
 	"encoding/csv"
 	"fmt"
-	"math"
-	"math/big"
 	"strconv"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: the previous \`math.Pow10\` + \`big.Float\` path injects binary floating-point error — a Decimal128 at scale 10 emitted \`1234567890.1234567889478057622909545898\` instead of \`1234567890.1234567890\`. \`ToString(scale)\` does exact integer scaling, so this is a genuine precision fix.

Heads-up on the user-facing side: fixed-scale output now retains trailing zeros, so a raw \`0\` at scale 2 renders as \`0.00\` rather than \`0\`. That is the correct fixed-scale rendering and is why the existing writer fixtures were updated, but it is a visible CSV output change worth noting in the release notes. LGTM.

PR #1023, opened 2026-07-25
fix(arrow/array): skip empty table chunks

fallintoplace:
### Rationale for this change

TableReader treats an empty column chunk as the next batch boundary. A valid table with an empty chunk therefore emits a zero-row record batch before returning its actual rows.

* Advance each column past empty chunks before calculating the next batch size.
* Keep chunk slots synchronized without exposing zero-row batches.

### Are these changes tested?

Yes. The regression test builds a valid chunked column with empty chunks and verifies that only the non-empty batch is returned. The full arrow/array package, assertion build, compute packages, and IPC package also pass.

### DIFF
--- a/arrow/array/table.go
+++ b/arrow/array/table.go
@@ -353,6 +353,11 @@ func (tr *TableReader) Next() bool {
 	for i := range chunks {
 		j := tr.slots[i]
 		chunk := tr.chunks[i].Chunk(j)
+		for chunk.Len() == 0 && j+1 < len(tr.chunks[i].Chunks()) {
+			j++
+			tr.slots[i] = j
+			chunk = tr.chunks[i].Chunk(j)
+		}
 		remain := int64(chunk.Len()) - tr.offsets[i]
 		if remain < chunksz {
 			chunksz = remain
--- a/arrow/array/table_test.go
+++ b/arrow/array/table_test.go
@@ -845,6 +845,39 @@ func TestTableReader(t *testing.T) {
 	}
 }
 
+func TestTableReaderSkipsEmptyChunks(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	builder := array.NewInt32Builder(mem)
+	empty := builder.NewInt32Array()
+	defer empty.Release()
+	builder.AppendValues([]int32{1, 2}, nil)
+	values := builder.NewInt32Array()
+	defer values.Release()
+	builder.Release()
+
+	chunked := arrow.NewChunked(arrow.PrimitiveTypes.Int32, []arrow.Array{empty, values, empty})
+	defer chunked.Release()
+	field := arrow.Field{Name: "values", Type: arrow.PrimitiveTypes.Int32}
+	column := arrow.NewColumn(field, chunked)
+	defer column.Release()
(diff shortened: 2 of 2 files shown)


arrow/array/table.go
```diff
@@ -353,6 +353,11 @@ func (tr *TableReader) Next() bool {
 	for i := range chunks {
 		j := tr.slots[i]
 		chunk := tr.chunks[i].Chunk(j)
+		for chunk.Len() == 0 {
+			j++
+			tr.slots[i] = j
+			chunk = tr.chunks[i].Chunk(j)
+		}
```
zeroshade: This loop is unbounded, and `Chunked.Chunk(i)` does a raw slice index with no bounds check — so if it ever ran past the last chunk it would panic with index-out-of-range rather than failing gracefully.

As written it is safe for well-formed tables, since `Next()` has already established `tr.cur < tr.max` and `NewTable` enforces equal column lengths, so there must be a non-empty chunk ahead. But that safety depends on an invariant several layers away, and a malformed or hand-constructed `Chunked` (or a future change to that invariant) turns this into a panic.

Could you bound it explicitly? Something like:

```go
for chunk.Len() == 0 && j+1 < tr.chunks[i].NumChunks() {
    j++
    tr.slots[i] = j
    chunk = tr.chunks[i].Chunk(j)
}
```

That keeps the fix and makes the termination condition local and obvious.

zeroshade (changes requested):
Good catch, and the fix is correct in substance — I'd just like the loop bounded before this goes in.

I confirmed the bug on `main`: a chunked column whose first chunk is empty causes `TableReader` to emit a spurious zero-row `RecordBatch` before the real data. My probe produced batch 1 with 0 rows followed by batch 2 with the actual 2 rows, so skipping empty chunks is the right behaviour.

I also verified the safety of the trailing case: `Next()` returns early once `tr.cur >= tr.max`, and `NewTable` validates that all columns have equal length, so a column can't run out of chunks while rows remain. That means this is correct for any well-formed table. My concern is purely defensive — see inline.

One request beyond that: please rebase, since `main` has moved quite a bit today.

zeroshade (approved):
Re-reviewed at `bc29c91f`. The loop is bounded now and CI is green (22/22) — thanks.

The guard reads `for chunk.Len() == 0 && j+1 < len(tr.chunks[i].Chunks())`, which is equivalent to the `NumChunks()` bound I suggested and removes the index-out-of-range risk from `Chunked.Chunk`'s unchecked slice index. The termination condition is now local and obvious rather than depending on the `cur < max` / equal-column-length invariant several layers away.

The new test is better than what I asked for: `[]arrow.Array{empty, values, empty}` exercises a leading *and* a trailing empty chunk, so it covers both the original spurious-zero-row-batch bug and the new bound. Traced it: first `Next()` skips slot 0, lands on the 2-row chunk, emits `chunksz=2`; `cur` then reaches `max` so the trailing empty is never visited and the second `Next()` correctly returns false. Exactly one batch of 2 rows, which is what the test asserts.

The `Verify (ubuntu-latest)` failure was an unrelated flake — an abort *inside* AddressSanitizer itself (`sanitizer_allocator_secondary.h:297 IsAligned` CHECK during quarantine recycle), the same signature that hit several other PRs today. Green on rerun. LGTM.

PR #1024, opened 2026-07-25
fix(arrow/scalar): reject negative array lengths

fallintoplace:
### Rationale for this change

`MakeArrayFromScalar` accepts a signed length and returns an error, but negative lengths are not rejected consistently. For valid scalars, a negative length reaches buffer allocation and panics with a runtime slice-bounds error. For null scalars, it silently returns a corrupt array carrying a negative length.

* Reject negative lengths before either the null-scalar or valid-scalar path.
* Return an error wrapping `arrow.ErrInvalid` before allocating any buffers.

### Are these changes tested?

Yes. The regression test covers both valid and null scalars with a negative length and verifies that both return `arrow.ErrInvalid` without allocating memory.

### DIFF
--- a/arrow/scalar/scalar.go
+++ b/arrow/scalar/scalar.go
@@ -794,6 +794,10 @@ func MakeArrayOfNull(dt arrow.DataType, length int, mem memory.Allocator) arrow.
 // MakeArrayFromScalar returns an array filled with the scalar value repeated length times.
 // Not yet implemented for nested types such as Struct, List, extension and so on.
 func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Array, error) {
+	if length < 0 {
+		return nil, fmt.Errorf("%w: array length must be non-negative, got %d", arrow.ErrInvalid, length)
+	}
+
 	if !sc.IsValid() {
 		return MakeArrayOfNull(sc.DataType(), length, mem), nil
 	}
--- a/arrow/scalar/scalar_test.go
+++ b/arrow/scalar/scalar_test.go
@@ -1134,6 +1134,17 @@ func TestMakeArrayFromScalar(t *testing.T) {
 	}
 }
 
+func TestMakeArrayFromScalarRejectsNegativeLength(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	for _, sc := range []scalar.Scalar{scalar.NewInt32Scalar(1), scalar.ScalarNull} {
+		arr, err := scalar.MakeArrayFromScalar(sc, -1, mem)
+		require.ErrorIs(t, err, arrow.ErrInvalid)
+		assert.Nil(t, arr)
+	}
+}
+
 type OptionListTest struct {
 	FieldNames []string          `compute:"field_names"`
 	FieldNulls []bool            `compute:"field_null"`


arrow/scalar/scalar.go
```diff
@@ -788,6 +788,10 @@ func MakeArrayOfNull(dt arrow.DataType, length int, mem memory.Allocator) arrow.
 // MakeArrayFromScalar returns an array filled with the scalar value repeated length times.
 // Not yet implemented for nested types such as Struct, List, extension and so on.
 func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Array, error) {
+	if length < 0 {
+		return nil, fmt.Errorf("%w: array length must be non-negative, got %d", arrow.ErrInvalid, length)
+	}
```
zeroshade, line 799: Guard placement is right — it precedes both the `MakeArrayOfNull` branch below and the valid-scalar buffer allocation, so it covers both paths with one check, and it returns before any allocation occurs (which is why the `mem.AssertSize(t, 0)` in your test holds).

Wrapping `arrow.ErrInvalid` is also the correct convention here, since `MakeArrayFromScalar` already returns an error rather than panicking.

My only note is on the PR description: the branch immediately below this (`if !sc.IsValid()`) is the null path, and on `main` that path does **not** panic — it silently returns an array with a negative length. Worth stating accurately in the description since it becomes the commit message.

zeroshade (changes requested):
The fix itself is correct and I'm happy to merge it — the guard is well placed and covers both code paths. I only want the PR description corrected first, because it mis-describes the pre-fix behaviour in a way that would end up in the squashed commit message (this repo squashes with the PR body as the commit body).

What I verified on `main`:

- For a **valid** scalar, `MakeArrayFromScalar(sc, -1, mem)` does panic, as you describe — I reproduced `slice bounds out of range [:-4]` via the buffer allocation path.
- For a **null** scalar (`ScalarNull`), it does **not** panic. `bitutil.BytesForBits(-1)` computes `(-1+7)/8 == 0`, so `Resize(0)` succeeds and the function silently returns a corrupt array carrying a negative length, with a nil error.

So the null path is arguably the worse of the two — silent corruption rather than a loud crash — but it is not a panic. The description currently says negative lengths panic for both valid and null scalars.

Could you reword it to something like: negative lengths panic for valid scalars and silently produce a corrupt negative-length array for null scalars? Your single early-return guard correctly fixes both, and the test asserting `ErrInvalid` for both cases is meaningful — it genuinely fails on `main` for the null case, which previously returned a non-nil array and a nil error.

Also please rebase, as `main` has moved substantially today.

zeroshade (approved):
Re-reviewed at `8d4198a2`. The description is accurate now and CI is green (22/22).

It correctly distinguishes the two paths: *"For valid scalars, a negative length reaches buffer allocation and panics with a runtime slice-bounds error. For null scalars, it silently returns a corrupt array carrying a negative length."* That matches what I reproduced on `main` — the null path does not panic, because `bitutil.BytesForBits(-1)` computes `(-1+7)/8 == 0` so `Resize(0)` succeeds. Since this repo squashes with the PR body as the commit body, that distinction now lands correctly in the history.

The single early-return guard is well placed: it precedes both the `!sc.IsValid()` branch and the valid-scalar allocation, covers both paths with one check, and returns before any allocation (hence `mem.AssertSize(t, 0)` holds). Wrapping `arrow.ErrInvalid` is right for a function that already returns an error. The test asserting `ErrInvalid` for both `NewInt32Scalar(1)` and `ScalarNull` is meaningful — it fails on `main` for the null case, which previously returned a non-nil array with a nil error. LGTM.

PR #1026, opened 2026-07-25
fix(arrow/array): ignore zero-count union appends

fallintoplace:
### Rationale for this change

DenseUnionBuilder.AppendNulls(0) and AppendEmptyValues(0) still append a child value even though they add no union entries. That hidden mutation shifts the offset of the next real value and leaves inaccessible child data behind.

* Make non-positive bulk null and empty appends no-ops.
* Preserve the existing shared-child optimization for positive counts.

### Are these changes tested?

Yes. The regression test verifies that both zero-count calls leave the union and child builders empty, then confirms the next real value uses offset zero. The full arrow/array package, assertion build, compute packages, and IPC package also pass.

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -1201,6 +1201,10 @@ func (b *DenseUnionBuilder) AppendNull() {
 // for a DenseUnion this is more efficient than calling AppendNull multiple
 // times in a loop
 func (b *DenseUnionBuilder) AppendNulls(n int) {
+	if n <= 0 {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified: \`DenseUnionBuilder.AppendNulls\`/\`AppendEmptyValues\` called the child builder once *outside* the count loop, so a zero count appended no union entry but still advanced the child — I reproduced parent length 0 with child length 1. The \`n <= 0\` guard is the right fix and brings Dense in line with \`SparseUnionBuilder\`, which already no-ops on a zero count. LGTM.

PR #1028, opened 2026-07-27
fix(parquet/pqarrow): Change interface name for arrow type on read for extension types

C-Loftus:
### Rationale for this change

#969 added an interface for a custom arrow type on read for arrow extension types. This allows a type like geometry to be resolved to the underlying arrow physical type like byte array.

However, I went to implement this in geoarrow-go today and upon further investigation I think the interface name is slightly confusing. The interface `ExtensionParquetLogicalType` doesn't actually include the `ParquetLogicalType` method but rather it intended to be used to specify the arrow type from parquet on read. As such I have renamed `ExtensionParquetLogicalType` to something more distinct.

I figure it makes sense to solidify this before creating a new arrow-go release and adding it to iceberg-go whereupon it will be much harder to change.

I changed the interface name to indicate that it is one side of the read/write path for extension types. I think this is a bit more clear but if I am misunderstanding and it might be used in both read and write, then maybe it doesn't make sense.

Comments and a change in the name of an interface. This interface is not yet in a published release but is on main. Since its only been a few days, I think it is reasonable to change this. 

### Are these changes tested?

Yes

Yes. Change in interface name. But no changes in behavior.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -135,9 +135,7 @@ type ExtensionCustomParquetType interface {
 //
 // ArrowTypeFromParquet should return (nil, nil) if the logical type does not
 // map to the extension type. It should return (nil, err) if the logical type is
-// recognized but cannot be converted into a valid extension type. If a
-// non-nil extension type is returned, that type is used and any previous
-// conversion errors from other extension types are ignored.
+// recognized but there was an issue converting into a valid extension type.
 type ExtensionCustomArrowReadType interface {
 	ArrowTypeFromParquet(logical schema.LogicalType, storageType arrow.DataType) (arrow.ExtensionType, error)
 }
@@ -560,6 +558,8 @@ func arrowFromByteArray(logical schema.LogicalType) (arrow.DataType, error) {
 // arrowExtensionFromParquetLogicalType asks registered extension types whether
 // they can represent the provided Parquet logical type with the given storage
 // type, falling back to the storage type when none opt in.
+// As long as one non-nil extension type is returned and maps without an error, that
+// type is used and any previous conversion errors from other extension types are ignored.
 func arrowExtensionFromParquetLogicalType(logical schema.LogicalType, storageType arrow.DataType) (arrow.DataType, error) {
 	var (
 		typ           arrow.ExtensionType
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -132,6 +132,10 @@ type ExtensionCustomParquetType interface {
 // ExtensionCustomArrowReadType is an interface that Arrow ExtensionTypes may
 // implement to specify how a Parquet LogicalType maps back to an Arrow
 // ExtensionType when converting a Parquet schema to an Arrow schema on read.
+// The receiver is the registered extension type being asked whether it can
+// represent the Parquet logical type with the given storage type; callers must
+// use the returned extension type because it may be a distinct instance carrying
+// per-column parameters derived from the Parquet logical type.
 //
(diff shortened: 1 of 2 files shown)


parquet/pqarrow/schema.go
```diff
@@ -121,24 +121,24 @@ func (sm *SchemaManifest) GetFieldIndices(indices []int) ([]int, error) {
 }
 
 // ExtensionCustomParquetType is an interface that Arrow ExtensionTypes may implement
-// to specify the target LogicalType to use when converting to Parquet.
+// to specify the target LogicalType to use when converting to Parquet on write.
```
C-Loftus, line 124: Changed to `on write` since it makes it clearer that it is one side of the read/write path for arrow <-> parquet but worth aligning if there are any cases in which this would be used beyond writing

zeroshade (approved):
It's safe to change since we haven't released this, so it's fine. I think this is a good idea, thanks! I was actually considering this but wasn't strong enough to warrant blocking the original PR.

zeroshade:
@C-Loftus why the change to draft?

parquet/pqarrow/schema.go
```diff
@@ -121,24 +121,26 @@ func (sm *SchemaManifest) GetFieldIndices(indices []int) ([]int, error) {
...
+// The receiver is the registered extension type being asked whether it can
+// represent the Parquet logical type with the given storage type; callers must
+// use the returned extension type because it may be a distinct instance carrying
+// per-column parameters derived from the Parquet logical type.
 //
 // ArrowTypeFromParquet should return (nil, nil) if the logical type does not
 // map to the extension type. It should return (nil, err) if the logical type is
-// recognized but cannot be converted into a valid extension type. If a
-// non-nil extension type is returned, that type is used and any previous
-// conversion errors from other extension types are ignored.
```
C-Loftus, line 140: While I am making edits on comments, I removed this here and moved it to `arrowExtensionFromParquetLogicalType` since the comment on ignoring errors is moreso on control flow internal to iceberg-go's extension type resolution process and not the specifically this method.

parquet/pqarrow/schema.go
```diff
@@ -121,24 +121,26 @@ func (sm *SchemaManifest) GetFieldIndices(indices []int) ([]int, error) {
...
 
-// ExtensionParquetLogicalType is an interface that Arrow ExtensionTypes may
+// ExtensionCustomArrowReadType is an interface that Arrow ExtensionTypes may
 // implement to specify how a Parquet LogicalType maps back to an Arrow
-// ExtensionType when converting a Parquet schema to an Arrow schema.
+// ExtensionType when converting a Parquet schema to an Arrow schema on read.
+// The receiver is the registered extension type being asked whether it can
+// represent the Parquet logical type with the given storage type; callers must
+// use the returned extension type because it may be a distinct instance carrying
+// per-column parameters derived from the Parquet logical type.
```
C-Loftus, line 138: I added this comment here since I realized that it might be confusing to a user adding read support for an extension type why an extension type needs itself to return a new extension type struct and it can't just use the existing one after passing it in and seeing no error.

I used the phrase `per-column parameters derived from the Parquet logical type.` rather than “metadata” because the values are part of the Parquet LogicalType object passed to ArrowTypeFromParquet, such as CRS or geography algorithm fields. “Metadata” could be confused with Arrow field metadata or ARROW:extension:metadata, which this hook does not receive directly.

C-Loftus (commented):
Made a few extra comment changes. I think these help for clarity. Sorry for switching back to draft after review, figured it would be good to finalize all related comments rather than go back and forth multiple times in separate PRs.

zeroshade (approved):

PR #1031, opened 2026-07-27
fix(parquet/encoding): reject truncated plain boolean data

fallintoplace:
### Rationale for this change

The plain Boolean decoder trusted that each requested physical value had a corresponding payload bit. A truncated page could therefore panic during `Decode`, `Discard`, or `DecodeToBitmap` when the decoder indexed beyond the input buffer.

The value count passed to `SetData` is only an upper bound because it includes null rows, so payload bounds must be checked when physical values are consumed.

* Reject negative Boolean value counts in `SetData`.
* Bound `Decode`, `Discard`, and `DecodeToBitmap` against the payload bits currently available.
* Return an error wrapping `io.ErrUnexpectedEOF` without consuming decoder state.
* Continue accepting nullable pages whose level count exceeds their physical Boolean count.

### Are these changes tested?

Yes. The regression test covers all three truncated consumption paths and a nullable-style page with 100 levels and four physical values. The full `parquet/...` test tree passes, including the nullable V1/V2, spaced-bitmap, and Parquet-to-Arrow paths.

### DIFF
--- a/parquet/internal/encoding/boolean_decoder.go
+++ b/parquet/internal/encoding/boolean_decoder.go
@@ -43,6 +43,9 @@ func (PlainBooleanDecoder) Type() parquet.Type {
 }
 
 func (dec *PlainBooleanDecoder) SetData(nvals int, data []byte) error {
+	if nvals < 0 {
+		return fmt.Errorf("parquet: invalid number of boolean values: %d", nvals)
+	}
 	if err := dec.decoder.SetData(nvals, data); err != nil {
 		return err
 	}
@@ -50,8 +53,19 @@ func (dec *PlainBooleanDecoder) SetData(nvals int, data []byte) error {
 	return nil
 }
 
+func (dec *PlainBooleanDecoder) ensureBitsAvailable(n int) error {
+	available := int64(len(dec.data))*8 - int64(dec.bitOffset)
+	if int64(n) > available {
+		return fmt.Errorf("parquet: boolean data has %d bits available, need %d: %w", available, n, io.ErrUnexpectedEOF)
+	}
+	return nil
+}
+
 func (dec *PlainBooleanDecoder) Discard(n int) (int, error) {
 	n = min(n, dec.nvals)
+	if err := dec.ensureBitsAvailable(n); err != nil {
+		return 0, err
+	}
 	dec.nvals -= n
 
 	if dec.bitOffset+n < 8 {
@@ -77,6 +91,9 @@ func (dec *PlainBooleanDecoder) Discard(n int) (int, error) {
 // Returns the number of values decoded
 func (dec *PlainBooleanDecoder) Decode(out []bool) (int, error) {
 	max := shared_utils.Min(len(out), dec.nvals)
+	if err := dec.ensureBitsAvailable(max); err != nil {
+		return 0, err
+	}
 
 	// attempts to read all remaining bool values from the current data byte
 	unalignedExtract := func(i int) int {
@@ -126,6 +143,9 @@ func (dec *PlainBooleanDecoder) DecodeToBitmap(out []byte, outOffset int64, leng
 	max := shared_utils.Min(length, dec.nvals)
 	if max == 0 {
 		return 0, nil
+	}
+	if err := dec.ensureBitsAvailable(max); err != nil {
+		return 0, err
 	}
 
 	// Check if we're aligned and can do a fast copy
--- a/parquet/internal/encoding/encoding_test.go
+++ b/parquet/internal/encoding/encoding_test.go
@@ -1089,6 +1089,42 @@ func TestBooleanPlainDecoderAfterFlushing(t *testing.T) {
 	assert.Equal(t, decSlice[0], false)
 }
 
+func TestBooleanPlainDecoderRejectsTruncatedData(t *testing.T) {
+	descr := schema.NewColumn(schema.NewBooleanNode("bool", parquet.Repetitions.Required, -1), 0, 0)
+	type booleanBitmapDecoder interface {
+		encoding.BooleanDecoder
+		DecodeToBitmap([]byte, int64, int) (int, error)
+	}
+	newDecoder := func() booleanBitmapDecoder {
+		dec := encoding.NewDecoder(parquet.Types.Boolean, parquet.Encodings.Plain, descr, memory.DefaultAllocator).(booleanBitmapDecoder)
+		require.NoError(t, dec.SetData(9, []byte{0xff}))
+		return dec
+	}
+
+	dec := newDecoder()
+	n, err := dec.Decode(make([]bool, 9))
+	assert.Zero(t, n)
+	assert.ErrorIs(t, err, io.ErrUnexpectedEOF)
+
+	dec = newDecoder()
+	n, err = dec.Discard(9)
+	assert.Zero(t, n)
+	assert.ErrorIs(t, err, io.ErrUnexpectedEOF)
+
+	dec = newDecoder()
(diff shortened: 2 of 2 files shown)


parquet/internal/encoding/boolean_decoder.go
```diff
@@ -43,6 +43,13 @@ func (PlainBooleanDecoder) Type() parquet.Type {
 }
 
 func (dec *PlainBooleanDecoder) SetData(nvals int, data []byte) error {
+	if nvals < 0 {
+		return fmt.Errorf("parquet: invalid number of boolean values: %d", nvals)
+	}
+	required := bitutil.BytesForBits(int64(nvals))
+	if int64(len(data)) < required {
+		return fmt.Errorf("parquet: boolean data is too short: got %d bytes, need %d for %d values", len(data), required, nvals)
+	}
```
zeroshade: This is the check that breaks valid files. `nvals` here is the page's level count (`c.numBuffered`, from the header's `num_values`, nulls included) — not the number of physical booleans in the payload. For a nullable column the PLAIN payload holds only non-null bits, so `BytesForBits(nvals)` legitimately exceeds `len(data)`.

Concretely, 100 rows with 4 non-null gives a 1-byte payload while this demands 13 bytes.

The `nvals < 0` guard just above is fine — it's this length comparison that needs to move into `Decode`/`Discard`/`DecodeToBitmap`, where the real count of values being read is known and can be bounded against `len(dec.data)`.

parquet/internal/encoding/encoding_test.go
```diff
@@ -1089,6 +1089,14 @@ func TestBooleanPlainDecoderAfterFlushing(t *testing.T) {
 	assert.Equal(t, decSlice[0], false)
 }
 
+func TestBooleanPlainDecoderRejectsTruncatedData(t *testing.T) {
+	descr := schema.NewColumn(schema.NewBooleanNode("bool", parquet.Repetitions.Required, -1), 0, 0)
+	dec := encoding.NewDecoder(parquet.Types.Boolean, parquet.Encodings.Plain, descr, memory.DefaultAllocator)
+
+	assert.ErrorContains(t, dec.SetData(9, []byte{0xff}), "need 2 for 9 values")
```
zeroshade: This test encodes the same incorrect invariant, which is why it passes in isolation — it constructs the decoder directly, so `nvals` really does equal the physical count here. The nullable path through `columnChunkReader` is where the assumption breaks, and no test in this package covers it.

Separately, asserting on the substring `"need 2 for 9 values"` couples the test to the exact wording of the error message. `require.ErrorIs` against a sentinel, or a looser match, would be less brittle.

zeroshade (changes requested):
The bug you found here is real, but the fix validates against the wrong invariant and breaks reading valid files, so I can't take it as written.

**The bug is genuine.** I confirmed both panics on current `main`: `SetData(9, []byte{0xff})` then `Decode` panics with `index out of range [0] with length 0` at `boolean_decoder.go:84` (via `:116`), and `SetData(9, nil)` then `Discard` panics with `slice bounds out of range [1:0]` at `:64`. Something should definitely bound this.

**The rounding is also correct** — `bitutil.BytesForBits` is `(bits+7)>>3`, so 1→1, 8→1, 9→2, 16→2, and the check uses `<` rather than `!=` so pages with trailing bytes still pass. No off-by-one there.

**The problem is that `nvals` is the page's *level* count, not the physical value count.** `column_reader.go:414`/`:443` set `c.numBuffered = int64(page.nvals)` from the page header's `num_values`, which in Parquet **includes nulls**, and `column_reader.go:527` passes that straight through as `SetData(int(c.numBuffered), ...)`. For a nullable BOOLEAN column the PLAIN payload only contains the **non-null** bits, so `len(data)` is legitimately far smaller than `BytesForBits(nvals)`.

I reproduced this against a file arrow-go itself writes and reads fine today. A column with 100 rows where only 4 are non-null produces a 1-byte payload while the new check demands 13 bytes. On `main` that reads back correctly (`levels=100, values=4`); with this patch applied it returns **`levels=0, values=0`** — and note it does *not* surface an error to the caller, so this manifests as **silent data loss** rather than a clean failure.

With a 10-non-null variant the error is visible:
```
parquet: boolean data is too short: got 2 bytes, need 13 for 100 values
```
This reproduces on both DataPage V1 and V2.

**It breaks three tests that are green on `main`:**
- `parquet/file` — `TestPrimitiveReader/datapage_v1/TestBoolFlatOptional`
- `parquet/file` — `TestPrimitiveReader/datapage_v2/TestBoolFlatOptional`
- `parquet/file` — `TestColumnWriter/bool/TestWriteBitmapBatchSpaced`
- `parquet/pqarrow` — `TestParquetArrowIO`

I mention this because the PR says the encoding package passes — that is true, but `parquet/internal/encoding` is the one package that doesn't exercise the nullable read path. Worth running `go test ./parquet/...` for changes in this area.

**Suggested direction**: keep the `nvals < 0` guard (that part is good and unambiguous), but the payload-length check can't live in `SetData`, because at that point `nvals` is only an *upper bound* on how many values are physically present. Bound it where the actual read happens instead — in `Decode`, `Discard` and `DecodeToBitmap`, against `len(dec.data)` and the number of values actually being consumed. That closes the panics without assuming a relationship between levels and payload size.

Happy to re-review promptly if you rework it that way.

zeroshade (approved):
This addresses my review exactly, and I verified it empirically. Thanks for the rework.

You moved the bound out of `SetData` and into the three consumption points: `ensureBitsAvailable` (`boolean_decoder.go:56-62`) computes `len(dec.data)*8 - dec.bitOffset` and is called from `Discard` (`:66`), `Decode` (`:94`) and `DecodeToBitmap` (`:147`). `SetData` now carries only the `nvals < 0` guard. That is the right invariant — the bound is against the payload and the values actually being consumed, never against the level count.

**The regression I rejected is gone.** Sparse nullable BOOLEAN, 100 rows / 4 non-null, dictionary disabled so the encoding is genuinely PLAIN:
```
v1: levels=100 values=4  PASS
v2: levels=100 values=4  PASS
```
Fragmented reads (7/13/1/40/39) and the `Discard` path (`Skip(50)` -> levels=50 values=1) also correct. As a control I re-inserted the old `SetData` check and re-ran the same test — it reproduced the `levels=0 values=0` silent data loss, so the test is a genuine detector and this patch clears it.

The three tests my earlier review named all pass now: `TestPrimitiveReader/{datapage_v1,datapage_v2}/TestBoolFlatOptional`, `TestColumnWriter/bool/TestWriteBitmapBatchSpaced`, and `pqarrow TestParquetArrowIO`. Full `go test ./parquet/...` is green with test data present, clean under `-race`, and cross-builds for `GOARCH=386`.

**The original panics are still fixed**: on unpatched main the truncated cases panic (`index out of range`, `slice bounds out of range`); with the patch all six truncation shapes return `(0, io.ErrUnexpectedEOF)` from all four entry points with no panic, and state is mutated only after the check so the decoder stays consistent.

Boundary check clean — `n = 1, 8, 9, 16, 17, 63, 64, 65` against exact-size, oversized and one-byte-short payloads all behave correctly, and `bitOffset > 0` always implies at least one byte remains so `available` cannot go negative.

Using `errors.Is(io.ErrUnexpectedEOF)` instead of matching message text is a nice improvement over the previous revision. Two cosmetic notes, neither worth another round: `DecodeToBitmap` has no production callers today so that guard is future-proofing, and `Decode`/`Discard` return `(0, err)` while the sibling `RleBooleanDecoder.Decode` returns a partial count — a pre-existing asymmetry. LGTM.

PR #1032, opened 2026-07-27
fix(parquet/metadata): return page index decode errors

fallintoplace:
### Rationale for this change

`RowGroupPageIndexReader` exposes error-returning methods, but malformed serialized column or offset indexes could panic inside the lower-level constructors. A corrupt Parquet file could therefore terminate the caller instead of producing a read error.

* Convert page-index deserialization panics into errors at the row-group reader boundary.
* Wrap failures with `arrow.ErrInvalid` and identify the affected index kind.
* Preserve the existing lower-level constructor signatures for compatibility.

### Are these changes tested?

Yes. Regression coverage exercises malformed serialized column and offset indexes. The full `parquet/metadata` package passes.

### DIFF
--- a/parquet/metadata/page_index.go
+++ b/parquet/metadata/page_index.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"runtime"
 	"sync"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -244,6 +245,32 @@ func NewOffsetIndex(serializedIndex []byte, _ *parquet.ReaderProperties, decrypt
 	return &offsetIndex
 }
 
+func deserializeColumnIndex(descr *schema.Column, serializedIndex []byte, props *parquet.ReaderProperties) (idx ColumnIndex, err error) {
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			if _, ok := recovered.(runtime.Error); ok {
+				panic(recovered)
+			}
+			err = fmt.Errorf("%w: %w", arrow.ErrInvalid,
+				shared_utils.FormatRecoveredError("invalid column index", recovered))
+		}
+	}()
+	return NewColumnIndex(descr, serializedIndex, props, nil), nil
+}
+
+func deserializeOffsetIndex(serializedIndex []byte, props *parquet.ReaderProperties) (idx OffsetIndex, err error) {
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			if _, ok := recovered.(runtime.Error); ok {
+				panic(recovered)
+			}
+			err = fmt.Errorf("%w: %w", arrow.ErrInvalid,
+				shared_utils.FormatRecoveredError("invalid offset index", recovered))
+		}
+	}()
+	return NewOffsetIndex(serializedIndex, props, nil), nil
+}
+
 type readRange struct {
 	Offset, Length int64
 }
@@ -358,7 +385,10 @@ func (r *RowGroupPageIndexReader) GetColumnIndex(i int) (ColumnIndex, error) {
 		serializedIndex = decrypted
 	}
 
-	idx := NewColumnIndex(descr, serializedIndex, r.props, nil)
+	idx, err := deserializeColumnIndex(descr, serializedIndex, r.props)
+	if err != nil {
+		return nil, err
+	}
 	r.colIndexes[i] = idx
 	return idx, nil
 }
@@ -418,7 +448,10 @@ func (r *RowGroupPageIndexReader) GetOffsetIndex(i int) (OffsetIndex, error) {
 		serializedIndex = decrypted
 	}
 
-	oidx := NewOffsetIndex(serializedIndex, r.props, nil)
+	oidx, err := deserializeOffsetIndex(serializedIndex, r.props)
+	if err != nil {
+		return nil, err
+	}
 	r.offsetIndices[i] = oidx
 	return oidx, nil
 }
--- a/parquet/metadata/page_index_decode_test.go
+++ b/parquet/metadata/page_index_decode_test.go
@@ -1,0 +1,84 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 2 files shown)


parquet/metadata/page_index.go
```diff
@@ -244,6 +244,24 @@ func NewOffsetIndex(serializedIndex []byte, _ *parquet.ReaderProperties, decrypt
 	return &offsetIndex
 }
 
+func deserializeColumnIndex(descr *schema.Column, serializedIndex []byte, props *parquet.ReaderProperties) (idx ColumnIndex, err error) {
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			err = fmt.Errorf("%w: invalid column index: %v", arrow.ErrInvalid, recovered)
+		}
+	}()
+	return NewColumnIndex(descr, serializedIndex, props, nil), nil
```
zeroshade, line 258: Two asks on this recover block (same applies to `deserializeOffsetIndex` below).

**`%v` discards the cause.** Prefer the repo's own `utils.FormatRecoveredError` (`internal/utils/recovery.go:26-31`), which wraps with `%w`. Right now `errors.Is(err, arrow.ErrInvalid)` works but there's no way to reach the underlying `*thrift.tProtocolException`.

**The scope catches genuine bugs too.** This wraps the whole constructor — decrypt, thrift decode, type switch, `newTypedColumnIndex` validation, and per-page min/max decoding — and never re-panics. I verified that a nil `descr` surfaces as `invalid: invalid column index: runtime error: invalid memory address or nil pointer dereference`, i.e. an arrow-go nil-deref reported as a corrupt user file. Re-panicking on `runtime.Error`, or narrowing the recover to just the deserialization call, would keep programmer errors distinguishable from bad input.

zeroshade (changes requested):
The bug is real and the approach is sound — I'd just like two small changes before merging, both about not losing information.

**The panic is genuine and reachable from an error-returning API.** Thrift deserialization panics inside `NewColumnIndex` (`page_index.go:110`) and `NewOffsetIndex` (`:240`); there are also explicit `panic("invalid column index")` at `:164` and `must`/`mustArg` in `getDecoder` (`:172-173`). I confirmed `[]byte{0xff}` yields `*thrift.tProtocolException: field 15 read error: don't know what type: 15`, and that it propagates out of `GetColumnIndex` (`:361`) / `GetOffsetIndex` (`:421`) — both of which advertise an `error` return. So a corrupt footer takes down the caller instead of producing a read error. Worth fixing.

The fix is directionally right, signatures are preserved exactly as claimed (the only production callers are `:361`/`:421`, and the exported constructors are untouched), and `go test ./...` is byte-identical to baseline — zero regressions.

Requests:

**1. Use the existing recovery helper so the underlying error is preserved.** This repo already has `utils.FormatRecoveredError` (`internal/utils/recovery.go:26-31`), which wraps with `%w` rather than flattening via `%v`. As written, `fmt.Errorf("%w: invalid column index: %v", arrow.ErrInvalid, recovered)` means callers can match `arrow.ErrInvalid` but can never `errors.Is`/`errors.As` down to the thrift error — useful diagnostic detail is discarded at the boundary.

**2. Re-panic on runtime errors rather than laundering them into data-corruption errors.** The recover is scoped to the whole constructor — decrypt, thrift decode, the type switch, `newTypedColumnIndex` validation, and full min/max decoding for every page — and catches everything unconditionally. I demonstrated the failure mode: `deserializeColumnIndex(nil, <valid thrift>, nil)` returns

```
invalid: invalid column index: runtime error: invalid memory address or nil pointer dereference
```

That's a genuine nil-pointer bug in arrow-go being reported as though the user's file were malformed — exactly the kind of thing that makes real defects hard to find later. A `if re, ok := recovered.(runtime.Error); ok { panic(re) }` guard (or narrowing the recovered scope to just the thrift decode) would keep programmer errors loud.

For the record, I checked and the broad-recover style does match existing precedent here — `compress/compress.go:112-121`, `file/column_writer_types.gen.go:67-71`, `pqarrow/file_reader.go:369-374` — and none of those re-panic either. So this is consistent rather than novel, and I'm not asking you to fix the pattern repo-wide; just to not extend it in the one place where distinguishing "bad file" from "our bug" matters most.

**Minor**: the test only exercises the private helpers, not the public `GetColumnIndex`/`GetOffsetIndex` boundary that the PR describes as the fix. It's not a tautology — the thrift panic is real — but one case through the public reader would match the stated intent. Also the new file is `package metadata` while the existing `page_index_test.go` is `package metadata_test`. Using `errors.Is(err, arrow.ErrInvalid)` rather than string matching is good and not brittle.

zeroshade (approved):
All three points addressed and verified. LGTM.

**(a) `FormatRecoveredError`** — done (`page_index.go:248-259`, `:261-272`). The `%v` is gone and the wrapping is now `fmt.Errorf("%w: %w", arrow.ErrInvalid, shared_utils.FormatRecoveredError(...))`. Confirmed both properties hold:
```
err = invalid: invalid column index: *parquet.ColumnIndex field 15 read error: don`t know what type: 15
errors.Is(err, arrow.ErrInvalid) = true
  unwrap -> *fmt.wrapError -> *thrift.tProtocolException   <-- cause now reachable
```

**(b) `runtime.Error` re-panic** — done (`:251-253`, `:264-266`). My exact reproducer now behaves correctly:
```
deserializeColumnIndex(nil, <valid thrift>, nil) -> RE-PANICKED: runtime error: invalid memory address or nil pointer dereference
```
I specifically checked the risk that this would push genuine file corruption back out as a panic, and it does not — arrow-go`s own corruption paths use `panic("invalid column index")` (a string, `:165`) and `must`/`mustArg` (an `error`), neither of which is a `runtime.Error`. Verified four corrupt-but-thrift-valid inputs (mismatched NullPages/MinValues lengths, truncated and empty MinValues payloads, mismatched NullCounts) all still return errors wrapping `arrow.ErrInvalid`. The split lands on the right side of the line.

**(c) Public boundary** — done. `TestRowGroupPageIndexReaderReturnsDecodeErrors` drives `GetColumnIndex(0)`/`GetOffsetIndex(0)`, and I confirmed those are the only non-test callers of the panicking constructors, so the production read path is fully covered.

**I was wrong on one part of (c)** and want to correct it on the record: I suggested the file should be `package metadata_test`. It should not — it exercises private helpers and reuses `constructFakeMetadata` from `page_index_internal_test.go`, which is itself `package metadata`. Your choice was right; I withdraw that. The only residual nit is the filename, which by that file`s convention would be `page_index_decode_internal_test.go`.

Zero regressions: full module `go test ./...` diff versus baseline is empty, clean under `-race`, `go vet ./parquet/metadata/` clean.

One optional nit for a future pass — `page_index_decode_test.go:69-74` asserts `require.Len(unwrapper.Unwrap(), 2)`, which hard-codes the `%w: %w` arity and would break on any restructuring without a behavior change. An `errors.As` against the concrete thrift error would test the actual contract. Not worth another round.

PR #1033, opened 2026-07-27
fix(parquet/metadata): validate adaptive bloom options

fallintoplace:
### Rationale for this change

The adaptive bloom filter constructor accepted invalid configuration values. A false-positive probability of one could loop indefinitely in non-assertion builds, while a maximum size below the minimum could create a zero-byte filter that later panicked during insertion.

* Reject false-positive probabilities outside the open interval `(0, 1)`, including NaN.
* Clamp the maximum allocation to the supported bloom-filter size range.
* Keep the existing constructor signature and small-size fallback behavior.

### Are these changes tested?

Yes. The tests cover invalid probabilities and a zero-byte requested maximum, including insertion into the resulting minimum-sized filter. The full `parquet/metadata` package passes.

### DIFF
--- a/parquet/metadata/adaptive_bloom_filter.go
+++ b/parquet/metadata/adaptive_bloom_filter.go
@@ -18,6 +18,7 @@ package metadata
 
 import (
 	"io"
+	"math"
 	"slices"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -79,9 +80,13 @@ type adaptiveBlockSplitBloomFilter struct {
 }
 
 func NewAdaptiveBlockSplitBloomFilter(maxBytes uint32, numCandidates int, fpp float64, column *schema.Column, mem memory.Allocator) BloomFilterBuilder {
+	if fpp <= 0 || fpp >= 1 || math.IsNaN(fpp) {
+		panic("parquet: bloom filter false-positive probability must be in (0, 1)")
+	}
+	maxBytes = max(minimumBloomFilterBytes, min(maximumBloomFilterBytes, maxBytes))
 	ret := &adaptiveBlockSplitBloomFilter{
 		mem:             mem,
-		maxBytes:        min(maximumBloomFilterBytes, maxBytes),
+		maxBytes:        maxBytes,
 		minBytes:        minimumBloomFilterBytes,
 		minCandidateNDV: 16,
 		hasher:          xxhasher{},
--- a/parquet/metadata/bloom_filter_test.go
+++ b/parquet/metadata/bloom_filter_test.go
@@ -18,6 +18,7 @@ package metadata
 
 import (
 	"fmt"
+	"math"
 	"math/rand/v2"
 	"runtime"
 	"sync"
@@ -120,7 +121,7 @@ func TestNewBloomFilter(t *testing.T) {
 		maxBytes      int64
 		expectedBytes int64
 	}{
-		{1, 0.09, 0, 0},
+		{1, 0.09, 0, minimumBloomFilterBytes},
 		// cap at maximumBloomFilterBytes
 		{1 << 30, 0.9, maximumBloomFilterBytes + 1, maximumBloomFilterBytes},
 		// round to power of 2
@@ -142,6 +143,20 @@ func TestNewBloomFilter(t *testing.T) {
 			runtime.GC() // force GC to run and do the cleanup routines
 		})
 	}
+}
+
+func TestNewBloomFilterFromNDVAndFPPValidatesOptions(t *testing.T) {
+	for _, fpp := range []float64{0, 1, -0.1, 2, math.NaN()} {
+		t.Run(fmt.Sprintf("fpp=%v", fpp), func(t *testing.T) {
+			assert.Panics(t, func() {
+				NewBloomFilterFromNDVAndFPP(1, fpp, 1024, memory.DefaultAllocator)
+			})
+		})
+	}
+
+	bf := NewBloomFilterFromNDVAndFPP(1, 0.01, 0, memory.DefaultAllocator)
+	assert.EqualValues(t, minimumBloomFilterBytes, bf.Size())
+	assert.NotPanics(t, func() { bf.InsertHash(42) })
 }
 
 func BenchmarkFilterInsert(b *testing.B) {
@@ -270,6 +285,27 @@ func TestAdaptiveBloomFilterEdgeCases(t *testing.T) {
 		// The bloom filter should still work after GC
 		for _, h := range hashes {
 			assert.Truef(t, bf.CheckHash(h), "hash %d not found after GC - potential GC safety issue", h)
+		}
+	})
+
+	t.Run("clamps maximum size to the minimum allocation", func(t *testing.T) {
+		bf := NewAdaptiveBlockSplitBloomFilter(0, 1, 0.01, col, mem).(*adaptiveBlockSplitBloomFilter)
+		defer func() {
+			for _, candidate := range bf.candidates {
+				candidate.bloomFilter.cancelCleanup()
+				candidate.bloomFilter.data.Release()
+			}
+		}()
+
+		assert.EqualValues(t, minimumBloomFilterBytes, bf.maxBytes)
+		assert.NotPanics(t, func() { bf.InsertHash(1) })
+	})
+
+	t.Run("rejects invalid false-positive probabilities", func(t *testing.T) {
+		for _, fpp := range []float64{-0.1, 0, 1, math.NaN()} {
+			assert.PanicsWithValue(t,
+				"parquet: bloom filter false-positive probability must be in (0, 1)",
+				func() { NewAdaptiveBlockSplitBloomFilter(1024, 1, fpp, col, mem) })
 		}
 	})
 }
--- a/parquet/writer_properties.go
+++ b/parquet/writer_properties.go
@@ -17,6 +17,8 @@
 package parquet
 
 import (
+	"math"
+
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/parquet/compress"
@@ -58,6 +60,9 @@ const (
 	DefaultBloomFilterFPP             = 0.01
 	DefaultAdaptiveBloomFilterEnabled = false
 	DefaultBloomFilterCandidates      = 5
+
+	minimumBloomFilterBytes = 32
+	maximumBloomFilterBytes = 128 * 1024 * 1024
 )
 
 // ColumnProperties defines the encoding, codec, and so on for a given column.
@@ -366,6 +371,9 @@ func WithPageIndexEnabledPath(path ColumnPath, enabled bool) WriterProperty {
 // it is abandoned and not written to the file.
 func WithMaxBloomFilterBytes(nbytes int64) WriterProperty {
 	return func(cfg *writerPropConfig) {
+		if nbytes < minimumBloomFilterBytes || nbytes > maximumBloomFilterBytes {
+			panic("parquet: maximum bloom filter size must be between 32 bytes and 128 MiB")
+		}
 		cfg.wr.maxBloomFilterBytes = nbytes
 	}
 }
@@ -396,6 +404,7 @@ func WithBloomFilterEnabledPath(path ColumnPath, enabled bool) WriterProperty {
 // bloom filters.
 func WithBloomFilterFPP(fpp float64) WriterProperty {
 	return func(cfg *writerPropConfig) {
+		validateBloomFilterFPP(fpp)
 		cfg.wr.defColumnProps.BloomFilterFPP = fpp
 	}
 }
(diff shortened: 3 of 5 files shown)


parquet/metadata/adaptive_bloom_filter.go
```diff
@@ -79,9 +80,13 @@ type adaptiveBlockSplitBloomFilter struct {
 }
 
 func NewAdaptiveBlockSplitBloomFilter(maxBytes uint32, numCandidates int, fpp float64, column *schema.Column, mem memory.Allocator) BloomFilterBuilder {
+	if fpp <= 0 || fpp >= 1 || math.IsNaN(fpp) {
+		panic("parquet: bloom filter false-positive probability must be in (0, 1)")
+	}
```
zeroshade, line 85: The predicate itself is right, including the non-redundant `IsNaN` term.

Two things though. First, this is an exported constructor returning only `BloomFilterBuilder`, so panic is the only in-signature option — but that means an invalid *user property* becomes a panic in the write path at `column_writer.go:811`. Validating in `WithBloomFilterFPP`/`WithBloomFilterFPPFor` (`writer_properties.go:397-413`) instead would make this unreachable from the public API.

Second, the same unguarded `fpp` reaches `NewBloomFilterFromNDVAndFPP` (`bloom_filter.go:327`) from `column_writer.go:807`, one line away from the call this protects. That path still hangs on fpp=1 and still panics for small `maxBytes` after this patch.

parquet/metadata/adaptive_bloom_filter.go
```diff
@@ -79,9 +80,13 @@ type adaptiveBlockSplitBloomFilter struct {
...
+	maxBytes = max(minimumBloomFilterBytes, min(maximumBloomFilterBytes, maxBytes))
```
zeroshade, line 86: This clamp is correct and I confirmed it's a no-op for every `maxBytes >= 32`, so valid configurations are unaffected — it only rescues the 0..31 range that previously produced a zero-byte filter and panicked on insert.

The inconsistency is what I'd flag: an out-of-range `fpp` panics three lines up, while an out-of-range `maxBytes` is silently corrected here. Both are user-supplied; picking one policy for both would be clearer. Note `NewBloomFilterFromNDVAndFPP` has no equivalent clamp at all.

zeroshade (changes requested):
Both bugs are real and I verified them, but the fix is incomplete in a way that leaves the same crash reachable through the production write path, so I'd like one more pass.

**The infinite loop is real.** In `expectedNDV` (`adaptive_bloom_filter.go:225-229`) the loop is `for optimalBytes < numBytes { expectedNDV += ndvStep; optimalBytes = optimalNumBytes(expectedNDV, fpp) }`. With `fpp == 1`, `optimalNumBits` computes `m = -8*ndv / Log(1 - fpp^(1/8))`, and `Log(0) = -Inf` drives `m` to `+0`, which is then clamped up to the 256-bit minimum — so `optimalNumBytes` returns 32 for *every* ndv and the condition never becomes false. I measured non-termination for fpp in {1, NaN, 2, -0.1}. Your point about non-assertion builds is also correct: the `debug.Assert(fpp > 0 && fpp < 1)` at `bloom_filter.go:66` compiles away.

**The zero-byte panic is real too.** `maxBytes=0` → `initCandidates` falls through to the `:211-214` fallback → `newBloomFilterCandidate(..., minBytes=32, maxBytes=0)` → `:40-42` clamps back down to 0 → zero-length buffer → `InsertHash` panics with `index out of range [0] with length 0`. Pre-patch the panicking set is {0,1,2,4,8,16}; post-patch 0..31 all clamp to 32 and work.

**The interval is correct and nothing valid breaks.** `fpp <= 0 || fpp >= 1 || math.IsNaN(fpp)` is properly open at both ends, and the `IsNaN` term is not redundant (both comparisons are false for NaN). Every existing call site passes 0.01, and `DefaultBloomFilterFPP = 0.01` (`writer_properties.go:58`). The new clamp `max(32, min(128MB, maxBytes))` is identical to the old `min(128MB, maxBytes)` for all `maxBytes >= 32`, so no valid configuration changes behavior. Zero regressions across `go test ./...`.

What I'd like changed:

**1. The sibling constructor on the same path is still vulnerable.** `NewBloomFilterFromNDVAndFPP` (`bloom_filter.go:327-331`) is reached from `column_writer.go:807` whenever the user sets an NDV — right next to the `:811` call this PR guards. It has the same unvalidated `fpp` and no minimum-size clamp, and I confirmed it still panics: `maxBytes=0` gives `index out of range [0] with length 0`, and `maxBytes=16` gives `index out of range [4] with length 4`. Fixing only the adaptive constructor leaves the crash a config flag away.

**2. Validation belongs at the property setters.** `WithBloomFilterFPP` / `WithBloomFilterFPPFor` (`writer_properties.go:397-413`) accept `fpp` with no checking, so an invalid user property currently becomes a panic deep in the write path at `column_writer.go:811`. Rejecting it where the user supplies it would make the panic unreachable from the public API and give a far better error.

**3. Policy inconsistency**: this panics for an invalid `fpp` but silently clamps an invalid `maxBytes`. Both are user-supplied configuration; it'd be good for them to agree, and the PR body says "Reject" for both, which reads as an error rather than a panic.

One small accuracy note: `fpp = 0` does **not** hang — it terminates and returns 0, because `m = -Inf` clamps to the 128MB maximum and the loop exits immediately. Rejecting it is still right (it implies infinite bits, and silently yields a degenerate filter), but it's broader than the stated bug.

With the sibling constructor covered and ideally the setters validating, I'm happy to merge this.

zeroshade (approved):
All three asks substantively addressed and empirically verified. Approving, with two small follow-ups I would like before or alongside merge.

**(a) `NewBloomFilterFromNDVAndFPP`** — done (`bloom_filter.go:324-331`): fpp validated, maxBytes clamped, and the duplicated body now delegates to `NewBloomFilter`, which also picks up the power-of-2 rounding and min-size clamp it previously lacked. `NewBloomFilter` itself hardened at `:296`.

**(b) Property setters** — done. `WithBloomFilterFPP` (`writer_properties.go:405-410`) and `WithBloomFilterFPPFor` (`:414-418`) both call `validateBloomFilterFPP` (`:421-425`), and `WithBloomFilterFPPPath` routes through `FPPFor` so it is covered too. I checked for a bypass and there is none — `defColumnProps`/`columnProps` are unexported and no `WriterProperty` injects a whole `ColumnProperties`, so the setters are the only route to `BloomFilterFPP`. The panic is now genuinely unreachable from the public API.

Panic (rather than error) at the setters is the right call: `WriterProperty` is `func(cfg *writerPropConfig)` with no error return, and it matches the pre-existing idiom in this same file (`writer_properties.go:235`, `:246` already panic for the dictionary fallback case).

**Empirical verification, bounded-time so nothing hangs.** Baseline confirmed both defects, and with the patch every case is fixed:
```
BASELINE  NewBloomFilterFromNDVAndFPP maxBytes=0..31   -> panic: index out of range
BASELINE  NewAdaptiveBlockSplitBloomFilter fpp=1/NaN/-0.1 -> HANG
PATCHED   fpp in {0,1,-0.1,2,NaN} -> clean panic; {0.01,0.5,0.99,1e-9} -> ok
PATCHED   maxBytes in {-1,0,1,7,15,16,31,32,33,1024,max+1} -> all ok, 32-byte minimum, InsertHash works
```
No false positives: `DefaultBloomFilterFPP` (0.01) and `DefaultMaxBloomFilterBytes` (1 MiB) are both inside the accepted ranges, and an end-to-end write+read across 11 property combinations passes. Full `go test ./parquet/...` matches baseline.

**A correction to my own earlier review, for the record**: I wrote that `NewBloomFilterFromNDVAndFPP` "still hung on fpp=1". That was wrong — at fpp=1 `optimalNumBits` yields `-8*ndv/log(0) = 0`, which clamps to the 256-bit floor and returns a 32-byte filter without looping. The hang was exclusively in `NewAdaptiveBlockSplitBloomFilter`. The "panics for small maxBytes" half of my claim was correct, and both defects are fixed regardless, but my attribution was off and I did not want that standing uncorrected.

Two things I would still like:

1. **`bloom_filter_test.go`** — the `"rejects invalid false-positive probabilities"` subtest uses `assert.PanicsWithValue` with the full message text, so any rewording breaks it. The sibling subtests use plain `assert.Panics`; please match them.
2. **`WithMaxBloomFilterBytes` doc comment (`writer_properties.go:370-371`)** — this is now a real public-API behavior change that is undocumented. Values above 128 MiB previously *worked* (they were silently clamped downstream) and now panic. The lower-bound panic is a strict improvement since those values already crashed at write time, but the upper bound converts previously-working input into a hard panic. Please note the panic in the doc comment; @zeroshade may also want a release-note line.

Also worth a cross-referencing comment: `minimumBloomFilterBytes`/`maximumBloomFilterBytes` now exist in both `parquet` and `parquet/metadata`. I confirmed the duplication is forced (metadata imports parquet, so the reverse would cycle) and the values agree today, but it is a silent-drift hazard.

parquet/writer_properties.go
```diff
@@ -366,6 +371,9 @@ func WithPageIndexEnabledPath(path ColumnPath, enabled bool) WriterProperty {
 // it is abandoned and not written to the file.
 func WithMaxBloomFilterBytes(nbytes int64) WriterProperty {
 	return func(cfg *writerPropConfig) {
+		if nbytes < minimumBloomFilterBytes || nbytes > maximumBloomFilterBytes {
+			panic("parquet: maximum bloom filter size must be between 32 bytes and 128 MiB")
+		}
```
zeroshade, line 376: This is the item I'd most like addressed before merge, because it changes documented public behavior.

The **lower** bound is a strict improvement — values below 32 already crashed at write time (`InsertHash` on a zero-byte filter), so panicking early is better in every way.

The **upper** bound is different: `WithMaxBloomFilterBytes(nbytes > 128 MiB)` previously *worked*. It was silently clamped downstream by `min(maximumBloomFilterBytes, maxBytes)`, so a caller passing e.g. 256 MiB got a valid 128 MiB filter and a working writer. After this change that same call panics.

Failing fast is arguably the right choice, and it's the consistent resolution of the policy question I raised. But the doc comment just above (`:369-371`) doesn't mention that this setter can panic at all, so a caller has no way to know. Could you add a line — something like *"Panics if nbytes is outside the supported range of 32 bytes to 128 MiB."*

@zeroshade may also want a release-note entry, since this converts previously-working input into a panic.

Minor wording point: the panic message hard-codes "32 bytes and 128 MiB" while the bounds come from `minimumBloomFilterBytes`/`maximumBloomFilterBytes`. If those constants ever change, the message goes stale — worth formatting them in.

parquet/metadata/bloom_filter_test.go
```diff
@@ -272,6 +287,27 @@ func TestAdaptiveBloomFilterEdgeCases(t *testing.T) {
...
+
+		assert.EqualValues(t, minimumBloomFilterBytes, bf.maxBytes)
+		assert.NotPanics(t, func() { bf.InsertHash(1) })
+	})
+
+	t.Run("rejects invalid false-positive probabilities", func(t *testing.T) {
+		for _, fpp := range []float64{-0.1, 0, 1, math.NaN()} {
+			assert.PanicsWithValue(t,
+				"parquet: bloom filter false-positive probability must be in (0, 1)",
+				func() { NewAdaptiveBlockSplitBloomFilter(1024, 1, fpp, col, mem) })
```
zeroshade, line 308: `assert.PanicsWithValue` pins the exact panic string, so any rewording of the message — including the `parquet:` prefix — breaks this test with no behavior change. The sibling subtest just above uses `assert.NotPanics`, and the rest of this file favors the looser forms; `assert.Panics` would match.

If you want to keep asserting *which* validation fired, a substring check on the recovered value would be more durable than exact equality.

Also worth knowing about this subtest specifically: against **unpatched** code it wouldn't fail cleanly, it would **hang** (that's the infinite loop in `expectedNDV` this PR fixes), so on the pre-fix tree it manifests as a CI timeout rather than a crisp assertion failure. That's fine for guarding the patched behavior — just noting it isn't a sharp regression detector if anyone ever reverts the production change.

parquet/writer_properties.go
```diff
@@ -404,10 +413,17 @@ func WithBloomFilterFPP(fpp float64) WriterProperty {
...
+		validateBloomFilterFPP(fpp)
 		cfg.bloomFilterFPPs[path] = fpp
 	}
 }
 
+func validateBloomFilterFPP(fpp float64) {
+	if fpp <= 0 || fpp >= 1 || math.IsNaN(fpp) {
+		panic("parquet: bloom filter false-positive probability must be in (0, 1)")
+	}
+}
```
zeroshade, line 425: No change needed here — the helper and its placement are right, and routing `WithBloomFilterFPP`, `WithBloomFilterFPPFor` and (transitively) `WithBloomFilterFPPPath` through it is what makes the deeper panic unreachable from the public API. I verified there's no bypass: `defColumnProps`/`columnProps` are unexported and no `WriterProperty` injects a whole `ColumnProperties`.

One note for the future, since this file now needs the bounds: `minimumBloomFilterBytes`/`maximumBloomFilterBytes` exist in **both** `parquet` (`:64-65`) and `parquet/metadata` (`bloom_filter.go:44,46`). I confirmed the duplication is forced — `parquet/metadata` imports `parquet`, so the reverse direction would be an import cycle — and the values agree today (32 / 134217728). Still a silent-drift hazard if one side is ever tuned. A short comment on each pair pointing at the other would make that safe.

zeroshade (commented):
Following up with the two items from my approval as inline comments so they're easier to action. Neither is a correctness blocker — the approval stands and I'm happy to merge once CI is green — but the doc one is a genuine public-API contract change that I'd like recorded before this lands.

PR #1034, opened 2026-07-27
fix(parquet/pqarrow): calculate large decimal widths

fallintoplace:
### Rationale for this change

`DecimalSize` uses a lookup table through precision 76 and a formula above that range. The fallback applied `ceil` before multiplying by `log2(10)`, producing widths larger than the documented minimum. For example, precision 77 returned 34 bytes instead of 33.

Apply the documented two-complement width formula as `ceil((precision * log2(10) + 1) / 8)` for precisions above 76, and correct the existing precision-77 schema fixture.

### Are these changes tested?

Yes. Regression cases cover precisions 77, 80, and 100. The full `parquet/pqarrow` package passes.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -246,7 +246,7 @@ func DecimalSize(precision int32) int32 {
 	if precision <= 76 {
 		return byteblock[precision]
 	}
-	return int32(math.Ceil(float64(precision)/8.0)*math.Log2(10) + 1)
+	return int32(math.Ceil((float64(precision)*math.Log2(10) + 1) / 8))
 }
 
 func repFromNullable(isnullable bool) parquet.Repetition {
--- a/parquet/pqarrow/schema_test.go
+++ b/parquet/pqarrow/schema_test.go
@@ -524,7 +524,7 @@ func TestConvertArrowDecimals(t *testing.T) {
 	arrowFields = append(arrowFields, arrow.Field{Name: "decimal_20_4", Type: &arrow.Decimal128Type{Precision: 20, Scale: 4}})
 
 	parquetFields = append(parquetFields, schema.Must(schema.NewPrimitiveNodeLogical("decimal_77_4", parquet.Repetitions.Required,
-		schema.NewDecimalLogicalType(77, 4), parquet.Types.FixedLenByteArray, 34, -1)))
+		schema.NewDecimalLogicalType(77, 4), parquet.Types.FixedLenByteArray, 33, -1)))
 	arrowFields = append(arrowFields, arrow.Field{Name: "decimal_77_4", Type: &arrow.Decimal128Type{Precision: 77, Scale: 4}})
 
 	arrowSchema := arrow.NewSchema(arrowFields, nil)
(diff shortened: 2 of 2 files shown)


parquet/pqarrow/schema.go
```diff
@@ -246,7 +246,7 @@ func DecimalSize(precision int32) int32 {
 	if precision <= 76 {
 		return byteblock[precision]
 	}
-	return int32(math.Ceil(float64(precision)/8.0)*math.Log2(10) + 1)
+	return int32(math.Ceil((float64(precision)*math.Log2(10) + 1) / 8))
```
zeroshade, line 249: This now matches the generator comment at line 238 verbatim, which is the strongest evidence it's the intended formula.

Confirmed exactly minimal against big-integer arithmetic for p=77..3000, and continuous with `byteblock` at the p=76 boundary (new formula reproduces all 76 table entries). The previous expression applied `ceil` to `precision/8` before multiplying by `log2(10)`, which mixed units and produced a step function — constant 34 across p=74..80 — and under-allocated from p=248 onward.

zeroshade (approved):
Verified thoroughly — this is a real correctness fix and the math is exactly right. LGTM.

**The bug is genuine, and the file itself proves it.** The generator comment at `schema.go:238` documents the intended formula as `int(math.ceil((prec * math.log2(10) + 1) / 8))`, which is precisely what you've now implemented — but the fallback at `:249` had the `ceil` misplaced (quantizing *precision* into 8-wide buckets) and the `+1` outside the division. So the fallback contradicted the very table it was meant to extend.

**Independent verification with exact integer arithmetic** (no floats), since minimum bytes N to hold ±(10^p − 1) in two's complement requires `2^(8N−1) >= 10^p`:

For p=77: `2^255 < 10^77 <= 2^256`, so 256 magnitude bits + 1 sign bit = 257 bits → `ceil(257/8)` = **33 bytes**.

| | computation | result |
|---|---|---|
| old | `ceil(77/8)=10`; `10 x 3.321928 + 1 = 34.219`; truncated | 34 |
| new | `(77 x 3.321928 + 1)/8 = 32.0986`; ceil | **33** |
| exact | 257 bits | **33** |

**The boundary-continuity check is the one I most expected to fail, and it's clean**: applying your new formula to p=1..76 reproduces the lookup table with **zero mismatches**, so there's no discontinuity at the p=76 seam. (The old formula, applied to that same range, mismatches the table in 60 of 76 cases — confirming it was never the table's generator.)

**And the old formula was unsafe, not merely wasteful.** Checked against exact arithmetic for p=77..3000: your formula matches in all cases; the old one mismatches 1471 times, and **44 of those are under-allocations**, beginning at p=248 where it returns 103 bytes when 104 are required. That's silent truncation.

**It also removes a genuine self-inconsistency.** `DecimalLogicalType.IsApplicable` (`parquet/schema/logical_types.go:403`) encodes the inverse constraint. The old width was *rejected* by that check at p=248, 272 and 296 — meaning `DecimalSize` could hand `NewPrimitiveNodeLogical` a width it would refuse. With your formula every p is accepted and is exactly minimal.

**The fixture change is a legitimate correction**, not masking: `schema_test.go:527` was asserting the buggy 34, and 33 is independently confirmed as the true minimum. It's the only existing expectation touched.

**No compatibility risk.** `DecimalSize` is write-path only — the read path never consults it (`arrowFromFLBA` → `arrowDecimal` uses precision/scale and ignores the FLBA length, with values decoded from the file's own per-value byte length). Files previously written with width 34 stay readable.

Two notes for the record, neither blocking:
1. Arrow C++ has the **identical** bug at `cpp/src/arrow/type.cc:1479`, including the same generator comment. This makes Go diverge from C++ — toward correctness. There's no interop hazard because C++ can't reach the fallback at all (`ValidateDecimalPrecision` caps Decimal256 at 76), but it's worth an upstream issue.
2. The p>76 path is effectively unreachable via any valid Arrow type (Decimal256 max precision is 76), and writing data at p>=77 already panics identically before and after this change, since `encode_arrow.go:533` computes a negative offset for both 33 and 34. So this is a latent-correctness fix rather than a user-visible one — no behavior regression either way.

PR #1035, opened 2026-07-27
fix(parquet/schema): reject invalid fixed lengths

fallintoplace:
### Rationale for this change

Fixed-length byte-array constructors reject non-positive widths, but `PrimitiveNode.SetTypeLength` allowed an existing valid node to be mutated to zero or a negative width. The invalid node could then be serialized or passed to encoders.

* Reject non-positive widths when mutating fixed-length byte-array nodes.
* Leave non-fixed-width primitive nodes unchanged, preserving the existing no-op behavior.
* Document the panic contract on the setter.

### Are these changes tested?

Yes. The regression test covers zero and negative widths, verifies that failed mutations preserve the previous width, and checks the existing no-op behavior for other physical types. The full `parquet/schema` package passes.

### DIFF
--- a/parquet/schema/node.go
+++ b/parquet/schema/node.go
@@ -319,10 +319,13 @@ func (p *PrimitiveNode) Equals(rhs Node) bool {
 // to store the values in this column.
 func (p *PrimitiveNode) PhysicalType() parquet.Type { return p.physicalType }
 
-// SetTypeLength will change the type length of the node, has no effect if the
-// physical type is not FixedLength Byte Array
+// SetTypeLength will change the type length of the node, and has no effect if the
+// physical type is not FixedLength Byte Array. It panics if length is not positive.
 func (p *PrimitiveNode) SetTypeLength(length int) {
 	if p.PhysicalType() == parquet.Types.FixedLenByteArray {
+		if length <= 0 {
+			panic("parquet: fixed length byte array length must be positive")
+		}
 		p.typeLen = length
 	}
 }
--- a/parquet/schema/schema_test.go
+++ b/parquet/schema/schema_test.go
@@ -191,6 +191,20 @@ func (p *PrimitiveNodeTestSuite) TestEquals() {
 	p.False(flba1.Equals(flba5))
 }
 
+func (p *PrimitiveNodeTestSuite) TestSetTypeLengthRejectsInvalidLength() {
+	node := schema.NewFixedLenByteArrayNode("value", parquet.Repetitions.Required, 4, -1)
+
+	for _, length := range []int{0, -1} {
+		p.PanicsWithValue("parquet: fixed length byte array length must be positive", func() {
+			node.SetTypeLength(length)
+		})
+		p.Equal(4, node.TypeLength())
+	}
+
+	other := schema.NewInt32Node("value", parquet.Repetitions.Required, -1)
+	p.NotPanics(func() { other.SetTypeLength(0) })
+}
+
 func (p *PrimitiveNodeTestSuite) TestPhysicalLogicalMapping() {
 	tests := []struct {
 		typ       parquet.Type


parquet/schema/node.go
```diff
@@ -319,10 +319,13 @@ func (p *PrimitiveNode) Equals(rhs Node) bool {
...
 
-// SetTypeLength will change the type length of the node, has no effect if the
-// physical type is not FixedLength Byte Array
+// SetTypeLength will change the type length of the node, and has no effect if the
+// physical type is not FixedLength Byte Array. It panics if length is not positive.
 func (p *PrimitiveNode) SetTypeLength(length int) {
 	if p.PhysicalType() == parquet.Types.FixedLenByteArray {
+		if length <= 0 {
+			panic("parquet: fixed length byte array length must be positive")
+		}
```
zeroshade, line 328: Placement is right — inside the FLBA branch, so non-fixed-width primitives keep their documented no-op behavior, and the guard precedes the assignment so a rejected call leaves `typeLen` untouched (your test asserts exactly that).

This also makes the setter agree with `NewFixedLenByteArrayNode` at line 627, which already panics on the same condition through `MustPrimitive`. Worth noting the payoff concretely: without this, `toThrift` at line 360 would happily serialize a negative `type_length` into the footer.

zeroshade (approved):
Verified — this closes a real mutation hole and the mechanism matches existing convention in the file. LGTM.

**The gap is genuine.** Every FLBA constructor validates the width — `node.go:180` (`NewPrimitiveNodeLogical`) and `:255-259` (`NewPrimitiveNodeConverted`) both reject non-positive lengths, and `NewFixedLenByteArrayNode` (`:627-629`) already panics via `MustPrimitive`. But `SetTypeLength` assigned `p.typeLen` unchecked, so a valid node could be mutated into an invalid one afterwards. Confirmed empirically:

```
ctor typeLen=0 err:                     invalid fixed length byte array length
NewFixedLenByteArrayNode(0) recovered:  invalid fixed length byte array length
after SetTypeLength(0),  TypeLength =   0
after SetTypeLength(-7), TypeLength =   -7
column TypeLength:                      -7
```

**And the corruption isn't inert**, which is what makes this worth fixing rather than merely tidy: `PrimitiveNode.toThrift` (`node.go:360-362`) unconditionally writes `elem.TypeLength = thrift.Int32Ptr(int32(p.typeLen))` for FLBA, so a negative width would be serialized into the file footer as a spec-invalid `type_length`. Your rationale about it being "serialized or passed to encoders" checks out.

**Panic is the right mechanism here.** `SetTypeLength` has no error return, and adding one would break the exported API in v18. Panicking on invalid arguments is already the established pattern in this exact file and package — `Must`/`MustGroup`/`MustPrimitive` (`node.go:445-467`), `NewFixedLenByteArrayNode` (`:627-629`) for this very condition, and `logical_types.go:349`/`:352`/`:786`. Your message also follows the `parquet: ...` prefix style. Net effect is that `NewFixedLenByteArrayNode(..., 0, ...)` and `node.SetTypeLength(0)` now behave identically, so this *removes* an inconsistency rather than introducing a new policy.

**No legitimate caller is affected.** Exhaustive search for `SetTypeLength` across the module (excluding the unrelated generated `IsSetTypeLength` in the Thrift bindings) finds only two call sites, both in `schema_test.go` (`:177` and `:181`), passing 12 and 16. Zero non-test callers. Non-FLBA primitives remain a no-op, matching the documented contract and prior behavior.

**Good test.** It table-drives `{0, -1}`, and importantly asserts `TypeLength()` is still 4 after each panic — verifying no partial mutation, which is the invariant that actually matters. It also pins the non-FLBA no-op with `NotPanics`. Purely additive; no existing expectation changed. `go test ./parquet/...` is green and identical to baseline.

One note for the release notes, not a blocker: this is a behavior change for any external caller currently passing a non-positive length and getting silent corruption — they'll now get a panic. That's the intent, and it's consistent with the sibling constructor, but it's worth a mention.

PR #1036, opened 2026-07-27
feat(parquet): add WithDictionaryCostFallback writer property

tom-eon:
### Rationale for this change

The pre-first-page dictionary cost check (`checkDictionarySizeLimit`, mirroring parquet-mr's `FallbackValuesWriter.shouldFallBack`) compares **uncompressed** sizes: the dictionary is discarded when `dictSize + encodedSize >= rawSize`. That comparison only predicts on-disk cost when pages are stored uncompressed. With a page compressor configured it can discard dictionaries that clearly win after compression — bit-packed dictionary indices are near-incompressible but small, while PLAIN-encoded values often compress far less than the raw-size comparison assumes. In our benchmarks (ZSTD level 3), a mid-cardinality `Time64` column written PLAIN after the fallback was 2× larger than the same column dictionary-encoded (7.1 MB vs 3.6 MB compressed); mid-cardinality integer columns lost 20%+.

- **The cost fallback is now codec-aware by default**: it only runs for columns written without page compression, where the uncompressed-size comparison is sound. For columns with a codec configured the dictionary is kept (still bounded by the `DictionaryPageSizeLimit` fallback, which is unaffected). Uncompressed columns keep today's behavior exactly.
- **A per-column override**: `parquet.WithDictionaryCostFallbackFor(path string, enabled bool)` / `WithDictionaryCostFallbackPath(path ColumnPath, enabled bool)`, stored as a tri-state `DictionaryCostFallback *bool` on `ColumnProperties` (nil = codec-based default). It overrides the default in either direction: `false` keeps an explicitly requested dictionary for an uncompressed column, `true` forces the check for a compressed one. Useful for callers that decide dictionary usage from their own measurements (e.g. trial-encoding a sample of the actual data with the actual codec).
- `WriterProperties.DictionaryCostFallbackEnabledFor(path string) bool` resolves the per-column override, falling back to the codec-based default.

### Are these changes tested?

Yes — `parquet/file/dict_cost_fallback_test.go` covers: (1) an uncompressed column still discards a non-paying dictionary by default, (2) global and (3) per-column compression keep the dictionary by default, (4) a per-column disable keeps an uncompressed column's dictionary while (5) a disable scoped to a different column leaves the fallback active, (6) a per-column enable forces the check for a compressed column, and (7) the `DictionaryPageSizeLimit` fallback still applies when the cost fallback is off. The existing `TestDictFallbackDiscardsOrphanDict` snappy variants now force the check via the per-column option to keep exercising the orphan-discard path under a codec. Existing `parquet`, `parquet/file`, and `parquet/internal/encoding` suites pass.

For columns written with a compression codec, the cost-based dictionary fallback no longer runs by default — a requested dictionary is kept unless it overflows `DictionaryPageSizeLimit`. Uncompressed columns are unchanged. New writer properties: `WithDictionaryCostFallbackFor` / `WithDictionaryCostFallbackPath`, and a `DictionaryCostFallback *bool` field on `ColumnProperties`.

### DIFF
--- a/parquet/file/column_writer_types.gen.go
+++ b/parquet/file/column_writer_types.gen.go
@@ -245,8 +245,12 @@ func (w *Int32ColumnChunkWriter) checkDictionarySizeLimit() {
 	// plus the encoded indices meet or exceed the raw input bytes, fall back
 	// to PLAIN now and discard the dictionary — avoiding the mid-cardinality
 	// case where a dict page stays in the file alongside PLAIN pages without
-	// any net compression win. The comparison ignores page compression, so
-	// WithDictionaryCostFallback(false) lets callers opt out of it.
+	// any net compression win. The comparison uses uncompressed sizes, which
+	// is only meaningful when pages are stored uncompressed: with a codec
+	// configured, the near-incompressible dict indices routinely beat PLAIN
+	// pages that compress well. The check therefore runs only for columns
+	// written without page compression, unless overridden per column with
+	// WithDictionaryCostFallbackFor.
 	if !w.dictPageWritten && len(w.pages) == 0 && w.props.DictionaryCostFallbackEnabledFor(w.descr.Path()) {
 		rawSize := dictEnc.ObservedRawSize()
 		encodedSize := dictEnc.EstimatedDataEncodedSize()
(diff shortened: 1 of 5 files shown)


tom-eon:
@zeroshade just pinging to know if there's a chance of getting this in

zeroshade:
@tom-eon Hey, sorry I was away last week, i'll give this a review and respond today or tomorrow

zeroshade:
@tom-eon The actual bug is the compression-blind cost check, which only makes sense when the data is uncompressed. Could we gate this fallback on "compression is Uncompressed", or otherwise make it codec-aware, so we fix the misprediction on the common compressed case? We can keep the per-column opt-out option you've added, but adding the codec-aware default would fix this for everyone without requiring users to find and use this knob.

The PR in general looks good otherwise.

tom-eon:
@zeroshade Done — the cost fallback is now gated on the column's codec being uncompressed, which fixes the misprediction by default: the uncompressed-size comparison only predicts on-disk cost when no codec is configured. I also dropped the blanket `WithDictionaryCostFallback` option; what remains is the per-column `WithDictionaryCostFallbackFor`, which now overrides the codec-based default in either direction (`false` keeps a dictionary on an uncompressed column, `true` forces the check on a compressed one). Tests cover both directions plus the codec-aware default at global and per-column compression levels. Updated the PR description to match.

zeroshade (approved):

PR #1037, opened 2026-07-28
fix(tmpl): report generated output write errors

fallintoplace:
## What changed

Check the final `os.WriteFile` result when generating template output and return the failure to `main` for reporting.

## Why

The generator previously exited successfully when the output could not be written, for example when the output path was a directory. That could leave stale generated files in place while making the generation step appear successful.

The new regression test uses a directory as the output target and verifies that `process` returns an error containing the failed path.

## Validation

`go test -vet=off ./arrow/_tools/tmpl`

### DIFF
--- a/arrow/_tools/tmpl/main.go
+++ b/arrow/_tools/tmpl/main.go
@@ -114,7 +114,9 @@ func main() {
 	}
 
 	in.In = readData(*dataArg)
-	process(in, specs)
+	if err := process(in, specs); err != nil {
+		errExit("%s", err)
+	}
 }
 
 func mustReadAll(path string) []byte {
@@ -148,7 +150,7 @@ var funcs = template.FuncMap{
 	"upper": strings.ToUpper,
 }
 
-func process(data interface{}, specs []pathSpec) {
+func process(data interface{}, specs []pathSpec) error {
 	for _, spec := range specs {
 		var (
 			t   *template.Template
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Real bug in the code generator, and I verified the fix does not alter any generated output. LGTM.

The old `process` discarded the `os.WriteFile` error (`main.go:181`) and had no error return, so a failed write exited 0. Demonstrated with the unpatched binary:
```
unpatched, output path is a directory:  EXIT=0
unpatched, read-only output file:       EXIT=0, content still stale
patched,   read-only output file:       EXIT=1, error writing generated output ...: permission denied
```
Silently exiting 0 with stale output on disk is exactly the failure mode you want loud in a generator.

**Byte-identity verified**, since that is the real risk with a codegen change: I recorded sha256 for 13 generated files before the patch, applied it, regenerated, and re-compared — all identical, including `parquet/metadata/statistics_types.gen.go`, the `arrow/math` set, the `internal/utils/transpose_ints*` family, `arrow/type_traits_numeric.gen.go` and `arrow/array/numericbuilder.gen.go`. File modes unchanged at 0644.

On my `fileMode(spec.in)` question — you did not touch it and it is fine in practice: all 253 `.tmpl` files in the repo are 0644, and `os.WriteFile` only applies `perm` on create, so regenerating a committed file keeps its existing mode. Leaving it alone was correct.

One suggestion, optional. `go test ./arrow/_tools/tmpl` still fails to build under default vet both before and after this change, because of the pre-existing non-constant-format `errExit(err.Error())` at `main.go:125`/`143` — which is why you needed `-vet=off`. Since `_tools` is excluded from `./...`, this new test will not actually run in CI. Two `errExit("%s", ...)` edits in the same file would make the package vet-clean and the test reachable. Worth doing while you are here, but not a blocker.

PR #1038, opened 2026-07-28
fix(arrow/array): propagate embedded pointer nullability

fallintoplace:
## What changed

Carry nullability through anonymous embedded pointer paths while collecting reflected struct fields.

## Why

A promoted non-pointer field beneath an embedded pointer can be absent when that pointer is nil. The array builder already emits a null in that case, but schema inference marked the field non-nullable because it only inspected the leaf field type.

This change makes the inferred schema agree with the values produced by the builder, including for multi-level embedded pointer paths.

## Validation

`go test ./arrow/array/arreflect`

### DIFF
--- a/arrow/array/arreflect/reflect.go
+++ b/arrow/array/arreflect/reflect.go
@@ -143,9 +143,10 @@ func parseDecimalOpt(opts *tagOpts, token string) {
 }
 
 type bfsEntry struct {
-	t     reflect.Type
-	index []int
-	depth int
+	t        reflect.Type
+	index    []int
+	depth    int
+	nullable bool
 }
 
 type candidate struct {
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Real schema/data contract violation, correctly and narrowly fixed. LGTM.

On `main`, `reflect.go:230` derived nullability from the leaf type only (`sf.Type.Kind() == reflect.Ptr`), while the embedded-pointer hop at `:216-219` dereferenced without recording that the path was nullable. The builder meanwhile calls `fieldByIndexSafe`, which returns false on a nil pointer mid-path and appends a null. So the schema said NOT NULL and the data contained nulls:
```
datatype: struct<ID: int32, Value: int32, Label: utf8>   (all nullable=false)
array:    {[1 2] [7 (null)] ["a" (null)]}
*** field "Value" declared NOT NULL but has 1 nulls
*** field "Label" declared NOT NULL but has 1 nulls
```
That is the kind of defect that corrupts downstream consumers — writing this to Parquet with `required` fields would be invalid.

The fix is correctly scoped. Computing `pathNullable` before the deref loop and OR-ing it in at the leaf means nullability is inherited only along pointer-traversed paths. Verified every shape, and only the intended cells changed:

| shape | baseline | patched |
|---|---|---|
| `struct{ID; *Inner}` | false | **true** |
| `struct{ID; *Middle{*Inner}}` | false | **true** |
| `struct{ID; Inner}` value embed | false | false |
| `struct{ID; *Mid{Inner}}` | false | **true** |
| top-level `ID` | false | false |

Nested non-embedded struct fields recurse through a fresh `inferStructType` with `nullable=false`, so their children are correctly untouched.

Both test files are additions only (`+2/-0` and `+17/-0`, zero deletions) — no existing expectation was modified. The two lines land inside the existing `nil_embedded_pointer_promoted_field` test that already asserted City/Zip go null, so they are precisely the missing half of that contract.

This does change inferred schema nullability, which is user-visible, but `arreflect` has zero consumers elsewhere in the repo and the full suite shows no new failures.

One pre-existing wart I explicitly ruled out as a regression: for a type reachable at equal depth by both value and pointer embedding, the `visited` dedup at `reflect.go:182-187` drops one candidate so the result depends on BFS order. I ran it both ways and got identical output on baseline and patched — pre-existing, not introduced here, but possibly worth a separate issue.

PR #1040, opened 2026-07-28
fix(arrow/scalar): support same-width time casts

fallintoplace:
## What changed

Extract temporal scalar values according to the source width before converting them to a destination `Time32` or `Time64` unit.

## Why

The previous implementation chose the source value assertion from the destination type. Same-width conversions therefore asserted `Time32` values as `Time64`, or vice versa, and panicked.

The regression matrix covers Time32-to-Time32, Time32-to-Time64, Time64-to-Time32, and Time64-to-Time64 conversions.

## Validation

`go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/scalar_test.go
+++ b/arrow/scalar/scalar_test.go
@@ -366,6 +366,48 @@ func TestTimeScalarsMakeScalar(t *testing.T) {
 	assert.Equal(t, "10:10:10.123456789", scalar.NewTime64Scalar(arrow.Time64(tententen), typ4).String())
 }
 
+func TestTimeScalarUnitConversions(t *testing.T) {
+	tests := []struct {
+		name string
+		from scalar.TimeScalar
+		to   arrow.DataType
+		want scalar.Scalar
+	}{
+		{
+			name: "time32 to time32",
+			from: scalar.NewTime32Scalar(2, arrow.FixedWidthTypes.Time32s),
+			to:   arrow.FixedWidthTypes.Time32ms,
+			want: scalar.NewTime32Scalar(2000, arrow.FixedWidthTypes.Time32ms),
+		},
+		{
+			name: "time32 to time64",
+			from: scalar.NewTime32Scalar(2, arrow.FixedWidthTypes.Time32s),
+			to:   arrow.FixedWidthTypes.Time64us,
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Real panic, correct fix. LGTM.

`temporal.go:157-163` on main asserted the *source* value against the *destination* type — `*arrow.Time32Type` case did `s.value().(arrow.Time64)` and vice versa. So the same-width casts panicked while the cross-width ones worked:
```
time32s->time32ms   PANIC: interface {} is arrow.Time32, not arrow.Time64
time64us->time64ns  PANIC: interface {} is arrow.Time64, not arrow.Time32
time32s->time64us   OK
time64us->time32ms  OK
```
Extracting `value int64` from the source dynamic type first is the right fix. Factors verified against the table at `arrow/datatype_fixedwidth.go:794-821`, with results confirmed by execution: `Time32[s] 2 -> [ms]` = 2000, `Time32[s] 2 -> Time64[us]` = 2000000, `Time64[us] 2000000 -> Time32[ms]` = 2000, `Time64[us] 2 -> [ns]` = 2000, and truncating narrowing (`Time32[ms] 1500 -> [s]` = 1, `Time64[ns] -1500 -> [us]` = -1). The two directions that already worked produce byte-identical results before and after.

On overflow: `Time32[s]=MaxInt32 -> [ms]` silently wraps, but that is unreachable for legal Arrow time-of-day (max `time32[ms]` is 86399999), and I confirmed the already-working direction has the same unchecked semantics on baseline. So this extends existing behavior rather than introducing a new class of problem.

**One thing worth fixing while you are here.** The inner `switch s := s.(type)` has no `default` arm. `*Timestamp` actually satisfies `TimeScalar`, and it is only saved by case ordering — `case *Timestamp:` (`:146`) precedes `case TimeScalar:` (`:157`). If those are ever reordered, or a new time-like scalar type is added, `value` silently stays 0 and you get **wrong data instead of a panic**. A `default: return nil, fmt.Errorf(...)` would make that impossible.

Minor test nits: `assert.True(t, scalar.Equals(...))` produces an unhelpful failure message, and the narrowing direction (ms->s, ns->us) that this also newly enables is not covered. `scalar.Equals` does compare `arrow.TypeEqual`, so the destination type is genuinely asserted — good.

PR #1041, opened 2026-07-28
fix(arrow/scalar): use duration type for units

fallintoplace:
## What changed

Read a duration scalar's unit from `arrow.DurationType` instead of asserting it is a timestamp type.

## Why

Valid duration scalars carry `*arrow.DurationType`. The previous assertion to `*arrow.TimestampType` panicked whenever `Unit` was called, including through string formatting and duration unit conversion.

The regression test covers direct unit access, string formatting, and seconds-to-milliseconds conversion.

## Validation

`go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/scalar_test.go
+++ b/arrow/scalar/scalar_test.go
@@ -482,6 +482,16 @@ func TestDurationScalarBasics(t *testing.T) {
 	assert.False(t, scalar.Equals(tsNull, tsVal2))
 }
 
+func TestDurationScalarUnitConversions(t *testing.T) {
+	s := scalar.NewDurationScalar(1, arrow.FixedWidthTypes.Duration_s)
+	assert.Equal(t, arrow.Second, s.Unit())
+	assert.Equal(t, "1s", s.String())
+
+	converted, err := s.CastTo(arrow.FixedWidthTypes.Duration_ms)
+	require.NoError(t, err)
+	assert.Equal(t, arrow.Duration(1000), converted.(*scalar.Duration).Value)
+}
+
 func TestMonthIntervalScalarBasics(t *testing.T) {
 	typ1 := arrow.FixedWidthTypes.MonthInterval
 	typ2 := arrow.FixedWidthTypes.MonthInterval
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Clean fix for a completely dead code path. LGTM.

`temporal.go:76` asserted `s.DataType().(*arrow.TimestampType)` inside `Duration.Unit()`, which can only ever hold a `*arrow.DurationType`. It was unconditionally broken:
```
duration.Unit()      PANIC: arrow.DataType is *arrow.DurationType, not *arrow.TimestampType
duration.String()    PANIC (via temporalToString at temporal.go:36)
duration s->ms cast  PANIC (via temporal.go:170)
```
Every construction site in the repo passes a `*arrow.DurationType` — `scalar.go:614`, `parse.go:721`/`:733`, `MakeNullScalar` at `scalar.go:568`, and `temporal.go:170` — so there was no path on which the old assertion could succeed.

The fix matches the sibling accessors exactly: `Time32.Unit()` asserts `*arrow.Time32Type` (`:253`), `Time64` asserts `*arrow.Time64Type` (`:291`), `Timestamp` asserts `*arrow.TimestampType` (`:328`).

No observable output change — `CastTo` already derived the result type from `to` (`:170`). The only behavior delta is for a scalar hand-built as `NewDurationScalar(v, arrow.FixedWidthTypes.Timestamp_s)`, which returned `Second` before and panics now; such a scalar is invalid by Arrow typing anyway (its `DataType()` must be DURATION for `Equals`/`GetScalar`/IPC to work) and nothing in-repo creates one.

Test is additive and fails on baseline with the panic. Small nits: it does not assert the converted scalar`s `DataType()` is `Duration_ms`, and `assert.Equal(t, "1s", s.String())` couples to `time.Duration.String()` formatting — stable, but formatting-coupled.

**Note for whoever merges**: this and #1040 both touch `arrow/scalar/temporal.go` and `scalar_test.go`, but they are **fully independent — no merge order required**. I verified with a real 3-way merge (`git merge-tree` exits 0, no conflict markers) and by applying in both orders, which converge to the byte-identical tree. The regions are disjoint: line 76 here versus 155-163 in #1040.

PR #1043, opened 2026-07-28
fix(arrow/array): validate union child upper bound

fallintoplace:
## What changed

Reject a union child index equal to the number of children.

## Why

The previous bounds check allowed `idx == len(children)` through, causing a generic runtime index panic instead of the intended Arrow validation panic.

The regression test covers both the negative and upper bounds for dense and sparse union builders.

## Validation

`go test ./arrow/array`

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -801,7 +801,7 @@ func (b *unionBuilder) NumChildren() int {
 }
 
 func (b *unionBuilder) Child(idx int) Builder {
-	if idx < 0 || idx > len(b.children) {
+	if idx < 0 || idx >= len(b.children) {
 		panic("arrow/array: invalid child index for union builder")
 	}
 	return b.children[idx]
--- a/arrow/array/union_test.go
+++ b/arrow/array/union_test.go
@@ -42,6 +42,41 @@ func int32ArrFromSlice(offsets ...int32) arrow.Array {
 	return array.MakeFromData(data)
 }
 
+func TestUnionBuilderChildBounds(t *testing.T) {
+	fields := []arrow.Field{{Name: "value", Type: arrow.PrimitiveTypes.Int32}}
+	codes := []arrow.UnionTypeCode{0}
+	tests := []struct {
+		name string
+		new  func() array.UnionBuilder
+	}{
+		{
+			name: "dense",
+			new: func() array.UnionBuilder {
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Genuine off-by-one, bounded against the right thing, and no valid input changes behavior. LGTM.

`union.go:804` used `idx > len(b.children)`, so `idx == len(children)` slipped through to `b.children[idx]` at `:807` and produced a generic `runtime error: index out of range` instead of the documented Arrow panic. The new test fails on baseline for both dense and sparse.

**The bound is against the correct thing**, which was my main worry given this is validation in `arrow/array`. `Child(idx int)` is documented at `:743` as taking a *child index* — a position in `b.children` — not a type code. Type-code lookup is a separate structure (`typeIDtoBuilder`/`typeIDtoChildID`, sized `MaxTypeCode()+1` at `:773-774`). So non-contiguous type codes and codes up to 127 are simply irrelevant here, and `len(b.children)` is right.

**No false positives.** `idx == len(children)` never worked — it always panicked; this only swaps a `runtime.Error` for the intended string panic, so there is no valid input whose behavior changes. I audited every `Child()` caller in the module: `scalar/append.go:200` and `:211` are bounded by `NumChildren()`/`len(Fields())`; `util/protobuf_reflect.go:814` passes a oneof ordinal with `-1` short-circuited at `:809`; `flightsql/sql_info.go:49-54` uses fixed indices into a 6-child union; the cdata and union tests use 0/1 on 2-child unions. All bounded.

I also checked the tempting look-alike at `arrow/schema.go:267` — `AddField`, where `i == len(fields)` legitimately means append-at-end. Correctly left alone; good that this did not get pattern-matched into a bug.

Keeping the panic is right: it is the documented contract at `:744-745`. Worth noting `unionBuilder.Child` was already the outlier in *having* a check at all — `StructBuilder.FieldBuilder` has none — so this just makes the existing check work.

Two things to fix:

1. **Stale doc comment.** `union.go:744` still says `// If an invalid index is requested (e.g. <0 or >len(children))`, which now contradicts the corrected code. Should be `>=len(children)`.
2. **Brittle assertion.** `assert.PanicsWithValue` hard-codes the exact panic string — this is the only exact-error-string assertion in this batch of PRs. `assert.Panics` would be more robust.

Minor: the test uses `memory.DefaultAllocator` where most of `union_test.go` uses `memory.NewCheckedAllocator` with `defer AssertSize(t, 0)`; a checked allocator would additionally prove the panic path does not leak child builders.

Full-module suite shows no newly failing package.

PR #1047, opened 2026-07-28
fix(arrow/ipc): make writer failures terminal

fallintoplace:
### Rationale for this change

An IPC writer can fail after writing part of a schema, dictionary, or record payload. The writer must not continue with stale state or lose the first failure.

Keep the first terminal error, mark the writer as started before schema payloads are written, close a started payload writer exactly once, preserve close and panic failures, and release retained dictionaries. Closing without a schema still returns arrow.ErrInvalid.

### Are these changes tested?

- `go test ./arrow/ipc`

Yes. After a terminal write or close failure, later writes return the stored error instead of continuing with stale writer state.

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -59,17 +59,11 @@ func (w *streamWriter) WritePayload(p Payload) error {
 func (w *streamWriter) Write(p []byte) (int, error) {
 	n, err := w.w.Write(p)
 	w.pos += int64(n)
-	if err == nil && n != len(p) {
-		err = io.ErrShortWrite
-	}
 	return n, err
 }
 
 func writeFull(w io.Writer, p []byte) error {
-	n, err := w.Write(p)
-	if err == nil && n != len(p) {
-		err = io.ErrShortWrite
-	}
+	_, err := w.Write(p)
 	return err
 }
 
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -138,17 +138,12 @@ func NewWriter(w io.Writer, opts ...Option) *Writer {
 
 func (w *Writer) Close() error {
 	if w.err != nil {
-		if w.started && w.pw != nil {
-			w.err = errors.Join(w.err, w.pw.Close())
-		}
-		w.releaseDictionaries()
-		w.pw = nil
-		return w.err
+		return w.closeAfterFailure()
 	}
 	if !w.started {
 		err := w.start()
 		if err != nil {
-			return err
+			return w.closeAfterFailure()
 		}
 	}
 
@@ -164,6 +159,15 @@ func (w *Writer) Close() error {
 	}
 
 	return nil
+}
+
+func (w *Writer) closeAfterFailure() error {
+	if w.started && w.pw != nil {
+		w.err = errors.Join(w.err, w.pw.Close())
+	}
+	w.releaseDictionaries()
+	w.pw = nil
+	return w.err
 }
 
 func (w *Writer) releaseDictionaries() {
@@ -229,7 +233,7 @@ func (w *Writer) Write(rec arrow.RecordBatch) (err error) {
 
 	enc.reset()
 	if err := enc.Encode(&data, rec); err != nil {
-		return fmt.Errorf("arrow/ipc: could not encode record to payload: %w", err)
+		return w.fail(fmt.Errorf("arrow/ipc: could not encode record to payload: %w", err))
 	}
 
 	if err := w.pw.WritePayload(data); err != nil {
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -157,7 +157,7 @@ func (w *Writer) Close() error {
 	w.pw = nil
 	w.releaseDictionaries()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


arrow/ipc/writer.go
```diff
@@ -291,10 +330,11 @@ func (w *Writer) start() error {
 	for _, data := range ps {
 		err := w.pw.WritePayload(data)
 		if err != nil {
-			return err
+			return w.fail(err)
 		}
 	}
 
+	w.started = true
```
zeroshade: we should probably move `w.started = true` above the for loop here since if even one succeeds writing then we technically started. The way this is currently written, if we succeed on one payload, but then fail on a subsequent payload `w.started` will remain `false`.

arrow/ipc/writer.go
```diff
@@ -148,24 +165,38 @@ func (w *Writer) Close() error {
...
 	}
+	w.lastWrittenDicts = nil
+}
 
-	return nil
+func (w *Writer) fail(err error) error {
+	if w.err == nil {
+		w.err = err
+	}
+	return w.err
```
zeroshade, line 184:
```suggestion
    w.err = errors.Join(w.err, err)
	return w.err
```

otherwise we can lose the error if we end up with multiple errors

arrow/ipc/writer.go
```diff
@@ -59,6 +59,17 @@ func (w *streamWriter) WritePayload(p Payload) error {
...
+	return n, err
+}
+
+func writeFull(w io.Writer, p []byte) (int, error) {
+	n, err := w.Write(p)
+	if err == nil && n != len(p) {
+		err = io.ErrShortWrite
+	}
 	return n, err
 }
```
zeroshade: we ignore the n on every call to `writeFull`, should we bother even returning n here?

zeroshade (changes requested):

arrow/ipc/writer.go
fallintoplace: I kept first-error-wins intentionally. Once w.err is set, Write and Close return immediately, so later operations are not attempted. Returning the original error also keeps the terminal error stable. I can revisit this if we expect cleanup to produce another error that should be preserved.

arrow/ipc/writer.go
```diff
@@ -136,6 +148,11 @@ func NewWriter(w io.Writer, opts ...Option) *Writer {
...
 
 func (w *Writer) Close() error {
+	if w.err != nil {
+		w.releaseDictionaries()
+		w.pw = nil
+		return w.err
```
zeroshade: in this scenario, aren't we leaking the payload writer by not calling Close on it if we have already started? we'd want to use `errors.Join(w.err, w.pw.Close())` as the error we will return

arrow/ipc/writer.go
zeroshade: that's fair, we can leave it for now

arrow/ipc/writer.go
fallintoplace: Added a regression test and pushed it in [75b990e3](https://github.com/fallintoplace/arrow-go/commit/75b990e33ba18bfc419f77d5bb25318d11e29dbb)

arrow/ipc/writer.go
```diff
@@ -136,6 +148,16 @@ func NewWriter(w io.Writer, opts ...Option) *Writer {
 }
 
 func (w *Writer) Close() error {
+	if w.err != nil {
+		if w.started && w.pw != nil {
+			if err := w.pw.Close(); err != nil {
+				w.err = errors.Join(w.err, err)
+			}
```
zeroshade:
```suggestion			
			w.err = errors.Join(w.err, err)			
```

You don't need to do the `err != nil` check here, `errors.Join` already does it

fallintoplace: Yep, you are right. Simplified it to `w.err = errors.Join(w.err, w.pw.Close())` and pushed it as 19fa153a. `go test ./arrow/ipc -count=1` passes. Thanks!

arrow/ipc/writer.go
```diff
@@ -59,9 +59,20 @@ func (w *streamWriter) WritePayload(p Payload) error {
...
+		err = io.ErrShortWrite
+	}
 	return n, err
 }
 
+func writeFull(w io.Writer, p []byte) error {
+	n, err := w.Write(p)
+	if err == nil && n != len(p) {
+		err = io.ErrShortWrite
+	}
```
zeroshade: So i was looking at https://pkg.go.dev/io#Writer and saw that Write will *always* return a non-nil error on a short write, so I don't think we need to do this at all or add the checks you've added.

arrow/ipc/writer.go
```diff
@@ -62,6 +62,11 @@ func (w *streamWriter) Write(p []byte) (int, error) {
 	return n, err
 }
 
+func writeFull(w io.Writer, p []byte) error {
+	_, err := w.Write(p)
+	return err
+}
```
zeroshade: at this point why even have the `writeFull` function instead of just putting everything back to being `_, err = w.Write(paddingBytes[:padding])` etc.?

zeroshade (approved):

PR #1048, opened 2026-07-28
fix(arrow/avro): stop reader workers before reuse

fallintoplace:
## Summary

- make decoded datum and record batch sends cancellation-aware
- wait for both reader workers before reuse or shutdown
- drain and release queued record batches during close
- synchronize reader errors and the decoded record counter
- stop workers and release the record builder on final release

## Testing

- `go test -race ./arrow/avro`

Regression coverage fills one-slot decode and record queues without consuming them, then verifies that `Close` returns and all checked allocations are released. It also reuses a partially consumed reader and verifies that no records from the previous file enter the new stream.

### DIFF
--- a/arrow/avro/loader.go
+++ b/arrow/avro/loader.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+
+	"github.com/apache/arrow-go/v18/arrow"
 )
 
 func (r *OCFReader) decodeOCFToChan() {
@@ -27,7 +29,7 @@ func (r *OCFReader) decodeOCFToChan() {
 	for {
 		select {
 		case <-r.readerCtx.Done():
-			r.err = fmt.Errorf("avro decoding cancelled, %d records read", r.avroDatumCount)
+			r.setErr(fmt.Errorf("avro decoding cancelled, %d records read", r.avroDatumCount.Load()))
 			return
 		default:
 			var datum any
@@ -36,17 +38,23 @@ func (r *OCFReader) decodeOCFToChan() {
 				if errors.Is(err, io.EOF) {
 					return
 				}
-				r.err = err
+				r.setErr(err)
 				return
 			}
-			r.avroChan <- datum
-			r.avroDatumCount++
+			select {
+			case r.avroChan <- datum:
+				r.avroDatumCount.Add(1)
+			case <-r.readerCtx.Done():
+				r.setErr(fmt.Errorf("avro decoding cancelled, %d records read", r.avroDatumCount.Load()))
+				return
+			}
 		}
 	}
 }
 
 func (r *OCFReader) recordFactory() {
 	defer close(r.recChan)
+	defer close(r.bldDone)
 	r.primed = true
 	recChunk := 0
 	switch {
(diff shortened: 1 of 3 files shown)


zeroshade (commented):
Thanks for this — I went through it carefully (including `go test -race` on your branch), and the concurrency work is solid. The design is exactly right, and it fixes real bugs that are still live on `main`:

- blocking sends on `avroChan` / `recChan` / `bldDone` leak the worker goroutines once the consumer stops (e.g. after `Close`);
- queued record batches and the `RecordBuilder` are never released;
- `r.err` / `avroDatumCount` are written by both workers and read by `Next()` / `Err()` without synchronization;
- `Reuse()` swaps the channels out from under workers that may still be running.

I traced the lifecycle and I'm satisfied on correctness:

- `Close()` can't deadlock — after cancel, the decoder's ctx-aware send returns and closes `avroChan`, `recordFactory` drains and its defers close `recChan` / `bldDone`, so `readWG.Wait()` returns before the drain-range (which then terminates on the closed channel);
- the `bldDone` send→`close` change still delivers each buffered record exactly once through `Next()` and terminates cleanly;
- no double-`Release` — `r.cur` is consumer-only and nil'd, buffered batches are drained exactly once by `Close`, the in-flight batch is released in `sendRecord`, and `Close` is idempotent via the `readCancel == nil` guard;
- `AssertSize(0)` in the three lifecycle tests confirms leak-completeness.

**The one blocker is a rebase.** This is based on the pre-migration reader, and #830 (hamba → twmb) has since landed, so the branch is currently `CONFLICTING`:

- `loader.go` — merges cleanly; the 3-way result is exactly right (the `for {}` / twmb-`Decode` loop with your ctx-aware send + `setErr` + atomic counter). Nothing to do here.
- `reader.go` — a mechanical conflict (both #830 and this PR touched imports / `Reuse` / `NewOCFReader` / the `avroDatumCount` field).
- `reader_test.go` — the real work: your four new tests use hamba's `ocf.NewEncoder(schema, &buf)`, but post-#830 the package uses `twmb/avro/ocf`, whose writer is `ocf.NewWriter(&buf, schema, ocf.WithSchema(schemaJSON))`. They'll need porting to that API (and the imports updated) or they won't compile.

Could you rebase onto current `main` and re-run `go test -race ./arrow/avro`? Apologies for the churn — that's a consequence of #830 merging ahead of you.

Two small non-blocking notes for while you're in there:

1. After a clean, fully-consumed read, `Close()` now makes `Err()` return `context.Canceled` (via `if Err()==nil { setErr(readerCtx.Err()) }`). It matches the old behavior and `Reuse` clears it, so it's harmless in practice — but you might consider not recording `context.Canceled` there so a successful read stays error-free.
2. If the underlying `io.Reader` blocks inside `Decode()`, `Close()`'s `Wait()` blocks with it. That's inherent (a blocking read isn't cancellable) and pre-existing — maybe worth a one-line comment noting it.

Happy to re-review promptly once it's rebased.

zeroshade (approved):
Re-reviewed after the rebase. The one blocker (rebase onto post-#830 main) is resolved: head is on recent main, uses twmb/avro, and the four tests are ported to `ocf.NewWriter`. Verified locally: `go test -race ./arrow/avro` passes, and the concurrency/lifecycle tests pass at `-race -count=30` with no flakiness.

The two extra commits are good improvements: `reset builder during reuse` (releases the old RecordBuilder + reinit, covered by TestOCFReaderReuseDiscardsPartialBuilderState) and `preserve first reader error` (setErr keeps the first non-nil error, so Close's context.Canceled can't clobber a real decode error). Non-blocking note #1 is effectively mitigated by the latter.

LGTM — merging once CI is green.

PR #1050, opened 2026-07-28
fix(parquet): reject short page writes

fallintoplace:
## Summary

- return `io.ErrShortWrite` when a direct dictionary or data page body write accepts fewer bytes than requested
- avoid updating page counters and indexes after an incomplete body write
- cover both direct page paths with a short-writing sink

## Testing

- `go test ./parquet/file`

### DIFF
--- a/parquet/file/page_writer.go
+++ b/parquet/file/page_writer.go
@@ -337,6 +337,9 @@ func (pw *serializedPageWriter) WriteDictionaryPage(page *DictionaryPage) (int64
 	if err != nil {
 		return 0, err
 	}
+	if written != len(data) {
+		return int64(written), io.ErrShortWrite
+	}
 
 	written += headerSize
 
@@ -399,6 +402,9 @@ func (pw *serializedPageWriter) WriteDataPage(page DataPage) (int64, error) {
 	written, err := pw.sink.Write(data)
 	if err != nil {
 		return int64(written), err
+	}
+	if written != len(data) {
+		return int64(written), io.ErrShortWrite
 	}
 	written += headerSize
 
--- a/parquet/file/page_writer.go
+++ b/parquet/file/page_writer.go
@@ -334,9 +334,6 @@ func (pw *serializedPageWriter) WriteDictionaryPage(page *DictionaryPage) (int64
 	if err != nil {
 		return 0, err
 	}
-	if written != len(data) {
-		return int64(written), io.ErrShortWrite
-	}
 
 	written += headerSize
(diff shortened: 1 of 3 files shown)


parquet/file/page_writer.go
```diff
@@ -337,8 +334,14 @@ func (pw *serializedPageWriter) WriteDictionaryPage(page *DictionaryPage) (int64
 	if err != nil {
 		return 0, err
 	}
+	if written != len(data) {
+		return int64(written), io.ErrShortWrite
+	}
```
zeroshade: this is unnecessary, Write will always return a non-nil error if `written < len(data)`

parquet/file/page_writer.go
```diff
@@ -400,7 +400,13 @@ func (pw *serializedPageWriter) WriteDataPage(page DataPage) (int64, error) {
 	if err != nil {
 		return int64(written), err
 	}
+	if written != len(data) {
+		return int64(written), io.ErrShortWrite
+	}
```
zeroshade: same comment as above, err will always be non-nil if written < len(data)

parquet/internal/thrift/helpers.go
```diff
@@ -76,12 +76,23 @@ func (t *Serializer) Serialize(msg thrift.TStruct, w io.Writer, enc encryption.E
...
 
 	if enc == nil {
-		return w.Write(b)
+		n, err := w.Write(b)
+		if err != nil {
+			return n, err
+		}
+		if n != len(b) {
+			return n, io.ErrShortWrite
+		}
```
zeroshade: same comment as above, this isn't necessary since Write will return a non-nil error on a short write

parquet/internal/thrift/helpers.go
```diff
@@ -76,12 +76,23 @@ func (t *Serializer) Serialize(msg thrift.TStruct, w io.Writer, enc encryption.E
...
 	}
 
 	var cipherBuf bytes.Buffer
 	cipherBuf.Grow(enc.CiphertextSizeDelta() + len(b))
 	enc.Encrypt(&cipherBuf, b)
+	expected := cipherBuf.Len()
 	n, err := cipherBuf.WriteTo(w)
+	if err == nil && int(n) != expected {
+		err = io.ErrShortWrite
+	}
```
zeroshade: same as above, err will be non-nil on a short write, so you don't need to do this

zeroshade (changes requested):

zeroshade (approved):

PR #1052, opened 2026-07-29
fix(arrow/compute): support casting between extension types

twuebi:
Casts are dispatched on the target type, and the cast function registered for extension targets only accepted NULL inputs. Casting an extension array to a different extension type therefore failed with "function 'cast_extension' has no kernel matching input types" even when the cast between the two storage types is well defined, while the reverse direction (extension input, concrete target) has always been handled by the generic `CastFromExtension` unwrap kernel.

Register a generic extension -> extension kernel that unwraps the input array, casts its storage to the target's storage type honoring the caller's cast options, and rewraps the result as the target extension type. Re-typing between extension types that share a storage type is zero-copy since the inner storage cast is then a no-op.

Also correct the `GetCommonCastKernels` doc comment, which claimed to return kernels for casting from extension types; it only ever returned the cast-from-null kernel.

### DIFF
--- a/arrow/compute/cast.go
+++ b/arrow/compute/cast.go
@@ -555,6 +555,37 @@ func CastFromExtension(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.Exec
 	return nil
 }
 
+// CastToExtension casts an extension array to another extension type by
+// casting the input's storage to the target's storage type and wrapping
+// the result as the target extension type. Re-typing between extension
+// types that share a storage type is zero-copy since the inner storage
+// cast is then a no-op.
+func CastToExtension(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResult) error {
+	opts := ctx.State.(kernels.CastState)
+
+	toType, ok := opts.ToType.(arrow.ExtensionType)
+	if !ok {
+		return fmt.Errorf("%w: cast target %s is not an extension type", arrow.ErrInvalid, opts.ToType)
+	}
+
+	arr := batch.Values[0].Array.MakeArray().(array.ExtensionArray)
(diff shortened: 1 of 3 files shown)


zeroshade:
> Register a generic extension -> extension kernel that unwraps the input array, casts its storage to the target's storage type honoring the caller's cast options, and rewraps the result as the target extension type. Re-typing between extension types that share a storage type is zero-copy since the inner storage cast is then a no-op.

This isn't technically valid at all since the semantics of the different extension types might be different even though the underlying storage type is compatible. For example, consider a cast from UUID (fixed-size-binary [16]) to JSON (String):

- If we're just unwrapping the input and casting the storage that means we do a fsb[16] -> utf8 cast
- If the bytes for the UUID *happen* to be valid utf8, it'll create a gibberish string rather than create a string version of the UUID
- Even if it created a valid UUID string, that wouldn't be valid JSON and thus would create an invalid JSON extension array.

Essentially, casting between extension types MUST take into account the semantics, so we can't create a generic extension type cast like this which just performs a cast of the underlying storage and wrapping it in the new extension. What is the specific case you're trying to support with this generic kernel? If it's casting between extension types that are defined in arrow-go, we can add specific kernels for those types in here. Otherwise, it might make more sense to just implement and register the kernels you need inside the package you want to use it in.

twuebi:
Oh, yes, you're right. Storage cannot decide this, going to close this.

Tried to fix an issue that popped up with iceberg-go geo type casts, where the same geoarrow.wkb type with different CRS metadata (SRID vs authority code) on the file and table sides. Just different labels for the same thing, [apache/iceberg-go#1599](https://github.com/apache/iceberg-go/issues/1599) solved this on the correct layer already, trying to push it into this layer was wrong.

PR #1053, opened 2026-07-29
fix(parquet/file): close file on OpenParquetFile errors

fallintoplace:
### Rationale for this change

`OpenParquetFile` opens a file or mmap before constructing the reader. If initialization fails while parsing metadata, the function returned without closing that owned source.

Require the internally opened source to implement `ReaderAt`, `Seeker`, and `Closer`, then close it when `NewParquetReader` fails. Preserve a simultaneous close failure with `errors.Join`.

### Are these changes tested?

Yes. A close-tracking regression test verifies that malformed input closes the owned source, and the mmap signature cross-compiles on Windows.

`go test ./parquet/file -run TestOpenParquetFileClosesSourceOnError`

Invalid files still return an error, but their file descriptor or mmap is released promptly.

### DIFF
--- a/parquet/file/file_reader.go
+++ b/parquet/file/file_reader.go
@@ -103,10 +103,7 @@ func OpenParquetFile(filename string, memoryMap bool, opts ...ReadOption) (*Read
 func newParquetReaderFromFile(source readerAtSeekerCloser, opts ...ReadOption) (*Reader, error) {
 	rdr, err := NewParquetReader(source, opts...)
 	if err != nil {
-		if closeErr := source.Close(); closeErr != nil {
-			return nil, errors.Join(err, closeErr)
-		}
-		return nil, err
+		return nil, errors.Join(err, source.Close())
 	}
 	return rdr, nil
 }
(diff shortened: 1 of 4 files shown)


parquet/file/file_reader.go
```diff
@@ -92,7 +97,18 @@ func OpenParquetFile(filename string, memoryMap bool, opts ...ReadOption) (*Read
...
+	return newParquetReaderFromFile(source, opts...)
+}
+
+func newParquetReaderFromFile(source readerAtSeekerCloser, opts ...ReadOption) (*Reader, error) {
+	rdr, err := NewParquetReader(source, opts...)
+	if err != nil {
+		if closeErr := source.Close(); closeErr != nil {
+			return nil, errors.Join(err, closeErr)
+		}
+		return nil, err
```
zeroshade: Can't we simplify to this?

```suggestion
		return nil, errors.Join(err, source.Close())
```

zeroshade (changes requested):

zeroshade (approved):

PR #1054, opened 2026-07-29
fix(arrow/ipc): reject non-dictionary initial messages

fallintoplace:
### Rationale for this change

When an IPC schema declares dictionaries, the reader expects dictionary batches before records. It detected an unexpected message but continued into `readDictionary`, which could replace the useful protocol error and consume the wrong message.

Stop immediately, mark the reader done, and preserve the expected-dictionary error.

### Are these changes tested?

Yes. A regression test provides a dictionary schema followed directly by a record batch.

`go test ./arrow/ipc -run TestReaderRejectsRecordBatchBeforeInitialDictionary`

Malformed streams now fail deterministically with the specific dictionary-ordering error.

### DIFF
--- a/arrow/ipc/reader.go
+++ b/arrow/ipc/reader.go
@@ -223,7 +223,6 @@ func (r *Reader) getInitialDicts() bool {
 		}
 
 		if msg.Type() != MessageDictionaryBatch {
-			r.done = true
 			r.err = fmt.Errorf("arrow/ipc: IPC stream did not have the expected (%d) dictionaries at the start of the stream", numDicts)
 			return false
 		}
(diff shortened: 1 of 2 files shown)


arrow/ipc/reader.go
```diff
@@ -223,7 +223,9 @@ func (r *Reader) getInitialDicts() bool {
 		}
 
 		if msg.Type() != MessageDictionaryBatch {
+			r.done = true
```
zeroshade: do we need to add the `r.done` if we're setting `r.err`? I believe we check `r.done || r.err != nil` right?

zeroshade (approved):

PR #1059, opened 2026-07-29
fix(arrow/csv): release Reader record builders

fallintoplace:
### Rationale for this change

Every CSV reader owns a `RecordBuilder`, but final `Reader.Release` did not release it. Header processing with an explicit schema also replaced the initial builder without releasing it, leaking two builders on that path.

Release the old builder before header-driven replacement, and release the active builder when the reader reference count reaches zero.

### Are these changes tested?

Yes. The regression test uses a supported custom extension builder that reserves allocator-backed storage when constructed and counts releases. It covers explicit-schema readers with and without headers, proving that one builder is released on the direct path and both builders are released on the header-replacement path.

`go test ./arrow/csv -run TestReaderReleaseFreesRecordBuilders`

No API changes. Releasing a CSV reader now frees every builder it owns.

### DIFF
--- a/arrow/csv/reader_test.go
+++ b/arrow/csv/reader_test.go
@@ -496,6 +496,22 @@ rec[2]["date64"]: [(null)]
 	}
 }
 
+func TestReaderReleaseFreesRecordBuilder(t *testing.T) {
+	for _, withHeader := range []bool{false, true} {
+		t.Run(fmt.Sprintf("header=%t", withHeader), func(t *testing.T) {
+			mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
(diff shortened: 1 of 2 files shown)


arrow/csv/reader_test.go
```diff
@@ -496,6 +496,22 @@ rec[2]["date64"]: [(null)]
 	}
 }
 
+func TestReaderReleaseFreesRecordBuilder(t *testing.T) {
```
serramatutu: I reverted your changes and ran the test on top of main. It passes without reproducing any bug.

Image

serramatutu (changes requested):

zeroshade (approved):

PR #1060, opened 2026-07-29
fix(parquet/pqarrow): release extension arrays after chunking

fallintoplace:
### Rationale for this change

The extension reader creates arrays and passes them to NewChunked, which retains each chunk. The original references were not released, and partially built chunks could also leak when a later chunk failed.

Release each extension array after NewChunked takes ownership, and release already-built chunks while unwinding an error.

### Are these changes tested?

- `go test ./parquet/pqarrow -run TestExtensionReaderBuildArrayReleasesPartialChunksOnPanic`

### DIFF
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -135,14 +135,16 @@ func (er *extensionReader) BuildArray(boundedLen int64) (*arrow.Chunked, error) 
 	extType := er.fieldWithExt.Type.(arrow.ExtensionType)
 
 	newChunks := make([]arrow.Array, len(chkd.Chunks()))
+	defer func() {
+		for _, c := range newChunks {
+			if c != nil {
+				c.Release()
+			}
+		}
+	}()
 	for i, c := range chkd.Chunks() {
 		newChunks[i] = array.NewExtensionArrayWithStorage(extType, c)
 	}
-	defer func() {
-		for _, c := range newChunks {
-			c.Release()
-		}
-	}()
 
 	return arrow.NewChunked(extType, newChunks), nil
 }
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -138,6 +138,11 @@ func (er *extensionReader) BuildArray(boundedLen int64) (*arrow.Chunked, error) 
 	for i, c := range chkd.Chunks() {
 		newChunks[i] = array.NewExtensionArrayWithStorage(extType, c)
 	}
+	defer func() {
+		for _, c := range newChunks {
+			c.Release()
+		}
+	}()
 
 	return arrow.NewChunked(extType, newChunks), nil
 }
--- a/parquet/pqarrow/file_reader_extension_test.go
+++ b/parquet/pqarrow/file_reader_extension_test.go
@@ -35,6 +35,35 @@ type mismatchingExtensionArray struct {
 	array.ExtensionArrayBase
 }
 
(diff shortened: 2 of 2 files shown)


parquet/pqarrow/file_reader.go
```diff
@@ -135,6 +135,13 @@ func (er *extensionReader) BuildArray(boundedLen int64) (*arrow.Chunked, error)
 	extType := er.fieldWithExt.Type.(arrow.ExtensionType)
 
 	newChunks := make([]arrow.Array, len(chkd.Chunks()))
+	defer func() {
```
serramatutu, line 138: Can we add a unit test to repro this using the checked allocator?

serramatutu (changes requested):

zeroshade (commented):
Thanks @fallintoplace — the fix itself is correct. I verified both leaks it addresses against a checked allocator:

- **Success (normal read) path** — every output chunk leaks one extension-array reference. Since `NewExtensionArrayWithStorage` shares the storage buffers via `NewData` (which retains them), those buffers are never freed after the caller releases the returned `Chunked`. Reverting the fix leaks ~256 B on a 2-chunk read; with the fix it returns to 0.
- **Partial/panic path** — covered by the new `TestExtensionReaderBuildArrayReleasesPartialChunksOnPanic` (leaks 128 B without the fix, 0 with it).

One request before merge: the added test only exercises the **panic** path. The headline leak here is the **success** path — the common case for every extension-column read — and it isn't covered by this test or any existing one. Could you add a success-path checked-allocator test? It's small; you can reuse the `chunkedColumnReader` fake with a stable extension type, e.g.:

```go
type stableExtType struct{ arrow.ExtensionBase }
type stableExtArray struct{ array.ExtensionArrayBase }

func (*stableExtType) StorageType() arrow.DataType { return arrow.PrimitiveTypes.Int32 }
func (*stableExtType) ArrayType() reflect.Type     { return reflect.TypeFor[stableExtArray]() }
func (*stableExtType) ExtensionName() string       { return "test.stable" }
func (*stableExtType) Serialize() string           { return "" }
func (*stableExtType) ExtensionEquals(o arrow.ExtensionType) bool { _, ok := o.(*stableExtType); return ok }
func (*stableExtType) Deserialize(arrow.DataType, string) (arrow.ExtensionType, error) {
	return &stableExtType{arrow.ExtensionBase{Storage: arrow.PrimitiveTypes.Int32}}, nil
}

func TestExtensionReaderBuildArrayReleasesChunks(t *testing.T) {
	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
	defer mem.AssertSize(t, 0)

	b := array.NewInt32Builder(mem)
	b.Append(1)
	first := b.NewInt32Array()
	b.Append(2)
	second := b.NewInt32Array()
	b.Release()
	chunks := arrow.NewChunked(arrow.PrimitiveTypes.Int32, []arrow.Array{first, second})
	first.Release()
	second.Release()

	extType := &stableExtType{arrow.ExtensionBase{Storage: arrow.PrimitiveTypes.Int32}}
	r := extensionReader{colReaderImpl: &chunkedColumnReader{chunks: chunks}, fieldWithExt: arrow.Field{Name: "extension", Type: extType}}

	out, err := r.BuildArray(0)
	require.NoError(t, err)
	out.Release()
}
```

I confirmed this fails on `main` (leaks 256 B) and passes with your fix. With a success-path test added I'm happy to merge.

zeroshade (approved):

PR #1061, opened 2026-07-29
fix(parquet/pqarrow): release partially constructed field readers

fallintoplace:
### Rationale for this change

`GetFieldReaders` constructs readers concurrently. If one fails after another succeeds, the successful reader remains in a private slice and is never returned, so the caller cannot release it.

When construction fails, release every non-nil reader that completed before returning the original error.

### Are these changes tested?

Yes. The regression test injects a read failure at the second column chunk after the first reader has been constructed. Without the cleanup, the checked allocator reports the first page reader's retained buffer; with the fix, allocator usage returns to zero.

`go test ./parquet/pqarrow -run TestGetFieldReadersReleasesPartialReadersOnError`

No API changes. Failed multi-column reader construction now cleans up partial results.

### DIFF
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -262,6 +262,11 @@ func (fr *FileReader) GetFieldReaders(ctx context.Context, colIndices, rowGroups
 		})
 	}
 	if err = g.Wait(); err != nil {
+		for _, rdr := range out {
+			if rdr != nil {
+				rdr.Release()
+			}
+		}
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


parquet/pqarrow/file_reader.go
```diff
@@ -262,6 +262,11 @@ func (fr *FileReader) GetFieldReaders(ctx context.Context, colIndices, rowGroups
 		})
 	}
 	if err = g.Wait(); err != nil {
+		for _, rdr := range out {
```
serramatutu, line 265: Can we add a unit test to repro the bug?

serramatutu (changes requested):

zeroshade (approved):

PR #1063, opened 2026-07-29
fix(parquet/variant): validate metadata offset tables

fallintoplace:
## What changed

Validate the complete Variant metadata offset table before allocating the key slice or reading offsets. The parser now also rejects a nonzero initial offset, decreasing offsets, and string offsets beyond the input.

## Why

The previous size check omitted the final offset entry. Truncated input could therefore reach an out-of-bounds slice, while decreasing offsets could underflow during key-size calculation. A large untrusted dictionary count was also used to allocate before checking whether the offset table could fit in the input.

Malformed metadata now consistently returns an error wrapping `ErrInvalidMetadata`.

## Testing

- `go test ./parquet/variant`

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -183,32 +183,38 @@ func (m *Metadata) loadDictionary(offsetSz uint8) error {
 	}
 
 	dictSize := readLEU32(m.data[hdrSizeBytes : hdrSizeBytes+offsetSz])
-	m.keys = make([][]byte, dictSize)
-
 	if dictSize == 0 {
+		m.keys = nil
 		return nil
 	}
 
-	// first offset is always 0
-	offsetStart, offsetPos := uint32(0), hdrSizeBytes+offsetSz
-	valuesStart := hdrSizeBytes + (dictSize+2)*uint32(offsetSz)
-	if hdrSizeBytes+int(dictSize+1)*int(offsetSz) > len(m.data) {
-		return fmt.Errorf("%w: offset out of range: %d > %d",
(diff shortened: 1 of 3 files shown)


parquet/variant/variant.go
```diff
@@ -183,32 +183,38 @@ func (m *Metadata) loadDictionary(offsetSz uint8) error {
...
 	}
 
-	// first offset is always 0
-	offsetStart, offsetPos := uint32(0), hdrSizeBytes+offsetSz
-	valuesStart := hdrSizeBytes + (dictSize+2)*uint32(offsetSz)
-	if hdrSizeBytes+int(dictSize+1)*int(offsetSz) > len(m.data) {
-		return fmt.Errorf("%w: offset out of range: %d > %d",
-			ErrInvalidMetadata, (dictSize+hdrSizeBytes)*uint32(offsetSz), len(m.data))
+	valuesStart := uint64(hdrSizeBytes) + (uint64(dictSize)+2)*uint64(offsetSz)
+	if valuesStart > uint64(len(m.data)) {
```
serramatutu, line 192: Shouldn't this be `>=`?

serramatutu (commented):
Ran the test locally before and after the fix and it all passes. LGTM to me except for a the small mistake `>=` instead of `>` when checking the range.

parquet/variant/variant.go
fallintoplace: valuesStart == len(m.data) is valid when the offset table ends exactly at the input boundary and the dictionary’s strings contain zero total bytes.

Changing > to >= would incorrectly reject that valid encoding. The subsequent check already rejects any nonzero final offset because valuesStart + end would exceed the input length.

zeroshade (approved):

PR #1064, opened 2026-07-29
fix(arrow/array): make dictionary suffix null indexes relative

fallintoplace:
## What changed

Clear the validity bit using the null index relative to `startOffset` when exporting a dictionary suffix.

## Why

`GetDictArrayData` builds a new dictionary whose bitmap starts at the requested offset, but it previously cleared the absolute memo-table index. For nonzero offsets this could mark the wrong dictionary entry valid while still reporting one null.

The regression test exports a suffix containing a null and verifies that the corresponding relative entry is null.

## Testing

- `go test ./arrow/array`


serramatutu (approved):
Ran the test before/after the fix, it does repro the issue. LGTM.

zeroshade (approved):

PR #1065, opened 2026-07-29
fix(arrow/array): always release JSON reader builder

fallintoplace:
## What changed

Release the JSON reader record builder whenever the final reader reference is released, independently of whether a current record batch exists.

## Why

The builder release was nested under the current-record check. A decode error can leave `cur` nil while the builder still owns buffers from values parsed before the error, so releasing the reader did not free those buffers.

The regression test parses an object far enough to allocate builder storage, triggers a duplicate-key error before a record batch is produced, and verifies with a checked allocator that `Release` frees the partial builder state.

## Testing

- `go test ./arrow/array`
- `go test -race ./arrow/array -run TestJSONReaderReleaseBuilderAfterPartialReadError`

### DIFF
--- a/arrow/array/json_reader_test.go
+++ b/arrow/array/json_reader_test.go
@@ -95,6 +95,38 @@ func TestJSONReaderAll(t *testing.T) {
 	assert.False(t, rdr.Next())
 }
 
+func TestJSONReaderReleaseBuilder(t *testing.T) {
+	schema := arrow.NewSchema([]arrow.Field{
+		{Name: "value", Type: arrow.PrimitiveTypes.Int64},
+	}, nil)
+
+	tests := []struct {
+		name    string
+		input   string
+		consume bool
+	}{
+		{name: "before reading", input: `{"value": 1}`},
+		{name: "after empty input", consume: true},
+		{name: "after reaching EOF", input: `{"value": 1}`, consume: true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+			rdr := array.NewJSONReader(strings.NewReader(tt.input), schema, array.WithAllocator(mem))
+
+			if tt.consume {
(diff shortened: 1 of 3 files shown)


arrow/array/json_reader_test.go
```diff
@@ -95,6 +95,38 @@ func TestJSONReaderAll(t *testing.T) {
 	assert.False(t, rdr.Next())
 }
 
+func TestJSONReaderReleaseBuilder(t *testing.T) {
```
serramatutu: I reverted your change to `json_reader.go` and ran the test you introduced. The test did not break, i.e it does not reproduce any memory leak.

Image

serramatutu (changes requested):

zeroshade (approved):
The fix is correct, and I want to close the loop on @serramatutu's earlier concern, since it was valid at the time.

At commit `1d4dfcf` the original test did not reproduce the leak — reverting `json_reader.go` and running it passed, exactly as @serramatutu found. The subsequent commits (`286afe9b`, `e5b98623` "replace non-reproducing lifecycle cases", `12cc58fc`) reworked it into the duplicate-key case now in `json_reader_release_error_test.go`, and that version *does* reproduce it. I re-ran the same experiment on the current head (`12cc58fc`):

- with the fix → `PASS`
- reverting `json_reader.go` to the pre-fix nested `Release()` → `FAIL: LEAK of 256 bytes` + `LEAK of 64 bytes`

Mechanism: `next1()` returns `false` when `readNext()`/`UnmarshalOne` errors on the duplicate key, but `value:1` has already been appended, so the builder holds buffers while `cur` stays nil. The old `Release()` gated `bldr.Release()` behind `cur != nil`, leaking those buffers; moving it out fixes it, with no change to the success path.

Verified locally: `go test ./arrow/array` and `go test -race ./arrow/array -run TestJSONReaderReleaseBuilderAfterPartialReadError` both pass; `go vet` clean.

@serramatutu thanks for catching the original non-reproducing test — the reworked one holds up. LGTM.

PR #1066, opened 2026-07-29
fix(arrow/flight/flightsql): avoid double-retaining parameter readers

fallintoplace:
## What changed

Remove the duplicate retain in `PreparedStatement.SetRecordReader`.

## Why

The setter retained the supplied reader twice, while parameter replacement and statement close released it only once. This left one reference permanently owned by the statement.

The regression test binds an allocator-backed record reader, drops the caller-owned references, then replaces the binding. The statement's single release must free the record buffers; the previous double retain leaves them allocated.

## Testing

- `go test ./arrow/flight/flightsql`
- `go test -race ./arrow/flight/flightsql -run TestPreparedStatementReleasesRecordReaderBindingOnce`

### DIFF
--- a/arrow/flight/flightsql/client_test.go
+++ b/arrow/flight/flightsql/client_test.go
@@ -43,6 +43,21 @@ type mockGrpcClientStream struct {
 	mock.Mock
 }
 
+type referenceCountingReader struct {
+	array.RecordReader
+	refs int
+}
+
+func (r *referenceCountingReader) Retain() {
+	r.refs++
+	r.RecordReader.Retain()
+}
+
+func (r *referenceCountingReader) Release() {
+	r.refs--
+	r.RecordReader.Release()
+}
+
 func (m *mockGrpcClientStream) Header() (metadata.MD, error)  { panic("unimplemented") }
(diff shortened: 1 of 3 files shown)


arrow/flight/flightsql/client_test.go
```diff
@@ -43,6 +43,21 @@ type mockGrpcClientStream struct {
 	mock.Mock
 }
 
+type referenceCountingReader struct {
```
serramatutu: This ref counting mechanism is used only in one place and is not a widespread architectural pattern. To me this is a signal that this is a workaround for a bug instead of an actual good fix.

I think the proper solution for readers is ensuring we do lifecycle management right (i.e call `Retain()` in the right places) instead of adding ref counting.

serramatutu (changes requested):
I don't think ref counting is the way for this. Also, can you add a test to repro the specific bug you're trying to fix?

arrow/flight/flightsql/client_test.go
zeroshade: Agreed, we should be calling Retain and Release in the correct places

zeroshade (approved):

PR #1075, opened 2026-07-30
fix(parquet): row-align V1 pages with offset indexes

### LINKED ISSUE #887: parquet: V1 pages ignore row boundaries when an OffsetIndex is present
### Describe the bug

The Parquet spec requires data pages to begin and end at row boundaries whenever an OffsetIndex is present, regardless of the page header version:

> If an OffsetIndex is present, a page must begin at a row boundary.
> Page indexes require pages to start and end at row boundaries, regardless of which page header is used.

`columnWriter.doBatches` only row-aligns batches for DataPageV2. For all V1 writes it falls back to the fixed-size `doBatches`, even when `PageIndexEnabled` is set for the column. When a repeated column is written as V1 with page index enabled and pages are small enough to split, the resulting OffsetIndex references pages that begin mid-row, i.e. an invalid OffsetIndex.

This predates and is separate from #883 (which fixed the same invariant for DataPageV2 spaced writes); it also affects the non-spaced `WriteBatch` path. It surfaced during review of #883 (thanks @zeroshade for flagging it).

### Reproduction

Writing a repeated `int32` column as V1 + `WithPageIndexEnabled(true)` with a batch size that is not a multiple of the row width and a small data page size produces **12 of 16** data pages beginning mid-row (first repetition level != 0):

```
V1+PageIndex: v1pages=16 midrow=12
```

(1000 rows x 4 values, `WithBatchSize(7)`, `WithDataPageSize(1024)`, `WithDataPageVersion(DataPageV1)`, `WithPageIndexEnabled(true)`.)

### Suggested direction

Route through the row-aware batching in `doBatches` when `DataPageVersion() == DataPageV2 || props.PageIndexEnabledFor(col.Path())`, rather than only for V2. That reuses the alignment path already covered by the #883 regression tests; the V1 + page-index case would want its own boundary test.

### Component(s)

Parquet

---
Filed by Truffle, an AI software agent (https://truffle.ghostwright.dev).

fallintoplace:
Fixes #887

## Problem

Parquet pages referenced by an OffsetIndex must begin at row boundaries. The writer already enforced this for repeated DataPageV2 columns, but repeated DataPageV1 columns still used fixed-size batching when page indexes were enabled. A logical row could consequently span two indexed V1 pages.

## Change

Reuse the existing repetition-level-aware batching for repeated V1 columns when page indexing is enabled. The same condition is applied to the generated variable-length writer paths. V1 writes without page indexes retain the existing fixed-size batching behavior.

## Coverage

The regression test uses indexed V1 pages, repeated three-value rows, and a batch size that does not divide the row width. It closes the writer to verify the final buffered page as well as intermediate pages.

## Validation

`go test ./parquet/file -run 'TestDataPageV1OffsetIndexRowBoundaries|TestDataPageV2RowBoundaries' -count=1`

### DIFF
--- a/parquet/file/column_writer.go
+++ b/parquet/file/column_writer.go
@@ -656,14 +656,9 @@ func (w *columnWriter) Close() (err error) {
 
 func (w *columnWriter) doBatches(total int64, repLevels []int16, action func(offset, batch int64)) {
 	batchSize := w.props.WriteBatchSize()
-	// if we're writing V1 data pages, have no replevels or the max replevel is 0 then just
-	// use the regular doBatches function.
-	//
-	// The spec also requires row-aligned pages for V1 when an OffsetIndex is
-	// present (PageIndexEnabled). That gap is on the WriteBatch path and is
-	// out of scope for the DataPageV2 row alignment here; it is tracked in
-	// https://github.com/apache/arrow-go/issues/887.
-	if w.props.DataPageVersion() == parquet.DataPageV1 || repLevels == nil || w.descr.MaxRepetitionLevel() == 0 {
+	requiresRowAlignment := w.props.DataPageVersion() != parquet.DataPageV1 ||
+		w.props.PageIndexEnabledFor(w.descr.Path())
+	if !requiresRowAlignment || repLevels == nil || w.descr.MaxRepetitionLevel() == 0 {
 		doBatches(total, batchSize, action)
 		return
 	}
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM — this is exactly the fix #887 called for, and it's clean.

- `doBatches` now routes through the row-aware path when `DataPageVersion() != V1 || PageIndexEnabledFor(path)`, and the generated `WriteBatch`/`WriteBatchSpacedWithError` mirror it with correctly-parenthesized precedence. V1 without a page index keeps the existing fixed-size batching, so there's no behavior change outside the indexed case.
- Generated-code hygiene checks out: `go generate ./parquet/file/` reproduces the committed `column_writer_types.gen.go` with zero diff, so the template and generated file are in sync.
- The regression test is a real reproduction: reverting the writer change makes it write a mid-row V1 page and the mock rejects it (`NumValues%3 != 0`); with the fix every page is row-aligned, and `Close()` covers the final buffered page.

Verified locally: `go test ./parquet/file` passes (with `PARQUET_TEST_DATA` set), `go vet` clean, no regressions on the V2 path.

PR #1079, opened 2026-08-01
fix(arrow): isolate schemas created by AddField

fallintoplace:
### Rationale for this change

Schema.AddField used append directly on the parent field slice when inserting at the end. If the parent had spare capacity, two schemas derived from it shared the same appended slot. Creating the second sibling then silently replaced the first sibling's field while its name index still described the original field.

For example, a derived schema created with field "first" can later report field "second" after another AddField call on the same parent.

Always allocate the result field slice before inserting the new field. Parent and sibling schemas no longer share writable backing storage. Field ordering, metadata, endianness, and index validation are unchanged.

### Are these changes tested?

Yes. The regression test builds a parent with spare field capacity, derives two sibling schemas, and verifies that creating the second sibling does not mutate the first.

- go test ./arrow
- go test -race ./arrow -run '^TestSchemaAddFieldDoesNotAliasParentStorage$' -count=1

No API changes. Schemas returned by AddField now remain immutable when their parent is reused.

### DIFF
--- a/arrow/schema.go
+++ b/arrow/schema.go
@@ -268,15 +268,10 @@ func (s *Schema) AddField(i int, field Field) (*Schema, error) {
 		return nil, fmt.Errorf("arrow: invalid field index %d", i)
 	}
 
-	var fields []Field
-	if i == len(s.fields) {
-		fields = append(s.fields, field)
-	} else {
-		fields = make([]Field, len(s.fields)+1)
-		copy(fields[:i], s.fields[:i])
-		fields[i] = field
-		copy(fields[i+1:], s.fields[i:])
-	}
+	fields := make([]Field, len(s.fields)+1)
+	copy(fields[:i], s.fields[:i])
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified the fix and the regression test.

- Root cause confirmed: `AddField` at the end used `append(s.fields, field)`, so when the parent's fields slice had spare capacity two derived schemas aliased the same appended slot. Reachable via the public API (append-growth yields spare capacity).
- Fix is correct: always `make([]Field, len+1)` + copy — no shared backing, and derived schemas are exact-capacity so the bug can't recur downstream. Ordering/metadata/endianness/index unchanged.
- Test has teeth: on `main` with the fix reverted it fails (`got="second", want="first"`); passes with the fix. Also passes under `-race`.
- No regressions: `go vet`, all `TestSchema*`, and full `./arrow` are green.

LGTM.

PR #1080, opened 2026-08-01
fix(parquet/variant): preserve metadata keys when cloning

fallintoplace:
### Rationale for this change

Cloning Variant metadata copies the key slice headers, but the keys can still point into the original metadata buffer. Mutating the source bytes can then change lookups on the clone.

Clone the metadata bytes and rebuild the parsed key slices against the cloned buffer so KeyAt and IdFor do not depend on the source.

### Are these changes tested?

- `go test ./parquet/variant`
- Added coverage for source-byte mutation and cloned key lookups.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -168,15 +168,18 @@ func NewMetadata(data []byte) (Metadata, error) {
 
 // Clone creates a deep copy of the metadata.
 func (m *Metadata) Clone() Metadata {
-	keys := make([][]byte, len(m.keys))
-	for i, key := range m.keys {
-		keys[i] = bytes.Clone(key)
+	clone := Metadata{data: bytes.Clone(m.data)}
+	if len(clone.data) > 0 && len(m.keys) > 0 {
+		if err := clone.loadDictionary(clone.OffsetSize()); err == nil {
+			return clone
+		}
 	}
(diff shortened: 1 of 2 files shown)


zeroshade:
Nothing should be mutating the keys or the data after it's been constructed like this, so is there really a need to do a full deep copy when it *should* be safe?

fallintoplace:
That is the intended usage, but the API currently does not enforce or document that ownership constraint. NewMetadata retains the caller-provided []byte, and Bytes() also exposes it directly, while Clone() is documented as creating a deep copy.

For metadata parsed by NewMetadata, the key slices point into the original input buffer. Reusing or overwriting that input after cloning can therefore change KeyAt and IdFor on the clone even though the clone’s serialized data was copied.

The normal path in this change does not copy the key payloads a second time. It keeps the existing bytes.Clone(m.data) and rebuilds the key slices to point into that cloned buffer. I can update the test to mutate the original input slice rather than calling metadata.Bytes() so the ownership case is clearer. @zeroshade

zeroshade:
> I can update the test to mutate the original input slice rather than calling metadata.Bytes() so the ownership case is clearer. @zeroshade

Yea, let's do that

zeroshade (approved):

PR #1083, opened 2026-08-01
fix(arrow/util): release protobuf enum dictionaries

fallintoplace:
## What changed

Release the temporary string builder and array used to initialize protobuf enum dictionaries.

## Why

`InsertStringDictValues` copies values from its input. The temporary builder and array remained owned by the conversion path and leaked after each enum append.

## Testing

- `go test ./arrow/util`
- Added checked-allocator coverage for enum conversion

### DIFF
--- a/arrow/util/protobuf_reflect.go
+++ b/arrow/util/protobuf_reflect.go
@@ -359,6 +359,7 @@ func (pdr protobufDictReflection) getDataType() arrow.DataType {
 func (pdr protobufDictReflection) getDictValues(mem memory.Allocator) arrow.Array {
 	enumValues := pdr.descriptor.Enum().Values()
 	bldr := array.NewStringBuilder(mem)
+	defer bldr.Release()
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM — clean, minimal leak fix.

- `getDictValues` now `defer bldr.Release()`s the temp string builder; since the `defer` runs after `bldr.NewArray()` computes the return value, the returned array keeps the transferred buffers and the builder's own allocation is freed.
- The DICTIONARY path releases `dictValues` after `InsertStringDictValues` copies from it, and before the `err` check, so the error path doesn't leak either.
- The regression test is a real reproduction: with the fix it passes, and reverting `protobuf_reflect.go` makes the checked allocator report `LEAK of 64 bytes` for the temp builder/array.

Verified locally: `go test ./arrow/util` passes with no regressions. (The `go vet` note on `protobuf_reflect.go:199` about copying a protobuf `Any` lock is pre-existing and unrelated to this change.)

PR #1085, opened 2026-08-01
test(arrow/util): cover zero-field protobuf records

fallintoplace:
### Rationale for this change

The zero-field record behavior is already present on main. When protobuf reflection produces a schema with no fields, one protobuf message still becomes one record with zero columns.

Add regression coverage for empty protobuf messages and exclusion policies that remove every field. This PR is test-only.

### Are these changes tested?

- `go test ./arrow/util`

No. This PR only adds regression coverage. The zero-field record behavior is already present on main.

### DIFF
--- a/arrow/util/protobuf_reflect_test.go
+++ b/arrow/util/protobuf_reflect_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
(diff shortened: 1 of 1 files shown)


zeroshade:
why 1 row instead of zero rows?

fallintoplace:
Record represents the single protobuf message passed to the reflection, so it should still produce one row when that message has no included fields. With zero columns, there are no arrays left to infer the row count, so I pass 1 explicitly. Returning zero rows would represent no records rather than one empty protobuf message.

zeroshade (approved):
that's fair. this LGTM

PR #1086, opened 2026-08-01
fix(arrow/util): handle invalid protobuf Any values

fallintoplace:
### Rationale for this change

An unknown protobuf Any type URL or invalid payload can leave the unpacked message nil. The schema conversion path then dereferences it instead of preserving the physical Any fields.

Handle unpacking errors and fall back to reflecting type_url and value. Avoid copying protobuf message state during conversion.

### Are these changes tested?

- `go test ./arrow/util`
- go vet ./arrow/util
- Added malformed Any schema and record coverage.


zeroshade (approved):
Just need to fix the merge conflicts

PR #1088, opened 2026-08-01
fix(arrow/util): return protobuf record conversion errors

fallintoplace:
### Rationale for this change

Record discards errors from AppendValueOrNull. An unknown protobuf enum value can then panic or leave builders with mismatched lengths instead of reporting the conversion failure.

Add RecordWithError with field context, reject undefined enum numbers safely, and clean up temporary builders and dictionary values on both success and failure. Keep Record with its existing panic-on-error behavior for compatibility.

### Are these changes tested?

- `go test ./arrow/util`
- Added checked-allocator coverage for undefined enum values and conversion cleanup.

Yes. RecordWithError provides an error-returning path while the existing panic-on-error Record API remains available.

### DIFF
--- a/arrow/util/protobuf_reflect.go
+++ b/arrow/util/protobuf_reflect.go
@@ -677,7 +677,7 @@ func (msg ProtobufMessageReflection) RecordWithError(mem memory.Allocator) (arro
 
 	schema := msg.Schema()
 	if schema.NumFields() == 0 {
-		return array.NewRecordBatch(schema, nil, 1)
+		return array.NewRecordBatch(schema, nil, 1), nil
 	}
 
 	recordBuilder := array.NewRecordBuilder(mem, schema)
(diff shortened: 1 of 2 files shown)


zeroshade (commented):
Thanks @fallintoplace — the error reporting here is a real improvement, and I confirmed all three fixes are still needed on `main`:

- `Record` still discards the `AppendValueOrNull` error (`protobuf_reflect.go:671`), so `RecordWithError` is genuinely new behavior, not a duplicate.
- The `EnumValue`/string path (`:773`) and the dictionary path (`:822`) both still do `…ByNumber(n).Name()` with no nil check, so an undefined enum number panics. Your `enumValue == nil` guards fix that, and the tests driving `999` through both handler paths (with `AssertSize(0)` on the error path) are exactly the right coverage.

The blocker is that the branch is currently `CONFLICTING`. It's based on a commit from before the sibling protobuf fixes landed, and it re-does two things that are now already on `main`:

- `getDictValues`' `defer bldr.Release()` and the `dictValues.Release()` in the DICTIONARY case — already there from #1083.
- The `recordBuilder.NewRecordBatch()` + `defer recordBuilder.Release()` + zero-field-schema handling — already there from #1084.

So the rebase should be small — keep what's unique to this PR and drop the rest:

1. Split `main`'s existing `Record` body into `RecordWithError`, returning `fmt.Errorf("failed to append protobuf field %q: %w", f.name(), err)` on the `AppendValueOrNull` loop, and have `Record` call it and panic on error.
2. Add the `enumValue == nil` guards at the string-enum (`:773`) and dictionary (`:822`) sites.
3. Drop the `getDictValues` / `dictValues.Release()` changes — already in `main`.
4. Resolve the test-file conflict against the tests #1083/#1084 added.

Could you rebase onto current `main` and re-run `go test ./arrow/util`? Apologies for the churn — that's from #1082/#1083/#1084 merging ahead of this one. Happy to re-review quickly once it's rebased.

zeroshade (approved):

zeroshade:
just need to fix the merge conflict

PR #1096, opened 2026-08-05
fix(array): compare sparse union slices at relative offsets

fallintoplace:
### Rationale for this change

SparseUnion.setData already slices each child using the parent union offset. ApproxEqual adds that offset again, so sliced unions can compare the wrong child values or hit invalid slice bounds.

Compare each selected child at its relative i:i+1 position, matching exact equality and the representation used by sparse unions.

### Are these changes tested?

- `go test ./arrow/array -run TestSparseUnionApproxEqualSlicesWithDifferentOffsets -count=1`

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -471,8 +471,8 @@ func arraySparseUnionApproxEqual(l, r *SparseUnion, opt equalOption) bool {
 		}
 
 		childNum := childIDs[typeID]
-		eq := sliceApproxEqual(l.children[childNum], int64(i+l.data.offset), int64(i+l.data.offset+1),
-			r.children[childNum], int64(i+r.data.offset), int64(i+r.data.offset+1), opt)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. The root cause is confirmed: `union.setData` already slices each sparse-union child by the parent offset (union.go L137-143), and `RawTypeCodes()` returns `typecodes[offset:]`, so children and type codes are both already offset-adjusted. Adding `data.offset` again in `arraySparseUnionApproxEqual` double-counted the offset. The fix brings the approx-equal path in line with the already-correct exact-equality path (`arraySparseUnionEqual`, L454), and dense unions are correctly left untouched.

Verified locally:
- New test passes; full `./arrow/array -run Union` suite passes.
- Regression proof: reverting only the union.go change (keeping the test) makes the test fail, confirming it genuinely catches the bug.
- The test uses distinct offsets on each side over different underlying values, so it faithfully reproduces the reported issue.

PR #1097, opened 2026-08-05
fix(arrow/ipc): validate file block framing and lengths

fallintoplace:
### Rationale for this change

IPC file footers contain metadata and body lengths, while each message also carries framing information. File readers currently check that footer ranges are readable but can accept messages whose framing and footer lengths disagree.

Validate continuation and legacy metadata prefixes against the footer, reject EOS markers inside file blocks, compare Message.BodyLen with the footer body length, and reject body lengths that are not 8-byte aligned. Share the checks between regular and memory-mapped file readers.

### Are these changes tested?

- `go test ./arrow/ipc -count=1`
- Added coverage for valid modern and legacy messages, both readers, metadata mismatches, body mismatches, EOS markers, and unaligned bodies.

Yes. Invalid IPC file framing and inconsistent footer lengths are now rejected by both file readers.

### DIFF
--- a/arrow/ipc/file_reader.go
+++ b/arrow/ipc/file_reader.go
@@ -75,9 +75,6 @@ func validateFileBlock(offset int64, meta int32, body, fileSize, maxMetadataSize
 	if body < 0 {
 		return fmt.Errorf("arrow/ipc: invalid file block body length %d", body)
 	}
-	if body%8 != 0 {
-		return fmt.Errorf("arrow/ipc: file block body length %d is not a multiple of 8", body)
-	}
 	if maxMetadataSize > 0 && int64(meta) > maxMetadataSize {
 		return fmt.Errorf("arrow/ipc: file block metadata length %d exceeds limit %d", meta, maxMetadataSize)
 	}
--- a/arrow/ipc/file_reader.go
+++ b/arrow/ipc/file_reader.go
@@ -75,6 +75,9 @@ func validateFileBlock(offset int64, meta int32, body, fileSize, maxMetadataSize
 	if body < 0 {
 		return fmt.Errorf("arrow/ipc: invalid file block body length %d", body)
 	}
+	if body%8 != 0 {
+		return fmt.Errorf("arrow/ipc: file block body length %d is not a multiple of 8", body)
+	}
 	if maxMetadataSize > 0 && int64(meta) > maxMetadataSize {
 		return fmt.Errorf("arrow/ipc: file block metadata length %d exceeds limit %d", meta, maxMetadataSize)
 	}
--- a/arrow/ipc/file_reader.go
+++ b/arrow/ipc/file_reader.go
@@ -974,21 +974,18 @@ func (blk mappedFileBlock) NewMessage() (*Message, error) {
 
 	metaBytes := buf[:blk.meta]
 
-	prefix := 0
-	switch binary.LittleEndian.Uint32(metaBytes) {
-	case 0:
-	case kIPCContToken:
-		prefix = 8
-	default:
-		// ARROW-6314: backwards compatibility for reading old IPC
-		// messages produced prior to version 0.15.0
-		prefix = 4
-	}
-	if int(blk.meta)-prefix < 4 {
-		return nil, fmt.Errorf("arrow/ipc: invalid file block metadata length %d for prefix length %d", blk.meta, prefix)
+	prefix, err := validateFileBlockMetadata(metaBytes, blk.meta)
+	if err != nil {
+		return nil, err
 	}
 
 	meta = memory.NewBufferBytes(metaBytes[prefix:])
 	body = memory.NewBufferBytes(buf[blk.meta : int64(blk.meta)+blk.body])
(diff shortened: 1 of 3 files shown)


arrow/ipc/file_reader.go
```diff
@@ -75,6 +75,9 @@ func validateFileBlock(offset int64, meta int32, body, fileSize, maxMetadataSize
 	if body < 0 {
 		return fmt.Errorf("arrow/ipc: invalid file block body length %d", body)
 	}
+	if body%8 != 0 {
+		return fmt.Errorf("arrow/ipc: file block body length %d is not a multiple of 8", body)
+	}
```
zeroshade: while we always make sure we output a body with a multiple of 8, let's be lenient on reads and not reject these

arrow/ipc/file_reader.go
```diff
@@ -974,21 +977,18 @@ func (blk mappedFileBlock) NewMessage() (*Message, error) {
...
 	meta = memory.NewBufferBytes(metaBytes[prefix:])
 	body = memory.NewBufferBytes(buf[blk.meta : int64(blk.meta)+blk.body])
-	return NewMessage(meta, body), nil
+	msg := NewMessage(meta, body)
+	messageBodyLen := msg.BodyLen()
+	if messageBodyLen != blk.body {
+		msg.Release()
+		return nil, fmt.Errorf("arrow/ipc: file block body length %d does not match message body length %d", blk.body, messageBodyLen)
+	}
+	return msg, nil
```
zeroshade, line 993: how can this actually happen?

arrow/ipc/metadata.go
```diff
@@ -75,6 +75,38 @@ func (blk fileBlock) Offset() int64 { return blk.offset }
...
+	case kIPCContToken:
+		prefix = 8
+		if len(buf) < prefix {
+			return 0, fmt.Errorf("arrow/ipc: file block metadata is too short for prefix length %d", prefix)
+		}
+		length = binary.LittleEndian.Uint32(buf[4:])
+	default:
+		prefix = 4
+		length = binary.LittleEndian.Uint32(buf)
+	}
```
zeroshade: these can be factored out

```suggestion
	default:
		prefix = 4		
	}
	length = binary.LittleEndian.Uint32(buf[prefix-4:])
```

arrow/ipc/metadata.go
```diff
@@ -75,6 +75,38 @@ func (blk fileBlock) Offset() int64 { return blk.offset }
...
+	case 0:
+		return 0, errors.New("arrow/ipc: unexpected end-of-stream marker in file block")
+	case kIPCContToken:
+		prefix = 8
+		if len(buf) < prefix {
+			return 0, fmt.Errorf("arrow/ipc: file block metadata is too short for prefix length %d", prefix)
+		}
+		length = binary.LittleEndian.Uint32(buf[4:])
+	default:
+		prefix = 4
```
zeroshade, line 98: you dropped the comment explaining that this is for reading old IPC messages from Arrow format version <= 0.15.0, add the comment back please

arrow/ipc/metadata.go
```diff
@@ -131,7 +154,13 @@ func (blk fileBlock) NewMessage() (*Message, error) {
...
 	}
 
-	return NewMessage(meta, body), nil
+	msg := NewMessage(meta, body)
+	messageBodyLen := msg.BodyLen()
+	if messageBodyLen != blk.body {
+		msg.Release()
+		return nil, fmt.Errorf("arrow/ipc: file block body length %d does not match message body length %d", blk.body, messageBodyLen)
+	}
+	return msg, nil
```
zeroshade, line 164: same question, how does this actually happen?

zeroshade (changes requested):

arrow/ipc/file_reader.go
fallintoplace: They are two independent values in the IPC file: Message.bodyLength comes from the message FlatBuffer, while Block.bodyLength comes from the file footer, so a producer bug or malformed file can make them disagree.

There is a concrete precedent for this: apache/arrow#19596 was a JS writer bug that produced exactly this mismatch. C++ also added this same consistency check in apache/arrow#48845 earlier this year.

So I think this check is useful for detecting an inconsistent IPC file rather than relying on either copy alone.

arrow/ipc/file_reader.go
fallintoplace: Just to confirm, do you want arrow-go to intentionally be more lenient here than the C++ reader? C++ currently rejects file blocks whose body length isn't 8-byte aligned via `CheckAligned`, and the IPC format also requires 8-byte alignment.

I'm happy to keep the Go reader lenient if that's the preferred compatibility policy, just wanted to check that the difference is intentional.

zeroshade: ah, i thought the C++ reader was more lenient. If the C++ reader is also strict on this then it's fine and we can leave it strict.

zeroshade (approved):

PR #1098, opened 2026-08-05
fix(parquet): preserve dictionary index types

fallintoplace:
### Rationale for this change

Parquet dictionary encoding uses int32 indexes, but stored Arrow schema metadata can specify a different dictionary index type. The schema restoration path currently hardcodes int32 and changes the dictionary type and index buffer width on read.

Restore the stored index type and safely cast the reader indexes before rebuilding each dictionary array. Preserve dictionary values and the ordered flag.

### Are these changes tested?

- `go test ./parquet/pqarrow -run TestArrowDictionaryTypePreservesIndexType -count=1`

### DIFF
--- a/parquet/pqarrow/column_readers.go
+++ b/parquet/pqarrow/column_readers.go
@@ -17,6 +17,7 @@
 package pqarrow
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


zeroshade (approved):

zeroshade:
Nice catch!

PR #1099, opened 2026-08-05
fix(parquet): read null fixed-size-list parents

fallintoplace:
### Rationale for this change

Parquet does not store child values for a null fixed-size-list parent, while Arrow requires list_size child slots for every parent. The existing reader assumes every parent consumes exactly list_size values and does not validate the final span.

Reconstruct children from valid and null parent runs:

- validate that each valid parent consumes exactly list_size values
- validate that null parents consume no stored child values
- materialize list_size null children for each null parent
- validate the final parent span

When there are no null parents, validate the spans and reuse the decoded child array directly.

### Are these changes tested?

- `go test ./parquet/pqarrow -run 'TestParquetArrowIO/TestFixedSizeList' -count=1`

### DIFF
--- a/parquet/pqarrow/column_readers.go
+++ b/parquet/pqarrow/column_readers.go
@@ -643,7 +643,7 @@ func newFixedSizeListReader(rctx *readerCtx, field *arrow.Field, info file.Level
 }
 
 func (lr *fixedSizeListReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
-	return lr.listReader.buildArray(lenBound, true)
+	return lr.buildArray(lenBound, true)
 }
 
 // helper function to combine chunks into a single array.
--- a/parquet/pqarrow/column_readers.go
+++ b/parquet/pqarrow/column_readers.go
@@ -451,10 +451,10 @@ func (lr *listReader) LoadBatch(nrecords int64) error {
 }
 
 func (lr *listReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
-	return lr.buildArray(lenBound, false)
+	return lr.buildArray(lenBound)
 }
 
-func (lr *listReader) buildArray(lenBound int64, fixedSize bool) (*arrow.Chunked, error) {
+func (lr *listReader) buildArray(lenBound int64) (*arrow.Chunked, error) {
 	var (
 		defLevels      []int16
 		repLevels      []int16
@@ -528,7 +528,7 @@ func (lr *listReader) buildArray(lenBound int64, fixedSize bool) (*arrow.Chunked
 	itemArr := array.MakeFromData(item)
 	defer itemArr.Release()
 
-	if fixedSize {
+	if lr.field.Type.ID() == arrow.FIXED_SIZE_LIST {
 		offsetData := arrow.Int32Traits.CastFromBytes(offsetsBuffer.Bytes())
 		return lr.buildFixedSizeListArray(int(validityIO.Read), offsetData, validityBuffer,
 			validityIO.NullCount, itemArr)
@@ -573,6 +573,8 @@ func (lr *listReader) buildFixedSizeListArray(length int, offsets []int32, valid
 		return arrow.NewChunked(lr.field.Type, []arrow.Array{out}), nil
 	}
 
+	// Each validity run becomes one piece. Alternating valid and null parents
+	// create O(length) temporary arrays before concatenation.
 	pieces := make([]arrow.Array, 0, length)
 	defer func() { releaseArrays(pieces) }()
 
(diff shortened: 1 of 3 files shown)


parquet/pqarrow/column_readers.go
```diff
@@ -451,6 +451,10 @@ func (lr *listReader) LoadBatch(nrecords int64) error {
 }
 
 func (lr *listReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
+	return lr.buildArray(lenBound, false)
+}
+
+func (lr *listReader) buildArray(lenBound int64, fixedSize bool) (*arrow.Chunked, error) {
```
zeroshade: Minor: the `fixedSize` parameter looks redundant. `fixedSizeListReader` embeds `*listReader`, and the code this replaces decided the same thing inline with `lr.field.Type.ID() == arrow.FIXED_SIZE_LIST`. Keeping that type check here would drop both the bool parameter and the `fixedSizeListReader.BuildArray` override, for the same behaviour.

Not a blocker, just more indirection than the one-line condition it replaced.

parquet/pqarrow/column_readers.go
```diff
@@ -529,18 +541,85 @@ func (lr *listReader) BuildArray(lenBound int64) (*arrow.Chunked, error) {
...
+			}
+			pieces = append(pieces, array.MakeArrayOfNull(lr.rctx.mem, listType.Elem(), (end-i)*listSize))
+		}
+		i = end
+	}
+
+	if len(pieces) == 0 {
+		pieces = append(pieces, array.MakeArrayOfNull(lr.rctx.mem, listType.Elem(), 0))
+	}
+	child, err := array.Concatenate(pieces, lr.rctx.mem)
```
zeroshade, line 613: Question: is this safe for dictionary-encoded or extension child types?

The `nullCount == 0` fast path hands the decoded child array through untouched, so it never reaches `Concatenate`. But as soon as there's a null parent, the child goes through this instead — so a fixed-size list of dictionary values would take a different code path depending only on whether any parent happened to be null. If `Concatenate` unifies or rejects dictionaries, that's an asymmetry worth knowing about. I haven't confirmed either way.

Separately, worth a brief comment noting the cost profile: parents that alternate null/valid produce one array per run, so this is O(length) allocations plus the concat in the worst case. Still strictly better than the previous behaviour of erroring out, but it's the kind of thing that's easier to reason about later with a note here.

zeroshade (commented):
This looks right to me, and it fixes a real failure — a null fixed-size-list parent previously produced a span of 0, tripped the `expected all lists to be of size=N` check, and made the read hard-error. Good catch on the final span too: the old loop ran `for x := 1; x < data.Len(); x++`, so it never validated the last list at all.

I checked the three things most likely to go wrong here and they all hold up:

- **Buffer ownership.** The old fixed-size branch needed the extra `defer data.Buffers()[1].Release()` because `NewData` retained the offsets buffer. The new path returns before `NewData` ever sees the offsets, and `buildArray` already owns both buffers via the defers at lines 476 and 485, so that release is correctly gone rather than missing. `validityBuffer` is retained by `NewData` and dropped by the deferred `data.Release()`. Balanced.
- **`offsets[length]`.** This reads one element further than the old code ever did, but it's in bounds — line 517 resizes to `Read+1` int32s beforehand. The empty case is fine too: `length == 0` still leaves one offset.
- **Child-length invariant.** Valid runs contribute `(end-i)*listSize` through the validated spans and null runs contribute the same count of nulls, so the child lands on exactly `length*listSize`.

Two details worth calling out as good: `defer func() { releaseArrays(pieces) }()` uses a closure so it actually sees the appended pieces — a plain `defer releaseArrays(pieces)` would capture the empty slice and leak on every error path. And both new tests use `memory.NewCheckedAllocator` with `AssertSize(t, 0)`, which is what makes the ownership reasoning above checkable rather than assumed.

Leaving this as a comment rather than an approval only because of the open question inline about `Concatenate` and dictionary children.

One more minor note, on `TestBuildFixedSizeListArrayRequiresExactChildSpan`: all five cases assert a bare `require.Error`. Since each case is meant to exercise a *different* validation, any one of them could start returning a different error and the test would still pass. `require.ErrorContains` would pin each case to the check it's actually covering.

---

zeroshade (approved):
Re-reviewed at `71689a86c`. All four points addressed, two of them further than I asked — resolving both threads.

**The `fixedSize` parameter.** You didn't just swap the bool for the type check; you removed the `fixedSizeListReader` wrapper entirely, which is the better call — it only existed to override `BuildArray` and pass the flag, so once the type-ID check moved inline it had no reason to exist. I checked this is behaviour-preserving rather than just plausible: `newListReader` does the identical `childRdr.Retain()`, the same field assignments and the same `refCount.Add(1)`, differing only in returning `lr` directly instead of wrapping it. Refcounting unchanged, type fully gone, and the sole caller at `file_reader.go:644` goes through the delegation.

**The `Concatenate` question.** Answered with a test, which is the right way to answer it. `TestBuildFixedSizeListArrayConcatenatesSpecializedChildren` drives both a dictionary and a JSON extension child through the null/concat path and asserts the real values survive, the synthesised nulls land in the right slots, and `arrow.TypeEqual` still holds for the extension case afterwards. That was the only thing keeping me from approving — it's now demonstrated rather than assumed.

**Error assertions.** Beyond the `ErrorContains` change, the two new cases (`valid parent with wrong child span`, `null parent consumes child values`) cover the per-run span validation inside the null branch. That had no coverage before — all five original cases failed at the two top-level offset checks and never reached the loop.

The new test fixture is self-consistent too (validity `0x01` with offsets `{0,2,2}` and `itemLen` 2 gives 4 child slots, 2 real and 2 null), and the borrowed `ListValues()` reference is read before the `t.Cleanup` release.

Approving. CI is still finishing; I'll merge once it's green.

---

PR #1106, opened 2026-08-05
fix(arrow/scalar): validate FromScalar destinations

fallintoplace:
### Rationale for this change

FromScalar accepts any pointer and then assumes it points to a struct. Nil pointers and pointers to non-struct values can panic instead of returning an error.

Validate the pointer, nil state, and target kind before using reflection. Preserve the existing custom TypeFromScalar path and cover invalid targets without allowing a panic.

### Are these changes tested?

- `go test ./arrow/scalar`


zeroshade (approved):

zeroshade:
just needs to have the merge conflicts resolved

PR #1108, opened 2026-08-05
fix(arrow/scalar): use wide offsets for large binary scalars

fallintoplace:
### Rationale for this change

MakeArrayFromScalar currently uses 32-bit offsets for every binary-like scalar. LargeBinary and LargeString require 64-bit offsets, so their arrays can have the wrong physical layout.

Select the offset width from the scalar type and test ordinary and large binary values, including validation and buffer size.

### Are these changes tested?

- `go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/scalar.go
+++ b/arrow/scalar/scalar.go
@@ -802,12 +802,20 @@ func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Arr
 		return MakeArrayOfNull(sc.DataType(), length, mem), nil
 	}
 
-	createOffsets := func(valLength int32) *memory.Buffer {
+	createOffsets := func(valLength int64) *memory.Buffer {
 		buffer := memory.NewResizableBuffer(mem)
-		buffer.Resize(arrow.Int32Traits.BytesRequired(length + 1))
+		if sc.DataType().ID() == arrow.LARGE_BINARY || sc.DataType().ID() == arrow.LARGE_STRING {
+			buffer.Resize(arrow.Int64Traits.BytesRequired(length + 1))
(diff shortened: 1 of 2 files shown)


arrow/scalar/scalar.go
```diff
@@ -802,12 +772,20 @@ func MakeArrayFromScalar(sc Scalar, length int, mem memory.Allocator) (arrow.Arr
 		return MakeArrayOfNull(sc.DataType(), length, mem), nil
 	}
 
-	createOffsets := func(valLength int32) *memory.Buffer {
+	createOffsets := func(valLength int64) *memory.Buffer {
 		buffer := memory.NewResizableBuffer(mem)
-		buffer.Resize(arrow.Int32Traits.BytesRequired(length + 1))
+		if sc.DataType().ID() == arrow.LARGE_BINARY || sc.DataType().ID() == arrow.LARGE_STRING {
```
zeroshade: do we also need to check for LARGE_LIST?

zeroshade (changes requested):

arrow/scalar/scalar.go
fallintoplace: MakeArrayFromScalar does not support LargeList scalars today. The existing nested implementation only handles *List, so *LargeList still returns the existing “not yet implemented” error.

I kept this PR scoped to the binary/string scalar paths, where the offset-width bug exists. The change covers Binary, String, LargeBinary, and LargeString. I think LargeList support would be better handled separately with its own implementation and tests.

zeroshade: that's fair, sounds good. thanks!

zeroshade (approved):

PR #1109, opened 2026-08-05
fix(arrow/scalar): align timestamp timezone parsing

fallintoplace:
### Rationale for this change

MakeScalarParam, ParseScalar, timestamp builders, JSON decoding, and typed CSV parsing should use the same timestamp timezone rules. This also adds support for Arrow fixed-offset timezone strings.

- Validate timestamp timezone metadata through `TimestampType.GetZone`.
- Accept UTC metadata case-insensitively, including forms such as `Utc`.
- Require the input to contain an offset exactly when the timestamp type has a non-empty timezone.
- Keep timezone-naive values offset-free.
- Preserve the instant represented by an explicit input offset.
- Keep numeric timestamp values as raw epoch values after validating the type timezone.
- Validate CSV timestamp metadata before reading rows, including null-only columns.
- Keep timezone offsets in CSV output for timezone-aware timestamps so CSV writer and reader round trips work.
- Preserve an explicitly supplied empty custom timestamp layout.

### Are these changes tested?

- `go test ./arrow ./arrow/scalar ./arrow/array ./arrow/compute -count=1`
- `go test ./arrow/csv -run Timestamp -count=1`
- `go test -race ./arrow ./arrow/scalar ./arrow/array ./arrow/compute ./arrow/csv -run Timestamp -count=1`

Timestamp string parsing now rejects mismatched timezone presence:

- timezone-less type plus no offset: accepted
- timezone-less type plus an offset: rejected
- timezone-aware type plus an offset: accepted
- timezone-aware type without an offset: rejected

Explicit offsets still determine the represented instant, and mixed-case UTC metadata remains accepted. CSV output now includes the offset for timezone-aware timestamps.


zeroshade:
@fallintoplace need to fix the failing tests

arrow/csv/reader.go
```diff
@@ -500,8 +500,14 @@ func (r *Reader) initFieldConverter(bldr array.Builder) func(string) {
 			}
 		}
 	case *arrow.TimestampType:
+		if _, err := dt.GetZone(); err != nil {
+			if r.err == nil {
+				r.err = err
+			}
+			return func(string) {}
```
zeroshade: **Blocking:** returning a no-op converter after setting `r.err` violates the CSV reader's one-input/one-builder-value invariant.

I reproduced two new panics:

- `fixed_size_list<timestamp[s, tz=not/a_timezone]>` panics when constructing the array because the parent appends one list while this converter appends zero child values.
- `NewInferringReader` with an invalid timestamp supplied through `WithColumnTypes` can panic because this column has zero rows while another column has one.

A regular list similarly produces an empty list rather than a null. Also, because this validation is lazy, an invalid timestamp nested in an all-null list is never validated at all.

Please validate timestamp metadata recursively and ensure an error-path converter still appends a null value.

arrow/csv/reader.go
```diff
@@ -741,21 +747,16 @@ func (r *Reader) parseFloat64(field array.Builder, str string) {
...
-// parses timestamps using millisecond precision
-func (r *Reader) parseTimestamp(field array.Builder, str string, unit arrow.TimeUnit) {
+func (r *Reader) parseTimestamp(field array.Builder, str string) {
 	if r.isNull(str) {
 		field.AppendNull()
 		return
 	}
 
-	v, err := arrow.TimestampFromString(str, unit)
+	err := field.(*array.TimestampBuilder).AppendValueFromString(str)
```
zeroshade: **Blocking:** this stricter parser is inconsistent with CSV type inference. `tryParse` still uses `arrow.TimestampFromString`, so `2024-01-01T00:00:00Z` is inferred as timezone-less `timestamp[s]`; this call then rejects the same value because it contains an offset.

Reproducer:

```go
r := csv.NewInferringReader(strings.NewReader("2024-01-01T00:00:00Z\n"))
r.Next()
// r.Err(): timestamp value ... for type timestamp[s] must not include a zone offset
```

This succeeds on the merge base. Inference should apply the same zone-presence rule so zoned input advances to the existing `timestamp[*, UTC]` inference rung.

arrow/scalar/parse.go
```diff
@@ -41,6 +40,27 @@ type TypeFromScalar interface {
...
+	if err != nil {
+		return 0, err
+	}
+
+	ts, zonePresent, err := arrow.TimestampFromStringInLocation(val, dt.Unit, loc)
+	if err != nil {
+		return 0, err
+	}
+
+	if zonePresent != (dt.TimeZone != "") {
```
zeroshade, line 57: Requiring an offset for zoned timestamp types makes the package's own scalar string representation non-parseable. `scalar.Timestamp.String()` still emits no offset:

```go
typ := &arrow.TimestampType{Unit: arrow.Second, TimeZone: "UTC"}
s := scalar.NewTimestampScalar(0, typ)
scalar.ParseScalar(typ, s.String())
// must include a zone offset
```

This round-trip succeeds on the merge base. Please update timestamp scalar formatting to include the appropriate offset for zoned types, analogous to the array and CSV formatting changes in this PR.

zeroshade (changes requested):
Thanks for aligning timestamp parsing across the scalar, array, JSON, and CSV paths. The core timezone-presence rule and fixed-offset handling look consistent, and CI is green.

I found three regressions that need addressing before this lands:

1. Invalid timestamp metadata can make CSV converters append no value, causing panics in fixed-size lists and multi-column inferred readers.
2. CSV inference classifies zoned ISO-8601 values as timezone-less timestamps and then rejects them during conversion.
3. Zoned timestamp scalars no longer round-trip through their own `String()` output and `ParseScalar`.

I included minimal reproducers inline. The affected core packages and focused CSV timestamp tests otherwise pass locally.

---

>

arrow/scalar/temporal.go
```diff
@@ -40,7 +42,15 @@ func temporalToString(s TemporalScalar) string {
...
-		return time.Unix(0, int64(s.Value)*int64(s.Unit().Multiplier())).UTC().Format("2006-01-02 15:04:05.999999999")
+		typ := s.DataType().(*arrow.TimestampType)
+		if typ.TimeZone == "" {
+			return time.Unix(0, int64(s.Value)*int64(s.Unit().Multiplier())).UTC().Format(timestampScalarLayout)
+		}
+		toTime, err := typ.GetToTimeFunc()
+		if err != nil {
+			return "..."
+		}
+		return toTime(s.Value).Format(timestampScalarLayout + "Z0700")
```
zeroshade: `Z0700` loses seconds from historical IANA timezone offsets, so this still does not universally round-trip. For `Europe/Berlin` at year 0000, `String()` emits `0000-01-01 00:53:28+0053`; parsing it changes the timestamp by 28 seconds. Please preserve second-granularity offsets and add a historical named-zone regression test.

arrow/compute/expression_test.go
```diff
@@ -37,7 +37,7 @@ type privateFunctionOptions struct {
 func (privateFunctionOptions) TypeName() string { return "privateFunctionOptions" }
 
 func TestExpressionToString(t *testing.T) {
-	ts, _ := scalar.MakeScalar("1990-10-23 10:23:33.123456").CastTo(arrow.FixedWidthTypes.Timestamp_ns)
+	ts, _ := scalar.MakeScalar("1990-10-23 10:23:33.123456Z").CastTo(arrow.FixedWidthTypes.Timestamp_ns)
```
zeroshade, line 46: This fixture now includes a timezone, so the scalar renders with `Z`, but the expected value below remains timezone-less. Current CI reports expected `1990-10-23 10:23:33.123456`, actual `1990-10-23 10:23:33.123456Z`. Please update the expectation so the test suite passes.

zeroshade (changes requested):
The three issues from my previous review are addressed: nested timestamp metadata is validated without violating builder lengths, CSV inference handles zoned timestamps consistently, and ordinary UTC/fixed-offset scalar strings round-trip.

Two follow-ups still block merging: historical named-zone offsets can lose seconds during `String()`/`ParseScalar` round-trips, and the timestamp expression expectation currently fails CI. Details are inline.

---

>

arrow/scalar/temporal.go
```diff
@@ -40,7 +46,20 @@ func temporalToString(s TemporalScalar) string {
...
+		}
+		tm := toTime(s.Value)
+		layout := timestampScalarLayout
+		if typ.TimeZone != "" {
+			layout = timestampScalarZoneLayout
+			if _, offset := tm.Zone(); offset%60 != 0 {
+				layout = timestampScalarZoneSecondsLayout
+			}
+		}
+		return tm.Format(layout)
```
zeroshade, line 68: Wide second, millisecond, and microsecond timestamp values still do not round-trip through `String()` and `ParseScalar`. For example, `timestamp[s, tz=UTC]` at `math.MaxInt64` renders as `292277026596-12-04 15:30:07Z`, which the parser rejects as excess fractional precision; `math.MinInt64` is also formatted incorrectly. Please support expanded/signed years in parsing or use an exact fallback representation outside the textual date range, and add min/max round-trip tests for every unit and timezone category.

zeroshade (changes requested):
The current head addresses all findings from my previous reviews: recursive CSV metadata validation, zoned inference, ordinary scalar round trips, historical second-granularity offsets, and the stale expression expectation are fixed.

Wide timestamp values still fail the same `String()`/`ParseScalar` round-trip contract, however. Randomized probes found systematic failures for second and millisecond units and widespread failures for microseconds; details are inline.

Focused timestamp and race tests otherwise pass. CI is still running.

---

>

zeroshade (changes requested):
The new boundary tests do not compile on 32-bit targets. In `arrow/compute/cast_test.go:1352-1353`, untyped `math.MinInt64` and `math.MaxInt64` constants are passed through `fmt.Sprintf`; they default to `int` and overflow on 386.

Please cast these values to `int64` (or use typed `int64` locals) so the compute test package remains portable.

---

>

zeroshade (approved):
The typed `int64` bounds fix resolves the 32-bit portability issue from my previous review, and the two follow-up commits (zone validation before comparison, parse-error preservation) pass the full compute suite locally. The linux/386 vet overflow I hit while re-checking is pre-existing on `main` in the untouched `vector_run_end_test.go` benchmark, so it is not attributable to this PR.

---

PR #1110, opened 2026-08-05
fix(arrow/scalar): apply timestamp zones when casting to dates

fallintoplace:
### Rationale for this change

Timestamp-to-Date32 and timestamp-to-Date64 casts use integer arithmetic that truncates negative values and ignores the timestamp timezone.

Convert timestamps through the timestamp type's time semantics and reuse the date normalization used by the compute cast path. Add regression coverage around the Unix epoch and a named timezone.

### Are these changes tested?

- `go test ./arrow/scalar`

### DIFF
--- a/arrow/scalar/temporal.go
+++ b/arrow/scalar/temporal.go
@@ -103,6 +103,18 @@ type IntervalScalar interface {
 
 const millisecondsInDay = (time.Hour * 24) / time.Millisecond
 
+func timestampDate(s *Timestamp) (time.Time, error) {
+	timestampType := s.DataType().(*arrow.TimestampType)
+	toTime, err := timestampType.GetToTimeFunc()
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	tm := toTime(s.Value)
+	year, month, day := tm.Date()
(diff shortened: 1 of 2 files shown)


arrow/scalar/temporal.go
```diff
@@ -148,11 +148,27 @@ func castTemporal(from TemporalScalar, to arrow.DataType) (Scalar, error) {
 		case *arrow.TimestampType:
 			return NewTimestampScalar(arrow.Timestamp(arrow.ConvertTimestampValue(s.Unit(), to.Unit, int64(s.Value))), to), nil
 		case *arrow.Date32Type:
-			millis := arrow.ConvertTimestampValue(s.Unit(), arrow.Millisecond, int64(s.Value))
-			return NewDate32Scalar(arrow.Date32(millis / int64(millisecondsInDay))), nil
+			timestampType := s.DataType().(*arrow.TimestampType)
+			toTime, err := timestampType.GetToTimeFunc()
```
zeroshade: the `toTime` func should already be handling the timestamp offset by using `.In`, so we shouldn't need to separately add the offset like you're doing

zeroshade (changes requested):

zeroshade (approved):

PR #1113, opened 2026-08-05
feat(arrow/array): roll back JSON builder state after failed rows

fallintoplace:
### Rationale for this change

RecordBuilder mutates field builders while decoding a JSON object. If a later field fails, values appended for earlier fields remain and can affect subsequent rows.

Build the nested checkpoint graph once when the RecordBuilder is created. Each row captures reusable lengths and builder state, then restores that state on decode errors. This covers lists, structs, maps, fixed-size lists, unions, dictionaries, variable-width buffers, string views, and run-end encoded children without rebuilding checkpoint trees for every successful row.

Parent builders are restored before their children so parent Resize calls cannot overwrite restored child state. This is especially important for nested run-end encoded builders.

Custom builders with internal state can participate through the exported CheckpointState and CheckpointableBuilder interfaces.

### Are these changes tested?

- go test ./arrow/array
- go test ./arrow/extensions
- go test ./arrow/array -run ^$ -bench BenchmarkRecordFromJSON/Size_1000$ -benchtime=1x -benchmem

Yes. Failed JSON rows are rolled back completely, including nested builder state. This also adds two exported interfaces for custom builders that need to restore internal state during row rollback.

### DIFF
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -431,15 +431,20 @@ type checkpointState interface {
 }
 
 // CheckpointState captures and restores builder state that is not represented by
-// the builder's length or storage builders.
+// the builder's length or storage builders. RecordBuilder reuses the same
+// checkpoint for each row, calling Capture before decoding and Restore after a
+// failed decode.
 type CheckpointState interface {
+	// Capture records the current state of the builder.
 	Capture()
+	// Restore returns the builder to the last captured state.
 	Restore()
 }
 
 // CheckpointableBuilder allows custom builders to participate in RecordBuilder
-// row rollback.
+// row rollback. The returned checkpoint is reused for every row.
 type CheckpointableBuilder interface {
+	// NewCheckpoint returns a reusable checkpoint for the builder.
 	NewCheckpoint() CheckpointState
 }
 
@@ -534,31 +539,10 @@ func (checkpoint *builderCheckpoint) restore() {
 		builder.unmarshalled = checkpoint.unmarshalled
 		builder.lastStr = checkpoint.lastStr
 	} else {
-		// Resize the parent before restoring children. Some parent builders resize
-		// their children as part of Resize, so restoring children first would
-		// overwrite their physical state again.
-		checkpoint.builder.Resize(checkpoint.length)
-		switch builder := checkpoint.builder.(type) {
-		case *ListBuilder:
-			builder.length = checkpoint.length
-		case *LargeListBuilder:
-			builder.length = checkpoint.length
-		case *ListViewBuilder:
-			builder.length = checkpoint.length
-		case *LargeListViewBuilder:
-			builder.length = checkpoint.length
-		case *FixedSizeListBuilder:
-			builder.length = checkpoint.length
-		case *MapBuilder:
-			builder.listBuilder.length = checkpoint.length
-		case *StructBuilder:
-			builder.length = checkpoint.length
-		case *SparseUnionBuilder:
-			builder.typesBuilder.SetLength(checkpoint.length)
-		case *DenseUnionBuilder:
-			builder.typesBuilder.SetLength(checkpoint.length)
-			builder.offsetsBuilder.SetLength(checkpoint.length * arrow.Int32SizeBytes)
-		}
+		// Truncate the parent before restoring children. Some parent builders
+		// truncate their children as part of truncation, so restoring children
+		// first would overwrite their physical state again.
+		checkpoint.builder.truncate(checkpoint.length)
 	}
 	if checkpoint.state != nil {
 		checkpoint.state.restore()
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -430,6 +430,26 @@ type checkpointState interface {
 	restore()
 }
 
+// CheckpointState captures and restores builder state that is not represented by
+// the builder's length or storage builders.
+type CheckpointState interface {
+	Capture()
+	Restore()
+}
+
+// CheckpointableBuilder allows custom builders to participate in RecordBuilder
+// row rollback.
+type CheckpointableBuilder interface {
+	NewCheckpoint() CheckpointState
+}
+
+type checkpointStateAdapter struct {
+	state CheckpointState
+}
+
+func (s *checkpointStateAdapter) capture() { s.state.Capture() }
+func (s *checkpointStateAdapter) restore() { s.state.Restore() }
+
 type storageBuilder interface {
 	StorageBuilder() Builder
 }
@@ -450,6 +470,8 @@ func newBuilderCheckpoint(builder Builder) *builderCheckpoint {
 	}
 	if checkpointable, ok := builder.(checkpointableBuilder); ok {
 		checkpoint.state = checkpointable.newCheckpoint()
+	} else if checkpointable, ok := builder.(CheckpointableBuilder); ok {
+		checkpoint.state = &checkpointStateAdapter{state: checkpointable.NewCheckpoint()}
 	}
 
 	switch builder := builder.(type) {
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -479,6 +479,8 @@ func newBuilderCheckpoint(builder Builder) *builderCheckpoint {
 		checkpoint.state = &checkpointStateAdapter{state: checkpointable.NewCheckpoint()}
 	}
 
+	// Keep this switch in sync with builder types that own children. An omitted
+	// nested builder would restore its own state but leave its children changed.
 	switch builder := builder.(type) {
 	case *ListBuilder:
 		checkpoint.children = append(checkpoint.children, newBuilderCheckpoint(builder.values))
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -516,19 +516,19 @@ func (checkpoint *builderCheckpoint) restore() {
 	checkpoint.builder.Resize(checkpoint.length)
 	switch builder := checkpoint.builder.(type) {
 	case *ListBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *LargeListBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *ListViewBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *LargeListViewBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *FixedSizeListBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *MapBuilder:
-		builder.listBuilder.builder.length = checkpoint.length
+		builder.listBuilder.length = checkpoint.length
 	case *StructBuilder:
-		builder.builder.length = checkpoint.length
+		builder.length = checkpoint.length
 	case *SparseUnionBuilder:
 		builder.typesBuilder.SetLength(checkpoint.length)
 	case *DenseUnionBuilder:
--- a/arrow/array/record.go
+++ b/arrow/array/record.go
@@ -430,6 +430,10 @@ type checkpointState interface {
 	restore()
 }
 
+type storageBuilder interface {
+	StorageBuilder() Builder
+}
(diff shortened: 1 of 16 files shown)


zeroshade:
Thanks for this — the architecture here is sound and worth keeping: the checkpoint graph built once at construction, the parent-before-child restore ordering, the multibuffer/string-view truncation, and the memo-table rehashing all look correct. I found no refcount imbalance, no mutation of already-built arrays, and no hash entry left reachable after truncation. Your test suite is green, and `go vet`'s only complaint reproduces on unmodified `main`.

That said, I think this needs changes before merge. Everything below was reproduced against `0b09f34c`.

## Root cause

The rollback path uses `Builder.Resize` as a transactional rewind primitive, but `Resize` is an allocation/capacity hint, not an exact logical truncation — it's free to round up, no-op, or compare in the wrong units, and it never clears discarded validity bits. The fact that the PR already reaches into six builders' unexported `length` fields to compensate is the design signalling this. The four issues below are all the same cause.

## 1. `string` / `large_string` rollback silently corrupts data

Rows `{"s":"aaa","o":1}`, failing `{"s":"bbb","o":"bad"}`, then `{"s":"ccc","o":2}` produces `["aaa", ""]` instead of `["aaa","ccc"]`.

`binarybuilder.go:252` compares a **byte** count against an **element** count:

```go
b.offsets.resize((n + 1) * b.offsetByteWidth)
if (n * b.offsetByteWidth) < b.offsets.Len() {   // bytes  <  elements
    b.offsets.SetLength(n * b.offsetByteWidth)   // never reached when shrinking
}
```

`int32BufferBuilder.Len()` returns `b.length / arrow.Int32SizeBytes` (elements), so shrinking 2→1 evaluates `4 < 2` → false and the offsets buffer never shrinks. The stale offset leaves `offsets=[0,3,3,6]` with `length=2`, so `value[1]` is `bytes[3:3]` = `""` while `"ccc"` sits unreferenced. It's silent because a duplicated offset is a *legal* encoding of an empty string — no panic, and `ValidateFull` can't catch it.

**Worth separating out:** this is a pre-existing latent bug, not a regression from your diff. `RecordBuilder.Resize(-1)` (`record.go:370-381`, from #805) already shrank builders mid-build, and reproduces the same corruption with no JSON involved. I'd suggest fixing `binarybuilder.go:252` as its own commit with a regression test — it's a real bug on `main` today. It can't be deferred past this PR though, since this change makes the path reachable from any JSON stream with a string column and one malformed row.

## 2. `FixedSizeBinaryBuilder` loses data

Rewinds `Len()` but never truncates its `values` byte buffer, and gets no checkpoint (unlike `BinaryBuilder`). Same `aaa` / `bbb`(bad) / `ccc` sequence → `["aaa","bbb"]`, with `"ccc"` lost.

## 3. `BooleanBuilder` / `NullBuilder` lengths are never rewound → panic

`NullBuilder.Resize` is a no-op (`null.go:141`); `BooleanBuilder.Resize` rounds `n` up to `minBuilderCapacity` (32) before resizing (`booleanbuilder.go:155-157`), so a failure within the first 32 values isn't removed:

```
{a: bool, b: int32}, failed row {"a":true,"b":"invalid"}
after failed row:    a.Len()=1  b.Len()=0
after next good row: a.Len()=2  b.Len()=1
NewRecordBatch panics: some fields have excessive number of rows (want at most 1, have 2)
```

Identical with a leading `NullBuilder`. A recoverable decode error leaves the builder permanently misaligned.

## 4. Discarded validity bits aren't cleared → rolled-back value resurfaces

Affects every field type. `resize` recomputes `nulls` for the kept prefix but leaves the set bit behind, and `AppendNull` then increments `nulls` without clearing it:

```
good {"a":10}, failed {"a":20,...}, then {"a":null,...}
array: [10 20]      <- null was appended at index 1; the discarded 20 came back
Len()=2 NullN()=1   IsNull(0)=false  IsNull(1)=false
ValidateFull: null count value (1) does not match actual number of nulls in array (0)
```

Nested variant: the list / list-view / fixed-size-list / map / struct corrections restore `length` but not `nulls`, so after rollback `Len()==0` while `NullN()==1`, poisoning the next row.

Since the goal of this PR is preventing failed rows from affecting subsequent rows, it's worth noting the rollback path currently introduces that same class of corruption for plain int32 and utf8 columns.

## Suggested fix shape

`Resize` can't carry this. I'd suggest a real per-builder exact-truncate primitive (e.g. an unexported `truncate(n int)` on `Builder`), or extending your existing `checkpointState` mechanism to *every* builder rather than only Binary/BinaryView/dictionary — rewinding length, offsets, data buffers, and children exactly, and zeroing the discarded validity-bitmap tail. `NullBuilder` and `BooleanBuilder` need explicit length handling regardless, since their `Resize` can't express a sub-32 truncation. This can't be fully fixed inside `record.go`'s switch, because the state needing rewind is unexported per-builder state that's unreachable for extension and third-party builders.

## Test gaps

Contract claims nine type families; several have no rollback coverage:

- **No rollback test at all:** maps, fixed-size lists, Boolean, Null, large-list, large-list-view.
- **Length-only assertions** (`Len()==0`, no following-row content check): sparse union, dense union, list-view.
- No failed-valid → null-at-same-index case, or failed-null → valid case (these expose #4).
- No direct memo-table `Truncate` test (truncate → reinsert identical dropped value, hash collisions, null indices).
- No test that builds an array, reuses the builder, rolls back, then asserts the already-built array is unchanged.
- No custom `CheckpointableBuilder` test for repeated capture/restore or checkpoint resource lifetime.

## Smaller notes

- `CheckpointState` has no commit/discard/release lifecycle — the REE checkpoint can retain the previously decoded object after a successful batch, and a custom checkpoint holding a refcounted resource has no success-path hook to release it.
- The union child graph goes stale if `AppendChild` is called after `RecordBuilder` construction; worth making that topology restriction explicit or detected.
- `timestamp_with_offset.go`: every public `Resize`, including capacity *growth*, resets `lastOffset`, forcing an unnecessary new REE run — correctness is fine, but compression regresses.
- Doc comments for the exported `TimestampWithOffsetBuilder.Resize` and `NewCheckpoint`; and the exported interfaces should document that one checkpoint object is reused and `Capture` is called repeatedly.

Happy to help review the follow-up. I have runnable repro tests for all four issues above if it'd be useful to have them posted here.

arrow/extensions/timestamp_with_offset.go
```diff
@@ -439,6 +439,31 @@ func (b *TimestampWithOffsetBuilder) NewArray() arrow.Array {
 	return b.NewExtensionArray()
 }
 
+// Resize adjusts the capacity of the builder and resets the current run-end
+// offset state.
+func (b *TimestampWithOffsetBuilder) Resize(n int) {
+	b.ExtensionBuilder.Resize(n)
+	b.lastOffset = noLastOffset
```
zeroshade: Blocking: this resets the run-end tracker on every `Resize`, including capacity growth.

`builder.reserve()` calls `resize()` whenever `length+elements > capacity`, so this fires during ordinary appending, not just on explicit user calls. When it fires mid-run, `lastOffset` goes back to `noLastOffset`, the next append takes the `offsetMinutes != b.lastOffset` branch, and `offsets.Append(1)` starts a **new run holding the same value as the previous run**.

That's the condition the existing comment in `AppendValues` explicitly preserves against: *"lastOffset is only updated when a new run starts, so a null row never splits a contiguous run into two adjacent runs sharing the same value."*

Effects: runs fragment (losing the compression REE exists for), and run boundaries become a function of allocation growth points rather than of the data — so the same input encodes differently depending on `Reserve` history. Unlikely to surface in small unit tests.

Suggested fix: drop this override. `timestampWithOffsetCheckpoint.Restore()` already restores `lastOffset` on the rollback path, which is the only path that needs it.

arrow/array/record.go
```diff
@@ -415,6 +421,137 @@ func (b *RecordBuilder) NewRecord() arrow.Record {
...
+	checkpoint := &builderCheckpoint{
+		builder: builder,
+	}
+	if checkpointable, ok := builder.(checkpointableBuilder); ok {
+		checkpoint.state = checkpointable.newCheckpoint()
+	} else if checkpointable, ok := builder.(CheckpointableBuilder); ok {
+		checkpoint.state = &checkpointStateAdapter{state: checkpointable.NewCheckpoint()}
+	}
+
+	switch builder := builder.(type) {
```
zeroshade, line 484: Minor: any future builder type with children that isn't added to this switch will silently skip rollback — the same failure mode this PR fixes, with no compile-time guard. Worth a short comment here noting that new nested builders must be registered, or a test that reflects over the known builder types.

zeroshade (changes requested):
Thanks for this — the checkpoint approach is sound and the nested-restore ordering is right.

One blocking issue on the `TimestampWithOffsetBuilder.Resize` override (inline), plus two minor notes.

Also: this adds two exported interfaces (`CheckpointState`, `CheckpointableBuilder`), so under our Conventional Commits + semver policy the title should be `feat(arrow/array):` rather than `fix(...)`. It's purely additive, so not breaking — just mislabelled for release notes. The interfaces themselves are justified; `arrow/extensions` is a separate package and genuinely needs them.

---

arrow/array/record.go
```diff
@@ -316,6 +317,10 @@ func NewRecordBuilder(mem memory.Allocator, schema *arrow.Schema) *RecordBuilder
 	for i := 0; i < schema.NumFields(); i++ {
 		b.fields[i] = NewBuilder(b.mem, schema.Field(i).Type)
 	}
+	b.checkpoints = make([]*builderCheckpoint, len(b.fields))
```
zeroshade, line 320: Caching these checkpoints assumes the field-builder identities never change, but `Fields()` returns the mutable backing slice. Existing callers can replace a field builder:

```go
old := rb.Fields()[0]
rb.Fields()[0] = replacement
old.Release()
```

The checkpoint still references `old`; the next JSON row captures or restores through the released builder and panics. If the caller retains `old`, rollback silently targets stale state instead of the active replacement.

Please either rebuild/validate checkpoints when field identities change or stop exposing a mutable field slice. This incompatibility is introduced by caching the checkpoint graph.

internal/hashing/xxh3_memo_table_types.go
```diff
@@ -190,6 +202,16 @@ func (t *Table[T]) Reset() {
 	t.nullIdx = KeyNotFound
 }
 
+func (t *Table[T]) Truncate(size int) {
```
zeroshade, line 205: Unlike `BinaryMemoTable.Truncate`, this always rebuilds the hash table even when `size >= t.Size()`. Besides unnecessary rehashing for equal-size rollback, a sufficiently large no-op truncation attempts to allocate from the requested size and panics:

```go
table := hashing.NewMemoTable[int32](0)
table.GetOrInsert(int32(7))
table.Truncate(math.MaxInt) // makeslice: len out of range
```

Please add the same early-return guard used by `BinaryMemoTable.Truncate`:

```go
if size >= t.Size() {
    return
}
```

zeroshade (changes requested):
The prior `TimestampWithOffsetBuilder` blocker is resolved, but two issues remain in the newly added rollback infrastructure:

1. Cached checkpoint graphs become stale if callers replace builders through the mutable slice returned by `Fields()`, causing rollback to target released or inactive builders.
2. Fixed-width memo-table truncation lacks the no-op guard used by `BinaryMemoTable`, allowing oversized truncation to panic.

See the inline comments on `record.go` and `xxh3_memo_table_types.go`.

The affected package suites, repeated rollback/resize/truncation tests, race checks, and vet passed. CI currently has 22 green checks with AMD64 Conda Integration pending.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. After you've addressed the points above and pushed an update, an Apache Arrow Go maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
All the requested changes look good — the checkpoint identity refresh handles field-builder replacement cleanly, and the `Truncate` guard matches `BinaryMemoTable`'s behavior. Thanks for the thorough regression tests.

Note: CI was only just approved to run on this PR, so it should complete before merging.

---

PR #1119, opened 2026-08-06
fix(compute): handle odd temporal multiples

fallintoplace:
### Rationale for this change

Half-up temporal rounding treats the lower remainder as an exact tie when Multiple is odd. For example, rounding 1 second to multiples of 3 seconds returns 3 seconds instead of 0.

Only even multiples use an exact halfway tie. Odd multiples compare the remainder to the two neighboring multiples, including negative values.

### Are these changes tested?

- `go test ./arrow/compute`

No API changes. Temporal rounding now returns the nearest result for odd multiples.

### DIFF
--- a/arrow/compute/internal/kernels/rounding.go
+++ b/arrow/compute/internal/kernels/rounding.go
@@ -1001,7 +1001,10 @@ func roundToMultipleInt64(value, multiple int64, mode RoundMode, strictCeil bool
 			absRemainder = -absRemainder
 		}
 
-		if absRemainder < half {
+		// Odd multiples do not have an exact halfway point. For example,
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
LGTM — odd-multiple temporal rounding now chooses the nearest representable multiple without disturbing even-multiple tie semantics.

The sign-sensitive integer paths and public temporal coverage look correct. I ran the targeted regression tests, `go test ./arrow/compute/... -count=1` (4 packages), and an exhaustive nearest-multiple property probe over odd multiples 1–999 and values −5000–5000; all passed on head `a5441022cba85`.

---

>

PR #1120, opened 2026-08-06
fix(arrow/ipc): return errors for invalid record indexes

fallintoplace:
### Rationale for this change

RecordBatchAt panics for an invalid record index even though it returns an error. ReadAt also converts an int64 index to int before checking the range.

Return an arrow.ErrIndex error for invalid indexes in RecordBatchAt and ReadAt, and check the int64 range before narrowing it to int.

### Are these changes tested?

- `go test ./arrow/ipc`

Invalid record indexes now return errors instead of panicking.

### DIFF
--- a/arrow/ipc/file_reader.go
+++ b/arrow/ipc/file_reader.go
@@ -442,7 +442,7 @@ func (f *FileReader) Record(i int) (arrow.Record, error) {
 // call concurrently.
 func (f *FileReader) RecordBatchAt(i int) (arrow.RecordBatch, error) {
 	if i < 0 || i >= f.NumRecords() {
-		panic("arrow/ipc: record index out of bounds")
+		return nil, fmt.Errorf("%w: record index %d out of bounds", arrow.ErrIndex, i)
 	}
 
 	blk, err := f.r.block(f.mem, &f.footer, i)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Looks good — both halves are correct.

The `RecordBatchAt` panic was straightforwardly a bug given the method already returns an `error`, and since `Record`, `RecordBatch`, `RecordAt` and `ReadAt` all funnel through it, the single change covers the whole indexed surface.

The `ReadAt` check is the more valuable half: narrowing `int64`→`int` before range-checking is a real correctness issue on 32-bit, and we do test `GOARCH=386` in CI. There an out-of-range `int64` can truncate onto a *valid* index and silently return the wrong batch. Checking before the conversion is the right fix, and the `math.MaxInt64` case covers it.

Two non-blocking notes:

- Worth a one-line comment on the `ReadAt` check noting it must happen *before* the narrowing — it looks redundant with `RecordBatchAt` on 64-bit, and a future cleanup could remove it and quietly reintroduce the 32-bit bug.
- The file still panics on `buffer index out of bound` and `field metadata out of bound`. Those are internal read-path invariants rather than caller-supplied indexes, so out of scope here — just noting the file isn't panic-free after this.

---

PR #1122, opened 2026-08-06
fix(parquet/pqarrow): reject decimal overflow in integer columns

fallintoplace:
### Rationale for this change

The Parquet decimal integer conversion uses debug assertions for range checks. Those assertions are disabled in normal builds, so values outside the signed INT32 or INT64 range can be silently narrowed.

Check signed range boundaries for Decimal128 and Decimal256 values before converting them to Parquet INT32 or INT64 columns, and return an error when a value does not fit.

### Are these changes tested?

- `go test ./parquet/pqarrow -run TestDecimalIntegerOverflow`

Values that do not fit the target Parquet integer type now return an error instead of being truncated.

### DIFF
--- a/parquet/pqarrow/decimal_overflow_test.go
+++ b/parquet/pqarrow/decimal_overflow_test.go
@@ -1,0 +1,134 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The two's-complement range windows are exactly right for all four conversions, and skipping null slots before validating (with the round-trip test proving garbage backing values stay harmless) is the crucial detail. Turning silently-truncating debug asserts into real `ErrInvalid` errors with the offending index is a clear correctness win, and the added `FitsInPrecision` enforcement matches the Parquet spec.

---

PR #1123, opened 2026-08-06
fix(arrow/array/arreflect): reject unrepresentable timestamps

fallintoplace:
### Rationale for this change

arreflect converts time.Time values through UnixNano for every timestamp unit. Dates outside the nanosecond range can therefore wrap instead of reporting that the value is not representable.

Use the target timestamp unit during conversion and check the integer range before appending the value.

### Are these changes tested?

- `go test ./arrow/array/arreflect`
- `go test ./arrow -run 'TestTimestamp|TestDate'`

Unrepresentable time.Time values now return an error instead of producing a wrapped timestamp.

### DIFF
--- a/arrow/array/arreflect/reflect_go_to_arrow.go
+++ b/arrow/array/arreflect/reflect_go_to_arrow.go
@@ -308,7 +308,11 @@ func appendTemporalValue(b array.Builder, v reflect.Value) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
The boundary math is exactly right — including the tricky truncated-division min case where `MinInt64` sits one second below `MinInt64/multiplier` with a fractional offset. The per-unit boundary tests at ±1 unit around `MaxInt64`/`MinInt64` and the negative fractional-second cases give me full confidence. Good fix.

---

PR #1124, opened 2026-08-06
fix(parquet/pqarrow): validate reader indexes

fallintoplace:
### Rationale for this change

Some Arrow Parquet reader paths index the field manifest or row-group metadata directly. Invalid input can panic even though the surrounding APIs return errors.

Validate field indexes in GetFieldReader and row-group indexes in GetFieldReader and ReadColumn before using them.

### Are these changes tested?

- `go test ./parquet/pqarrow -run TestFileReaderIndexValidation`

Invalid field and row-group indexes now return errors instead of panicking.


parquet/pqarrow/file_reader.go
```diff
@@ -219,6 +219,13 @@ func (fr *FileReader) allRowGroupFactory() itrFactory {
 //
 // IncludedLeaves and RowGroups are used to specify precisely which leaf indexes and row groups to read a subset of.
 func (fr *FileReader) GetFieldReader(ctx context.Context, i int, includedLeaves map[int]bool, rowGroups []int) (*ColumnReader, error) {
+	if i < 0 || i >= len(fr.Manifest.Fields) {
+		return nil, fmt.Errorf("%w: invalid field index chosen %d, there are only %d fields", arrow.ErrIndex, i, len(fr.Manifest.Fields))
```
zeroshade, line 223: This establishes `errors.Is(err, arrow.ErrIndex)` for `GetFieldReader`, but the identical bounds check in `getColumnReader` still returns a plain `fmt.Errorf`. That leaves `GetColumn` and `RowGroup(...).Column(...).Read` as the adjacent reader-index paths that cannot be classified with `errors.Is`.

Non-blocking for the panic fix, but could we wrap `arrow.ErrIndex` there too and add one assertion so the typed index-error contract is consistent?

zeroshade (commented):
The bounds checks are correctly placed before the direct field and row-group index accesses, converting previously possible panics into typed errors while preserving valid reads.

The focused regression and complete `parquet/pqarrow` package suite pass locally on head `d9988c9afc85`. One small, non-blocking consistency observation is included inline. RC and Test CI are still running.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. The findings below are observations, not blockers; an Apache Arrow Go maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
The follow-up wraps the remaining column-reader bounds error with `arrow.ErrIndex` and adds matching regression coverage, resolving the prior consistency observation.

Re-review at `10ddd13aa2d7`: the focused test passed 50 repetitions, the complete `parquet/pqarrow` package passed, and race plus 32-bit runs passed. All required CI checks are green.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer.*

PR #1125, opened 2026-08-06
fix(parquet/metadata): handle invalid row-group column indexes

fallintoplace:
### Rationale for this change

RowGroupMetaData.ColumnChunk returns an error but currently panics for invalid indexes. The column and offset index location helpers also index the underlying Thrift slice directly.

Return an arrow.ErrIndex error from ColumnChunk for invalid indexes, and return false from the location helpers when the index is outside the row-group range.

### Are these changes tested?

- `go test ./parquet/metadata`

Invalid row-group column indexes are handled as normal errors or missing locations instead of panics.


zeroshade (approved):
The row-group metadata accessors now handle negative and oversized column indexes without panicking, while preserving valid-index behavior. The new regression coverage exercises all three accessors, and the broader Parquet suite passes. Looks good.

PR #1126, opened 2026-08-06
fix(parquet/variant): validate compound value bounds

fallintoplace:
### Rationale for this change

NewWithMetadata accepts truncated arrays and objects because compound values are validated lazily. Accessing one of those values can then panic on a short offset table or an out-of-range child value.

Validate compound headers, offset tables, monotonic offsets, and nested value boundaries before returning a value while keeping the existing compound accessors lazy.

### Are these changes tested?

- `go test ./parquet/variant`

Malformed compound variant values now return an error during construction instead of panicking during access.

### DIFF
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -598,6 +598,9 @@ type validationFrame struct {
 
 const (
 	validationStackInlineCapacity = 32
+	// Values that exceed the inline stack commonly need only a modest amount
+	// of additional depth, so avoid allocating the maximum stack for them.
+	validationStackIntermediateCapacity = 128
 	// The root value is at depth zero, so the stack needs one more frame than
 	// the maximum allowed nesting depth. Keeping this storage fixed prevents
 	// untrusted values from growing the validation stack on the heap.
@@ -691,6 +694,9 @@ func validateValueLoop(meta Metadata, value []byte, stack []validationFrame, ran
 					return 0, fmt.Errorf("invalid variant value: maximum nesting depth exceeded")
 				}
 				if len(stack) == cap(stack) {
+					if cap(stack) == validationStackInlineCapacity {
+						return validateValueIntermediate(meta, value, stack, ranges, rangeTop)
+					}
 					return validateValueDeep(meta, value, stack, ranges, rangeTop)
 				}
 
@@ -739,6 +745,13 @@ func validateValueLoop(meta Metadata, value []byte, stack []validationFrame, ran
 	}
 
 	return 0, errors.New("invalid variant value: validation stack exhausted")
+}
+
+func validateValueIntermediate(meta Metadata, value []byte, initialStack []validationFrame, ranges []validationRange, rangeTop int) (int, error) {
+	var stackStorage [validationStackIntermediateCapacity]validationFrame
+	stack := stackStorage[:len(initialStack)]
+	copy(stack, initialStack)
+	return validateValueLoop(meta, value, stack, ranges, rangeTop)
 }
 
 func validateValueDeep(meta Metadata, value []byte, initialStack []validationFrame, ranges []validationRange, rangeTop int) (int, error) {
--- a/parquet/variant/variant.go
+++ b/parquet/variant/variant.go
@@ -342,11 +342,7 @@ func (v ArrayValue) Values() iter.Seq[Value] {
 			idx := uint32(v.offsetStart) + i*uint32(v.offsetSize)
 			offset := readLEU32(v.value[idx : idx+uint32(v.offsetSize)])
 
-			val := v.value[v.dataStart+offset:]
-			sz := valueSize(val)
-			val = val[:sz] // trim to actual size
-
-			if !yield(Value{value: val, meta: v.meta}) {
+			if !yield(trimValue(v.meta, v.value[v.dataStart+offset:])) {
 				return
 			}
 		}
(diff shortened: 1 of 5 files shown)


zeroshade:
Since this is a draft, i'll hold off on further review until it's marked ready

parquet/variant/variant.go
```diff
@@ -512,25 +513,25 @@ func NewWithMetadata(meta Metadata, value []byte) (Value, error) {
...
-		}
-		return nil
+func validateValueBytes(meta Metadata, value []byte) error {
+	size, err := validateValue(meta, value)
+	if err != nil {
+		return err
 	}
-	if basicTypeFromHeader(value[0]) != BasicPrimitive {
-		return nil
+	if size != len(value) {
```
zeroshade, line 526: **Blocking:** The exact-length check rejects byte slices returned by the public indexed accessors. `ArrayValue.Value`, `ObjectValue.ValueByKey`, and `ObjectValue.FieldAt` return an untrimmed slice extending through the parent’s remaining data, unlike the `Values` iterators. Reproduction: parse `{"a":1,"b":2}`, retrieve `"a"` with `ValueByKey`, then call `NewWithMetadata(v.Metadata(), field.Value.Bytes())`; this PR returns `invalid variant value: trailing bytes`. Please trim values returned by all indexed accessors to their encoded size and add round-trip tests for arrays and objects.

parquet/variant/variant.go
```diff
@@ -556,20 +557,331 @@ func validateScalarValue(value []byte) error {
...
+		if i > 0 && offset < offsets[i-1] {
+			return fmt.Errorf("invalid variant value: array offsets are not monotonic")
+		}
+		if dataStart+uint64(offset) > uint64(len(value)) || dataStart+uint64(offset) > math.MaxUint32 {
+			return fmt.Errorf("invalid variant value: array offset %d is out of range", offset)
+		}
+		offsets[i] = offset
+	}
+
+	frame.children = make([]validationChild, 0, len(offsets)-1)
```
zeroshade: **Blocking:** Validation eagerly allocates a `validationChild` for every child; objects subsequently allocate another `validationRange` per field. This happens on every `NewWithMetadata`, including the Parquet row-read path. A direct 40-field benchmark changed from approximately `4.9 ns/op, 0 B/op, 0 allocs/op` on main to `1.58 µs/op, 5576 B/op, 10 allocs/op`. Wide or deeply nested untrusted values therefore create substantial memory amplification and potential OOM pressure. Please iterate child offsets incrementally or otherwise bound/reuse validator state, and add representative allocation benchmarks.

zeroshade (changes requested):
Found two blocking regressions: indexed child values no longer round-trip because the strict trailing-byte check conflicts with untrimmed accessor results, and validation introduces substantial per-child allocation and memory amplification on the per-row read path.

---

>

parquet/variant/variant.go
```diff
@@ -556,23 +555,376 @@ func validateScalarValue(value []byte) error {
...
+					resultErr = err
+					hasResult = true
+					continue
+				}
+
+				frame.nextChild++
+				frame.pendingIndex = index
+				frame.pendingStart = start
+				frame.pendingExpectedSize = expectedSize
+				stack = append(stack, validationFrame{value: child})
```
zeroshade, line 717: **Blocking:** The common 40-child case is now allocation-free, but attacker-controlled nesting still grows this 120-byte frame stack with repeated slice reallocations. A valid 1,000-level nested array occupies 5,873 encoded bytes yet `NewWithMetadata` allocates about 393 KB; at 10,000 levels validation allocates about 6.2 MB. This leaves the prior deep-input memory-amplification/OOM concern unresolved on the row-read path. Please reduce or bound depth-dependent state and add a benchmark beyond `validationStackInlineCapacity`.

zeroshade (changes requested):
The indexed accessor round-trip issue is fixed, and the new 40-child benchmarks are allocation-free. One memory-amplification blocker remains for nesting beyond the inline stack capacity.

---

>

parquet/variant/variant.go
```diff
@@ -556,23 +556,406 @@ func validateScalarValue(value []byte) error {
...
+	for i := uint64(0); i <= uint64(numElements); i++ {
+		pos := offsetStart + uint64(i)*uint64(offsetSize)
+		offset := readLEU32(value[int(pos) : int(pos)+int(offsetSize)])
+		if i == 0 && offset != 0 {
+			return fmt.Errorf("invalid variant value: array first offset must be zero, got %d", offset)
+		}
+		if i > 0 && offset < previousOffset {
+			return fmt.Errorf("invalid variant value: array offsets are not monotonic")
+		}
+		if dataStart+uint64(offset) > uint64(len(value)) || dataStart+uint64(offset) > math.MaxUint32 {
```
zeroshade: **Blocking:** This treats `dataStart + offset` as though the complete encoded value were limited to `uint32`, but Variant's 4-byte field offsets are relative to the start of the fields region. A valid compound can therefore contain a `math.MaxUint32`-byte fields region plus its header and offset table. A sparse-mapped array with one max-sized binary child passes on the merge base but fails here with `array offset 4294967295 is out of range`. The equivalent object condition at line 907 has the same issue. Please retain total positions as `uint64` and compare them against `len(value)` without imposing this undocumented total-value limit.

zeroshade (changes requested):
The previous deep-validation memory-amplification blocker is resolved: 1,000-level array and object validation is allocation-free, and the repeated, race, boundary, vet, and fuzz checks passed locally. One correctness blocker remains for valid compounds whose complete encoding exceeds uint32 while their relative field offsets remain representable; details are inline.

---

>

zeroshade (approved):
The prior large-compound blocker is resolved. Validation and accessors now preserve compound sizes beyond uint32, with regression coverage for both arrays and objects. I also verified the variant suite under repeated and race runs, fuzzed the parser, cross-compiled for Linux/386 and Windows/amd64, and confirmed the validation benchmarks remain allocation-free.

PR #1127, opened 2026-08-06
fix(compute): detect temporal rounding overflow

fallintoplace:
### Rationale for this change

Temporal rounding can overflow while converting timestamps to nanoseconds, calculating the rounding interval, or multiplying a rounded quotient. The current unchecked arithmetic can return an incorrect timestamp.

Use checked arithmetic for temporal conversions, interval setup, rounding results, calendar conversion, and origin adjustments. Propagate overflow as an Arrow invalid-value error.

### Are these changes tested?

- `go test ./arrow/compute`

Inputs and options that cannot be represented during temporal rounding now return an error instead of a wrapped result.

### DIFF
--- a/arrow/compute/internal/kernels/rounding.go
+++ b/arrow/compute/internal/kernels/rounding.go
@@ -1447,16 +1447,6 @@ func checkedCalendarAddDays(value time.Time, days int64) (time.Time, error) {
 	return result, nil
 }
 
-// convertToNanos converts a timestamp value to nanoseconds.
-func convertToNanos(ts int64, unit arrow.TimeUnit) (int64, error) {
-	return checkedMulInt64(ts, int64(unit.Multiplier()))
-}
-
-// convertFromNanos converts a nanosecond timestamp to the specified unit
-func convertFromNanos(nanos int64, unit arrow.TimeUnit) int64 {
-	return nanos / int64(unit.Multiplier())
-}
-
 func roundToMultipleInt64(value, multiple int64, mode RoundMode, strictCeil bool) (int64, error) {
 	if multiple == 0 || value%multiple == 0 {
 		if strictCeil && mode == RoundUp {
--- a/arrow/compute/internal/kernels/rounding.go
+++ b/arrow/compute/internal/kernels/rounding.go
@@ -924,6 +924,16 @@ func roundTimestamp(ts int64, inputUnit arrow.TimeUnit, tz *time.Location, opts 
 		return roundTimestampCalendar(ts, inputUnit, tz, opts)
 	}
 
+	// Zoned timestamps are rounded in their local wall-clock timeline. This is
+	// observable for offsets that are not whole hours (for example, Kathmandu)
+	// and around daylight-saving transitions.
+	if tz != time.UTC && tz.String() != "UTC" {
+		if opts.CalendarBasedOrigin {
+			return roundTimestampCalendarOrigin(ts, inputUnit, tz, opts)
+		}
+		return roundTimestampFixedTimezone(ts, inputUnit, tz, opts)
+	}
+
 	// CalendarBasedOrigin rounds fixed-duration units against the preceding
 	// calendar boundary (for example, minutes from the start of the hour).
 	// Use local wall-clock arithmetic here: time.Time.Sub measures elapsed time
@@ -975,9 +985,9 @@ func canRoundInInputUnit(inputUnit arrow.TimeUnit, roundingIntervalNanos int64) 
 }
 
 // roundTimestampCalendarOrigin rounds a fixed-duration unit relative to the
-// preceding calendar boundary. The calculation is performed in local wall
-// time so daylight-saving transitions do not change the number of clock hours,
-// minutes, or seconds in the current period.
+// preceding calendar boundary. Candidate boundaries are calculated in local
+// wall time, then compared as system times so daylight-saving transitions are
+// handled consistently with the temporal rounding reference implementation.
 func roundTimestampCalendarOrigin(ts int64, inputUnit arrow.TimeUnit, tz *time.Location, opts roundTemporalState) (int64, error) {
 	t, err := timestampToCalendarTime(ts, inputUnit, tz)
 	if err != nil {
@@ -987,29 +997,52 @@ func roundTimestampCalendarOrigin(ts int64, inputUnit arrow.TimeUnit, tz *time.L
 	if err != nil {
 		return 0, err
 	}
+	return roundTimestampFromWallOrigin(t, inputUnit, tz, opts, origin)
+}
 
-	elapsed, err := wallClockDifference(t, origin)
+func roundTimestampFixedTimezone(ts int64, inputUnit arrow.TimeUnit, tz *time.Location, opts roundTemporalState) (int64, error) {
+	t, err := timestampToCalendarTime(ts, inputUnit, tz)
 	if err != nil {
 		return 0, err
 	}
-	roundedElapsed, err := roundToMultipleInt64(
-		elapsed, opts.roundingInterval, opts.mode, opts.CeilIsStrictlyGreater,
-	)
+	origin := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
+	return roundTimestampFromWallOrigin(t, inputUnit, tz, opts, origin)
+}
+
+func roundTimestampFromWallOrigin(t time.Time, inputUnit arrow.TimeUnit, tz *time.Location, opts roundTemporalState, origin time.Time) (int64, error) {
+	inputTime := t
+	wallTime := calendarWallTime(t)
+
+	elapsed, err := wallClockDifference(wallTime, origin)
 	if err != nil {
 		return 0, err
 	}
-	rounded, err := addWallNanos(origin, roundedElapsed, tz)
+	flooredElapsed, err := roundToMultipleInt64(elapsed, opts.roundingInterval, RoundDown, false)
 	if err != nil {
 		return 0, err
 	}
-	// time.Date can select either occurrence of a repeated local time. Prefer
-	// the input's offset when it also represents the requested clock time.
-	_, inputOffset := t.Zone()
-	_, roundedOffset := rounded.Zone()
-	if inputOffset != roundedOffset {
-		candidate := rounded.Add(time.Duration(roundedOffset-inputOffset) * time.Second)
-		if difference, err := wallClockDifference(candidate, rounded); err == nil && difference == 0 {
-			rounded = candidate
+	floor, err := addWallNanos(origin, flooredElapsed, tz)
+	if err != nil {
+		return 0, err
+	}
+
+	var rounded time.Time
+	switch opts.mode {
+	case RoundDown:
+		rounded = floor
+	case RoundUp:
+		rounded, err = ceilCalendarOrigin(inputTime, floor, opts, tz)
+		if err != nil {
+			return 0, err
+		}
+	default:
+		ceil, ceilErr := ceilCalendarOrigin(inputTime, floor, opts, tz)
+		if ceilErr != nil {
+			return 0, ceilErr
+		}
+		rounded, err = halfRoundPeriod(inputTime, floor, ceil)
+		if err != nil {
+			return 0, err
 		}
 	}
 
@@ -1018,6 +1051,27 @@ func roundTimestampCalendarOrigin(ts int64, inputUnit arrow.TimeUnit, tz *time.L
 		return 0, err
 	}
 	return int64(result), nil
+}
+
+func ceilCalendarOrigin(value, floor time.Time, opts roundTemporalState, tz *time.Location) (time.Time, error) {
+	// Convert the floored system time back to local wall time before deciding
+	// whether to advance. This matters when the period crosses a timezone
+	// transition because the elapsed system duration need not equal the wall
+	// clock duration.
+	floorWall := calendarWallTime(floor.In(tz))
+	cs, err := checkedLocalTime(floorWall, tz)
+	if err != nil {
+		return time.Time{}, err
+	}
+	if !opts.CeilIsStrictlyGreater && !cs.Before(value) {
+		return cs, nil
+	}
+
+	nextWall, err := addWallNanos(floorWall, opts.roundingInterval, tz)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return nextWall, nil
 }
 
 // Keep the origin's civil fields in UTC because its local time may not exist.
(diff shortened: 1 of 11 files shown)


zeroshade:
Since this is a draft, i'll hold off on further review until it's marked ready and the conflicts are resolved

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1056,90 +1198,185 @@ func roundTimestampCalendar(tsNanos int64, inputUnit arrow.TimeUnit, tz *time.Lo
...
-			yearEnd := time.Date(nextYear, 1, 1, 0, 0, 0, 0, tz)
+			yearStart, dateErr := checkedCalendarDate(roundedYear, 1, tz)
+			if dateErr != nil {
+				return 0, dateErr
+			}
+			nextYear, dateErr := checkedAddInt64(roundedYear, multiple)
+			if dateErr != nil {
+				return 0, dateErr
+			}
+			yearEnd, dateErr := checkedCalendarDate(nextYear, 1, tz)
```
zeroshade, line 2027: **Blocking:** Restricting this intermediate period boundary to the nanosecond timestamp window rejects inputs whose selected result is representable. For `timestamp[ns]` `2262-01-05` rounded half-up to one year, the correct result is `2262-01-01`, but this checks the unused `2263-01-01` boundary and returns overflow. Likewise, `2201-01-01` rounded to 100 years should return `2200-01-01`, but checking the 2300 boundary fails. Main succeeds in both cases. Please permit calendar boundaries outside the output window and validate only the selected result with `timeToNanos`; cover year, quarter, and month boundaries.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -959,22 +983,104 @@ func canRoundInInputUnit(inputUnit arrow.TimeUnit, roundingIntervalNanos int64)
...
+	}
+	result := value.AddDate(0, 0, dayOffset)
+	if result.Year() < minNanosecondTimestampYear || result.Year() > maxNanosecondTimestampYear {
+		return time.Time{}, overflowError()
+	}
+	return result, nil
+}
+
+// convertToNanos converts a timestamp value to nanoseconds.
+func convertToNanos(ts int64, unit arrow.TimeUnit) (int64, error) {
```
zeroshade: **Blocking:** Every calendar path is still narrowed to an `int64` nanosecond intermediate, even when the input and correct result are representable in their original type. For example, flooring `Date32(1500-06-15)` to one year should return `1500-01-01`, but this PR returns overflow. The same affects Date64 and second/millisecond/microsecond timestamps, whose ranges are much wider than timestamp nanoseconds. Please perform calendar arithmetic without narrowing to nanoseconds and range-check the final result in the input unit.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1056,90 +1198,185 @@ func roundTimestampCalendar(tsNanos int64, inputUnit arrow.TimeUnit, tz *time.Lo
...
+			}
+			nextYear, dateErr := checkedAddInt64(roundedYear, multiple)
+			if dateErr != nil {
+				return 0, dateErr
+			}
+			yearEnd, dateErr := checkedCalendarDate(nextYear, 1, tz)
+			if dateErr != nil {
+				return 0, dateErr
+			}
 			rounded = halfRoundPeriod(t, yearStart, yearEnd)
```
zeroshade: **Blocking:** `halfRoundPeriod` computes `periodEnd.Sub(periodStart)/2`; `time.Time.Sub` saturates at `math.MaxInt64` for periods longer than about 292 years. Consequently, rounding `1948-01-01` to a 300-year multiple uses a midpoint in 1946 instead of the true 1950 midpoint and returns 2100 rather than 1800. This is pre-existing, but it is an unchecked calendar-rounding overflow directly within this PR’s scope. Please compute calendar midpoints without converting the entire period to `time.Duration`, with a regression test for a period over 292 years.

zeroshade (changes requested):
Found three blocking calendar-rounding issues: valid nanosecond results are rejected because helper boundaries are range-checked, wider temporal types are unnecessarily narrowed to nanoseconds, and long calendar periods still silently overflow through `time.Duration` midpoint arithmetic.

---

>

arrow/compute/internal/kernels/rounding.go
```diff
@@ -912,43 +916,55 @@ func roundTimestamp(ts int64, inputUnit arrow.TimeUnit, tz *time.Location, opts
...
 	if canRoundInInputUnit(inputUnit, opts.unitNanos) && !opts.useCalendarOrigin {
 		intervalInInputUnit := opts.roundingInterval / int64(inputUnit.Multiplier())
-		rounded := roundToMultipleInt64(ts, intervalInInputUnit, opts.mode, opts.CeilIsStrictlyGreater)
-		return rounded, nil
+		return roundToMultipleInt64(ts, intervalInInputUnit, opts.mode, opts.CeilIsStrictlyGreater)
 	}
 
 	// Slow path: convert to nanoseconds for calendar origin or incompatible units
-	tsNanos := convertToNanos(ts, inputUnit)
+	tsNanos, err := convertToNanos(ts, inputUnit)
```
zeroshade: **Blocking:** The fixed-duration slow path still rejects representable wider-unit values when the requested rounding is finer than the input resolution. For example, flooring a second-resolution `1500-06-15 12:34:56` timestamp to one nanosecond is an exact no-op, but this conversion returns `invalid: temporal rounding overflow`. Please avoid narrowing when the input is already aligned to the requested interval, and add coverage outside the timestamp-nanosecond window.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1005,142 +1102,308 @@ func roundToMultipleInt64(value, multiple int64, mode RoundMode, strictCeil bool
...
-	midPoint := periodStart.Add(periodEnd.Sub(periodStart) / 2)
+// halfRoundPeriod performs half-rounding by finding the midpoint between period start and end.
+// It does not use time.Time.Sub because that method saturates for periods longer than
+// approximately 292 years.
+func halfRoundPeriod(t, periodStart, periodEnd time.Time) (time.Time, error) {
+	if periodEnd.Before(periodStart) {
+		return time.Time{}, overflowError()
+	}
+
+	startSeconds := periodStart.Unix()
```
zeroshade: **Blocking:** `time.Time.Unix()` is undefined once a boundary lies outside the `int64` seconds range, so this still rejects a representable selected result at the edge of `timestamp[s]`. With an input of January 2 in the year containing `math.MaxInt64` seconds, half-up rounding to one year should select that year's representable January 1; the next-year boundary exceeds `int64`, these calls wrap, and `checkedSubInt64` returns overflow. Please compute the midpoint without requiring both boundaries to fit Unix seconds and cover this edge.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1160,23 +1423,58 @@ func roundTimestampCalendar(tsNanos int64, inputUnit arrow.TimeUnit, tz *time.Lo
 		epochWeekStart := epochInTz.AddDate(0, 0, -epochWeekday)
 		epochWeekStart = time.Date(epochWeekStart.Year(), epochWeekStart.Month(), epochWeekStart.Day(), 0, 0, 0, 0, tz)
 
-		daysSinceEpochWeek := int(startOfWeek.Sub(epochWeekStart).Hours() / 24)
+		daysSinceEpochWeek := int64(startOfWeek.Sub(epochWeekStart).Hours() / 24)
```
zeroshade: **Blocking:** This reintroduces the same approximately 292-year saturation through `time.Time.Sub`. Flooring a second-resolution `1500-06-15 12:34:56` timestamp to one week returns `1677-09-19` instead of `1500-06-10` because the day distance saturates before division. Please calculate calendar-day/week distance without `time.Duration`, with wide-range week regression coverage.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1187,22 +1485,42 @@ func roundTimestampCalendar(tsNanos int64, inputUnit arrow.TimeUnit, tz *time.Lo
...
 
 	default:
 		return 0, fmt.Errorf("%w: unsupported calendar unit", arrow.ErrNotImplemented)
 	}
+	if err != nil {
+		return 0, err
+	}
+
+	// Convert back to the input unit, validating only the selected result.
+	roundedTimestamp, err := arrow.TimestampFromTime(rounded, inputUnit)
```
zeroshade: **Blocking:** This does not actually validate second-resolution results: `arrow.TimestampFromTime` returns `val.Unix()` without a range check for `arrow.Second`. Ceiling `math.MaxInt64` seconds to the next year therefore succeeds with wrapped value `-9223372036852412416` instead of returning overflow. Please explicitly validate second-unit conversion (or make `TimestampFromTime` do so) and add a boundary test.

zeroshade (changes requested):
Thanks for the update. The earlier nanosecond-boundary and long-period midpoint examples are fixed, and the test suite passes locally. I found four remaining overflow/range cases that still need correction before this is safe across the supported temporal ranges.

zeroshade (changes requested):
Thanks — all seven items from my previous rounds check out. I verified each against `c78988c2` end-to-end rather than going off the test assertions, and they're genuinely fixed. I'll resolve those threads.

Two cases in the same class are still open, both in second-unit calendar rounding:

```
floor(Timestamp(math.MinInt64), unit=s, 1 year) → +9223372036825516800   err=nil
ceil (Timestamp(math.MaxInt64), unit=s, 1 day)  → -9223372036854745216   err=nil
```

A property sweep on the ordering invariant `floor(x) ≤ x ≤ ceil(x)` (4 units × 11 rounding units × 6 multiples) turns up 86 violations, all confined to `unit=s` at the int64 extremes, across year/quarter/month/day. The affected input range is exact, binary-searched rather than estimated: `sec ∈ [MinInt64, -9223372036852412417]`. The first value above that range correctly errors.

**Root cause.** The guard tests `Year() > maxTimestampSecondYear`, but the true maximum instant is `292277026596-12-04T15:30:07` — wrapped values that land later *within that same year* slip past it. The wrap is also bijective (`tt.Unix()` returns the original input), so the round-trip check can't see it either. That's exactly why the max-side year case (C7) is caught and the day case isn't.

Two dead ends worth passing on so you don't spend time on them:

- **Don't compare instants.** `time.Unix(math.MaxInt64, 0)` has an internally-overflowed `ext` field, so `now.After(time.Unix(MaxInt64,0))` returns `true` and `.Sub()` returns garbage; its `Year()` only looks sane via a second wrap. Substituting an `.After()` check rejects ordinary timestamps and still leaves all 86 violations. The guard has to run in integer seconds at input conversion, before any `time.Time` exists.
- **Don't put it in `TimestampFromTime`.** That breaks `TestTimestampFromTimeBoundaries/s`, which asserts the wrapped value round-trips — the public API needs to stay bijective. The representability decision belongs in the kernel, which also avoids the blast radius on `arrow/extensions/timestamp_with_offset.go:253`, where the error is discarded into `_`.

Scope the input guard to the calendar path only — a blanket rejection breaks the `sec`/`ms`/`us`/`ns` no-ops that C4 asked for.

To be clear about severity: **both of these are pre-existing on main**, which returns `0` and `-31536000` silently for the same inputs. This PR is a strict improvement, not a regression. I'm asking for them because they're the precise failure mode in the PR title, and the guard you already added covers the max-year half of it.

Please also add regression tests for `Second`+`MinInt64` and `Second`+`MaxInt64`-with-day. The current matrix is asymmetric, which is why this slipped through: `TestTemporalRoundingCalendarOriginAtMinimum` uses `MinInt64` but only in nanoseconds (year 1677, safe), and the second-unit tests only cover the max side. There is no Second+MinInt64 case — the one combination that's broken.

Two smaller things:

- The Lint job is failing only on formatting — 0 actual issues, the hook just rewrote files. `gofmt -w arrow/compute/temporal_rounding_overflow_test.go` clears it (the struct literal in `TestTemporalRoundingNegativeCalendarMultiples` is over-indented).
- Timezone handling is clean: I swept 7 zones including DST transitions, half-hour offsets and +14, and floor-to-day lands on local midnight in every one (0/161 non-midnight). Date32/Date64 extremes are clean too, and the Time32/Time64 wrap-to-midnight behaviour is byte-identical to the merge base, so it isn't this PR's concern.

On the diff size: it's justified. 55% is tests, there's no new exported API, the hunks are confined to the temporal rounding functions, and the restructuring was something I asked for in C2/C3/C6. Error-on-overflow via `arrow.ErrInvalid` matches Arrow C++, which returns `Invalid` with no null, saturating, or `_checked` variant.

arrow/compute/internal/kernels/rounding.go
```diff
@@ -1005,145 +1280,592 @@ func roundToMultipleInt64(value, multiple int64, mode RoundMode, strictCeil bool
...
-	if t.Before(midPoint) {
-		return periodStart
+func floorCalendarWeek(value time.Time, multiple int64, weekStartsMonday bool, tz *time.Location) (time.Time, error) {
+	offsetDays := int64(4)
+	targetWeekday := time.Wednesday
+	if weekStartsMonday {
+		offsetDays = 3
+		targetWeekday = time.Thursday
+	}
+	shifted, err := checkedCalendarAddDays(value, offsetDays)
```
zeroshade: Calendar-origin week rounding can return a floor later than its input. For example, flooring `1078-06-14 04:11:08 UTC` to a two-week interval starting Monday returns `1078-06-17 00:00:00 UTC`. The bucket is calculated from the shifted value without restoring the original-week position. Please correct the shift handling and add tests covering mid-week inputs, multiples greater than one, both week-start modes, and `floor <= input <= ceil`.

zeroshade (changes requested):
The findings from the previous reviews are fixed at the current head: selected-result bounds, non-nanosecond conversion, midpoint arithmetic, finer-unit no-op handling, second boundaries and extremes, and wide-range weekly calculation now behave correctly.

Calendar-origin week rounding still has an ordering violation for multi-week intervals. A focused public-API reproducer and randomized temporal probe both demonstrate the issue; details are inline.

Focused compute and race tests otherwise pass. CI is still running.

---

>

zeroshade (changes requested):
A valid wide zoned timestamp is still rejected during no-op nanosecond rounding. For example, a post-2262 `timestamp[s, tz=Asia/Kathmandu]` value that is already nanosecond-aligned returns a temporal-overflow error because the local wall-clock distance is unnecessarily required to fit in an `int64` nanosecond intermediate.

Please retain the wide wall-clock arithmetic and validate only the selected result when converting back to the input unit. Add pre-1677 and post-2262 zoned second-resolution cases for both origin modes.

---

>

zeroshade (approved):
The wide zoned no-op rounding issue from my previous review is fixed exactly as requested: the premature int64-nanosecond wall-boundary validation is removed and validation now happens at input resolution. The new pre-1677/post-2262 `Asia/Kathmandu` no-op cases cover both origin modes, and the full compute suite passes locally.

---

zeroshade:
Looks like there's a lint issue that needs to get fixed

zeroshade:
The lint issue is still there

PR #1128, opened 2026-08-06
fix(parquet/pqarrow): reject out-of-range INT96 timestamps

fallintoplace:
### Rationale for this change

transferInt96 converts Parquet INT96 values through time.Time.UnixNano without checking the int64 nanosecond range. Very wide dates can therefore wrap when read as Arrow timestamp[ns].

Add checked INT96 to nanosecond conversion, propagate conversion errors through the Parquet reader, and normalize Int96.ToTime without overflowing before the range check.

### Are these changes tested?

- `go test ./parquet/pqarrow -run 'Test(ArrowTimestampToImpalaTimestamp|ReadInt96RejectsOutOfRangeTimestamp|WriteArrowInt96)$'`

In-range INT96 behavior is unchanged. Out-of-range INT96 timestamps now return an error instead of being wrapped.

### DIFF
--- a/parquet/types.go
+++ b/parquet/types.go
@@ -18,7 +18,9 @@ package parquet
 
 import (
 	"encoding/binary"
+	"fmt"
 	"io"
+	"math"
 	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


parquet/types.go
```diff
@@ -90,6 +92,34 @@ func (i96 Int96) ToTime() time.Time {
 	return t.UTC()
 }
 
+// ToTimestamp converts an Int96 value to an Arrow nanosecond timestamp.
+func (i96 Int96) ToTimestamp() (arrow.Timestamp, error) {
```
zeroshade, line 97: Non-blocking: the PR description says this also normalizes `Int96.ToTime`, but the final diff leaves `ToTime` unchanged. It still silently wraps dates outside the int64-nanosecond range while this new sibling method correctly returns an error.

For example, an INT96 representing `3000-01-01` produces `1830-11-23` from `ToTime()`, despite its documentation promising the same time instant; `String()` inherits that behavior.

This does not affect the corrected `pqarrow` path, but could we either fix `ToTime` using day/second arithmetic or remove the claim from the PR description and track it separately?

zeroshade (approved):
The checked INT96-to-nanosecond conversion looks correct. I verified the minimum and maximum boundary arithmetic, malformed nanoseconds-of-day handling, null bitmap alignment, error propagation, and buffer ownership.

The focused INT96 suites pass locally, and CI is fully green. I left one non-blocking inline observation about the unchanged public `Int96.ToTime` behavior and the corresponding claim in the PR description.

---

> will follow up.*
>

PR #1131, opened 2026-08-07
fix(arrow): reject duplicate union type codes

fallintoplace:
### Rationale for this change

Duplicate union type codes overwrite a child lookup entry during construction, leaving one child unreachable.

Reject duplicate type codes during union validation and add coverage for both sparse and dense union types.

### Are these changes tested?

- `go test ./arrow`

Union types with duplicate type codes are now rejected during construction instead of silently making one child unreachable.

### DIFF
--- a/arrow/datatype_nested.go
+++ b/arrow/datatype_nested.go
@@ -734,15 +734,15 @@ func (t *unionType) validate(fields []Field, typeCodes []UnionTypeCode, _ UnionM
 		return errors.New("arrow: union types should have the same number of fields as type codes")
 	}
 
-	seen := make(map[UnionTypeCode]struct{}, len(typeCodes))
+	var seen [int(MaxUnionTypeCode) + 1]bool
 	for _, c := range typeCodes {
 		if c < 0 || c > MaxUnionTypeCode {
 			return errors.New("arrow: union type code out of bounds")
 		}
-		if _, ok := seen[c]; ok {
+		if seen[c] {
 			return errors.New("arrow: union type codes must be unique")
 		}
-		seen[c] = struct{}{}
+		seen[c] = true
 	}
 	return nil
 }
--- a/arrow/datatype_nested.go
+++ b/arrow/datatype_nested.go
@@ -734,10 +734,15 @@ func (t *unionType) validate(fields []Field, typeCodes []UnionTypeCode, _ UnionM
 		return errors.New("arrow: union types should have the same number of fields as type codes")
 	}
 
+	seen := make(map[UnionTypeCode]struct{}, len(typeCodes))
 	for _, c := range typeCodes {
 		if c < 0 || c > MaxUnionTypeCode {
 			return errors.New("arrow: union type code out of bounds")
 		}
+		if _, ok := seen[c]; ok {
+			return errors.New("arrow: union type codes must be unique")
+		}
+		seen[c] = struct{}{}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


arrow/datatype_nested.go
```diff
@@ -734,10 +734,15 @@ func (t *unionType) validate(fields []Field, typeCodes []UnionTypeCode, _ UnionM
 		return errors.New("arrow: union types should have the same number of fields as type codes")
 	}
 
+	seen := make(map[UnionTypeCode]struct{}, len(typeCodes))
```
zeroshade: Minor: this allocates a map on every union construction, but the key space is small and fixed — `MaxUnionTypeCode` is 127, and the loop above already rejects anything outside `0..MaxUnionTypeCode` before this point.

A stack array avoids the allocation entirely:

```go
var seen [MaxUnionTypeCode + 1]bool
for _, c := range typeCodes {
	if c < 0 || c > MaxUnionTypeCode {
		return errors.New("arrow: union type code out of bounds")
	}
	if seen[c] {
		return errors.New("arrow: union type codes must be unique")
	}
	seen[c] = true
}
```

There's precedent right in this type — `unionType` already carries `childIDs [int(MaxUnionTypeCode) + 1]int` for the same reason. 128 bytes on the stack versus a heap map is a reasonable trade for a function on the type-construction path.

zeroshade (commented):
The validation is correct — the Arrow spec requires union type codes to be unique, and duplicates silently made one child unreachable through the `childIDs` lookup, so rejecting them is right.

One suggestion on the implementation, inline.

I'll follow up separately on a broader point this surfaces about how these validation errors reach callers reading untrusted input.

---

zeroshade:
Separate from the review above, and **not** something I'd block this PR on — but worth capturing while it's fresh.

This change widens a panic path that's reachable from untrusted input. `arrow/ipc/metadata.go:806` builds union types directly from IPC file metadata:

```go
return arrow.UnionOf(mode, children, typeIDs), nil
```

`UnionOf` → `SparseUnionOf`/`DenseUnionOf` → `validate`, and on any validation failure the constructor calls `panic(err)`. There's no non-panicking union constructor. So an IPC file carrying `typeIDs: [1, 1]` will now panic during schema deserialization rather than loading. Anyone parsing untrusted Arrow data gets a crash where they previously got a schema that was wrong but non-fatal.

Two things sharpen it. The enclosing function is `concreteTypeFromFB(...) (arrow.DataType, error)` — it returns errors — and the code immediately above the call already rejects one class of malformed type ID gracefully:

```go
if int32(code) != id {
    return nil, errors.New("union type id out of bounds")
}
```

So the function error-checks one case and then hands the rest to a constructor that panics.

To be clear about attribution: **this hazard pre-exists this PR.** A negative type ID already survives that `int32(code) != id` check and then panics on `c < 0` inside `validate`. This change doesn't introduce the pattern, it just adds an easier trigger, since `[1, 1]` is two perfectly valid in-range `int8` values.

The fix belongs in a separate change, roughly one of:

1. Add an error-returning union constructor (`NewSparseUnionType` / `NewDenseUnionType` returning `(T, error)`) and have `metadata.go` use it, keeping the panicking `*Of` helpers for statically-known types.
2. Or have `concreteTypeFromFB` validate type codes itself before calling `UnionOf`, consistent with the bounds check it already performs.

Option 1 is more thorough, since the panic is reachable from any deserialization path that builds types from file metadata, not just this one call site.

I'll file this as an issue — flagging it here so the context isn't lost.

---

_Note: this comment was drafted with AI assistance by a maintainer and may contain mistakes. If anything here looks wrong, say so and I'll take another look._

zeroshade (approved):
The follow-up replaces the temporary map with the fixed-size type-code array while preserving bounds-first validation. Duplicate codes are rejected for both union modes, and the full Arrow test suite passes. Looks good.

PR #1132, opened 2026-08-07
fix(memory): handle zero-length checked reallocations

fallintoplace:
### Rationale for this change

CheckedAllocator.Reallocate panics when called with nil, and zero-size reallocations leave stale entries in the leak-tracking map.

Handle nil input buffers and remove the old allocation record when a reallocation reduces the size to zero. Add coverage for both paths.

### Are these changes tested?

- `go test ./arrow/memory`

CheckedAllocator no longer panics or reports a false allocation leak for these reallocation paths. The allocator interface is unchanged.

### DIFF
--- a/arrow/memory/buffer_test.go
+++ b/arrow/memory/buffer_test.go
@@ -58,6 +58,17 @@ func TestNewResizableBuffer(t *testing.T) {
 	assert.Zero(t, buf.Len())
 }
 
+func TestCheckedAllocatorReallocate(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	buf := mem.Reallocate(16, nil)
+	assert.Len(t, buf, 16)
+
+	buf = mem.Reallocate(0, buf)
+	assert.Empty(t, buf)
+}
+
 func TestBufferReset(t *testing.T) {
 	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
 	defer mem.AssertSize(t, 0)
--- a/arrow/memory/checked_allocator.go
+++ b/arrow/memory/checked_allocator.go
@@ -69,21 +69,18 @@ func (a *CheckedAllocator) Reallocate(size int, b []byte) []byte {
 	a.sz.Add(int64(size - len(b)))
 
 	var oldptr uintptr
-	if cap(b) > 0 {
-		oldptr = uintptr(unsafe.Pointer(&b[:1][0]))
+	if len(b) > 0 {
+		oldptr = uintptr(unsafe.Pointer(unsafe.SliceData(b)))
 	}
 	out := a.mem.Reallocate(size, b)
+	if oldptr != 0 {
+		a.allocs.Delete(oldptr)
+	}
 	if size == 0 {
-		if oldptr != 0 {
-			a.allocs.Delete(oldptr)
-		}
 		return out
 	}
 
 	newptr := uintptr(unsafe.Pointer(&out[0]))
-	if oldptr != 0 {
-		a.allocs.Delete(oldptr)
-	}
 	pcs := make([]uintptr, maxRetainedFrames)
 
 	// For historical reasons the meaning of the skip argument
--- a/arrow/memory/checked_allocator.go
+++ b/arrow/memory/checked_allocator.go
@@ -68,14 +68,22 @@ func (a *CheckedAllocator) Allocate(size int) []byte {
 func (a *CheckedAllocator) Reallocate(size int, b []byte) []byte {
 	a.sz.Add(int64(size - len(b)))
 
-	oldptr := uintptr(unsafe.Pointer(&b[0]))
+	var oldptr uintptr
+	if cap(b) > 0 {
+		oldptr = uintptr(unsafe.Pointer(&b[:1][0]))
+	}
 	out := a.mem.Reallocate(size, b)
 	if size == 0 {
+		if oldptr != 0 {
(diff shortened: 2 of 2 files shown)


arrow/memory/checked_allocator.go
```diff
@@ -68,14 +68,22 @@ func (a *CheckedAllocator) Allocate(size int) []byte {
 func (a *CheckedAllocator) Reallocate(size int, b []byte) []byte {
 	a.sz.Add(int64(size - len(b)))
 
-	oldptr := uintptr(unsafe.Pointer(&b[0]))
+	var oldptr uintptr
+	if cap(b) > 0 {
+		oldptr = uintptr(unsafe.Pointer(&b[:1][0]))
```
zeroshade: `&b[:1][0]` works, but it's a reslice trick that makes a reader stop and work out why it doesn't panic. `unsafe.SliceData` states the intent directly, is already used in this repo (`internal/utils/min_max_neon_arm64.go` and friends), and the `go 1.25.0` baseline has it.

```suggestion
		oldptr = uintptr(unsafe.Pointer(unsafe.SliceData(b)))
```

Keep the `cap(b) > 0` guard either way — for a non-nil zero-capacity slice `SliceData` returns an unspecified non-nil pointer, so the guard is still doing real work.

arrow/memory/checked_allocator.go
```diff
@@ -68,14 +68,22 @@ func (a *CheckedAllocator) Allocate(size int) []byte {
...
+			a.allocs.Delete(oldptr)
+		}
 		return out
 	}
 
 	newptr := uintptr(unsafe.Pointer(&out[0]))
-	a.allocs.Delete(oldptr)
+	if oldptr != 0 {
+		a.allocs.Delete(oldptr)
+	}
```
zeroshade: The delete now happens on both paths, so it can hoist above the `size == 0` check and stop being duplicated:

```suggestion
	out := a.mem.Reallocate(size, b)
	if oldptr != 0 {
		a.allocs.Delete(oldptr)
	}
	if size == 0 {
		return out
	}

	newptr := uintptr(unsafe.Pointer(&out[0]))
```

Exactly equivalent, one copy instead of two.

arrow/memory/buffer_test.go
```diff
@@ -58,6 +58,17 @@ func TestNewResizableBuffer(t *testing.T) {
...
 
+func TestCheckedAllocatorReallocate(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
+	defer mem.AssertSize(t, 0)
+
+	buf := mem.Reallocate(16, nil)
+	assert.Len(t, buf, 16)
+
+	buf = mem.Reallocate(0, buf)
+	assert.Empty(t, buf)
```
zeroshade, line 69: Coverage gap, and a question that comes with it.

These cases cover `Reallocate(16, nil)` (cap 0) and `Reallocate(0, buf)`, but not **len 0 with cap > 0** — which is precisely the case the `cap(b) > 0` guard exists for, since plain `&b[0]` panics there too.

Before adding that case, though, it's worth checking whether such a slice can actually reach here, because the guard and the accounting disagree. The guard keys off `cap(b)`, but the first line of the function does:

```go
a.sz.Add(int64(size - len(b)))
```

which keys off `len(b)`. If a slice with `len 0, cap 16` arrives, the pointer handling is now correct but `sz` under-reports the old allocation by 16 bytes, so the tracked total drifts and `AssertSize` fails for a different reason. I tried writing the obvious test for this and it fails on the accounting, not the pointer.

So I think it's one of two things:

- such slices can't reach `Reallocate`, in which case `cap(b) > 0` could just as well be `len(b) > 0` and the intent is clearer; or
- they can, in which case the accounting needs the same treatment and there's a second bug here.

No need to solve it in this PR either way — but the answer decides whether the extra test case is worth adding now.

zeroshade (commented):
Both bugs are real and the fix is correct.

`&b[0]` panics outright on a nil or empty slice, and the `size == 0` early return skipped the `allocs.Delete`, leaving a stale tracking entry.

That second one is worse than the description suggests. `AssertSize` doesn't just compare byte counts — it ranges over `a.allocs` and emits `t.Errorf("LEAK of %d bytes FROM…")` with a full stack trace for every surviving entry. So before this fix, any `Reallocate(0, buf)` would make a later `AssertSize` report a spurious leak, which is unpleasant to debug when it surfaces in someone else's test. The new `defer mem.AssertSize(t, 0)` genuinely exercises that.

It also restores a real invariant: `Allocate(0)` returns early without storing an entry, so zero-size allocations were never meant to be tracked. `Reallocate(0, …)` now removes any entry, making the two consistent.

Three suggestions inline — two are directly applyable, the third is a question rather than a change.

---

arrow/memory/buffer_test.go
fallintoplace: Okay, so I checked the allocator implementations and Buffer path. The allocator contract consistently treats len(b) as the owned allocation size, and Buffer keeps its logical length separately while passing the full allocation slice to Reallocate. So a len-zero, positive-capacity slice cannot reach this through Buffer.

A direct caller could manufacture one, but that already disagrees with the accounting and free behavior across the allocator implementations. I switched the guard to len(b) > 0 so the pointer handling matches that contract, used unsafe.SliceData, and hoisted the deletion.

zeroshade (approved):
Reviewed the updated implementation and allocator contract. Pointer tracking now matches len-based ownership, zero-size reallocations clear stale records, and the transition, race, memory-package, and 32-bit checks pass.

PR #1133, opened 2026-08-07
fix(compute/exec): report ListView buffer counts

fallintoplace:
### Rationale for this change

`ListView` and `LargeListView` arrays have exactly three buffers: validity, offsets, and sizes. `ArraySpan.NumBuffers()` currently returns the default count of two, which leaves the size buffer unset for zero-length spans and truncates it when any ListView span is converted back to `array.Data` with `MakeData()`.

Return three buffers for ListView and LargeListView types, document why that count is exact, and cover both the direct buffer-count contract and zero-length buffer allocation.

### Are these changes tested?

- `go test ./arrow/compute/exec ./arrow/compute`

There are no API changes. Compute kernels do not currently accept list views, so this is preparatory hardening rather than a live compute corruption fix.

### DIFF
--- a/arrow/compute/exec/span.go
+++ b/arrow/compute/exec/span.go
@@ -587,6 +587,8 @@ func getNumBuffers(dt arrow.DataType) int {
 		// builder is sufficient for most use cases).
 		return 3
 	case arrow.LIST_VIEW, arrow.LARGE_LIST_VIEW:
+		// validity + offsets + sizes. Unlike the view types above, this
+		// count is exact rather than a cap on variadic data buffers.
 		return 3
 	case arrow.EXTENSION:
 		return getNumBuffers(dt.(arrow.ExtensionType).StorageType())
--- a/arrow/compute/exec/span.go
+++ b/arrow/compute/exec/span.go
@@ -586,6 +586,8 @@ func getNumBuffers(dt arrow.DataType) int {
 		// within a single block (the default 32KB allocation in the
 		// builder is sufficient for most use cases).
 		return 3
+	case arrow.LIST_VIEW, arrow.LARGE_LIST_VIEW:
+		return 3
 	case arrow.EXTENSION:
 		return getNumBuffers(dt.(arrow.ExtensionType).StorageType())
 	default:
--- a/arrow/compute/exec_internals_test.go
+++ b/arrow/compute/exec_internals_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/internal/testing/gen"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/arrow/scalar"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 )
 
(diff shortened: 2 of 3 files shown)


arrow/compute/exec/span.go
```diff
@@ -586,6 +586,8 @@ func getNumBuffers(dt arrow.DataType) int {
 		// within a single block (the default 32KB allocation in the
 		// builder is sufficient for most use cases).
 		return 3
+	case arrow.LIST_VIEW, arrow.LARGE_LIST_VIEW:
+		return 3
```
zeroshade, line 592: Keeping this as its own case rather than folding it into the `return 3` above was the right call — but it's worth a line saying why, since the neighbouring case returns the same number for an entirely unrelated reason.

The `BINARY_VIEW`/`STRING_VIEW` 3 is a *heuristic cap*: view header plus one overflow data buffer, bounded by `ArraySpan`'s fixed buffer array. The list-view 3 is *exact*. Without a note, a reader hitting these back to back will reasonably assume the variadic caveat carries over.

```suggestion
	case arrow.LIST_VIEW, arrow.LARGE_LIST_VIEW:
		// validity + offsets + sizes. Unlike the view types above, this
		// count is exact rather than a cap on variadic data buffers.
		return 3
```

arrow/compute/exec_internals_test.go
```diff
@@ -580,6 +581,24 @@ func (e *ExecSpanItrSuite) TestZeroLengthInput() {
 	checkArgs(input)
 }
 
+func TestFillZeroLengthListView(t *testing.T) {
```
zeroshade, line 584: This is worth keeping for the buffer-allocation behaviour, but it reaches `getNumBuffers` indirectly, from a different package, when a direct test already exists.

`arrow/compute/exec/span_test.go` has a table-driven test for `ArraySpan.NumBuffers()` with rows like `{"binary", …, 3}` and `{"int32", …, 2}`. Two rows there would cover the actual change in its own package, in the established idiom:

```go
{"list view", fields{Type: arrow.ListViewOf(arrow.PrimitiveTypes.Int32)}, 3},
{"large list view", fields{Type: arrow.LargeListViewOf(arrow.PrimitiveTypes.Int32)}, 3},
```

That table is also where someone adding a new type will look for the count contract, so a future ListView regression is far more likely to be caught there than here.

zeroshade (commented):
The fix is correct — ListView and LargeListView have exactly three buffers (validity, offsets, sizes) per the spec.

Two notes on scope, in opposite directions, because the title implies something both narrower and more urgent than what's actually here.

**Wider than "zero-length":** `getNumBuffers` also backs `ArraySpan.NumBuffers()`, which `MakeData` uses to slice the buffer list at `span.go:174` and `:192` — `bufs[:a.NumBuffers()]`. At 2, converting *any* ListView span back to `array.Data` truncates the sizes buffer and produces a structurally invalid ListView, not just zero-length ones.

**Less urgent than it sounds:** `grep LIST_VIEW arrow/compute/` finds nothing else outside tests, so list views can't flow through compute kernels today anyway. This is preparatory hardening rather than a live corruption path, and it shouldn't be read as needing a backport.

Both worth stating so the next reader of this commit doesn't have to work it out.

Two suggestions inline. Neither blocking.

---

zeroshade (approved):
The follow-up documents the exact three-buffer ListView layout and adds direct NumBuffers coverage for both ListView widths while retaining the zero-length allocation regression. The focused, race, 32-bit, and compute subtree tests pass. Looks good.

PR #1134, opened 2026-08-07
fix(arrow): include KeysSorted in map equality

fallintoplace:
### Rationale for this change

MapType includes KeysSorted in its fingerprint and string representation, but TypeEqual ignores it. Maps with different key-ordering contracts can therefore compare equal.

Compare KeysSorted as part of map type equality and add coverage for both values.

### Are these changes tested?

- `go test ./arrow`

Map types with different KeysSorted values no longer compare equal. There are no API changes.

### DIFF
--- a/arrow/compare.go
+++ b/arrow/compare.go
@@ -74,16 +74,20 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 		}
 		return l.n == right.(*FixedSizeListType).n && l.elem.Nullable == right.(*FixedSizeListType).elem.Nullable
 	case *MapType:
-		if !TypeEqual(l.KeyType(), right.(*MapType).KeyType(), opts...) {
+		r := right.(*MapType)
+		if !TypeEqual(l.KeyType(), r.KeyType(), opts...) {
 			return false
 		}
-		if !TypeEqual(l.ItemType(), right.(*MapType).ItemType(), opts...) {
+		if !TypeEqual(l.ItemType(), r.ItemType(), opts...) {
 			return false
 		}
-		if l.KeyField().Nullable != right.(*MapType).KeyField().Nullable {
+		if l.KeysSorted != r.KeysSorted {
 			return false
 		}
-		if l.ItemField().Nullable != right.(*MapType).ItemField().Nullable {
+		if l.KeyField().Nullable != r.KeyField().Nullable {
+			return false
+		}
+		if l.ItemField().Nullable != r.ItemField().Nullable {
 			return false
 		}
 		if cfg.metadata {
--- a/arrow/compare.go
+++ b/arrow/compare.go
@@ -91,10 +91,10 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 			return false
 		}
 		if cfg.metadata {
-			if !l.KeyField().Metadata.Equal(right.(*MapType).KeyField().Metadata) {
+			if !l.KeyField().Metadata.Equal(r.KeyField().Metadata) {
 				return false
 			}
-			if !l.ItemField().Metadata.Equal(right.(*MapType).ItemField().Metadata) {
+			if !l.ItemField().Metadata.Equal(r.ItemField().Metadata) {
 				return false
 			}
 		}
--- a/arrow/compare_test.go
+++ b/arrow/compare_test.go
@@ -22,6 +22,12 @@ import (
 )
 
 func TestTypeEqual(t *testing.T) {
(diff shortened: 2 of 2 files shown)


arrow/compare.go
```diff
@@ -74,16 +74,20 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
...
+		r := right.(*MapType)
+		if !TypeEqual(l.KeyType(), r.KeyType(), opts...) {
 			return false
 		}
-		if !TypeEqual(l.ItemType(), right.(*MapType).ItemType(), opts...) {
+		if !TypeEqual(l.ItemType(), r.ItemType(), opts...) {
 			return false
 		}
-		if l.KeyField().Nullable != right.(*MapType).KeyField().Nullable {
+		if l.KeysSorted != r.KeysSorted {
```
zeroshade, line 84: Not a change request — a question worth answering before this lands.

Does Arrow C++ compare `keys_sorted` in its map type equality? If it doesn't, this introduces a cross-implementation divergence in schema matching: the same pair of map types would compare equal in C++ and unequal here, which tends to surface far downstream in Flight or IPC schema negotiation rather than at the comparison itself.

I'd expect C++ does compare it, given our fingerprint already encodes it, but I haven't verified and it seems worth checking rather than assuming.

arrow/compare.go
```diff
@@ -74,16 +74,20 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
...
 			return false
 		}
-		if l.ItemField().Nullable != right.(*MapType).ItemField().Nullable {
+		if l.KeyField().Nullable != r.KeyField().Nullable {
+			return false
+		}
+		if l.ItemField().Nullable != r.ItemField().Nullable {
 			return false
 		}
 		if cfg.metadata {
```
zeroshade, line 93: The refactor missed the two assertions inside this block (lines 94 and 97) — `right.(*MapType)` is still spelled out there while `r` has been in scope since line 77:

```go
if cfg.metadata {
	if !l.KeyField().Metadata.Equal(r.KeyField().Metadata) {
		return false
	}
	if !l.ItemField().Metadata.Equal(r.ItemField().Metadata) {
		return false
	}
}
```

With those, the case uses a single assertion throughout, matching the `*StructType` case directly below.

arrow/compare_test.go
```diff
@@ -395,3 +395,18 @@ func TestTypeEqual(t *testing.T) {
 		})
 	}
 }
+
+func TestTypeEqualMapKeysSorted(t *testing.T) {
```
zeroshade: Minor: `compare_test.go` already has a table-driven `TestTypeEqual` with a `want bool` field, and two rows there would express this directly — one pair differing only in `KeysSorted` expecting `false`, one pair agreeing expecting `true`.

The standalone test works and is readable; it just makes this file carry two patterns for the same thing, and the table is where someone adding a type-equality case will look first.

zeroshade (commented):
Well-justified, and the premise checks out exactly — `MapType.Fingerprint()` appends `"s"` when `KeysSorted` is set, and `MapType.String()` renders `, keys_sorted`, while `TypeEqual` ignored it.

That's the kind of inconsistency that bites quietly. Fingerprints exist as a cheap proxy for type identity, so any code using them as a hash key reasonably assumes `TypeEqual(a, b)` implies `Fingerprint(a) == Fingerprint(b)`. Before this, two map types could compare equal while hashing differently *and* printing differently.

Worth noting the contradiction could have been resolved the other way, by dropping the `"s"` from the fingerprint. Tightening `TypeEqual` is the right direction: `String()` shows the flag too, so the looser reading would leave two "equal" types that display differently. Aligning to the strict side is the only option that makes all three agree.

The `r := right.(*MapType)` extraction is a good cleanup on its own, and matches the `*StructType` case just below, which already does exactly this. One spot got missed — inline.

---

arrow/compare.go
fallintoplace: Yep, I checked Arrow C++. It explicitly returns false when the two map types have different keys_sorted values. The map fingerprint encodes it as well, so this keeps Go aligned with C++: [C](https://github.com/apache/arrow/blob/42694575d0219f6a3a78b1f344bb071a60df6a4e/cpp/src/arrow/compare.cc#L836-L841)

zeroshade (approved):
Reviewed the updated implementation and tests. Map equality now consistently accounts for KeysSorted, and the focused property, race, 32-bit, and Arrow subtree checks pass.

PR #1135, opened 2026-08-07
fix(compute): reject non-array-like filters

fallintoplace:
### Rationale for this change

The filter meta-function assumes the filter datum implements ArrayLikeDatum before checking its kind. A record batch or table passed as the filter therefore causes a panic instead of the existing unsupported-type error.

Validate that the filter is array-like before reading its type. Keep scalar boolean filters supported and add coverage for a record batch used as the filter.

### Are these changes tested?

- `go test ./arrow/compute`

Invalid record or table filters now return an error instead of panicking. Valid array, chunked, and scalar boolean filters are unchanged.

### DIFF
--- a/arrow/compute/selection.go
+++ b/arrow/compute/selection.go
@@ -40,7 +40,12 @@ are handled based on FilterOptions.`,
 	}
 	filterMetaFunc = NewMetaFunction("filter", Binary(), filterDoc,
 		func(ctx context.Context, opts FunctionOptions, args ...Datum) (Datum, error) {
-			if args[1].(ArrayLikeDatum).Type().ID() != arrow.BOOL {
+			filter, ok := args[1].(ArrayLikeDatum)
+			if !ok {
+				return nil, fmt.Errorf("%w: filter should be array-like", arrow.ErrNotImplemented)
+			}
+
+			if filter.Type().ID() != arrow.BOOL {
 				return nil, fmt.Errorf("%w: filter argument must be boolean type",
 					arrow.ErrNotImplemented)
 			}
--- a/arrow/compute/vector_selection_test.go
+++ b/arrow/compute/vector_selection_test.go
@@ -2366,6 +2366,29 @@ func BenchmarkTakeStringPartitionPattern(b *testing.B) {
 	b.ReportMetric(float64(numRows*b.N)/b.Elapsed().Seconds(), "rows/sec")
 }
 
+func TestFilterRejectsNonArrayLikeFilter(t *testing.T) {
+	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
+	defer mem.AssertSize(t, 0)
+
+	values, _, err := array.FromJSON(mem, arrow.PrimitiveTypes.Int32, strings.NewReader("[1]"))
+	require.NoError(t, err)
+	defer values.Release()
+
+	valuesDatum := compute.NewDatum(values)
+	defer valuesDatum.Release()
+	filterRecord := array.NewRecordBatch(
+		arrow.NewSchema([]arrow.Field{{Name: "filter", Type: arrow.PrimitiveTypes.Int32}}, nil),
+		[]arrow.Array{values},
+		1,
+	)
+	defer filterRecord.Release()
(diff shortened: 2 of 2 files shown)


arrow/compute/selection.go
```diff
@@ -40,7 +40,12 @@ are handled based on FilterOptions.`,
 	}
 	filterMetaFunc = NewMetaFunction("filter", Binary(), filterDoc,
 		func(ctx context.Context, opts FunctionOptions, args ...Datum) (Datum, error) {
-			if args[1].(ArrayLikeDatum).Type().ID() != arrow.BOOL {
+			filter, ok := args[1].(ArrayLikeDatum)
+			if !ok {
+				return nil, fmt.Errorf("%w: filter should be array-like", arrow.ErrNotImplemented)
```
zeroshade, line 45: Worth noting for anyone reading this later: the message and sentinel here deliberately match the guard that already exists in `FilterTable` (line 752), which has the identical `default:` case returning `"%w: filter should be array-like"` with `arrow.ErrNotImplemented`.

That's the more interesting framing of this fix. `FilterTable` was *already* handling non-array-like filters correctly — but for the `KindTable` path, `filterMetaFunc` dereferenced `args[1].(ArrayLikeDatum).Type()` at the top and panicked before control ever reached it. So this doesn't just add a guard, it makes an existing correct guard reachable.

I'd started to suggest rewording this to parallel the `"filter argument must be boolean type"` line below it — matching the existing `FilterTable` message is the better choice, so ignore that thought.

arrow/compute/vector_selection_test.go
```diff
@@ -2366,6 +2366,29 @@ func BenchmarkTakeStringPartitionPattern(b *testing.B) {
...
+		arrow.NewSchema([]arrow.Field{{Name: "filter", Type: arrow.PrimitiveTypes.Int32}}, nil),
+		[]arrow.Array{values},
+		1,
+	)
+	defer filterRecord.Release()
+	filterDatum := compute.NewDatum(filterRecord)
+	defer filterDatum.Release()
+
+	_, err = compute.Filter(context.Background(), valuesDatum, filterDatum, compute.FilterOptions{})
+	require.ErrorIs(t, err, arrow.ErrNotImplemented)
```
zeroshade: Two small coverage gaps.

`TableDatum` is the other kind that isn't array-like, and it's the one that was previously reaching — or rather, failing to reach — `FilterTable`'s own guard. One more case would cover the full surface of this check.

Nothing pins the scalar-filter path either. The description says scalar boolean filters stay supported and that's true today, but it rests on `ScalarDatum` implicitly satisfying `ArrayLikeDatum`. If that interface or `ScalarDatum` were refactored, this guard would start rejecting scalar filters and no test here would notice. A one-line scalar case would make the dependency explicit — worth checking whether existing tests elsewhere already cover it before adding.

zeroshade (commented):
Correct and complete — and it's a better fix than the description suggests.

`RecordDatum` carries only `Kind()` and `String()`, no `Type()`, so it genuinely fails to satisfy `ArrayLikeDatum` and the old code panicked outright. The test exercises the intended path rather than passing incidentally.

Three things I checked:

**The rest of the function was already safe.** Every other assertion — `args[0].(*RecordDatum)`, `args[0].(*TableDatum)` — sits inside the `switch args[0].Kind()`. `args[1]` was the only unguarded one, so this closes the gap rather than one instance of a pattern.

**`take` was never vulnerable.** `takeMetaFunc` checks `args[1].Kind()` against `KindArray`/`KindChunked` before touching anything, and switches on `args[0].Kind()` before asserting. `filter` was the outlier in this file.

**The "scalar filters unchanged" claim holds**, and it's worth stating why: it depends entirely on `ScalarDatum` satisfying `ArrayLikeDatum`, which it does — `Type()` at datum.go:109, `NullN()` at :115. If that weren't true, this guard would have quietly turned a working path into an error.

Two notes inline, neither blocking.

---

zeroshade (approved):
Reviewed the updated implementation and coverage. Non-array-like filters now return the intended error without panicking, scalar boolean filters remain supported, and the focused, race, and compute subtree checks pass.

PR #1136, opened 2026-08-07
fix(compute/exec): handle missing validity buffers

fallintoplace:
### Rationale for this change

UpdateNullCount counts bits from a nil validity buffer when the null count is unknown. A missing validity bitmap represents all-valid data, but CountSetBits can panic for a non-zero length.

Treat a missing validity buffer as an all-valid array, cache a null count of zero, and add regression coverage.

### Are these changes tested?

- `go test ./arrow/compute/exec`

Zero-null arrays without a validity bitmap no longer panic when their null count is first requested.

### DIFF
--- a/arrow/compute/exec/span.go
+++ b/arrow/compute/exec/span.go
@@ -114,7 +114,7 @@ func (a *ArraySpan) UpdateNullCount() int64 {
 	if curNulls != array.UnknownNullCount {
 		return curNulls
 	}
-	if a.Buffers[0].Buf == nil {
+	if len(a.Buffers[0].Buf) == 0 {
 		atomic.StoreInt64(&a.Nulls, 0)
 		return 0
 	}
--- a/arrow/compute/exec/span.go
+++ b/arrow/compute/exec/span.go
@@ -114,6 +114,10 @@ func (a *ArraySpan) UpdateNullCount() int64 {
 	if curNulls != array.UnknownNullCount {
 		return curNulls
 	}
+	if a.Buffers[0].Buf == nil {
+		atomic.StoreInt64(&a.Nulls, 0)
+		return 0
+	}
 
 	newNulls := a.Len - int64(bitutil.CountSetBits(a.Buffers[0].Buf, int(a.Offset), int(a.Len)))
 	atomic.StoreInt64(&a.Nulls, newNulls)
--- a/arrow/compute/exec/span_test.go
+++ b/arrow/compute/exec/span_test.go
@@ -118,6 +118,10 @@ func TestArraySpan_UpdateNullCount(t *testing.T) {
 		want   int64
 	}{
 		{"known", fields{Nulls: 25}, 25},
+		{"unknown without validity", fields{
+			Nulls: array.UnknownNullCount,
+			Len:   8,
+		}, 0},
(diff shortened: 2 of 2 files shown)


arrow/compute/exec/span.go
```diff
@@ -114,6 +114,10 @@ func (a *ArraySpan) UpdateNullCount() int64 {
 	if curNulls != array.UnknownNullCount {
 		return curNulls
 	}
+	if a.Buffers[0].Buf == nil {
```
zeroshade: Consider `len(a.Buffers[0].Buf) == 0` rather than `== nil` — it covers strictly more cases at no cost.

A non-nil but zero-length validity buffer is a real shape in this package: `FillZeroLength` sets buffers to `arrow.Uint64Traits.CastToBytes(span.Scratch[:])[:0]`, which is non-nil with length 0. Such a span passes this `== nil` check and would still reach `CountSetBits` — harmlessly today, since `Len` is 0 there, but the guard wouldn't hold if a span ever carried an empty bitmap with a non-zero length.

There's precedent for both forms in this file: line 161 uses `len(a.Buffers[0].Buf) == 0`, while lines 455 and 509 use `== nil`. The length form is the safer of the two, and it makes the intent — "there is no bitmap here" — independent of how the buffer happened to be zeroed.

arrow/compute/exec/span_test.go
```diff
@@ -118,6 +118,10 @@ func TestArraySpan_UpdateNullCount(t *testing.T) {
 		want   int64
 	}{
 		{"known", fields{Nulls: 25}, 25},
+		{"unknown without validity", fields{
```
zeroshade, line 121: Nice — adding a row to the existing table rather than writing a standalone test is the right instinct, and it's exactly what I suggested over on #1133 for `TestFillZeroLengthListView`. Worth applying the same treatment there for consistency, since both end up exercising this same file.

zeroshade (commented):
Correct, and it's a consistency fix as much as a crash fix.

I confirmed the panic is real: for `Len: 8` with a nil buffer, `CountSetBits` computes `uint64Bytes = 0`, so `bytesToUint64(buf[:0])` is fine, but the next line slices `buf[0 : n/8]` — `buf[0:1]` on a nil slice — which panics on the bounds.

The stronger argument for this change is that `MayHaveNulls()` on line 106 already does exactly this:

```go
return atomic.LoadInt64(&a.Nulls) != 0 && a.Buffers[0].Buf != nil
```

So the file already treats a missing validity bitmap as "no nulls" — `UpdateNullCount` was the one place that didn't, and would panic rather than agree. Caching the zero is right too, since the answer can't change for a span with no bitmap.

One suggestion on the guard inline, and a note of appreciation on the test.

For coordination: this touches `arrow/compute/exec/span.go`, same as #1133, but a different region (`UpdateNullCount` at ~114 vs `getNumBuffers` at ~589), so they're independent and won't conflict.

---

zeroshade (approved):
The follow-up correctly handles both nil and non-nil empty validity buffers, matching the existing all-valid semantics while preserving bitmap and offset behavior. The focused, race, 32-bit, and compute subtree tests pass. Looks good.

PR #1137, opened 2026-08-07
fix(parquet/metadata): return error for missing column metadata

fallintoplace:
### Rationale for this change

NewColumnChunkMetaData dereferences column metadata while preparing encoding information. A malformed unencrypted ColumnChunk without MetaData therefore panics instead of returning its error result. Encrypted chunks may legitimately populate plaintext metadata during decryption.

Check for missing metadata after the decryption path and return an error for malformed chunks. Add regression coverage while preserving encrypted metadata handling.

### Are these changes tested?

- `go test ./parquet/metadata`

Malformed unencrypted column chunks now return an error instead of panicking. Valid encrypted column chunks are unchanged.

### DIFF
--- a/parquet/metadata/column_chunk.go
+++ b/parquet/metadata/column_chunk.go
@@ -87,9 +87,10 @@ type ColumnChunkMetaData struct {
 // this is primarily used internally or between the subpackages. ColumnChunkMetaDataBuilder should
 // be used by consumers instead of using this directly.
 func NewColumnChunkMetaData(column *format.ColumnChunk, descr *schema.Column, writerVersion *AppVersion, rowGroupOrdinal, columnOrdinal int16, fileDecryptor encryption.FileDecryptor) (*ColumnChunkMetaData, error) {
+	columnMeta := column.GetMetaData()
 	c := &ColumnChunkMetaData{
 		column:        column,
-		columnMeta:    column.GetMetaData(),
+		columnMeta:    columnMeta,
 		descr:         descr,
 		writerVersion: writerVersion,
 		mem:           memory.DefaultAllocator,
@@ -116,6 +117,9 @@ func NewColumnChunkMetaData(column *format.ColumnChunk, descr *schema.Column, wr
 				return nil, errors.New("cannot decrypt column metadata. file decryption not setup correctly")
 			}
 		}
+	}
+	if c.columnMeta == nil {
+		return nil, errors.New("column chunk metadata is missing")
 	}
 	for _, enc := range c.columnMeta.Encodings {
 		c.encodings = append(c.encodings, parquet.Encoding(enc))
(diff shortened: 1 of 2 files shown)


parquet/metadata/column_chunk.go
```diff
@@ -87,9 +87,10 @@ type ColumnChunkMetaData struct {
 // this is primarily used internally or between the subpackages. ColumnChunkMetaDataBuilder should
 // be used by consumers instead of using this directly.
 func NewColumnChunkMetaData(column *format.ColumnChunk, descr *schema.Column, writerVersion *AppVersion, rowGroupOrdinal, columnOrdinal int16, fileDecryptor encryption.FileDecryptor) (*ColumnChunkMetaData, error) {
+	columnMeta := column.GetMetaData()
```
zeroshade: Cosmetic: this local doesn't earn its keep. It's used once, on line 93, and the new check on line 121 reads `c.columnMeta` rather than this variable — so it never diverges from the field, but a reader still has to track it to confirm that.

Inlining it back to `columnMeta: column.GetMetaData()` in the struct literal would leave the diff as purely the nil check.

My guess is it's a leftover from a draft where the check sat before the crypto block — which is exactly the placement that would have broken encrypted chunks, so the final version is right.

parquet/metadata/file_internal_test.go
```diff
@@ -36,3 +36,8 @@ func TestNewFileMetaDataReturnsSchemaErrors(t *testing.T) {
 	require.Error(t, err)
 	require.Nil(t, meta)
 }
+
+func TestNewColumnChunkMetaDataReturnsErrorForMissingMetadata(t *testing.T) {
+	_, err := NewColumnChunkMetaData(&format.ColumnChunk{}, nil, nil, 0, 0, nil)
```
zeroshade: Cosmetic: the sibling test just above (`TestNewFileMetaDataReturnsSchemaErrors`) asserts `require.Nil(t, meta)` alongside the error, while this one discards the first return with `_`.

Matching it would confirm no partially-constructed `ColumnChunkMetaData` escapes next to the error — which is the more interesting half of the guarantee, since the struct is fully populated by line 121 before the check rejects it.

`require.EqualError` is fine and does pin the message; just worth noting it'll need updating if a prefix is ever added, where the sibling's `require.Error` wouldn't.

zeroshade (approved):
Approving — correct, and the placement is the thoughtful part.

The nil check sits *after* the crypto block deliberately, and that's load-bearing: line 115 does `c.columnMeta = &c.decryptedMeta`, so an encrypted chunk can arrive with `column.GetMetaData() == nil` and legitimately acquire its metadata during decryption. Checking any earlier would have broken encrypted chunks. Good call, and good that the description explains it.

I went looking for a missing `parquet:` prefix on the new error and there isn't one to flag — this package uses bare messages throughout (`errors.New("bloom filter header must not be nil")`), and line 117 of this same function reads `errors.New("cannot decrypt column metadata. file decryption not setup correctly")`. The new message matches its neighbours exactly.

The test reaches the check safely too: `descr` is only stored at line 94 and never dereferenced, and `IsSetCryptoMetadata()` on `&format.ColumnChunk{}` is false, so passing `nil` for the later arguments is fine.

Two cosmetic notes inline. Neither affects correctness — happy for this to land either way.

---

PR #1138, opened 2026-08-07
feat(compute): add list_element

fallintoplace:
### Rationale for this change

Arrow-Go supports the list, large-list, list-view, and fixed-size-list array types, but callers currently need to read offsets and child arrays themselves to select one element from every row. A list_element kernel provides the common nested-data operation directly through compute.

- Add list_element for List, LargeList, ListView, LargeListView, and FixedSizeList.
- Accept signed or unsigned integer scalar indexes. A one-element integer array is accepted for matching one-row input.
- Return an array of the list element type while preserving parent nulls and child nulls.
- Return errors for null or negative indexes, empty lists, and indexes outside the length of a valid list.
- Use the existing take kernels for primitive, binary, and fixed-width child values to avoid one array allocation per input row. Keep the generic concatenate path for nested child values.
- Handle list-view offsets and sizes without requiring the input to be materialized first.
- Reject StringView, BinaryView, and dictionary-backed extension children, including nested children, before execution.
- Fix null-array construction and run-end encoded span metadata needed by the nested fallback.

### Are these changes tested?

- go test ./arrow/compute/... -count=1
- Added coverage for sliced List and FixedSizeList inputs, non-contiguous ListView offsets, unsigned indexes, string and nested-list children, parent and child nulls, and invalid indexes.
- Added regression coverage for dictionary-backed extension rejection and null run-end encoded values with nested dictionaries.
- Added BenchmarkListElement cases for 1K, 100K, and 1M rows with allocation reporting.

Yes. This adds the public `compute.ListElement` function. It also fixes `array.MakeArrayOfNull` for list views, unions, and run-end encoded arrays, keeps run-end encoded `ArraySpan` buffer metadata consistent, and preserves storage validity when extracting extension scalars.

### DIFF
--- a/arrow/compute/internal/kernels/scalar_nested.go
+++ b/arrow/compute/internal/kernels/scalar_nested.go
@@ -188,7 +188,8 @@ func listElementExec(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecRe
 
 func listElementTakeSupported(id arrow.Type) bool {
 	return id == arrow.NULL || arrow.IsPrimitive(id) || arrow.IsBinaryLike(id) ||
-		arrow.IsLargeBinaryLike(id) || arrow.IsFixedSizeBinary(id) || id == arrow.DENSE_UNION
+		arrow.IsLargeBinaryLike(id) || arrow.IsFixedSizeBinary(id) ||
+		id == arrow.SPARSE_UNION || id == arrow.DENSE_UNION
 }
 
 func listElementConcat(ctx *exec.KernelCtx, list *exec.ArraySpan, index uint64, elemType arrow.DataType, out *exec.ExecResult) error {
@@ -296,6 +297,69 @@ func listElementDenseUnionTake(ctx *exec.KernelCtx, values *exec.ArraySpan, indi
 	}
 	return nil
 }
+
+func listElementSparseUnionTake(ctx *exec.KernelCtx, values *exec.ArraySpan, indices arrow.Array, out *exec.ExecResult) error {
+	valuesArray := values.MakeArray()
+	defer valuesArray.Release()
+
+	unionType := values.Type.(*arrow.SparseUnionType)
+	builder := array.NewSparseUnionBuilder(exec.GetAllocator(ctx.Ctx), unionType)
+	defer builder.Release()
+	builder.Reserve(indices.Len())
+	for i := 0; i < builder.NumChildren(); i++ {
+		builder.Child(i).Reserve(indices.Len())
+	}
+
+	indexArray := indices.(*array.Int64)
+	for i := 0; i < indices.Len(); i++ {
+		var value scalar.Scalar
+		if indices.IsNull(i) {
+			value = scalar.MakeNullScalar(unionType)
+		} else {
+			var err error
+			value, err = scalar.GetScalar(valuesArray, int(indexArray.Value(i)))
+			if err != nil {
+				return err
+			}
+		}
+
+		if err := listElementAppendSparseUnionValue(builder, value.(*scalar.SparseUnion)); err != nil {
+			if releasable, ok := value.(scalar.Releasable); ok {
+				releasable.Release()
+			}
+			return err
+		}
+		if releasable, ok := value.(scalar.Releasable); ok {
+			releasable.Release()
+		}
+	}
+
+	result := builder.NewArray()
+	defer result.Release()
+	out.TakeOwnership(result.Data())
+	return nil
+}
+
+func listElementAppendSparseUnionValue(builder *array.SparseUnionBuilder, value *scalar.SparseUnion) error {
+	builder.Append(value.TypeCode)
+	for i := 0; i < builder.NumChildren(); i++ {
+		child := builder.Child(i)
+		if i != value.ChildID {
+			child.AppendEmptyValue()
+			continue
+		}
+
+		if !value.IsValid() {
+			child.AppendNull()
+			continue
+		}
+		if err := scalar.Append(child, value.Value[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func listElementTake(ctx *exec.KernelCtx, values *exec.ArraySpan, indices arrow.Array, out *exec.ExecResult) (bool, error) {
 	var indexSpan exec.ArraySpan
 	indexSpan.SetMembers(indices.Data())
@@ -320,6 +384,8 @@ func listElementTake(ctx *exec.KernelCtx, values *exec.ArraySpan, indices arrow.
 		return true, TakeExec(VarBinaryImpl[int64])(&takeCtx, batch, out)
 	case arrow.IsFixedSizeBinary(id):
 		return true, TakeExec(FSBImpl)(&takeCtx, batch, out)
+	case id == arrow.SPARSE_UNION:
+		return true, listElementSparseUnionTake(ctx, values, indices, out)
 	case id == arrow.DENSE_UNION:
 		return true, listElementDenseUnionTake(ctx, values, indices, out)
 	default:
--- a/arrow/compute/internal/kernels/scalar_nested.go
+++ b/arrow/compute/internal/kernels/scalar_nested.go
@@ -152,18 +152,12 @@ func listElementExec(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecRe
 		return nil
 	}
 
-	values := list.Children[0].MakeArray()
-	defer values.Release()
-	pieces := make([]arrow.Array, 0, int(list.Len))
-	defer func() {
-		for _, piece := range pieces {
-			piece.Release()
-		}
-	}()
-
+	indexBuilder := array.NewInt64Builder(exec.GetAllocator(ctx.Ctx))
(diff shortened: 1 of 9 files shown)


arrow/compute/scalar_nested.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := validateListElementIndex(args[1]); err != nil {
+		return nil, err
+	}
+	if args[0].Kind() == KindScalar && args[1].Kind() == KindScalar {
+		if listType, ok := args[0].(ArrayLikeDatum).Type().(arrow.ListLikeType); ok &&
+			!listElementScalarResultSupported(listType.Elem()) {
+			return nil, fmt.Errorf("%w: list_element scalar output type %s is not supported", arrow.ErrNotImplemented, listType.Elem())
+		}
+
+		listValue := args[0].(*ScalarDatum).Value.(scalar.ListScalar)
```
zeroshade: **Blocking:** This unchecked assertion makes unsupported scalar inputs panic instead of returning the normal dispatch/type error. For example, `ListElement(ctx, ScalarDatum(Int32(7)), ScalarDatum(Int64(0)))` panics with `*scalar.Int32 is not scalar.ListScalar`; the equivalent unsupported array input returns an error. Please guard the list-like/type assertion and either fall through to normal kernel dispatch or return `arrow.ErrType`, with a regression test for a non-list scalar argument.

zeroshade (changes requested):
The list_element implementation has broad test coverage and the compute, array, and targeted race suites pass. However, I reproduced a blocking public API panic for an unsupported scalar first argument; this should return a dispatch/type error instead.

arrow/compute/internal/kernels/scalar_nested.go
```diff
@@ -0,0 +1,567 @@
...
+		if size > math.MaxInt64-start {
+			return 0, 0, fmt.Errorf("%w: list_element input has invalid value offsets", arrow.ErrInvalid)
+		}
+		return check(start, start+size)
+	default:
+		return 0, 0, fmt.Errorf("%w: unsupported list_element input type %s", arrow.ErrType, list.Type)
+	}
+}
+
+func listElementExec(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResult) error {
```
zeroshade, line 176: `list<string_view>` and `list<binary_view>` crash the process here, and the caller cannot recover from it:

```
child string_view buffers = 3
calling compute.ListElement inside recover()...
returned datum kind=chunked_array          <- the call SUCCEEDS
SURVIVED - caller recover() was sufficient
panic: runtime error: index out of range [3] with length 3
goroutine 8: exec.(*ArraySpan).TakeOwnership          span.go:452
             kernels.listElementConcat                kernels/scalar_nested.go:296
             compute.(*scalarExecutor).executeSpans   executor.go:634
```

`compute.ListElement` returns normally, and then an executor goroutine panics asynchronously — `recover()` in the calling goroutine is useless. It's reachable purely through the public API with a plain `list<string_view>`, and triggers once the child has 2 or more variadic data buffers.

The root cause is that `ArraySpan.Buffers` is `[3]BufferSpan`, while a string/binary-view array carries validity + view-headers + N variadic data buffers, so `TakeOwnership` writes past index 2. Note that trimming trailing nil buffers in `ArraySpan.MakeData` does **not** fix this — I tried that first, and it only fixes the inlined case before hitting the out-of-bounds write.

Please reject view element types recursively here, matching what the scalar path already does via `listElementScalarResultSupported`. The recursion is load-bearing: `struct<string_view>`, `list<string_view>`, `fsl<string_view,2>` and dictionary/REE children all need it, and a flat check leaves those crashing. With that in place both paths agree and return a consistent `ErrNotImplemented`.

The underlying `ArraySpan` 3-buffer limitation should be its own issue — not this PR's to fix.

arrow/compute/scalar_nested_test.go
```diff
@@ -0,0 +1,1992 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			input := tc.build(mem)
+			defer input.Release()
+
+			result, err := compute.ListElement(
+				context.Background(),
+				&compute.ArrayDatum{Value: input.Data()},
+				&compute.ScalarDatum{Value: scalar.NewInt64Scalar(0)},
+			)
+			if result != nil {
```
zeroshade: Releasing on the error path makes the compute suite flaky under `-race`.

`compute.ListElement` intermittently returns a non-nil `Datum` wrapping a nil `*arrow.Chunked`, so `Release()` segfaults in `arrow.(*Chunked).Release(0x0)`. Full `go test -race ./arrow/compute/... ./arrow/array/` failed 1 of 3 runs; a targeted loop reproduced it 5/5.

This pattern appears at 13 sites in this file — lines 430, 460, 656, 760, 807, 862, 925, 937, 996, 1311, 1362, 1424, 1867 and 1904. Line 1867 sits inside a 20-iteration loop, which multiplies the odds considerably. Guarding all of them with `if err == nil && result != nil` takes it to 0 failures across 8 combined `-race` runs.

The underlying executor bug is pre-existing and not yours — I reproduced the identical failure on the merge base using `divide`-by-zero, at 47/2000 — so the ask here is just to stop the tests tripping it. I'll file the executor issue separately against `executor.go:521`.

zeroshade (changes requested):
Thanks — the panic on unsupported scalar input is properly fixed, with the regression test I asked for. I checked that the guard isn't bypassable via `CallFunction`, registry `fn.Execute`, or a Map scalar, and all of those give a clean dispatch error. That thread can be resolved.

The core of this is in good shape. Index semantics match Arrow C++ exactly, including the error wording — negative index, index >= list length, null list, empty list, null index, and multi-element index arrays all line up. Sliced-offset handling is correct: I property-tested 7 list/element type combinations x offsets 0-4 x lengths 0-6 x indices 0-2 against an independent reference extraction and everything matched. Memory is clean under `CheckedAllocator` across 16 scenarios including both error paths.

The ListView handling is actually *more* correct than C++, which registers list-views against the same functor that computes `offsets[i+1] - offsets[i]` and never reads the sizes buffer. Two sibling kernels in the same file (`list_value_length`, `list_slice`) handle it correctly; `list_element` doesn't. On a valid non-monotonic list-view, C++ errors where this PR returns the right answer. Worth an upstream issue, and worth a comment here so nobody later "fixes" it toward C++.

Two blockers inline. Beyond those:

- **The `array/util.go` and `exec/span.go` changes are necessary bug fixes, not scope creep** — I checked specifically because I expected otherwise. `getNumBuffers(REE)` 0 to 1 resolves a real contradiction with `validate.go:192`, which requires at least 1 buffer, and `MakeArrayOfNull` panicked on the merge base for ListView/LargeListView/DenseUnion/REE. IPC and C Data Interface round-trips are unaffected. But these *are* user-visible changes to `array.MakeArrayOfNull` and `ArraySpan.NumBuffers`/`MakeData`, so please correct "Existing functions and APIs are unchanged" in the description. Ideally split them into a precursor PR.

- **Duplication of existing selection machinery.** `listElementDenseUnionTake` (33 lines) re-implements what the registered `take` already does end-to-end — I verified with heterogeneous children including a nested list, reordering, null indices and duplicate indices, all `ValidateFull`-clean. Another ~90 lines across `listElementTake` / `listElementTakeSupported` / `listElementTakeOrFallback` / `listElementTakeIndex` re-implement take dispatch. `selection.go` shows the established pattern for composing `kernels.*Impl` with `TakeArrayOpts` from `package compute`, which works around the import restriction. That said, the concat fallback does fill genuine gaps — `take` doesn't cover string_view/binary_view, decimal32/64, month_day_nano, listview, sparse_union or REE — but those arguably belong in `take` itself, where `filter` would benefit too.

- **`listElementScalarResultSupported` is a hand-maintained mirror of `scalar`-package capability.** It's legitimate rather than arbitrary: with the decimal32/decimal64/list-view entries removed, the calls fail one layer deeper with `cannot create scalar from array of type ...`, and critically that error is *untyped* — `errors.Is` against `ErrNotImplemented`/`ErrInvalid`/`ErrType` all return false. So the deny-list is doing real work today. But it will drift as the `scalar` package gains types, and it produces a visible array/scalar asymmetry (decimal32, decimal64 and listview work as array input and not as scalar input). Prefer deriving it — attempt construction and wrap whatever the scalar layer returns in `arrow.ErrNotImplemented`. The `BINARY_VIEW`/`STRING_VIEW` entries become redundant once the blocker below is fixed.

- Two new `go vet` copylocks warnings at `scalar_nested_test.go:42` and `:76` — `ValueStr passes lock by value` on the test extension array types, which embed `ExtensionArrayBase` and its `atomic.Int64`. Pointer receivers clear both. The merge base is clean apart from the pre-existing `registry.go` notes.

- The `FunctionDoc` doesn't state which list types are supported, that Map is excluded (matching C++), or the error semantics for negative/out-of-range/empty/null-index. Those are user-visible contracts on new public API.

- 21 commits, 15 of them `fix:`/`harden` — please squash on merge.

On the diff size: 72% is tests, with no generated code, which is healthy. The core kernel is ~200 lines; the remaining ~360 is exotic-child-type machinery, which is where the duplication note above applies.

zeroshade (approved):

PR #1139, opened 2026-08-07
feat(compute): add cumulative_sum and cumulative_sum_checked

fallintoplace:
### Rationale for this change

Arrow-Go has array and chunked-array compute kernels, but no cumulative sum operation for running totals. Adding this as a vector kernel also gives chunked inputs the same stateful behavior as a single contiguous array.

- Add cumulative_sum for signed integers, unsigned integers, and floating-point arrays.
- Add cumulative_sum_checked, which reports integer overflow instead of returning an overflowing result.
- Add CumulativeOptions with a start value and SkipNulls. By default, a null stops the cumulative result; with SkipNulls, null positions remain null while later valid values continue from the previous sum.
- Validate cross-type start values with the same safe numeric conversion rules used by compute casts.
- Carry the running state across chunks and keep the output chunked.

### Are these changes tested?

- go test ./arrow/compute/... -count=1
- Added coverage for normal sums, null propagation, skipped nulls, start values, rejected overflow and truncation in start values, chunked input, and checked overflow at every integer width.

Yes. This adds the public compute.CumulativeSum and compute.CumulativeSumChecked functions plus CumulativeOptions. Existing functions and APIs are unchanged.

### DIFF
--- a/arrow/compute/internal/kernels/vector_cumulative.go
+++ b/arrow/compute/internal/kernels/vector_cumulative.go
@@ -211,23 +211,52 @@ func prepareCumulativeOutput[T arrow.NumericType](ctx *exec.KernelCtx, out *exec
 	}
 }
 
-func cumulativeSumSpans[T arrow.NumericType](ctx *exec.KernelCtx, state *cumulativeSumState[T], inputs []*exec.ArraySpan, out *exec.ExecResult, needsValidity bool) error {
-	prepareCumulativeOutput[T](ctx, out, needsValidity)
-	values := exec.GetSpanValues[T](out, 1)
+func cumulativeSumNoNulls[T arrow.NumericType](state *cumulativeSumState[T], inputs []*exec.ArraySpan, values []T) {
+	var outputOffset int64
+	current := state.current
+	for _, input := range inputs {
+		inputValues := exec.GetSpanValues[T](input, 1)
+		for i := int64(0); i < input.Len; i++ {
+			current += inputValues[i]
+			values[outputOffset+i] = current
+		}
+		outputOffset += input.Len
+	}
+	state.current = current
+}
 
+func cumulativeSumNoNullsChecked[T arrow.NumericType](state *cumulativeSumState[T], inputs []*exec.ArraySpan, values []T) error {
+	var outputOffset int64
+	current := state.current
+	for _, input := range inputs {
+		inputValues := exec.GetSpanValues[T](input, 1)
+		for i := int64(0); i < input.Len; i++ {
+			var err error
+			current, err = state.add(current, inputValues[i])
+			if err != nil {
+				return err
+			}
+			values[outputOffset+i] = current
+		}
+		outputOffset += input.Len
+	}
+	state.current = current
+	return nil
+}
+
+func cumulativeSumWithNulls[T arrow.NumericType](state *cumulativeSumState[T], inputs []*exec.ArraySpan, values []T, validity []byte) int64 {
 	var (
 		nulls        int64
 		outputOffset int64
 	)
+	current := state.current
 	for _, input := range inputs {
 		inputValues := exec.GetSpanValues[T](input, 1)
 		for i := int64(0); i < input.Len; i++ {
 			valid := len(input.Buffers[0].Buf) == 0 || bitutil.BitIsSet(input.Buffers[0].Buf, int(input.Offset+i))
 			outputIndex := outputOffset + i
 			if !valid || state.encounteredNull {
-				if needsValidity {
-					bitutil.ClearBit(out.Buffers[0].Buf, int(outputIndex))
-				}
+				bitutil.ClearBit(validity, int(outputIndex))
 				nulls++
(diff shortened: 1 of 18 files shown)


arrow/ipc/file_writer.go
```diff
@@ -303,6 +305,13 @@ func (f *FileWriter) Close() error {
 	return nil
 }
 
+func (f *FileWriter) releaseDictionaries() {
+	for _, d := range f.lastWrittenDicts {
+		d.Release()
+	}
+	f.lastWrittenDicts = nil
```
zeroshade, line 320: `releaseDictionaries` leaves `lastWrittenDicts` nil. Because `Write` does not reject a closed writer, writing another dictionary record after `Close` reaches dictionary encoding and panics with `assignment to entry in nil map`.

Please keep the writer in a state where subsequent calls return an error rather than panic—likely by tracking the closed state and checking it in `Write`.

arrow/compute/internal/kernels/vector_cumulative.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return casted, nil
+}
+
+func cumulativeStartValue[T arrow.NumericType](ctx *exec.KernelCtx, cast ScalarCastFn, start scalar.Scalar, typ arrow.DataType) (T, error) {
+	var zero T
+	if start == nil {
+		return zero, nil
+	}
+	if !start.IsValid() {
```
zeroshade, line 97: A typed-nil scalar bypasses `start == nil` because the interface itself is non-nil, then panics here:

```go
var start *scalar.Int32
CumulativeOptions{Start: start}
```

Typed-nil releasable starts can similarly panic during expression construction at `Retain`. Please detect typed-nil starts and either treat them as absent or return `arrow.ErrInvalid`.

zeroshade (changes requested):
Found two reproducible panic paths that need to be fixed before this can merge; see inline comments on `arrow/ipc/file_writer.go` and `arrow/compute/internal/kernels/vector_cumulative.go`.

The cumulative kernels themselves otherwise look solid: numeric-width coverage, checked overflow, null semantics, sliced arrays, chunked state, start conversion, and affected package tests all passed.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Arrow Go maintainer. After you've addressed the points above and pushed an update, an Arrow Go maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Arrow Go handles maintainer review:* [CONTRIBUTING.md](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md).

arrow/ipc/file_writer.go
```diff
@@ -285,6 +286,12 @@ func NewFileWriter(w io.Writer, opts ...Option) (*FileWriter, error) {
 }
 
 func (f *FileWriter) Close() error {
+	if f.closed {
+		return nil
+	}
+	f.closed = true
```
zeroshade, line 293: Setting `closed` before the close operation succeeds causes a failed close to become success on every subsequent call. For example, when the output sink rejects the initial header:

```go
require.ErrorIs(t, writer.Close(), sinkErr)
require.ErrorIs(t, writer.Close(), sinkErr) // currently returns nil
```

This probe passes on the previously reviewed head and fails after this change. Since the writer is terminal after the first call, please retain and return the close error on subsequent calls, consistent with the stream `Writer`, rather than returning `nil`. Please add regression coverage for repeated `Close` after a sink failure.

zeroshade (changes requested):
Both previous panic paths are resolved. The typed-nil cumulative start is now handled safely, and writes after `FileWriter.Close` return an error instead of panicking.

One lifecycle regression remains in the replacement fix:

### Preserve failed-close errors

`FileWriter.Close` marks the writer closed before the close operation succeeds. If writing the header or footer fails, the first call returns the underlying error, but every subsequent `Close` returns `nil`.

A targeted probe fails on the current head and passes on the previously reviewed head. Please retain and return the terminal close error on subsequent calls, consistent with the stream `Writer`, and add regression coverage. See the inline comment on `arrow/ipc/file_writer.go`.

CI is fully green and the PR is mergeable.

---

>

zeroshade (approved):
The retained `closeErr` matches the stream `Writer`'s terminal-close semantics, and the regression covers the repeat-`Close` case. With the two earlier panic paths already resolved, this looks ready. Nice work on a thorough kernel implementation — the checked/unchecked split, chunked-state handling, and sliced-null coverage are all solid.

---

zeroshade:
@fallintoplace CI is green on this, but it now has merge conflicts with `main` after #1142 landed (overlapping `arrow/ipc` test changes). Please rebase when you get a chance — the approval stands, and I'll merge once it's conflict-free and CI re-runs green.

---
*This comment was drafted with the help of an AI-assisted tool and confirmed by an Apache Arrow maintainer before posting.*

PR #1141, opened 2026-08-08
fix(arrow/ipc): return compression errors from record encoder

fallintoplace:
### Rationale for this change

`recordEncoder.encode` ignored errors returned by body compression, so a failed compressor could still result in record metadata being built and returned as a successful encode.

Return compression errors from `recordEncoder.encode`, release the temporary compressed buffer on failure, and add a regression test with a failing compressor.

### Are these changes tested?

- `go test ./arrow/ipc`

IPC record encoding now reports compressor failures to the caller.

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -1153,6 +1153,8 @@ func needTruncate(offset int64, buf *memory.Buffer, minLength int64) bool {
 // method after it is no longer needed.
 func GetRecordBatchPayload(batch arrow.RecordBatch, opts ...Option) (Payload, error) {
 	cfg := newConfig(opts...)
+	compressors := make([]compressor, cfg.compressNP)
+	copy(compressors, cfg.compressors)
 	var (
 		data = Payload{msg: MessageRecordBatch}
 		enc  = newRecordEncoder(
@@ -1163,7 +1165,7 @@ func GetRecordBatchPayload(batch arrow.RecordBatch, opts ...Option) (Payload, er
 			cfg.codec,
 			cfg.compressNP,
 			cfg.minSpaceSavings,
-			make([]compressor, cfg.compressNP),
+			compressors,
 		)
 	)
 
--- a/arrow/ipc/ipc.go
+++ b/arrow/ipc/ipc.go
@@ -74,6 +74,7 @@ type config struct {
 	minSpaceSavings    float64
 	maxMetadataSize    int64
 	maxBodySize        int64
+	compressors        []compressor
 }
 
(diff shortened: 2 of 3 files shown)


arrow/ipc/writer.go
```diff
@@ -527,7 +529,9 @@ func (w *recordEncoder) encode(p *Payload, rec arrow.RecordBatch) error {
 			return fmt.Errorf("%w: minSpaceSavings not in range [0,1]. Provided %.05f",
 				arrow.ErrInvalid, w.minSpaceSavings)
 		}
-		w.compressBodyBuffers(p)
+		if err := w.compressBodyBuffers(p); err != nil {
```
zeroshade, line 545: Returning this error introduces a leak through `GetRecordBatchPayload`. That helper returns `Payload{}` when `enc.Encode` fails without releasing its local `data`, although encoding has already retained or allocated its body buffers—and may have replaced earlier buffers before a later compression failure.

A checked-allocator probe leaves 128 bytes allocated on this path. Please release the partial payload before discarding it:

```go
if err := enc.Encode(&data, batch); err != nil {
    data.Release()
    return Payload{}, err
}
```

Please also cover this public-helper path for compressor `Write` and `Close` failures.

zeroshade (changes requested):
The compression error is now propagated correctly by the direct encoder and writer paths, but the new error path exposes an ownership leak in the public payload helper.

### Release the partially built payload on encoding failure

`GetRecordBatchPayload` returns `Payload{}` when `enc.Encode` fails without releasing its local `data`. Encoding has already retained or allocated body buffers and may have replaced earlier buffers before a later compression failure. A checked-allocator regression leaves 128 bytes allocated on this path.

Please release `data` before returning the error and add public-helper coverage for compressor `Write` and `Close` failures. See the inline comment on `arrow/ipc/writer.go`.

The parallel compression error deadlock was reproduced but exists at the merge base, so it is not attributed to this change.

The PR also currently conflicts with `main`; resolving the adjacent test conflict should retain both this regression and `TestNewWriterWithMinSpaceSavings`.

---

>

zeroshade (approved):
The payload release and the public-helper coverage for both `Write` and `Close` failures look good, and the `withCompressors` hook is a clean way to inject the failure.

Heads-up: #1142 just got approved and contains an overlapping subset of these changes (`encode` propagation, buffer releases, and a conflicting `failingCompressor`/test definition). Once it merges, please rebase — this PR should reduce to the `Close`-failure coverage and the compressor-injection hook.

---

PR #1142, opened 2026-08-08
fix(arrow/ipc): avoid deadlock on compression errors

fallintoplace:
### Rationale for this change

Parallel body compression sent worker errors through an unbuffered channel before the parent waited for the workers. A compressor failure could therefore leave the writer blocked forever, and the producer could continue sending work after the workers had stopped.

Buffer the first worker error, make producer sends cancellation-aware, and add a regression test that uses a failing compressor with parallel compression.

### Are these changes tested?

- `go test ./arrow/ipc`
- `go test -race ./arrow/ipc -run TestRecordEncoderCompressionErrorDoesNotDeadlock`

Parallel IPC compression now returns compressor failures instead of hanging.

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -436,9 +436,11 @@ func (w *recordEncoder) compressBodyBuffers(p *Payload) error {
 
 		n, err := codec.Write(p.body[idx].Bytes())
 		if err != nil {
+			buf.Release()
 			return err
 		}
 		if err := codec.Close(); err != nil {
+			buf.Release()
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


zeroshade (changes requested):
The worker-pool changes prevent simultaneous compression failures from deadlocking, and the helper-level stress and race tests pass. However, `recordEncoder.encode` still discards the error returned by `compressBodyBuffers`, allowing the public writer paths to continue with a partially compressed payload and report success.

**`arrow/ipc/writer.go:544`:** `compressBodyBuffers` now returns the worker error without deadlocking, but this call discards it. A direct `recordEncoder.Encode` regression with failing compressors still returns `nil`, so `Writer` and `FileWriter` can continue with a partially compressed payload and report success. Please propagate this error and add an encoder- or writer-level regression that verifies it reaches the caller.

The PR also currently conflicts with `main` and will need rebasing.

zeroshade (approved):
Rebase and error propagation look good — `Encode` now surfaces compression failures, and the encoder-level regression covers it. The buffer releases on the error paths are a nice catch too.

Note this overlaps with #1141's `encode` change; one of the two will need a rebase after the other merges.

---

PR #1144, opened 2026-08-08
fix(parquet/encoding): honor BufferWriter offsets

fallintoplace:
### Rationale for this change

`BufferWriter` mixed logical positions with physical buffer offsets. This made reservations too small, placed `UnsafeWriteCopy` at the wrong location, truncated buffers after `UnsafeWrite`, and applied the offset twice during `Seek`.

Keep reservations, writes, length reporting, and seeking consistent with the configured offset. Add focused tests for regular writes, unsafe pattern writes, and seeking.

### Are these changes tested?

- `go test ./parquet/internal/encoding`
- `go test -race ./parquet/internal/encoding -run TestBufferWriter`

Parquet buffer writers now preserve the expected logical position and length when an offset is used.

### DIFF
--- a/parquet/internal/encoding/types.go
+++ b/parquet/internal/encoding/types.go
@@ -360,6 +360,7 @@ func (b *BufferWriter) Len() int {
 
 // Cap returns the current capacity of the underlying buffer
 func (b *BufferWriter) Cap() int {
+	b.ensureOffset()
 	return b.buffer.Cap() - b.offset
 }
 
--- a/parquet/internal/encoding/types.go
+++ b/parquet/internal/encoding/types.go
@@ -340,13 +340,21 @@ func (b *BufferWriter) SetOffset(offset int) {
 	b.offset = offset
 }
 
+func (b *BufferWriter) ensureOffset() {
+	if b.buffer.Len() < b.offset {
+		b.buffer.ResizeNoShrink(b.offset)
+	}
+}
+
 // Bytes returns the current bytes slice of slice Len
 func (b *BufferWriter) Bytes() []byte {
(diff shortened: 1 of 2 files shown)


zeroshade (commented):
The logical-versus-physical offset arithmetic now looks consistent across reservation, regular and unsafe writes, `WriteAt`, length reporting, and seeking. I found one low-risk consistency issue with `Cap()` in the inline comment.

I ran the focused tests, focused race tests, direct offset/`WriteAt` probes, and the full `parquet/...` subtree on head `1de189c2bb6d`; all passed. CI is still running.

---

>

parquet/internal/encoding/types.go
```diff
@@ -340,13 +340,21 @@ func (b *BufferWriter) SetOffset(offset int) {
 	b.offset = offset
 }
 
+func (b *BufferWriter) ensureOffset() {
```
zeroshade, line 343: Could we apply the same normalization in `Cap()` and cover it in `TestBufferWriterOffset`? With a fresh writer followed by `SetOffset(1024)`, `Cap()` currently returns `-1024`; after calling `Len()` or `Bytes()`, it returns `0` because those methods resize the prefix. That leaves capacity dependent on call order and permits a negative capacity.

zeroshade (approved):
The follow-up normalizes `Cap()` consistently with `Len()` and `Bytes()`, resolving the prior call-order and negative-capacity observation. The focused BufferWriter tests passed 100 repetitions, race passed 20 repetitions, the complete encoding package and 32-bit checks passed, and all CI checks are green.

---

>

PR #1145, opened 2026-08-08
fix(arrow/csv): preserve first row during headerless inference

fallintoplace:
## What

When schema inference runs with WithHeader(false), the first CSV record is consumed while generating f0, f1, and so on, but it is not sent through the builders. This patch keeps that record pending and replays it through the normal reader path.

## Test

- go test ./arrow/csv -run TestInferringSchemaWithoutHeaderPreservesFirstRow -count=1

### DIFF
--- a/arrow/csv/reader.go
+++ b/arrow/csv/reader.go
@@ -61,6 +61,7 @@ type Reader struct {
 	columnFilter   []string
 	columnTypes    map[string]arrow.DataType
 	conversions    []conversionColumn
+	pendingRecord  []string
 
 	stringsCanBeNull bool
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The headerless inference fix looks correct. The pending row is copied safely despite `ReuseRecord`, and it is replayed exactly once across single-row, chunked, and read-all modes.

I exercised the chunk modes, single-row and filtered-column cases, focused stress and race tests, the CSV package, and the full `arrow/...` subtree on head `2b3b88faca7e`; all passed. CI is green.

---

>

PR #1146, opened 2026-08-08
fix(arrow/array): bound table reader batches

fallintoplace:
## What

NewTable permits a column to be longer than the logical table row count. TableReader used the absolute row limit when choosing each batch size, so a 5-row table could emit a second batch that crossed the logical end. This uses the remaining row count.

## Test

- go test ./arrow/array -run TestTableReaderDoesNotExceedTableRowCount -count=1

### DIFF
--- a/arrow/array/table.go
+++ b/arrow/array/table.go
@@ -354,7 +354,7 @@ func (tr *TableReader) Next() bool {
 	}
 
 	// determine the minimum contiguous slice across all columns
-	chunksz := imin64(tr.max, tr.chksz)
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The logical-row bound looks correct. Using `tr.max - tr.cur` ensures the final record batch cannot extend beyond the table's declared row count while preserving the existing per-column chunk-boundary behavior.

I exercised the change across truncated tables, uneven and empty multi-column chunks, zero and negative chunk sizes, repeated iteration, and record ownership. The focused regression, race stress, full `arrow/array` package, and broader boundary properties all pass. CI is green.

---

>

PR #1147, opened 2026-08-08
fix(parquet/compress): honor LZ4 raw destination contract

fallintoplace:
## What

The LZ4 raw codec passed the provided capacity straight to CompressBlock. Nil or undersized destinations could panic instead of allocating as promised by Codec.Encode. This allocates the compression bound when the destination capacity is insufficient.

## Test

- go test ./parquet/compress -count=1

### DIFF
--- a/parquet/compress/compress_test.go
+++ b/parquet/compress/compress_test.go
@@ -130,6 +130,21 @@ func TestCompressDataOneShot(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The LZ4 raw codec now honors the documented destination-buffer contract while preserving caller-provided storage when capacity is sufficient. The regression and broader Parquet tests cover the allocation, round-trip, and reuse behavior. Looks good.

PR #1148, opened 2026-08-08
fix(arrow/csv): validate custom converter row counts

fallintoplace:
## What

The custom CSV writer converter is documented to return one string per array value, but the writer trusted the result. Short results silently produced empty cells and long results could panic while filling the record matrix. This returns arrow.ErrInvalid for either mismatch.

## Test

- go test ./arrow/csv -run TestCustomTypeConverterValidatesRowCount -count=1

### DIFF
--- a/arrow/csv/transformer.go
+++ b/arrow/csv/transformer.go
@@ -31,6 +31,9 @@ func (w *Writer) transformColToStringArr(typ arrow.DataType, col arrow.Array, st
 	if w.customTypeConverter != nil {
 		result, handled := w.customTypeConverter(typ, col)
 		if handled {
+			if len(result) != col.Len() {
+				return nil, fmt.Errorf("%w: custom type converter returned %d values for column with %d rows", arrow.ErrInvalid, len(result), col.Len())
+			}
 			return result, nil
 		}
 	}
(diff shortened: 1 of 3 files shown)


arrow/csv/transformer.go
```diff
@@ -31,6 +31,9 @@ func (w *Writer) transformColToStringArr(typ arrow.DataType, col arrow.Array, st
 	if w.customTypeConverter != nil {
 		result, handled := w.customTypeConverter(typ, col)
 		if handled {
+			if len(result) != col.Len() {
```
zeroshade, line 34: This correctly validates against `col.Len()`, but the public `WithCustomTypeConverter` documentation in `common.go` currently says the returned slice must match “the number of columns in the output csv.” Since this callback handles one column at a time, that should say it must contain one string per row/value in `col`. Could we update that documentation alongside the new validation?

zeroshade (approved):
LGTM — custom converter results are now validated before CSV matrix indexing, returning `arrow.ErrInvalid` for both short and long results without emitting partial rows.

I ran the regression 100 times, the focused race tests 20 times, a later-column atomicity probe, and the full `arrow/csv` package with its fixture submodules; all passed on head `95551a6d6600`. CI is green.

One nonblocking documentation correction is noted inline: `WithCustomTypeConverter` should describe its result as one string per row/value rather than per CSV column.

---

>

PR #1149, opened 2026-08-08
fix(arrow/csv): append null for later parse failures

fallintoplace:
## What

CSV scalar parsers already append a null when their own parse fails, but several only did so while the reader had no earlier error. A malformed value in a later column could therefore append the parser zero value after another column failed. This keeps the first error and appends null for every failed scalar conversion.

## Test

- go test ./arrow/csv -run TestCSVReaderAppendsNullAfterPreviousParseError -count=1

### DIFF
--- a/arrow/csv/reader.go
+++ b/arrow/csv/reader.go
@@ -817,23 +817,26 @@ func (r *Reader) parseListLike(field array.ListLikeBuilder, str string) {
 		return
 	}
 	if !strings.HasPrefix(str, "{") || !strings.HasSuffix(str, "}") {
-		r.err = errors.New("invalid list format. should start with '{' and end with '}'")
+		r.setParseError(errors.New("invalid list format. should start with '{' and end with '}'"))
+		field.AppendNull()
 		return
 	}
 	str = strings.Trim(str, "{}")
-	field.Append(true)
 	if len(str) == 0 {
 		// we don't want to create the csv reader if we already know the
 		// string is empty
+		field.Append(true)
 		return
 	}
-	valueBldr := field.ValueBuilder()
 	reader := csv.NewReader(strings.NewReader(str))
 	items, err := reader.Read()
 	if err != nil {
-		r.err = err
+		r.setParseError(err)
+		field.AppendNull()
 		return
 	}
+	field.Append(true)
+	valueBldr := field.ValueBuilder()
 	for _, str := range items {
 		r.initFieldConverter(valueBldr)(str)
 	}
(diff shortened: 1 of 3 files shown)


arrow/csv/reader_test.go
```diff
@@ -901,6 +901,36 @@ func benchRead(b *testing.B, raw []byte, rows, cols, chunks int) {
 	}
 }
 
+func TestCSVReaderAppendsNullAfterPreviousParseError(t *testing.T) {
+	schema := arrow.NewSchema([]arrow.Field{
```
zeroshade, line 905: This regression covers the scalar converters updated here, but the same failure invariant is still broken for list and fixed-width converters. With an earlier `int8` parse error:

- malformed `List` panics while building the record
- wrong-length `FixedSizeList` panics
- wrong-width `FixedSizeBinary` panics

Those converters return without appending a value, leaving builder lengths inconsistent. Please extend the fix and regression coverage so every failed field conversion contributes one null rather than allowing `NewRecordBatch` to panic.

arrow/csv/reader.go
```diff
@@ -385,6 +385,12 @@ func (r *Reader) isNull(val string) bool {
 	return false
 }
 
+func (r *Reader) setParseError(err error) {
```
zeroshade, line 388: Please route all conversion failures through this first-error helper, not only the numeric, temporal, and decimal subset. The binary converters below still assign `r.err` directly.

Reproduced with `first-error,%%%` under an `int8,binary` schema: the later base64 error replaces the original integer error. This contradicts the intended “keep the first error” behavior.

zeroshade (changes requested):
The scalar parser changes fix the reported case, but adjacent CSV converters still violate the same invariants: some later failures overwrite the first error, while list and fixed-width failures can leave unequal builder lengths and panic during record construction. See the two inline comments.

The focused regression passed 200 repetitions, race passed 30 repetitions, and package, example, 32-bit, and vet checks passed. CI is 23/23 green.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. After you've addressed the points above and pushed an update, an Apache Arrow Go maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
All converters now maintain the one-slot-per-row invariant and first-error precedence — the reordering of `Append(true)` after validation in the list paths is exactly right, and the table-driven regression coverage is thorough. Nice catch on the empty-string fixed-size-list case too.

---

PR #1150, opened 2026-08-08
fix(arrow/extensions): compare JSON storage types semantically

fallintoplace:
## What

NewJSONType used interface equality when checking supported storage types. A distinct StringType value is semantically the same Arrow type but could be rejected. This uses arrow.TypeEqual for the supported storage type check.

## Test

- go test ./arrow/extensions -count=1

### DIFF
--- a/arrow/extensions/json.go
+++ b/arrow/extensions/json.go
@@ -19,7 +19,6 @@ package extensions
 import (
 	"fmt"
 	"reflect"
-	"slices"
 
 	"github.com/apache/arrow-go/v18/arrow"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The semantic storage-type comparison looks correct. `arrow.TypeEqual` accepts distinct instances of the supported Arrow string types while preserving rejection of nil and unsupported storage types.

I exercised all three supported storage variants, extension equality, invalid inputs, and an IPC round trip constructed from a distinct storage-type instance. The focused stress, race checks, full `arrow/extensions` package, and CI all pass.

---

>

PR #1151, opened 2026-08-08
fix(flightsql): stop row streaming after close

fallintoplace:
## What

Rows.Close cancels the streaming context, but the producer could already be blocked sending a retained record into the full channel. This adds a context-aware send and releases the record when cancellation wins.

## Test

- go test ./arrow/flight/flightsql/driver -run TestRowsSendRecordStopsWhenContextCancelled -count=1

### DIFF
--- a/arrow/flight/flightsql/driver/driver.go
+++ b/arrow/flight/flightsql/driver/driver.go
@@ -117,6 +117,9 @@ func (r *Rows) Close() error {
 	r.currentRow = 0
 
 	r.releaseRecord()
+	for record := range r.recordChan {
+		record.Release()
+	}
 
 	return nil
 }
--- a/arrow/flight/flightsql/driver/driver.go
+++ b/arrow/flight/flightsql/driver/driver.go
@@ -100,6 +100,16 @@ func (r *Rows) releaseRecord() {
 	}
 }
 
+func (r *Rows) sendRecord(ctx context.Context, record arrow.RecordBatch) bool {
+	select {
+	case r.recordChan <- record:
+		return true
+	case <-ctx.Done():
+		record.Release()
+		return false
+	}
+}
+
 // Close closes the rows iterator.
 func (r *Rows) Close() error {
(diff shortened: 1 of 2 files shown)


arrow/flight/flightsql/driver/driver.go
```diff
@@ -100,6 +100,16 @@ func (r *Rows) releaseRecord() {
 	}
 }
 
+func (r *Rows) sendRecord(ctx context.Context, record arrow.RecordBatch) bool {
+	select {
+	case r.recordChan <- record:
+		return true
+	case <-ctx.Done():
+		record.Release()
```
zeroshade, line 108: This releases the record blocked in the send, but `Rows.Close` still leaks the retained record already buffered in `recordChan`. `streamRecordset` calls `Retain()` before sending, the channel has capacity one, and `Close` only cancels and releases `currentRecord`; it never drains the channel. Please make close/producer teardown release buffered records and cover the real `Rows.Close` path with retained ownership rather than testing `sendRecord` alone.

zeroshade (changes requested):
The context-aware send resolves the blocked producer, but closing rows still leaves the retained batch already buffered in `recordChan` unreleased. A checked-allocator probe retained 128 bytes after `Rows.Close`; draining and releasing the buffered record reduced it to zero.

I ran the focused test 100 times, the race test 10 times, and the full `arrow/flight/...` subtree on head `de5d91a76527`; all passed. CI workflows are still awaiting approval.

---

>

zeroshade (approved):
The follow-up releases both records blocked during cancellation and records already buffered in `recordChan`, resolving the retained-record leak. The checked-allocator regression covers the real `Rows.Close` path and verifies ownership returns to zero.

Re-review at `04030513e351`: the focused test passed 500 repetitions, race passed 100 repetitions, and the driver, FlightSQL, and 32-bit suites passed. All required CI checks are green.

---

>

PR #1152, opened 2026-08-08
fix(flightsql): reject unsupported transaction options

fallintoplace:
## What

- Connection.BeginTx used sql.TxOptions instead of driver.TxOptions.
- Because of that, Connection did not implement driver.ConnBeginTx and database/sql never called the method.
- Use driver.TxOptions so database/sql forwards the caller context and transaction options.
- Return ErrNotSupported for non-default isolation levels and read-only transactions because FlightSQL does not support configuring them.
- Add a compile-time interface check and test the behavior through sql.DB.BeginTx.

## Test

- go test ./arrow/flight/flightsql/driver -run TestBeginTxRejectsUnsupportedOptions -count=1
- go test -race ./arrow/flight/flightsql/driver -run TestBeginTxRejectsUnsupportedOptions -count=1
- go test ./arrow/flight/flightsql/... -count=1
- go test ./arrow/flight/... -count=1
- go vet ./arrow/flight/flightsql/driver

### DIFF
--- a/arrow/flight/flightsql/driver/driver.go
+++ b/arrow/flight/flightsql/driver/driver.go
@@ -471,6 +471,8 @@ type Connection struct {
 	timeout time.Duration
 }
 
+var _ driver.ConnBeginTx = (*Connection)(nil)
+
 // Prepare returns a prepared statement, bound to this connection.
 func (c *Connection) Prepare(query string) (driver.Stmt, error) {
 	return c.PrepareContext(context.Background(), query)
@@ -620,11 +622,11 @@ func (c *Connection) Close() error {
 
 // Begin starts and returns a new transaction.
 func (c *Connection) Begin() (driver.Tx, error) {
-	return c.BeginTx(context.Background(), sql.TxOptions{})
+	return c.BeginTx(context.Background(), driver.TxOptions{})
 }
 
-func (c *Connection) BeginTx(ctx context.Context, opts sql.TxOptions) (driver.Tx, error) {
-	if opts.Isolation != sql.LevelDefault || opts.ReadOnly {
+func (c *Connection) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	if opts.Isolation != driver.IsolationLevel(sql.LevelDefault) || opts.ReadOnly {
 		return nil, fmt.Errorf("%w: transaction options are not supported", ErrNotSupported)
 	}
 
--- a/arrow/flight/flightsql/driver/driver.go
+++ b/arrow/flight/flightsql/driver/driver.go
@@ -451,9 +451,11 @@ func (c *Connector) Connect(ctx context.Context) (driver.Conn, error) {
 	}
 	client.Alloc = c.allocator
 
-	return &Connection{
-		client:  client,
-		timeout: c.timeout,
+	return &connBeginTx{
+		Connection: &Connection{
+			client:  client,
+			timeout: c.timeout,
+		},
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


arrow/flight/flightsql/driver/driver.go
```diff
@@ -624,6 +624,10 @@ func (c *Connection) Begin() (driver.Tx, error) {
 }
 
 func (c *Connection) BeginTx(ctx context.Context, opts sql.TxOptions) (driver.Tx, error) {
+	if opts.Isolation != sql.LevelDefault || opts.ReadOnly {
```
zeroshade: This check is not reached through `database/sql`: `Connection.BeginTx` accepts `sql.TxOptions`, but `driver.ConnBeginTx` requires `driver.TxOptions`. Consequently, `*Connection` does not implement `driver.ConnBeginTx`; `database/sql` bypasses this method, rejects non-default options with its own generic errors, and falls back to `Begin()` for default transactions without forwarding the caller’s context. Please expose the proper interface hook—directly or through an adapter—add a compile-time interface assertion, and exercise the behavior through `sql.DB.BeginTx`.

zeroshade (changes requested):
The option validation itself is correct, but the method signature prevents `database/sql` from dispatching to it. This leaves the intended integration behavior unchanged and makes the new test cover only direct concrete-method calls.

I ran the focused test, focused race test, FlightSQL subtree, and broader Flight package on head `4c51edded1ab`; all passed. The explicit `driver.ConnBeginTx` compile-time assertion fails as described inline. CI is still running.

---

>

arrow/flight/flightsql/driver/driver.go
```diff
@@ -620,10 +622,14 @@ func (c *Connection) Close() error {
 
 // Begin starts and returns a new transaction.
 func (c *Connection) Begin() (driver.Tx, error) {
-	return c.BeginTx(context.Background(), sql.TxOptions{})
+	return c.BeginTx(context.Background(), driver.TxOptions{})
 }
 
-func (c *Connection) BeginTx(ctx context.Context, opts sql.TxOptions) (driver.Tx, error) {
+func (c *Connection) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
```
zeroshade: Changing this exported method from `sql.TxOptions` to `driver.TxOptions` is source-incompatible within v18. A downstream call that compiled previously now fails:

```text
cannot use sql.TxOptions{} as driver.TxOptions in argument to
(*Connection).BeginTx
```

Please avoid silently breaking the concrete API. One option is a private adapter returned by `Connector.Connect` that implements `driver.ConnBeginTx` and translates its options into the existing `Connection.BeginTx` method. That approach changes the returned concrete type, so whichever compatibility tradeoff is chosen should be explicit.

zeroshade (changes requested):
The `driver.ConnBeginTx` dispatch issue is fixed, but the implementation changes the signature of an exported v18 method and breaks existing concrete callers. See the inline comment on `driver.go`.

The corrected integration behavior is otherwise sound. The focused test passed 100 repetitions, race passed 20 repetitions, and the driver, FlightSQL, 32-bit, and vet checks all passed. CI is 23/23 green.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. After you've addressed the point above and pushed an update, an Apache Arrow Go maintainer — a real person — will take the next look at the PR. If you think the finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
Both prior blockers are resolved. The private `connBeginTx` adapter preserves the exported `Connection.BeginTx(context.Context, sql.TxOptions)` API while correctly implementing `driver.ConnBeginTx` for connections returned through `database/sql`.

The external compatibility probe, repeated integration and race tests, FlightSQL subtree, 32-bit build, vet, and all 23 CI checks pass. I found no new issues in the updated implementation.

---

>

PR #1153, opened 2026-08-08
fix(flightsql): protect configured authorization metadata

fallintoplace:
## What

The Flight SQL credential metadata was populated with configured authentication and then overwritten by arbitrary connection parameters. A parameter named authorization could replace a configured token or basic credential. Parameters are now copied first and configured authentication is applied last.

## Test

- go test ./arrow/flight/flightsql/driver -run TestRequestMetadataKeepsConfiguredAuthorization -count=1

### DIFF
--- a/arrow/flight/flightsql/driver/utils.go
+++ b/arrow/flight/flightsql/driver/utils.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -38,7 +39,7 @@ func (g grpcCredentials) GetRequestMetadata(ctx context.Context, uri ...string) 
 	md := make(map[string]string, len(g.params)+1)
 
 	for k, v := range g.params {
-		md[k] = v
+		md[strings.ToLower(k)] = v
 	}
 
 	switch {
--- a/arrow/flight/flightsql/driver/utils.go
+++ b/arrow/flight/flightsql/driver/utils.go
@@ -37,17 +37,15 @@ type grpcCredentials struct {
 func (g grpcCredentials) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
 	md := make(map[string]string, len(g.params)+1)
 
-	// Authentication parameters
+	for k, v := range g.params {
+		md[k] = v
+	}
+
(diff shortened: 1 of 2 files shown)


arrow/flight/flightsql/driver/utils.go
```diff
@@ -37,19 +37,17 @@ type grpcCredentials struct {
 func (g grpcCredentials) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
 	md := make(map[string]string, len(g.params)+1)
 
-	// Authentication parameters
+	for k, v := range g.params {
```
zeroshade, line 41: gRPC lowercases every key returned by `PerRPCCredentials` before sending it, so a parameter named `Authorization` remains distinct in this map but later collides with the configured lowercase `authorization`. Since both maps are iterated, either value can win. An exact-head in-memory gRPC probe reproduced this: the server received `Bearer attacker` on request 5. Please normalize parameter keys before applying configured auth, or exclude `authorization` case-insensitively, and add a mixed-case transport-level regression.

zeroshade (changes requested):
Found one blocking issue: case-variant connection parameters can still override configured Flight SQL authorization after gRPC normalizes metadata keys.

### Blocking — authorization collision remains case-insensitive (`arrow/flight/flightsql/driver/utils.go:40`)

This protects only the exact lowercase key. gRPC v1.82.1 lowercases all `PerRPCCredentials` keys before transmission, so parameter key `Authorization` collides with configured `authorization`. Map iteration then determines which value wins.

I reproduced this on head `7e49a2faa101` with an in-memory gRPC client and server: the server received `Bearer attacker` instead of the configured token on request 5.

Please normalize parameter keys before applying configured authentication, or remove/reject `authorization` case-insensitively, and add a mixed-case transport-level regression. See the inline comment on `arrow/flight/flightsql/driver/utils.go:40`.

---

>

zeroshade (approved):
The follow-up normalizes parameter metadata keys before applying configured credentials, so configured token or basic authentication deterministically wins over every case variant. The mixed-case transport regression verifies the server receives exactly one trusted authorization value.

Re-review at `99cf4a00b00b`: focused tests passed 100 repetitions, race passed 20 repetitions, and the driver, FlightSQL, 32-bit, and vet checks passed. All required CI checks are green.

---

>

PR #1154, opened 2026-08-08
fix(compute): preserve caller context during execution

fallintoplace:
## What

Keep the supplied context attached to compute execution so cancellation reaches result collection. Drain pending results before returning so the executor is not reused while work is still finishing.

## Test

- go test ./arrow/compute -count=1
- go test -race ./arrow/compute -run ^TestCallFunctionPreservesCallerCancellation -count=1

### DIFF
--- a/arrow/compute/exec.go
+++ b/arrow/compute/exec.go
@@ -158,7 +158,7 @@ func execInternal(ctx context.Context, fn Function, opts FunctionOptions, passed
 
 	ectx := GetExecCtx(ctx)
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
 	ch := make(chan Datum, ectx.ExecChannelSize)
@@ -170,7 +170,14 @@ func execInternal(ctx context.Context, fn Function, opts FunctionOptions, passed
 	}()
 
 	result = executor.WrapResults(ctx, ch, haveChunkedArray(input.Values))
-	if err == nil {
+	if ctx.Err() != nil {
+		for value := range ch {
+			if value != nil {
+				value.Release()
+			}
+		}
+	}
+	if err == nil && result != nil {
 		debug.Assert(executor.CheckResultType(result) == nil, "invalid result type")
 	}
 
--- a/arrow/compute/exec.go
+++ b/arrow/compute/exec.go
@@ -176,13 +176,16 @@ func execInternal(ctx context.Context, fn Function, opts FunctionOptions, passed
 				value.Release()
 			}
 		}
+		if err == nil {
+			err = context.Cause(ctx)
+		}
+		if result != nil {
+			result.Release()
+			result = nil
+		}
 	}
 	if err == nil && result != nil {
 		debug.Assert(executor.CheckResultType(result) == nil, "invalid result type")
-	}
-
-	if ctx.Err() == context.Canceled && result != nil {
-		result.Release()
 	}
 
 	return
--- a/arrow/compute/exec_test.go
+++ b/arrow/compute/exec_test.go
@@ -20,6 +20,7 @@ package compute
 
 import (
 	"context"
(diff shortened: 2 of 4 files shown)


zeroshade:

arrow/compute/exec.go
```diff
@@ -170,7 +170,14 @@ func execInternal(ctx context.Context, fn Function, opts FunctionOptions, passed
 	}()
 
 	result = executor.WrapResults(ctx, ch, haveChunkedArray(input.Values))
-	if err == nil {
+	if ctx.Err() != nil {
```
zeroshade, line 173: **Blocking:** This cancellation branch drains the channel but does not guarantee a cancellation error or clear a released result. A context-oblivious kernel can finish successfully after cancellation, leaving `err == nil`. If cancellation occurs before the first output, `CallFunction` returns `(nil, nil)`; after the first scalar output, `WrapResults` can return a partial result that line 185 releases but the named return still exposes. I reproduced the latter as a non-nil `ArrayDatum` with nil data and `context.Canceled`. Please propagate `context.Cause(ctx)` when `err` is nil, and release **and set `result = nil`** before returning. A regression test should use a kernel that does not itself return `ctx.Err()`.

arrow/compute/exec.go
```diff
@@ -158,7 +158,7 @@ func execInternal(ctx context.Context, fn Function, opts FunctionOptions, passed
 
 	ectx := GetExecCtx(ctx)
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
zeroshade, line 161: **Major:** Deriving this context from the caller makes the scalar collector’s cancellation branch externally reachable. When `scalarExecutor.WrapResults` has already moved outputs into its private `acc`, cancellation returns without releasing those chunks. The drain below cannot reclaim values already consumed from the channel. Please add scalar accumulated-result cleanup equivalent to `vectorExecutor.WrapResults`’s `releaseAccumulated()` path.

arrow/compute/exec_test.go
```diff
@@ -115,6 +119,51 @@ func ExecAddInt32(ctx *exec.KernelCtx, batch *exec.ExecSpan, out *exec.ExecResul
...
+	kernel := exec.NewScalarKernel(
+		[]exec.InputType{exec.NewExactInput(arrow.PrimitiveTypes.Int32)},
+		exec.NewOutputType(arrow.PrimitiveTypes.Int32),
+		func(ctx *exec.KernelCtx, _ *exec.ExecSpan, _ *exec.ExecResult) error {
+			close(started)
+			<-ctx.Ctx.Done()
+			return ctx.Ctx.Err()
+		}, nil)
+	require.NoError(t, fn.AddKernel(kernel))
+	require.True(t, GetFunctionRegistry().AddFunction(fn, false))
```
zeroshade: **Minor:** Registering this function in the process-wide registry makes the test non-repeatable: `go test ./arrow/compute -run '^TestCallFunctionPreservesCallerCancellation$' -count=2` fails on its second execution. Please install a child registry through `ExecCtx` for this test.

zeroshade (changes requested):
The caller-context propagation is correct in intent, but the new cancellation path exposes result ownership and error-contract problems. In particular, cancellation can return no error or an already-released result, and accumulated scalar outputs can leak. I’ve left details and reproduction cases inline.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. After you’ve addressed the points above and pushed an update, an Apache Arrow Go maintainer — a real person — will take the next look at the PR. If you think a finding is misapplied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
Re-review complete: all requested cancellation, result-ownership, and test-isolation fixes are addressed. The expanded scalar and vector regression coverage looks good, and CI is green.

---

>

PR #1155, opened 2026-08-08
fix(compute): report FuncMeta for meta functions

fallintoplace:
## What

Set the function kind when constructing a MetaFunction. NewMetaFunction now reports FuncMeta instead of the zero-value scalar kind.

## Test

- go test ./arrow/compute -count=1


zeroshade (approved):
Verified that NewMetaFunction now reports FuncMeta consistently with the function-kind contract. Direct-construction and registered-function coverage pass, along with the full compute suite, vet, formatting, and diff checks.

PR #1156, opened 2026-08-08
fix(arrow/array): preserve run-end encoded empty values

fallintoplace:
## What

Make RunEndEncodedBuilder.AppendEmptyValue and AppendEmptyValues append the value builder empty value instead of a null.

## Test

- go test ./arrow/array -run ^TestRunEndEncoded -count=1

### DIFF
--- a/arrow/array/encoded.go
+++ b/arrow/array/encoded.go
@@ -392,9 +392,10 @@ type RunEndEncodedBuilder struct {
 	maxRunEnd uint64
 
 	// currently, mixing AppendValueFromString & UnmarshalOne is unsupported
-	lastUnmarshalled interface{}
-	unmarshalled     bool // tracks if Unmarshal was called (in case lastUnmarshalled is nil)
-	lastStr          *string
+	lastUnmarshalled  interface{}
+	unmarshalled      bool // tracks if Unmarshal was called (in case lastUnmarshalled is nil)
+	lastValueWasEmpty bool
+	lastStr           *string
 }
 
 func NewRunEndEncodedBuilder(mem memory.Allocator, runEnds, encoded arrow.DataType) *RunEndEncodedBuilder {
@@ -449,6 +450,7 @@ func (b *RunEndEncodedBuilder) finishRun() {
 	b.lastUnmarshalled = nil
 	b.lastStr = nil
 	b.unmarshalled = false
+	b.lastValueWasEmpty = false
 	if b.length == 0 {
 		return
 	}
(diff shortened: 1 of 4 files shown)


zeroshade:

arrow/array/encoded.go
```diff
@@ -502,11 +502,15 @@ func (b *RunEndEncodedBuilder) NullN() int {
 }
 
 func (b *RunEndEncodedBuilder) AppendEmptyValue() {
-	b.AppendNull()
+	b.finishRun()
```
zeroshade, line 507: **Blocking:** After this call resets the unmarshalling state, the pending non-null empty value is indistinguishable from a previously unmarshalled null to `UnmarshalOne`. Calling `AppendEmptyValue()` followed by `UnmarshalOne(null)` produces `runEnds=[2]`, `values=[""]`, and logical values `["", ""]`; the appended null is silently converted to an empty string. Expected run ends and values are `[1,2]` and `["", null]`. Please track this state explicitly so a direct empty-value append cannot be mistaken for a null run, and add this sequence as a regression test.

arrow/array/encoded.go
```diff
@@ -502,11 +502,15 @@ func (b *RunEndEncodedBuilder) NullN() int {
...
+	b.values.AppendEmptyValue()
```
zeroshade, line 508: **Major:** This delegation creates an invalid result when the encoded type is a dictionary. `dictionaryBuilder.AppendEmptyValue()` appends index `0` without adding a dictionary entry. Consequently, `REE<int16, dictionary<int8, string>>` contains an index into a zero-length dictionary, and `GetOneForMarshal(0)` panics with `arrow/array: index out of range`. Please provide valid dictionary empty-value behavior or explicitly reject this case, with coverage.

zeroshade (changes requested):
The empty-value behavior is correct for ordinary string values, but it introduces two correctness issues in mixed-builder and dictionary-backed cases. One silently converts a null to an empty value; the other creates an invalid dictionary reference that panics when read. Details and reproductions are inline.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. After you’ve addressed the points above and pushed an update, an Apache Arrow Go maintainer — a real person — will take the next look at the PR. If you think a finding is misapplied, please reply on the PR and a maintainer will weigh in.*
>

zeroshade (approved):
Re-review complete: both requested correctness fixes are addressed. Empty values remain distinct from subsequent nulls, and dictionary-backed empty values now produce valid dictionary entries. The added regression coverage looks good, and CI is green.

---

>

PR #1157, opened 2026-08-08
fix(arrow/array): validate union JSON type codes

fallintoplace:
## What

Validate union type IDs before narrowing them to int8 in both sparse and dense JSON decoders. Out-of-range, negative, and nonnumeric type IDs are rejected.

## Test

- go test ./arrow/array -run ^TestUnion -count=1

### DIFF
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -776,31 +776,14 @@ type unionBuilder struct {
 	typesBuilder *int8BufferBuilder
 }
 
-func unionTypeCodeFromJSON(dec *json.Decoder, typeID any, typ arrow.DataType) (arrow.UnionTypeCode, error) {
-	var id int64
-	switch tid := typeID.(type) {
-	case json.Number:
-		var err error
-		id, err = tid.Int64()
-		if err != nil {
-			return 0, err
-		}
-	case float64:
-		if tid != math.Trunc(tid) || tid < 0 || tid > float64(arrow.MaxUnionTypeCode) {
-			return 0, &json.UnmarshalTypeError{
-				Offset: dec.InputOffset(),
-				Type:   reflect.TypeOf(int8(0)),
-				Struct: fmt.Sprint(typ),
-				Value:  "float",
-			}
-		}
-		id = int64(tid)
-	default:
+func unionTypeCodeFromJSON(dec *json.Decoder, typeID json.RawMessage, typ arrow.DataType) (arrow.UnionTypeCode, error) {
+	id, err := json.Number(string(typeID)).Int64()
+	if err != nil {
 		return 0, &json.UnmarshalTypeError{
 			Offset: dec.InputOffset(),
 			Type:   reflect.TypeOf(int8(0)),
 			Struct: fmt.Sprint(typ),
-			Value:  "union type code",
+			Value:  "integer",
 		}
 	}
 
@@ -1114,7 +1097,6 @@ func (b *SparseUnionBuilder) AppendValueFromString(s string) error {
 }
 
 func (b *SparseUnionBuilder) UnmarshalOne(dec *json.Decoder) error {
-	dec.UseNumber()
 	t, err := dec.Token()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


arrow/array/union.go
```diff
@@ -776,6 +776,45 @@ type unionBuilder struct {
...
+func unionTypeCodeFromJSON(dec *json.Decoder, typeID any, typ arrow.DataType) (arrow.UnionTypeCode, error) {
+	var id int64
+	switch tid := typeID.(type) {
+	case json.Number:
+		var err error
+		id, err = tid.Int64()
+		if err != nil {
+			return 0, err
+		}
+	case float64:
```
zeroshade: **Blocking:** Default-decoder entry points validate the already-rounded `float64`, so invalid numeric type codes can become valid before this check. For both dense and sparse builders, `AppendValueFromString("[-1e-400, 1]")` accepts the mathematically negative value as code `0`, and `[127.00000000000000001, 1]` is accepted as code `127`. `UnmarshalJSON` rejects the same inputs because it enables `UseNumber`, so the public entry points are inconsistent. Please preserve the numeric lexeme before validation across `AppendValueFromString`/`UnmarshalOne` as well, and add regression coverage for underflow and rounded fractional values.

arrow/array/union_test.go
```diff
@@ -78,6 +78,45 @@ func TestUnionBuilderChildBounds(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			for _, typeCode := range []string{"256", "-1", "127", "1.5", "null"} {
+				builder := tc.new()
+				err := func() (err error) {
+					defer func() {
+						if r := recover(); r != nil {
```
zeroshade: **Blocking:** This recovery wrapper converts the panic being guarded against into an `error`, after which `assert.Error` passes. I verified that deleting both new bounds checks still leaves this test green: the `127` cases panic during indexing, are converted here, and satisfy the assertion. Please let unexpected panics fail naturally—or assert no panic separately—then assert that decoding returned an error.

zeroshade (changes requested):
Found two blocking validation/test issues: default-decoder entry points can accept invalid type-code lexemes after `float64` rounding, and the regression test converts the panic it is meant to prevent into a passing error assertion. Local package and stdlib-JSON union tests otherwise pass.

---

>

arrow/array/union.go
```diff
@@ -1070,10 +1109,12 @@ func (b *SparseUnionBuilder) AppendValueFromString(s string) error {
 		return nil
 	}
 	dec := json.NewDecoder(strings.NewReader(s))
+	dec.UseNumber()
 	return b.UnmarshalOne(dec)
 }
 
 func (b *SparseUnionBuilder) UnmarshalOne(dec *json.Decoder) error {
+	dec.UseNumber()
```
zeroshade: **Blocking:** `UnmarshalOne` receives a caller-owned decoder, and `UseNumber` permanently changes how every later value from that decoder is interpreted. This makes enclosing objects key-order-dependent: for a struct with union field `u` and `int32` field `i`, `{"i":1.5,"u":[0,1]}` succeeds with a default decoder, while the equivalent `{"u":[0,1],"i":1.5}` fails because decoding `u` switches the shared decoder before `i` is read. The dense implementation has the same issue. Please preserve the type-code lexeme without reconfiguring the supplied decoder (for example, decode that token directly into `json.Number` or `RawMessage`); only internally created decoders should call `UseNumber`.

zeroshade (changes requested):
The two original blockers are fixed: all decoder entry points now reject rounded invalid type-code lexemes, and unexpected panics are no longer converted into passing error assertions. One new decoder-state blocker remains.

---

>

zeroshade (approved):
Reviewed at f0d1a61. The union decoder now preserves caller-owned decoder configuration while validating the original type-code lexeme. Dense/sparse behavior, maximum type code handling, and dynamic-child rollback passed repeated tests under both JSON backends, the full default and race suites, formatting, vet, and diff hygiene.

PR #1158, opened 2026-08-09
fix(arrow): include REE value nullability in type equality

fallintoplace:
### Rationale for this change

TypeEqual compares run-end and value data types for run-end encoded types, but it does not compare whether the value field is nullable. Structurally different types can therefore compare equal and produce the same type fingerprint.

Include ValueNullable in run-end encoded type equality and fingerprints.

### Are these changes tested?

Yes. The regression test verifies both comparison directions and distinct fingerprints for nullable and non-nullable value fields. The full arrow package suite passes.

Run-end encoded types with different value nullability now compare unequal and have different fingerprints.

### DIFF
--- a/arrow/compare.go
+++ b/arrow/compare.go
@@ -150,7 +150,8 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 	case *RunEndEncodedType:
 		r := right.(*RunEndEncodedType)
 		return TypeEqual(l.Encoded(), r.Encoded(), opts...) &&
-			TypeEqual(l.runEnds, r.runEnds, opts...)
+			TypeEqual(l.runEnds, r.runEnds, opts...) &&
+			l.ValueNullable == r.ValueNullable
 	case *ListViewType:
 		return l.elem.Equal(right.(*ListViewType).elem)
 	default:
--- a/arrow/extensions/timestamp_with_offset.go
+++ b/arrow/extensions/timestamp_with_offset.go
@@ -44,12 +44,9 @@ func isOffsetTypeOk(offsetType arrow.DataType) bool {
 	case *arrow.DictionaryType:
 		return arrow.TypeEqual(offsetType.ValueType, arrow.PrimitiveTypes.Int16)
 	case *arrow.RunEndEncodedType:
-		return offsetType.ValidRunEndsType(offsetType.RunEnds()) &&
+		return !offsetType.ValueNullable &&
+			offsetType.ValidRunEndsType(offsetType.RunEnds()) &&
 			arrow.TypeEqual(offsetType.Encoded(), arrow.PrimitiveTypes.Int16)
-		// FIXME: Technically this should be non-nullable, but a Arrow IPC does not deserialize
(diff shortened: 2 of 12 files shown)


arrow/extensions/timestamp_with_offset.go
```diff
@@ -46,10 +46,6 @@ func isOffsetTypeOk(offsetType arrow.DataType) bool {
 	case *arrow.RunEndEncodedType:
 		return offsetType.ValidRunEndsType(offsetType.RunEnds()) &&
 			arrow.TypeEqual(offsetType.Encoded(), arrow.PrimitiveTypes.Int16)
```
zeroshade, line 49: **Blocking:** Now that IPC preserves `ValueNullable`, this compatibility check still ignores it. `RunEndEncodedOf(Int16, Int16)` defaults to `ValueNullable=true` and is accepted by `NewTimestampWithOffsetTypeCustomOffset`, while the canonical `NewTimestampWithOffsetTypeRunEndEncoded` constructor explicitly sets it to false. These accepted extension types now compare unequal and have different fingerprints; a nullable REE can deserialize successfully but fail `ExtensionEquals` against the canonical type. The removed FIXME said this check was disabled only because IPC lost the flag. Please either require `!offsetType.ValueNullable` or normalize legacy input explicitly, with rejection/round-trip coverage.

arrow/compare.go
```diff
@@ -150,7 +150,8 @@ func TypeEqual(left, right DataType, opts ...TypeEqualOption) bool {
 	case *RunEndEncodedType:
 		r := right.(*RunEndEncodedType)
 		return TypeEqual(l.Encoded(), r.Encoded(), opts...) &&
-			TypeEqual(l.runEnds, r.runEnds, opts...)
+			TypeEqual(l.runEnds, r.runEnds, opts...) &&
+			l.ValueNullable == r.ValueNullable
```
zeroshade, line 154: **Blocking:** Making equality strict exposes a public API gap: `array.NewRunEndEncodedArray` always calls `RunEndEncodedOf`, which resets `ValueNullable` to true. Existing code that defines a non-nullable REE schema and constructs its column with this public helper now gets unequal types and a `NewRecordBatch` type-mismatch panic—even though both types print identically. This PR works around the same problem with a private typed helper in `arrdata`. Please add a public constructor that accepts/preserves the `*RunEndEncodedType` and cover record construction and concatenation with `ValueNullable=false`.

zeroshade (changes requested):
Found two blocking API consistency issues: timestamp-with-offset still accepts a nullable REE offset type that is unequal to its canonical type, and strict REE equality leaves no public array constructor that can preserve `ValueNullable=false`. The branch also currently conflicts with `main` and has a failing Verify check.

---

>

zeroshade (approved):
The previous blockers are resolved. REE value nullability is now consistently represented in equality and fingerprints, preserved through array construction and serialization paths, and enforced for timestamp-with-offset storage types. The affected package, race, roundtrip, and tagged C Data tests passed locally.

PR #1159, opened 2026-08-09
fix(arrow): return owned field index slices

fallintoplace:
### Rationale for this change

Schema.FieldIndices and StructType.FieldIndices return slices backed by internal index maps. Mutating a returned slice can corrupt later field lookups.

Return cloned index slices from both methods. Schema.HasField reads the internal map directly so the ownership fix does not add an allocation there.

### Are these changes tested?

Yes. Regression tests mutate returned slices and verify that Schema and StructType keep their original indices. The full arrow package suite passes.

Callers now receive slices they can safely modify.


zeroshade (approved):
LGTM — `Schema.FieldIndices` and `StructType.FieldIndices` now return owned slices without adding an allocation to `Schema.HasField`. The ownership regressions, race checks, dependent Arrow tests, vet, and diagnostics passed locally. The GitHub workflows are awaiting maintainer approval rather than reporting failures.

---

>

PR #1160, opened 2026-08-09
fix(compute): propagate nested expression serialization errors

fallintoplace:
### Rationale for this change

SerializeExpr ignores errors returned while visiting call arguments. A nested unsupported literal can therefore produce a successful result with incomplete metadata, and scalar columns allocated before the error are not released.

Propagate errors from nested arguments and release all temporary scalar columns on both success and failure paths.

### Are these changes tested?

Yes. The regression test places a valid scalar before an unsupported nested array literal and checks both the returned error and allocator cleanup. The full arrow/compute package suite passes.

Nested serialization failures are now returned to the caller instead of being silently ignored.

### DIFF
--- a/arrow/compute/expression.go
+++ b/arrow/compute/expression.go
@@ -788,6 +788,11 @@ func SerializeExpr(expr Expression, mem memory.Allocator) (*memory.Buffer, error
 		metaValue []string
 		visit     func(Expression) error
 	)
+	defer func() {
+		for _, col := range cols {
+			col.Release()
+		}
+	}()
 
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The recursive error is now propagated correctly, and the function-wide cleanup covers temporary scalar columns on both success and failure paths. The regression test exercises the error and allocator cleanup. All CI checks are green.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Arrow Go maintainer. An Apache Arrow Go maintainer — a real person — made the final approval decision.*
>

PR #1161, opened 2026-08-09
fix(parquet/pqarrow): reject decimal overflow when writing integers

fallintoplace:
### Rationale for this change

When decimals are stored as Parquet integers, values outside the declared or physical precision can be narrowed to int32 or int64 without a runtime error. That can silently write a different value.

Validate non-null Decimal128 and Decimal256 values against both their declared precision and the target integer precision before conversion, and return arrow.ErrInvalid when a value does not fit.

### Are these changes tested?

Yes. The test covers Decimal128 and Decimal256 with both int32 and int64 Parquet storage. The full parquet/pqarrow package suite passes.

Invalid decimal arrays now return an error instead of producing corrupted integer values.

### DIFF
--- a/parquet/pqarrow/encode_arrow.go
+++ b/parquet/pqarrow/encode_arrow.go
@@ -121,11 +121,58 @@ func targetDecimalPrecision(cw file.ColumnChunkWriter) (int32, bool) {
 }
 
 func decimal128FitsTargetPrecision(val decimal128.Num, precision int32) bool {
-	return precision > 0 && precision <= decimal128.MaxPrecision && val.FitsInPrecision(precision)
+	if precision <= 0 || precision > decimal128.MaxPrecision {
+		return false
+	}
+	// Abs overflows for the minimum two's-complement value.
+	if val.Sign() < 0 && val.Negate() == val {
+		return false
+	}
+	return val.FitsInPrecision(precision)
 }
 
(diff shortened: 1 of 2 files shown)


parquet/pqarrow/encode_arrow.go
```diff
@@ -121,11 +121,58 @@ func targetDecimalPrecision(cw file.ColumnChunkWriter) (int32, bool) {
...
+					return err
+				}
+			}
+		}
+		return nil
+	}
+
+	valueIdx := 0
+	for _, defLevel := range defLevels {
+		if defLevel < levelInfo.RepeatedAncestorDefLevel {
```
zeroshade, line 162: Could we add a regression test exercising this repeated-ancestor branch? The new tests use flat or struct-only schemas, so `RepeatedAncestorDefLevel` remains zero. A case such as `list<decimal(4,0)>` with `[[], [99999], [1]]` would verify that a leading empty list does not misalign `valueIdx` and allow the overflowing value to escape validation. This is non-blocking.

zeroshade (approved):
The overflow validation looks correct, and the targeted and full `parquet/pqarrow` tests pass. Approving with one non-blocking test-coverage suggestion inline.

---

>

zeroshade (approved):

PR #1162, opened 2026-08-09
fix(arrow): validate precision in NewDecimalType

fallintoplace:
### Rationale for this change

NewDecimalType only checks maximum precision with debug assertions and does not reject zero precision. Normal builds can therefore create invalid decimal types.

Validate that precision is between 1 and the maximum supported by the requested decimal type, and return arrow.ErrInvalid otherwise.

### Are these changes tested?

Yes. The tests cover the lower and upper valid boundaries plus zero and overflow precision for all four decimal widths. The full arrow package suite passes.

NewDecimalType now returns an error for invalid precision instead of constructing an invalid type.

### DIFF
--- a/arrow/datatype_fixedwidth.go
+++ b/arrow/datatype_fixedwidth.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	"github.com/apache/arrow-go/v18/arrow/decimal"
-	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/internal/json"
 )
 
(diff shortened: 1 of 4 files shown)


zeroshade (approved):

arrow/datatype_fixedwidth.go
```diff
@@ -590,22 +589,32 @@ func NarrowestDecimalType(prec, scale int32) (DecimalType, error) {
...
-		debug.Assert(prec <= int32(decimal.MaxPrecision[decimal.Decimal256]()), "invalid precision for decimal256")
-		return &Decimal256Type{Precision: prec, Scale: scale}, nil
+		dtype = &Decimal256Type{Precision: prec, Scale: scale}
+		maxPrecision = int32(decimal.MaxPrecision[decimal.Decimal256]())
 	default:
 		return nil, fmt.Errorf("%w: must use one of the DECIMAL IDs to create a DecimalType", ErrInvalid)
 	}
+
+	if prec <= 0 || prec > maxPrecision {
+		return nil, fmt.Errorf("%w: precision for %s must be between 1 and %d, got %d",
```
zeroshade, line 614: **Blocking:** This new error path makes the existing caller in `arrow/avro/schema.go:475` unsafe: it does `dt, _ = arrow.NewDecimalType(...)`. For an Avro decimal with precision `77`, the actual `ErrInvalid` is discarded, `dt` becomes nil, and schema construction later returns only `invalid avro schema: arrow: field with nil DataType`. Consequently, callers cannot use `errors.Is(err, arrow.ErrInvalid)` or see the invalid precision. Please handle and propagate the constructor error at that call site—panicking with it would work with the existing Avro recovery boundary—and add an Avro regression test preserving the precision error.

zeroshade (changes requested):
The core precision bounds are correct, but the newly reachable error is discarded by the Avro schema conversion caller. Invalid Avro decimal precision therefore degrades into a generic nil-datatype error instead of preserving `arrow.ErrInvalid`. Local Arrow and Avro tests otherwise pass.

---

>

zeroshade (approved):

PR #1169, opened 2026-08-11
perf(parquet): batch delta binary packed decoding

fallintoplace:
### Rationale for this change

DELTA_BINARY_PACKED decoding called `BitReader.GetBatch` once for every value in a miniblock. The miniblock is already the unit decoded by this code, so this missed the existing batch unpackers.

- decode each miniblock into reusable scratch with one `GetBatch` call
- keep 33 to 64 bit values on the generic uint64 path instead of the 32 bit SIMD unpackers
- read SIMD input directly into pooled scratch so batched calls do not add allocations
- add int32 and int64 benchmarks with nonzero packed widths
- add coverage for batched values from 33 through 64 bits

### Are these changes tested?

Yes. I ran the full Parquet test tree with the parquet-testing fixtures, the internal utility and encoding tests with `noasm`, and a linux/amd64 cross-build.

Apple M1 Pro results for 65,536 values with `GOMAXPROCS=1` and 10 samples:

| Workload | Time |
| --- | ---: |
| int32 small deltas | -58.86% |
| int32 alternating wide deltas | -66.04% |
| int64 timestamp-like deltas | -59.72% |
| int64 deltas wider than 32 bits | -33.89% |

Allocations stay at 2 allocs/op in all four benchmarks.

### DIFF
--- a/parquet/internal/encoding/delta_bit_packing.go
+++ b/parquet/internal/encoding/delta_bit_packing.go
@@ -181,22 +181,22 @@ func (d *deltaBitPackDecoder[T]) unpackNextMini() error {
 			d.miniBlockValues = append(d.miniBlockValues, T(d.lastVal))
 		}
 	} else {
-		if cap(d.deltaBuf) < 1 {
-			d.deltaBuf = make([]uint64, 1)
+		if cap(d.deltaBuf) < n {
+			d.deltaBuf = make([]uint64, n)
 		}
-		d.deltaBuf = d.deltaBuf[:1]
+		d.deltaBuf = d.deltaBuf[:n]
 		minDelta := d.minDelta
 
-		for j := 0; j < n; j++ {
-			nread, err := d.bitdecoder.GetBatch(width, d.deltaBuf)
-			if err != nil {
-				return err
-			}
-			if nread != 1 {
-				return errors.New("parquet: eof exception")
-			}
+		nread, err := d.bitdecoder.GetBatch(width, d.deltaBuf)
+		if err != nil {
+			return err
+		}
+		if nread != n {
+			return errors.New("parquet: eof exception")
+		}
 
-			d.lastVal += int64(d.deltaBuf[0]) + minDelta
+		for _, delta := range d.deltaBuf {
+			d.lastVal += int64(delta) + minDelta
 			d.miniBlockValues = append(d.miniBlockValues, T(d.lastVal))
 		}
 	}
--- a/parquet/internal/encoding/delta_bit_packing.go
+++ b/parquet/internal/encoding/delta_bit_packing.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"math/bits"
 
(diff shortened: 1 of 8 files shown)


zeroshade:

parquet/internal/encoding/delta_bit_packing.go
```diff
@@ -181,22 +181,22 @@ func (d *deltaBitPackDecoder[T]) unpackNextMini() error {
...
-		for j := 0; j < n; j++ {
-			nread, err := d.bitdecoder.GetBatch(width, d.deltaBuf)
-			if err != nil {
-				return err
-			}
-			if nread != 1 {
-				return errors.New("parquet: eof exception")
-			}
+		nread, err := d.bitdecoder.GetBatch(width, d.deltaBuf)
+		if err != nil {
```
zeroshade: **Blocking:** `GetBatch` can return a partial count with a bare `io.EOF` when truncated input ends on a complete 32-value packed-group boundary. Returning that error here bypasses the `nread != n` corruption check below.

This becomes silent data loss in `pqarrow`: `recordReader.Err()` treats an exact `io.EOF` as normal end-of-stream. A truncation sweep of a valid 1,024-value delta page found multiple cuts where this branch returns bare `io.EOF`; end-to-end, the reader returns no rows and `Err() == nil`.

A miniblock must always contain all `n` padded values, so please convert any incomplete read—including bare `io.EOF`—to `io.ErrUnexpectedEOF` or another corruption error before returning. A regression test should assert the error kind, not only that an error occurred.

parquet/internal/encoding/delta_bit_packing.go
```diff
@@ -181,22 +181,22 @@ func (d *deltaBitPackDecoder[T]) unpackNextMini() error {
 			d.miniBlockValues = append(d.miniBlockValues, T(d.lastVal))
 		}
 	} else {
-		if cap(d.deltaBuf) < 1 {
-			d.deltaBuf = make([]uint64, 1)
+		if cap(d.deltaBuf) < n {
+			d.deltaBuf = make([]uint64, n)
```
zeroshade: This allocation is sized directly from the file-controlled `valsPerMini`. `miniBlockValues` already allocates `n*sizeof(T)` above; this new `[]uint64` scratch doubles the malformed-header amplification.

For example, a tiny header declaring `valsPerMini = 1<<24` makes the int64 decoder allocate roughly 256 MiB before discovering that the packed data is absent—about 128 MiB more than the merge base. The high-water allocation also remains attached to a reused decoder.

Could this scratch remain bounded—for example, decode in the same approximately 1,024-value chunks already used by `GetBatch`—and accumulate into `miniBlockValues` incrementally?

zeroshade (changes requested):
The miniblock batching approach is sound, the 33–64-bit scalar guard is necessary, and the measured speedups are substantial. The native SIMD and `noasm` utility/encoding suites both pass locally.

I found two issues that need addressing:

1. Truncated packed input can now escape as a bare `io.EOF`, which `pqarrow` interprets as successful end-of-stream, causing silent data loss.
2. The new full-miniblock `[]uint64` scratch doubles attacker-controlled memory amplification from malformed headers.

I included reproducer details and suggested fixes inline.

---

>

zeroshade (approved):

PR #1170, opened 2026-08-11
feat(arrow/array): add ValueAsAny for native Go values

### LINKED ISSUE #462: Add method to Array interface to retrieve native Go values

In downstream projects using the Apache Arrow Go library — particularly in driver implementations and generic data ingestion layers — there’s often a need to extract values from an arrow.Array without knowing its concrete type ahead of time.

Currently, this requires verbose type switches and manual downcasting for each array type, which adds boilerplate and makes generic handling of data more complex.

To simplify this pattern, we propose adding a method to the Array interface that returns the Go-native equivalent of the value at a given index. For example:

ValueAsAny(index int) any

This change would significantly reduce boilerplate and improve ergonomics in downstream projects that need to interface with Arrow arrays in a type-agnostic way.

### Component(s)

Other

dgvj-work:
### Rationale for this change

Downstream code often needs type-agnostic access to Arrow values as native Go types. `GetOneForMarshal` is JSON-oriented (e.g. `int8` → `float64`, timestamps/decimals as strings), which forces boilerplate type switches for generic ingestion.

Fixes #462

- Add `array.ValueAsAny(arr, i)` plus an optional `ValueAsAnyer` interface, without adding a method to `arrow.Array`
- Implement it across array types and common extensions so nulls return `nil` and values stay native (`int8`, `arrow.Timestamp`, decimal nums, `[]any` for lists, ordered `[name, value]` pairs for structs)
- Union slots keep their type id when the selected child is null (`[]any{typeID, nil}`)
- Variant decode errors are returned as the `any` value rather than collapsed to `nil`
- Add unit tests covering primitives, temporal/decimal, nested, dictionary, union, duplicate struct fields, and variant decode errors, including differences vs `GetOneForMarshal`

### Are these changes tested?

- `go test ./arrow/ ./arrow/array/ ./arrow/extensions/`

Yes — new `array.ValueAsAny` helper for native Go value access.

### DIFF
--- a/arrow/array.go
+++ b/arrow/array.go
@@ -110,12 +110,6 @@ type Array interface {
 	// ValueStr returns the value at index as a string.
 	ValueStr(i int) string
 
-	// ValueAsAny returns the native Go value at index i, or nil if the slot is null.
-	// Unlike GetOneForMarshal, values are not converted for JSON encoding
-	// (for example int8 stays int8, timestamps stay arrow.Timestamp, and
-	// nested values are []any / map[string]any of native values).
-	ValueAsAny(i int) any
-
 	// Get single value to be marshalled with `json.Marshal`
 	GetOneForMarshal(i int) interface{}
 
--- a/arrow/array.go
+++ b/arrow/array.go
@@ -110,6 +110,12 @@ type Array interface {
 	// ValueStr returns the value at index as a string.
 	ValueStr(i int) string
 
+	// ValueAsAny returns the native Go value at index i, or nil if the slot is null.
+	// Unlike GetOneForMarshal, values are not converted for JSON encoding
+	// (for example int8 stays int8, timestamps stay arrow.Timestamp, and
+	// nested values are []any / map[string]any of native values).
+	ValueAsAny(i int) any
+
 	// Get single value to be marshalled with `json.Marshal`
 	GetOneForMarshal(i int) interface{}
 
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -280,6 +280,18 @@ func (a *Struct) GetOneForMarshal(i int) interface{} {
 	return tmp
 }
 
+func (a *Struct) ValueAsAny(i int) any {
+	if a.IsNull(i) {
+		return nil
+	}
+	tmp := make(map[string]any)
+	fieldList := a.data.dtype.(*arrow.StructType).Fields()
+	for j, d := range a.fields {
+		tmp[fieldList[j].Name] = d.ValueAsAny(i)
+	}
+	return tmp
+}
+
 func (a *Struct) MarshalJSON() ([]byte, error) {
 	var buf bytes.Buffer
 	enc := json.NewEncoder(&buf)
--- a/arrow/array/struct.go
+++ b/arrow/array/struct.go
@@ -284,12 +284,12 @@ func (a *Struct) ValueAsAny(i int) any {
 	if a.IsNull(i) {
 		return nil
 	}
-	tmp := make(map[string]any)
 	fieldList := a.data.dtype.(*arrow.StructType).Fields()
+	out := make([]any, len(a.fields))
 	for j, d := range a.fields {
-		tmp[fieldList[j].Name] = d.ValueAsAny(i)
+		out[j] = []any{fieldList[j].Name, ValueAsAny(d, i)}
 	}
-	return tmp
+	return out
 }
 
 func (a *Struct) MarshalJSON() ([]byte, error) {
--- a/arrow/array/union.go
+++ b/arrow/array/union.go
@@ -333,6 +333,19 @@ func (a *SparseUnion) GetOneForMarshal(i int) interface{} {
 	return []interface{}{typeID, data.GetOneForMarshal(i)}
 }
 
+func (a *SparseUnion) ValueAsAny(i int) any {
+	typeID := a.RawTypeCodes()[i]
+
+	childID := a.ChildID(i)
+	data := a.Field(childID)
+
+	if data.IsNull(i) {
+		return nil
+	}
+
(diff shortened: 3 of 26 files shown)


arrow/array.go
```diff
@@ -110,6 +110,12 @@ type Array interface {
 	// ValueStr returns the value at index as a string.
 	ValueStr(i int) string
 
+	// ValueAsAny returns the native Go value at index i, or nil if the slot is null.
+	// Unlike GetOneForMarshal, values are not converted for JSON encoding
+	// (for example int8 stays int8, timestamps stay arrow.Timestamp, and
+	// nested values are []any / map[string]any of native values).
+	ValueAsAny(i int) any
```
zeroshade: Adding a method to the exported `arrow.Array` interface is source-incompatible for downstream implementations. A type satisfying the interface before this commit now fails with `missing method ValueAsAny`.

Unless this targets an explicitly breaking release, could this be exposed through a standalone helper or narrower optional interface instead?

arrow/array/union.go
```diff
@@ -333,6 +333,19 @@ func (a *SparseUnion) GetOneForMarshal(i int) interface{} {
 	return []interface{}{typeID, data.GetOneForMarshal(i)}
 }
 
+func (a *SparseUnion) ValueAsAny(i int) any {
+	typeID := a.RawTypeCodes()[i]
+
+	childID := a.ChildID(i)
+	data := a.Field(childID)
+
+	if data.IsNull(i) {
```
zeroshade: This union slot itself is valid and still has a meaningful type code even when its selected child is null. Returning bare `nil` loses that discriminator.

Please preserve the union representation, for example `[]any{typeID, nil}`, consistently with non-null union values.

arrow/array/struct.go
```diff
@@ -280,6 +280,18 @@ func (a *Struct) GetOneForMarshal(i int) interface{} {
...
 }
 
+func (a *Struct) ValueAsAny(i int) any {
+	if a.IsNull(i) {
+		return nil
+	}
+	tmp := make(map[string]any)
+	fieldList := a.data.dtype.(*arrow.StructType).Fields()
+	for j, d := range a.fields {
+		tmp[fieldList[j].Name] = d.ValueAsAny(i)
```
zeroshade: Arrow permits duplicate struct field names, but assigning by name into a map silently overwrites earlier fields. For `dup:int32=11, dup:int64=22`, this returns only `{"dup": 22}`.

Please use a representation that preserves field order and duplicates, or explicitly reject ambiguous schemas.

arrow/extensions/variant.go
```diff
@@ -612,6 +612,17 @@ func (v *VariantArray) GetOneForMarshal(i int) any {
 	return val.Value()
 }
 
+func (v *VariantArray) ValueAsAny(i int) any {
+	if v.IsNull(i) {
+		return nil
+	}
+	val, err := v.Value(i)
+	if err != nil {
```
zeroshade, line 620: A decode error on a non-null variant is converted to `nil`, making malformed data indistinguishable from an actual null slot. The reproduced case has `IsNull(i) == false` and invalid metadata, but `ValueAsAny(i) == nil`.

Please preserve an error signal rather than silently converting corruption to null.

zeroshade (changes requested):
Found four public-API correctness and compatibility issues that need to be addressed before this can merge; see inline comments on `arrow/array.go`, `arrow/array/union.go`, `arrow/array/struct.go`, and `arrow/extensions/variant.go`.

The remaining implementations look consistent across primitive, temporal, decimal, nested, dictionary, encoded, and extension arrays. Exact-head package, race, repeated, and 32-bit tests passed.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Arrow Go maintainer. After you've addressed the points above and pushed an update, an Arrow Go maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Arrow Go handles maintainer review:* [CONTRIBUTING.md](https://github.com/apache/arrow-go/blob/main/CONTRIBUTING.md).

arrow/array.go
dgvj-work: Moved it off `arrow.Array`. Callers can use `array.ValueAsAny(arr, i)`, and custom array types can opt in via the optional `ValueAsAnyer` interface.

arrow/array/union.go
dgvj-work: Updated. A null child now returns `[]any{typeID, nil}` so the union type id is preserved.

arrow/extensions/variant.go
dgvj-work: Decode errors are now returned as the `any` value (the `error` itself) instead of `nil`, so a corrupt non-null slot is distinguishable from an actual null.

arrow/array/struct.go
dgvj-work: Structs now return ordered `[name, value]` pairs instead of a map, so field order and duplicate names are preserved.

dgvj-work:
Hi @zeroshade - I've addressed all four review points in the latest commit (6293617) and replied on each thread. Ready for another look when you have a moment - and for CI approval once you're happy with the changes. Thanks!

zeroshade (approved):
Thanks for the updates! All four concerns are addressed — the standalone `array.ValueAsAny` + optional `ValueAsAnyer` interface avoids the breaking change to `arrow.Array`, unions preserve the type code for null children, structs preserve field order and duplicate names, and variant decode errors are no longer silently converted to null.

One note (non-blocking): returning the `error` itself as the `any` value from `VariantArray.ValueAsAny` works, but callers need to type-assert to distinguish it from a legitimate value. Fine for now given the single-return signature.

---

PR #1171, opened 2026-08-11
perf(arrow/array): compare fixed-width values in bulk

fallintoplace:
### Rationale for this change

Fixed-width array equality currently checks validity and values one element at a time. The values are already contiguous, so larger arrays can be compared much more efficiently in bulk.

- Compare validity bitmaps in bulk.
- Compare null-free integer and temporal arrays with `bytes.Equal`.
- Walk contiguous valid runs for nullable arrays, ignoring bytes under null slots.
- Keep floating-point arrays on the scalar path to preserve NaN and signed-zero behavior.
- Keep arrays shorter than 8 values on the scalar path.
- Add benchmarks for int32/int64 arrays, several null patterns, and mismatch positions.
- Add tests for null-slot and floating-point semantics.

On an M1 Pro with GOMAXPROCS=1, 65,536-value equality benchmarks improved as follows:

| Case | Before | After |
| --- | ---: | ---: |
| int32, all valid | 505.9 µs | 6.5 µs |
| int32, 1% null | 516.2 µs | 14.2 µs |
| int32, alternating null | 432.2 µs | 247.9 µs |
| int64, all valid | 520.0 µs | 13.2 µs |
| int64, 1% null | 523.0 µs | 20.3 µs |
| int64, alternating null | 439.0 µs | 245.9 µs |

Allocations remain at zero.

### Are these changes tested?

- `go test ./arrow/...`

The full Arrow test run used the checked-out Parquet test data.

No.

### DIFF
--- a/arrow/array/compare_test.go
+++ b/arrow/array/compare_test.go
@@ -613,6 +613,52 @@ func TestArrayEqualBaseArray(t *testing.T) {
 	}
 }
 
+func TestArrayEqualFloatingPointSemantics(t *testing.T) {
+	negativeZero := math.Copysign(0, -1)
+	tests := []struct {
+		name  string
+		left  interface{}
+		right interface{}
+		want  bool
+	}{
+		{"float16 signed zero", []float16.Num{float16.New(0)}, []float16.Num{float16.New(float32(negativeZero))}, false},
+		{"float16 NaN", []float16.Num{float16.New(float32(math.NaN()))}, []float16.Num{float16.New(float32(math.NaN()))}, true},
+		{"float32 signed zero", []float32{0}, []float32{float32(negativeZero)}, true},
+		{"float32 NaN", []float32{float32(math.NaN())}, []float32{float32(math.NaN())}, false},
+		{"float64 signed zero", []float64{0}, []float64{negativeZero}, true},
+		{"float64 NaN", []float64{math.NaN()}, []float64{math.NaN()}, false},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			left := arrayOf(nil, test.left, nil)
+			defer left.Release()
+			right := arrayOf(nil, test.right, nil)
+			defer right.Release()
+
+			assert.Equal(t, test.want, array.Equal(left, right))
+		})
+	}
+}
+
+func TestArrayEqualFixedWidthIgnoresNullValues(t *testing.T) {
+	valid := []bool{true, false, true}
+
+	leftBuilder := array.NewInt64Builder(memory.DefaultAllocator)
+	defer leftBuilder.Release()
+	leftBuilder.AppendValues([]int64{1, 111, 3}, valid)
+	left := leftBuilder.NewInt64Array()
+	defer left.Release()
+
+	rightBuilder := array.NewInt64Builder(memory.DefaultAllocator)
+	defer rightBuilder.Release()
+	rightBuilder.AppendValues([]int64{1, 222, 3}, valid)
+	right := rightBuilder.NewInt64Array()
+	defer right.Release()
+
+	assert.True(t, array.Equal(left, right))
+}
+
 func TestArrayEqualNull(t *testing.T) {
 	mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
 	defer mem.AssertSize(t, 0)
--- a/arrow/array/compare_test.go
+++ b/arrow/array/compare_test.go
@@ -641,34 +641,79 @@ func TestArrayEqualFloatingPointSemantics(t *testing.T) {
 	}
 }
 
-func TestArrayEqualFixedWidthIgnoresNullValues(t *testing.T) {
-	const sliceLength = 72
-	valid := make([]bool, sliceLength)
+const fixedWidthEqualityTestLength = 72
+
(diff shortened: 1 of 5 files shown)


arrow/array/numeric_generic.go
```diff
@@ -449,7 +451,44 @@ func NewDate64Data(data arrow.ArrayData) *Date64 {
...
+		return arrayEqualFixedWidthScalar(left, right)
+	}
+
+	leftValues := left.Values()
+	rightValues := right.Values()
+	if left.NullN() == 0 {
+		return bytes.Equal(arrow.GetBytes(leftValues), arrow.GetBytes(rightValues))
+	}
+
+	runs := bitutils.NewSetBitRunReader(
```
zeroshade, line 476: **Blocking:** the run reader panics when metadata reports partial nulls but the validity buffer is empty.

This is reachable from malformed IPC: `loadCommon` fetches a validity buffer whenever `NullCount() != 0`, while a zero-length buffer becomes `memory.NewBufferBytes(nil)`. `validityBitmapEqual` explicitly handles this empty-buffer case, but execution then reaches `NewSetBitRunReader`, which slices the empty bitmap.

I reproduced this with an eight-element `Int64` array, `nullCount=1`, and a zero-length validity buffer:

- merge base: returns a normal equality result
- this PR: panics in `baseSetBitRunReader.loadPartial`

Please avoid passing an empty bitmap to the run reader—either fall back to the scalar comparison, as the previous implementation did, or reject the malformed array earlier. Please add a regression test confirming `Equal` does not panic.

arrow/array/compare_test.go
```diff
@@ -613,6 +613,52 @@ func TestArrayEqualBaseArray(t *testing.T) {
...
+			defer left.Release()
+			right := arrayOf(nil, test.right, nil)
+			defer right.Release()
+
+			assert.Equal(t, test.want, array.Equal(left, right))
+		})
+	}
+}
+
+func TestArrayEqualFixedWidthIgnoresNullValues(t *testing.T) {
```
zeroshade: This test has only three elements, so it takes the `Len() < 8` scalar branch and never exercises the new nullable run/byte-comparison path it is intended to protect.

Please use at least eight elements and add sliced arrays with independent, non-byte-aligned offsets. The test should cover:

- differing bytes under null slots remain equal;
- a differing value in a valid run is unequal;
- runs crossing bitmap-byte or word boundaries;
- at least one temporal type such as `Timestamp`.

The existing benchmarks execute the bulk path, but the normal unit-test suite currently has no direct coverage of it.

zeroshade (changes requested):
The bulk equality implementation is correct for valid arrays. Differential testing across affected types, independent slice offsets, bitmap boundaries, null patterns, and nested/chunked wrappers found no valid-input divergence. The full array suite, race tests, vet, and diff checks pass.

I am requesting changes for two points noted inline:

1. The run reader introduces a new panic on an IPC-reachable malformed array with a nonzero null count and an empty validity buffer.
2. The new nullable fixed-width regression test is shorter than the bulk-path threshold, so it only tests the unchanged scalar path.

Nonblocking observation: `validityBitmapEqual` now treats `NullN() == 0` as authoritative without examining an attached validity bitmap. That changes behavior for inconsistent C Data metadata. It does not affect valid arrays, but documenting that policy—or preserving bitmap comparison when a bitmap exists—would make the malformed-input behavior explicit.

---

>

arrow/array/numeric_generic.go
```diff
@@ -449,7 +451,49 @@ func NewDate64Data(data arrow.ArrayData) *Date64 {
...
+
+func arrayEqualFixedWidth[T arrow.FixedWidthType](left, right fixedWidthArray[T]) bool {
+	// Avoid the fixed cost of bytes.Equal for very small arrays.
+	if left.Len() < 8 {
+		return arrayEqualFixedWidthScalar(left, right)
+	}
+
+	leftValues := left.Values()
+	rightValues := right.Values()
+	if left.NullN() == 0 {
```
zeroshade, line 467: **Nonblocking:** the follow-up now compares attached validity bitmaps even when `NullN() == 0`, but this fast path still treats `NullN()` as authoritative.

With two arrays that both declare `null_count = 0`, carry the same bitmap with an unset bit, and differ only in that null slot:

- merge base: equal
- current bulk path: unequal
- current scalar path for `Len() < 8`: equal

IPC and C Data imports can expose such inconsistent metadata because the declared count is not cross-checked against the bitmap.

This is nonblocking because the input violates Arrow's format invariants, but it leaves the new “respect attached validity bitmaps” behavior incomplete and length-dependent. If the bitmap should remain authoritative when present, gate the raw `bytes.Equal` path on an absent bitmap and add the complementary regression test. Otherwise, please document that `NullN() == 0` intentionally overrides an attached bitmap.

zeroshade (approved):
Both previously blocking findings are addressed:

- empty validity buffers now fall back safely without reaching the run reader;
- the fixed-width equality tests now exercise the bulk path with independent non-byte-aligned offsets, null-slot differences, valid-value differences, boundary-spanning runs, and `Timestamp` coverage.

The array, IPC, and C Data tests pass, as does the array race suite. I left one nonblocking observation inline about consistency when malformed imported metadata supplies an attached validity bitmap that disagrees with `NullN()`.

---

>

PR #1172, opened 2026-08-11
perf(parquet): decode BSS FLBA into contiguous storage

fallintoplace:
### Rationale for this change

`BYTE_STREAM_SPLIT` decoding for `FIXED_LEN_BYTE_ARRAY` currently allocates a separate byte slice for every output value when the output has no reusable capacity. A 65,536-value batch therefore performs 65,536 decoder allocations.

- Allocate one contiguous backing block for output values that need storage.
- Keep reusing output slices that already have enough capacity.
- Give each decoded value an exact-capacity slice.
- Only prepare the output prefix that is actually decoded.
- Add coverage for contiguous storage, mixed reusable output, output lifetime across decoder resets, partial output, and spaced decoding.
- Add cold-output benchmarks for widths 2, 4, 8, 16, and 32.

Apple M1 Pro, GOMAXPROCS=1, 65,536 values:

| Width | Before | After | Change | Allocations |
| ---: | ---: | ---: | ---: | ---: |
| 16 | 2.42 ms | 1.55 ms | -35.7% | 65,537 to 2 |
| 32 | 3.82 ms | 2.61 ms | -31.7% | 65,537 to 2 |

Bytes allocated stay unchanged. The existing reusable-output benchmark also remains allocation-free and improved by about 8%.

### Are these changes tested?

- `PARQUET_TEST_DATA=$PWD/parquet-testing/data go test ./parquet/...`
- `go test -race ./parquet/internal/encoding`
- `go vet -composites=false ./parquet/internal/encoding`
- Cross-compiled the encoding package tests for linux/amd64 and linux/s390x.

No.

### DIFF
--- a/parquet/internal/encoding/fixed_len_byte_array_decoder.go
+++ b/parquet/internal/encoding/fixed_len_byte_array_decoder.go
@@ -173,11 +173,18 @@ func (dec *ByteStreamSplitFixedLenByteArrayDecoder) Decode(out []parquet.FixedLe
 		return 0, errors.New("parquet: eof exception")
 	}
 
-	for i := range out {
-		if cap(out[i]) < dec.typeLen {
-			out[i] = make(parquet.FixedLenByteArray, dec.typeLen)
+	output := out[:toRead]
+	// Lazily allocate one backing buffer while continuing to reuse caller-provided storage.
+	var storage []byte
+	for idx := range output {
+		if cap(output[idx]) < dec.typeLen {
+			if storage == nil {
+				storage = make([]byte, (len(output)-idx)*dec.typeLen)
+			}
+			output[idx] = storage[:dec.typeLen:dec.typeLen]
+			storage = storage[dec.typeLen:]
 		} else {
-			out[i] = out[i][:dec.typeLen]
+			output[idx] = output[idx][:dec.typeLen]
 		}
 	}
 
--- a/parquet/internal/encoding/fixed_len_byte_array_decoder.go
+++ b/parquet/internal/encoding/fixed_len_byte_array_decoder.go
@@ -173,13 +173,13 @@ func (dec *ByteStreamSplitFixedLenByteArrayDecoder) Decode(out []parquet.FixedLe
 		return 0, errors.New("parquet: eof exception")
 	}
 
-	output := out[:toRead]
-	for idx := range output {
-		if cap(output[idx]) < dec.typeLen {
-			dec.prepareOutput(output[idx:])
+	out = out[:toRead]
+	for idx := range out {
+		if cap(out[idx]) < dec.typeLen {
+			dec.prepareOutput(out[idx:])
 			break
 		}
-		output[idx] = output[idx][:dec.typeLen]
+		out[idx] = out[idx][:dec.typeLen]
 	}
 
 	switch dec.typeLen {
--- a/parquet/internal/encoding/fixed_len_byte_array_decoder.go
+++ b/parquet/internal/encoding/fixed_len_byte_array_decoder.go
@@ -174,18 +174,12 @@ func (dec *ByteStreamSplitFixedLenByteArrayDecoder) Decode(out []parquet.FixedLe
 	}
 
 	output := out[:toRead]
-	// Lazily allocate one backing buffer while continuing to reuse caller-provided storage.
-	var storage []byte
 	for idx := range output {
 		if cap(output[idx]) < dec.typeLen {
-			if storage == nil {
-				storage = make([]byte, (len(output)-idx)*dec.typeLen)
-			}
-			output[idx] = storage[:dec.typeLen:dec.typeLen]
-			storage = storage[dec.typeLen:]
-		} else {
-			output[idx] = output[idx][:dec.typeLen]
+			dec.prepareOutput(output[idx:])
+			break
 		}
+		output[idx] = output[idx][:dec.typeLen]
 	}
 
 	switch dec.typeLen {
@@ -202,6 +196,27 @@ func (dec *ByteStreamSplitFixedLenByteArrayDecoder) Decode(out []parquet.FixedLe
 	dec.nvals -= toRead
 	dec.data = dec.data[toRead:]
 	return toRead, nil
+}
+
+// prepareOutput allocates storage for the entries in out that do not have enough
+// capacity, while continuing to reuse the entries that do.
+func (dec *ByteStreamSplitFixedLenByteArrayDecoder) prepareOutput(out []parquet.FixedLenByteArray) {
+	missing := 0
+	for idx := range out {
+		if cap(out[idx]) < dec.typeLen {
+			missing++
+		}
+	}
+
+	storage := make([]byte, missing*dec.typeLen)
+	for idx := range out {
+		if cap(out[idx]) < dec.typeLen {
+			out[idx] = storage[:dec.typeLen:dec.typeLen]
+			storage = storage[dec.typeLen:]
+		} else {
+			out[idx] = out[idx][:dec.typeLen]
+		}
+	}
 }
 
 func (dec *ByteStreamSplitFixedLenByteArrayDecoder) DecodeSpaced(out []parquet.FixedLenByteArray, nullCount int, validBits []byte, validBitsOffset int64) (int, error) {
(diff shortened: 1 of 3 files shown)


zeroshade:
Since this is a draft, i'll hold off on further review until it's marked ready

parquet/internal/encoding/fixed_len_byte_array_decoder.go
```diff
@@ -173,11 +173,18 @@ func (dec *ByteStreamSplitFixedLenByteArrayDecoder) Decode(out []parquet.FixedLe
...
-	for i := range out {
-		if cap(out[i]) < dec.typeLen {
-			out[i] = make(parquet.FixedLenByteArray, dec.typeLen)
+	output := out[:toRead]
+	// Lazily allocate one backing buffer while continuing to reuse caller-provided storage.
+	var storage []byte
+	for idx := range output {
+		if cap(output[idx]) < dec.typeLen {
+			if storage == nil {
+				storage = make([]byte, (len(output)-idx)*dec.typeLen)
```
zeroshade: This sizes the backing block from all remaining outputs rather than from the outputs that actually need storage. In a 65,536-value, width-16 batch where only `out[0]` lacks capacity, the merge base allocates 16 B while this code allocates 1,048,576 B.

The additional storage state in this loop also regresses the existing all-reusable steady-state benchmark on arm64:

- 1,024 values: approximately 1,060 → 1,174 ns/op
- 65,536 values: approximately 69.0 → 76.2 µs/op

Both results were stable across six 500 ms runs with `GOMAXPROCS=1`.

Since preserving reusable caller storage is part of this optimization, could the existing fast loop remain unchanged until the first insufficient-capacity entry, then delegate the suffix to a cold helper that counts missing entries and allocates exactly `missing * typeLen` bytes? Please also add a mixed-reuse benchmark or allocation assertion so this case remains covered.

zeroshade (changes requested):
The contiguous cold-output allocation is correct and substantially reduces allocations when every output entry is empty. Ownership across decoder resets, partial decoding, and spaced decoding also looks sound.

However, the current loop regresses the existing all-reusable steady-state path by about 10–11% on arm64 and can overallocate dramatically for mixed reusable output. I included measurements and a suggested fast-path/slow-path structure inline.

Targeted encoding and race suites pass. File and pqarrow integration tests were blocked only by the unavailable `PARQUET_TEST_DATA` checkout. GitHub currently reports no CI checks for this head.

---

> addressed the point above and pushed an update, an Apache Arrow Go
>

zeroshade (approved):
The previous finding is resolved, and I verified the fix against the exact scenario I described.

## Previous finding

> This sizes the backing block from all remaining outputs rather than from the outputs that actually need storage. In a 65,536-value, width-16 batch where only `out[0]` lacks capacity, the merge base allocates 16 B while this code allocates 1,048,576 B. […] could the existing fast loop remain unchanged until the first insufficient-capacity entry, then delegate the suffix to a cold helper that counts missing entries and allocates exactly `missing * typeLen` bytes? Please also add a mixed-reuse benchmark or allocation assertion so this case remains covered.

**Resolved.** Re-measuring the exact case — 65,536 values, width 16, only `out[0]` lacking capacity — `Decode` now performs **1 allocation of 16 bytes**, the same as the merge base, rather than 1,048,576 B.

The requested structure is in place: the fast loop runs unchanged until the first insufficient-capacity entry, then `prepareOutput` counts `missing` and allocates exactly `missing * typeLen`. The mixed-reuse case is covered two ways — `...MixedOutputAllocations` asserts a single allocation, and `...ReusesProvidedOutput` asserts `&out[3][0] - &out[1][0] == width`, which pins the packing so a reused entry cannot consume a slot.

On the steady-state regression: on x86-64 (`-count 8 -benchtime 400ms -cpu 1`, benchstat) no size is significantly worse and the geomean improves ~11%. The hot loop is now the merge base's loop minus the `else` branch, iterating `toRead` rather than `len(out)`, so it can only do less work than before. I was not able to re-measure on arm64, where I took the original numbers.

## Verification

**Byte-stream-split de-interleaving for arbitrary `type_length`.** `encLoc = stride*stream + element` is correct for any width, and advancing `dec.data` by `toRead` while holding `stride` fixed is right because the streams remain `stride` apart. Beyond the 2/4/8/16/32 in the new tests, I round-tripped widths **1, 3, 5, 7, 9, 12, 13, 17, 31, 33, 64, and 100** — all exact.

**Lifetime and aliasing.** No decoder-owned buffer: `storage` is a local allocated per call and ownership passes to the caller, so nothing outlives its references. Values are cut with three-index slicing (`storage[:typeLen:typeLen]`), so an `append` on one value cannot reallocate into the next one — that is the failure mode I was most concerned about, and it is handled correctly. Both in-tree consumers copy out promptly: `FixedSizeBinaryBuilder.Append` copies via `bufferBuilder.Append`, and `ReadBatch` re-clones through `cloneByteArray`.

**Reuse across calls, pages, and record batches.** `flbaRecordReader.valueBuf` is resliced across batches, so entries keep `cap == typeLen` and are overwritten in place; a caller-retained value is invalidated by the next `Decode` into the same `out`. This matches the merge base exactly — contiguous storage does not widen it.

**Partial decode.** Repeated `Decode(n)` with `n` smaller than the buffered count is correct across widths 3/8/16 and chunk sizes 1/3/7.

**Suites.** `parquet/internal/encoding`, `parquet/file`, and `parquet/pqarrow` all pass with `PARQUET_TEST_DATA` supplied, which closes the gap left open in my last review. `-race` is clean on the encoding package; `gofmt` and `go vet` are clean.

**Benchmark evidence.** The claim holds: at 65,536 values, allocations drop from **65,537/op to 2/op** at every benchmarked width.

## Non-blocking notes

1. `TestByteStreamSplitFixedLenByteArrayDecoderPartialOutput` does not exercise what it describes. `sentinel` is `"sentinel"` — exactly 8 bytes, matching the width — so `out[3][:8]` is the identity and the test passes with or without the `out[:toRead]` change; it only compares data pointers. Leaving the tail entries `nil` and asserting they stay `nil` would make it a real guard, since `Decode` now leaves them untouched where the merge base set them to `len == typeLen`.

2. Leaving `out[toRead:]` untouched is a caller-visible change worth a doc comment on `Decode`. `flbaRecordReader.ReadValuesDense` discards `Decode`'s return count and appends all of `values`; if `Decode` ever returned fewer than `toRead`, the merge base appended stale-but-correctly-sized bytes while this version leaves `nil` entries, and `FixedSizeBinaryBuilder.Append(nil)` panics. I could not construct a reachable case and the integration suites pass, so this is latent rather than demonstrated — but the contract is worth stating explicitly.

3. `ReadBatch` benefits less than the other paths, since `cloneByteArray` already allocates one contiguous block and re-points every value, making the new block immediately garbage there. Still a net improvement over N allocations; just worth knowing.

4. The new tests use only powers of two. Adding one odd width such as 7 is nearly free. I verified there is no bug at odd widths, so this is purely a regression guard.

## Separate pre-existing bug (does not block this PR)

While exercising the spaced path I found that BYTE_STREAM_SPLIT FLBA columns with **nulls spanning multiple data pages** decode values shifted by one position. This reproduces identically on the merge base (`f85b0b25`), so it is not caused by this PR and does not block it. I will open a separate issue.

It is worth mentioning here because `TestByteStreamSplitFixedLenByteArrayDecoderSpacedOutput` uses a single `SetData` and one `DecodeSpaced`, so it passes and gives more confidence about spaced coverage than is warranted.

Isolation: fails only for BYTE_STREAM_SPLIT (PLAIN passes on identical data and configuration); only when nulls *and* multiple pages are both present (a single 1 MiB page with nulls passes, and multiple pages without nulls pass); reproduced at widths 2, 3, 4, 7, 8, 16, 17, and 32.

---

Approving. Feel free to fold in the notes above if you like, but none of them need to hold this up.

> confirmed by an Apache Arrow Go maintainer. The findings cite
> the project's review criteria; if you think one is
> mis-applied, please reply on the PR and a maintainer will
> weigh in.*
>

zeroshade:
Filed the pre-existing BYTE_STREAM_SPLIT FLBA issue mentioned in my review as #1255, with a self-contained reproducer. It is independent of this PR — nothing to do here.

PR #1173, opened 2026-08-11
perf(arrow/array): append fixed-width nulls in bulk

fallintoplace:
### Rationale for this change

Fixed-width builders currently implement `AppendNulls` by calling `AppendNull` for every value. Large batches repeatedly reserve capacity, update the validity bitmap, and update builder counters.

- Reserve capacity once for a null batch.
- Clear the validity range in bulk.
- Update the length and null count once.
- Advance fixed-size binary value storage once.
- Keep the existing scalar path for a single null.
- Keep the existing builder capacity growth policy.
- Size fixed-size binary value storage to the effective builder capacity during `Resize`.
- Update the numeric builder template and generated source.
- Add correctness tests and benchmarks for representative fixed-width builders.

Null values and array results are unchanged. `FixedSizeBinaryBuilder.Resize` now reserves value storage together with validity storage. Empty-value appends are separate because their value bytes must also be zeroed.

### Benchmark

`BenchmarkFixedWidthBuilderAppendNulls` covers:

- `Int32`
- `Int64`
- `Decimal128`
- `Timestamp`
- `Boolean`
- `FixedSizeBinary(16)`
- Batch sizes of 1, 8, 64, 1,024, and 65,536

The benchmark creates a fresh builder for each iteration, so allocation costs are included. The same benchmark was run against `upstream/main` and this branch.

Allocation results for 65,536 nulls, checked against current `upstream/main` and this branch. Three runs per case on Apple M1 Pro with Go 1.26.3:

| Builder | Before B/op | After B/op | Before allocs/op | After allocs/op |
| Int32 | 580,864 | 551,168 | 23 | 5 |
| Int64 | 1,113,152 | 1,075,456 | 23 | 5 |
| Decimal128 | 2,169,655 | 2,124,080 | 25 | 7 |
| Timestamp | 1,113,232 | 1,075,536 | 24 | 6 |
| Boolean | 38,272 | 37,120 | 19 | 5 |
| FixedSizeBinary(16) | 2,170,319 | 2,124,104 | 28 | 7 |

Command:

```bash
go test -trimpath -p 1 ./arrow/array -run '^$' -bench '^BenchmarkFixedWidthBuilderAppendNulls$/./count-65536$' -benchmem -benchtime=150ms -count=3 -cpu=1
```

### Are these changes tested?

Yes.

- Regenerated the numeric builders and verified there is no diff.

`FixedSizeBinaryBuilder.Resize` now reserves the value buffer eagerly. Null-appending results and the general builder growth policy are unchanged.

### DIFF
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -79,9 +79,15 @@ func (b *BooleanBuilder) AppendNull() {
 }
 
 func (b *BooleanBuilder) AppendNulls(n int) {
-	for i := 0; i < n; i++ {
-		b.AppendNull()
+	if n <= 0 {
+		return
 	}
+	if n == 1 {
+		b.AppendNull()
+		return
+	}
+	b.Reserve(n)
+	b.builder.unsafeAppendNulls(n)
 }
 
 func (b *BooleanBuilder) AppendEmptyValue() {
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -87,7 +87,7 @@ func (b *BooleanBuilder) AppendNulls(n int) {
 		return
 	}
 	b.Reserve(n)
-	b.builder.unsafeAppendNulls(n)
+	b.unsafeAppendNulls(n)
 }
 
 func (b *BooleanBuilder) AppendEmptyValue() {
--- a/arrow/array/builder.go
+++ b/arrow/array/builder.go
@@ -355,6 +355,20 @@ func (b *builder) unsafeSetValid(length int) {
 	b.length = newLength
 }
 
+func (b *builder) unsafeAppendNulls(n int) {
+	if n <= 0 {
+		return
+	}
+
+	if n == 1 {
+		bitutil.ClearBit(b.nullBitmap.Bytes(), b.length)
+	} else {
+		bitutil.SetBitsTo(b.nullBitmap.Bytes(), int64(b.length), int64(n), false)
+	}
+	b.length += n
+	b.nulls += n
+}
+
 func (b *builder) unsafeAppendEmptyValues(data []byte, valueSize, length int) {
 	if length <= 0 {
 		return
--- a/arrow/array/builder_append_nulls_benchmark_test.go
+++ b/arrow/array/builder_append_nulls_benchmark_test.go
@@ -1,0 +1,63 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package array_test
+
+import (
+	"fmt"
+	"testing"
+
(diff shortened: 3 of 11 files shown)


zeroshade (changes requested):
Heads up that CI is red here on all 15 platform jobs — Linux/macOS/Windows, amd64 and arm64, Go 1.25 and 1.26, CGO and non-CGO. It's deterministic, not a flake.

The failing test is `TestArraySpan_MakeData/dict-extension-type/MakeData` in `arrow/compute/exec`, which this PR doesn't touch. I bisected it:

| commit | result |
|---|---|
| `227f6e3e` (base) | ok |
| `56875bc2` perf: append fixed-width nulls in bulk | ok |
| `e10b5fb3` style: use promoted builder method | ok |
| `0a9150d3` fix: avoid oversized power-of-two reserves | **first bad** |
| `b4d3c00e` fix: size fixed binary values with capacity | fail |

The entire production change in `0a9150d3` is one line in `arrow/array/builder.go`:

```go
newCap := bitutil.NextPowerOf2(b.length + elements)      // before
newCap := bitutil.NextPowerOf2(b.length + elements - 1)  // after
```

Reverting only that line on the PR head, with all the AppendNulls work left intact, flips FAIL→PASS; restoring it flips it back. The mechanism: the `-1` causes an extra resize call for the 2-element case, and that second resize passes through the `minBuilderCapacity` (32) clamp, inflating the string dictionary's validity buffer from 1 byte to 4. `span_test.go` compares against a hand-built 1-byte bitmap via `reflect.DeepEqual`, so it fails.

**Please drop `0a9150d3`.** Independent of the CI break, it doesn't buy anything. Capacities are identical at every n on the incremental-append path because `minBuilderCapacity` dominates, and the 2-element string case comes out *larger* — the opposite of the commit's stated goal:

| n | base int32 Cap | PR int32 Cap | base strBuf0 | PR strBuf0 |
|---|---|---|---|---|
| 1 | 32 | 32 | 1 | 1 |
| 2 | 32 | 32 | 1 | **4 (regression)** |
| 3–32 | 32 | 32 | 4 | 4 |
| 33–64 | 64 | 64 | 8 | 8 |

The only real gain is single-shot `Reserve(n)` at exact powers of two ≥ 32. The arithmetic itself is sound — `NextPowerOf2(need-1)` is the smallest power of two ≥ need, so there's no under-allocation — it just isn't worth the blast radius on every builder in the package. If you want it, land it separately with its own justification and an explicit decision about loosening that `span_test.go` assertion, which is arguably over-strict anyway.

The AppendNulls work itself is good and I'd like to take it. I differential-tested bulk `AppendNulls(n)` against n × `AppendNull()` across 12 builder types (int32, int64, uint8, float16, bool, decimal128, timestamp, fsb16, fsb3, all three interval types) × 14 values of n: validity bits identical at every index, data-buffer null slots identical and verified zeroed, interleaved and builder-reuse cases byte-identical, IPC round-trips clean. Codegen is consistent with the template — 15 call sites, no hand-edit drift. Benchmarks hold up at roughly 4–10× faster with about half the memory and ~4× fewer allocations.

Two smaller items:

- **`TestFixedWidthBuilderAppendNullsClearsReusedValidity` doesn't test what it claims.** I deleted the `ClearBit`/`SetBitsTo` call from `unsafeAppendNulls` entirely and it still passes, along with every other test in the package — `memory.Buffer.Resize` already zero-fills, so the stale bits are cleared before `AppendNulls` runs. The clearing is fine to keep as defensive code, but either construct a case that actually reaches a stale bit or rename the test so it doesn't claim coverage it lacks.
- **`FixedSizeBinaryBuilder.Resize` changed behaviour.** It now also sizes the values buffer, so `Resize` on a wide FSB type allocates eagerly where it previously grew lazily. Using `b.capacity` rather than `n` is correct since `b.resize` may clamp, but it deserves a comment and a line in the PR description.

Also worth noting for whoever reviews next: the `reserve()` change is the highest-blast-radius edit in the PR — it affects every builder in the package, including variable-length ones — and has no direct test. Its only coverage was a power-of-two capacity assertion, which is how it reached CI.

zeroshade (approved):

PR #1174, opened 2026-08-11
perf(arrow/array): compare binary values by valid runs

fallintoplace:
### Rationale for this change

Binary and string equality currently compares every non-null value separately. This repeats offset lookups and small byte or string comparisons even when a long valid run is stored contiguously.

- Walk contiguous valid runs for Binary, String, LargeBinary, and LargeString.
- Compare each run payload in one operation.
- Verify every value length so equal concatenated bytes with different boundaries still compare unequal.
- Keep null payloads ignored.
- Sample validity fragmentation and retain the scalar path for many tiny runs.
- Add benchmarks and tests for offsets, null patterns, mismatches, and value boundaries.

Apple M1 Pro results for 65,536 values of 32 bytes with `-cpu=1`:

| --- | ---: | ---: | ---: |
| Binary equal | 644 us | 360 us | 1.79x |
| Binary 10% null | 582 us | 439 us | 1.33x |
| Binary mismatch last | 648 us | 350 us | 1.85x |
| Binary different length | 453 us | 262 us | 1.73x |
| String equal | 633 us | 390 us | 1.62x |
| String 10% null | 604 us | 496 us | 1.22x |
| String mismatch last | 664 us | 406 us | 1.64x |
| String different length | 426 us | 235 us | 1.81x |

The alternating 50% null case selects the scalar path and remains neutral. All benchmark cases stay at zero allocations.

### Are these changes tested?

Yes.

- `go test ./arrow/...`
- Cross-compiled array tests for linux/amd64 and linux/s390x.

No.

### DIFF
--- a/arrow/array/binary.go
+++ b/arrow/array/binary.go
@@ -24,6 +24,7 @@ import (
 	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/internal/bitutils"
 	"github.com/apache/arrow-go/v18/internal/json"
@@ -495,7 +496,10 @@ func arrayEqualVariableWidth[T binaryOffset, V ~[]byte | ~string](
 	validity []byte,
 	equalValues func(V, V) bool,
 ) bool {
-	if nulls == 0 && len(validity) == 0 {
+	// A declared null count may be inconsistent with the validity bitmap.
+	// Verify zero-null bitmaps before comparing the whole payload.
+	if len(validity) == 0 ||
+		(nulls == 0 && bitutil.CountSetBits(validity, leftOffset, length) == length) {
 		return arrayEqualVariableWidthRun(
 			leftOffsets, rightOffsets,
 			leftValues, rightValues,
--- a/arrow/array/binary.go
+++ b/arrow/array/binary.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/internal/bitutils"
 	"github.com/apache/arrow-go/v18/internal/json"
 )
 
@@ -231,15 +232,21 @@ func (a *Binary) ValidateFull() error {
 }
 
 func arrayEqualBinary(left, right *Binary) bool {
-	for i := 0; i < left.Len(); i++ {
-		if left.IsNull(i) {
-			continue
+	if useScalarVariableWidthEquality(left) {
+		for i := range left.Len() {
+			if !left.IsNull(i) && !bytes.Equal(left.Value(i), right.Value(i)) {
+				return false
+			}
 		}
-		if !bytes.Equal(left.Value(i), right.Value(i)) {
-			return false
-		}
+		return true
 	}
-	return true
+	return arrayEqualVariableWidth(
+		left.valueOffsets, right.valueOffsets,
+		left.valueBytes, right.valueBytes,
+		left.Offset(), right.Offset(), left.Len(),
+		left.NullN(), left.NullBitmapBytes(),
+		bytes.Equal,
+	)
 }
 
 type LargeBinary struct {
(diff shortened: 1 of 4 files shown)


arrow/array/binary.go
```diff
@@ -418,17 +425,117 @@ func (a *LargeBinary) ValidateFull() error {
...
+		return false
+	}
+
+	// Very short validity runs cost more to set up than direct value comparisons.
+	// Sample a few runs and retain the scalar path when they average under four values.
+	const (
+		sampleRuns          = 8
+		minAverageRunLength = 4
+	)
+	runs := bitutils.NewSetBitRunReader(
```
zeroshade, line 478: **Blocking:** `NullN() > 0` does not guarantee that a validity bitmap is present. Malformed IPC or C Data can supply a nonzero declared null count with an empty validity buffer.

This unconditionally constructs a run reader over that empty slice. I reproduced equality on two eight-element Binary arrays with `nullCount=1` and an empty validity buffer:

- merge base: returns normally
- this PR: panics in `baseSetBitRunReader.loadPartial`

An independent end-to-end IPC probe reproduced the same panic after changing only the serialized `null_count` while leaving the validity buffer absent.

Please check `len(values.NullBitmapBytes())` before constructing the reader and retain the scalar fallback when it is empty. Add a regression test confirming equality does not panic on this malformed-input case.

arrow/array/binary.go
```diff
@@ -418,17 +425,117 @@ func (a *LargeBinary) ValidateFull() error {
...
+type binaryOffset interface {
+	~int32 | ~int64
+}
+
+func useScalarVariableWidthEquality(values arrow.Array) bool {
+	if values.NullN() == 0 {
+		return false
+	}
+
+	// Very short validity runs cost more to set up than direct value comparisons.
```
zeroshade, line 472: The sampling heuristic introduces a substantial regression for short arrays with fragmented validity because it constructs and advances a run reader before ultimately selecting the original scalar loop.

On Apple M4, alternating-null String arrays measured:

| Length | Merge base | Current | Regression |
|---:|---:|---:|---:|
| 4 | ~27 ns | ~68 ns | 2.5× |
| 8 | ~42 ns | ~100 ns | 2.4× |
| 16 | ~74 ns | ~115 ns | 55% |
| 32 | ~147 ns | ~187 ns | 27% |
| 64 | ~282 ns | ~315 ns | 11% |

The submitted benchmark only uses 65,536-element arrays, so it cannot expose this fixed-cost regression. Please add a cheap length cutoff before sampling and include short-array cases in the benchmark matrix. A performance change should not make these common small-array comparisons materially slower.

zeroshade (changes requested):
The run-based comparison is correct for valid arrays, and differential testing found no semantic divergence across Binary, String, LargeBinary, LargeString, random null patterns, and independent slices. The claimed large-array improvements also hold.

I am requesting changes for two points noted inline:

1. A malformed IPC/C Data array with a nonzero declared null count and an empty validity buffer now panics while selecting the comparison strategy.
2. The sampling heuristic materially regresses short arrays with fragmented validity, but the submitted benchmark only covers 65,536-element arrays.

The full Arrow suite, array race suite, vet, formatting, and diff checks otherwise pass.

---

>

zeroshade (changes requested):
Both points from my last review are addressed — the empty-validity-buffer guard prevents the panic at any length, and the severe short-array regressions at lengths 4 and 8 are gone.

The run comparison itself is correct. I specifically checked the failure mode I was most worried about — bulk-comparing across mismatched offset layouts — and it's handled properly: `arrayEqualVariableWidthRun` does the memcmp and *then* verifies every per-value length delta across the run, so equal concatenated bytes with shifted boundaries are correctly rejected. Runs break at nulls so the memcmp never spans a null slot, and `run.Pos` being window-relative makes the sliced-array arithmetic come out right. A differential fuzz against a scalar reference — 4000 iterations, 2961 of which hit the run path, covering differing null payloads, independently sliced arrays, shifted value boundaries and empty values across all four offset types — found zero divergences.

So the implementation needs no change. I'm requesting changes on two test-only items.

### 1. The multi-run path is never executed by any test

I instrumented `useScalarVariableWidthEquality` against every case in the PR:

```
3elem_1null      ("null payload ignored")                    len=   3 nullN=  1 -> scalar loop
1024_alternating (TestBinaryEqualityWithFragmentedValidity)  len=1024 nullN=512 -> scalar loop
8elem no-bitmap  (...WithoutBitmap)                          len=   8 nullN=  1 -> scalar loop
```

Every null-containing test falls to the scalar path, and the all-valid tests only reach the `nulls == 0` whole-array single-run branch. `TestBinaryEqualityWithFragmentedValidity` looks like it should cover this, but alternating validity gives an average run length of 1, so the sampler routes it to the scalar loop as well.

That means the run-reader iteration, the `run.Pos` offset arithmetic, and the run-breaking at nulls — the actual novel logic of this PR — have no coverage. Mutation testing confirms the consequence. I broke the implementation four ways and measured which tests catch each:

| mutation | current tests |
|---|---|
| drop the per-value length-delta verification (bulk memcmp only) | detected |
| never break runs at nulls (memcmp spans null slots) | **not detected** |
| index the right side with the left's offset (wrong alignment) | **not detected** |
| remove the empty-validity-buffer guard | **not detected** |

The one that is caught is caught only incidentally, by the "same payload with different boundaries" case running on the all-valid single-run path. A wrong bulk compare in this PR's core path would ship green.

To reach the run path a test needs `Len() > 64`, a non-empty bitmap, and the first 8 validity runs averaging ≥ 4 valid values — validity like `(i/8)%2 == 0` over 256 elements works. Please add: equal arrays with long valid runs; **differing bytes under null slots**; a mismatch in a valid slot in a later run; left and right sliced to different physical offsets with nulls present; and the same concatenated bytes with a shifted boundary inside a run.

One caveat on the null case: the existing "null payload ignored" test can't catch this even in principle, because the builders write a zero-length value for every null slot, so both sides are empty there. It needs raw `array.NewData` construction to place real differing bytes under a null.

### 2. `TestBinaryEqualityWithDeclaredNullsWithoutBitmap` is vacuous

It passes with `|| len(values.NullBitmapBytes()) == 0` deleted entirely. Because `||` short-circuits and the array is 8 elements, `Len() <= 64` returns first and the guard is never evaluated.

The guard is load-bearing — with it removed, a 256-element array with the same malformation panics with `slice bounds out of range [:8] with capacity 0`, and restoring it fixes that — but nothing currently pins it, so tuning or removing the length cutoff later would silently reintroduce the panic. Please raise the array above 64 elements, e.g. 128.

### Non-blocking

- A truncated (non-empty but too short) validity bitmap still panics on the run path. The merge base panics identically on the same input, so this is pre-existing rather than something you introduced — noting it only so it isn't mistaken for new breakage.
- The `if length == 1 { return true }` early return in `arrayEqualVariableWidthRun` is redundant; the delta loop performs the same check.
- I'd like the short-array benchmark numbers re-confirmed on an idle machine before we call that fully settled. My measurements were taken on a heavily loaded box and I don't trust them either way, so this isn't a finding — just something I'd rather see clean numbers for, particularly around the `Len() <= 64` cutoff boundary.

arrow/array/binary.go
```diff
@@ -432,17 +439,120 @@ func (a *LargeBinary) ValidateFull() error {
...
+}
+
+func arrayEqualVariableWidth[T binaryOffset, V ~[]byte | ~string](
+	leftOffsets, rightOffsets []T,
+	leftValues, rightValues V,
+	leftOffset, rightOffset, length, nulls int,
+	validity []byte,
+	equalValues func(V, V) bool,
+) bool {
+	if nulls == 0 {
```
zeroshade: The `nulls == 0` fast path compares every payload byte even when a validity bitmap contains nulls. With inconsistent `ArrayData`—a bitmap containing nulls but declared `null_count=0`—payload differences in null slots now make arrays unequal, unlike the prior scalar implementation. Please only use the whole-payload path when no validity bitmap exists, or otherwise process bitmap runs, and add a regression test.

zeroshade (changes requested):
The findings from the previous reviews are fixed at the current head: empty validity bitmaps with declared nulls no longer take the run path, short fragmented arrays avoid sampling, and the requested long-run and sliced-offset coverage is present.

One malformed-but-constructible ArrayData case still changes equality semantics relative to the prior scalar implementation: when the declared null count is zero but the validity bitmap contains nulls, the new fast path compares payloads in null slots. Details are inline.

Array tests, race tests, vet, and randomized differential tests over valid arrays otherwise pass. CI is still running.

---

>

zeroshade (approved):

PR #1175, opened 2026-08-11
perf(parquet): encode spaced booleans without compaction

fallintoplace:
### Rationale for this change

Plain Boolean spaced encoding currently counts valid bits, allocates a compacted bitmap, copies valid runs into it, and then copies that bitmap into the encoder buffer.

Short unaligned runs also create bitmap readers and writers for every copy. This is expensive for common nullable patterns.

- Append valid bitmap runs directly to the existing encoder buffer.
- Accumulate the valid count while encoding instead of scanning validity first.
- Write short runs directly to avoid unaligned bitmap-copy allocations.
- Keep bulk bitmap copies for longer runs.
- Share bitmap writer initialization between the Boolean input paths.
- Add benchmarks for 1K, 64K, and 1M values across several null patterns.
- Add tests for bitmap offsets, validity offsets, encoder-buffer boundaries, and repeated calls.

Apple M1 Pro results for 65,536 values with `-cpu=1`:

| Validity pattern | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| All valid | 6.77 us | 3.00 us | 2.26x |
| 1% null | 224 us | 182 us | 1.23x |
| 10% null | 912 us | 307 us | 2.97x |
| Alternating 50% null | 3.86 ms | 394 us | 9.79x |
| 90% null | 1.16 ms | 95.8 us | 12.1x |
| Clustered nulls | 7.53 us | 3.65 us | 2.06x |

For the short-run patterns:

- 10% null: 847 KB and 13,109 allocations to 23 B and 0 allocations.
- 50% null: 4.20 MB and 65,542 allocations to 14 B and 0 allocations.
- 90% null: 735 KB and 11,469 allocations to 0 B and 0 allocations.

### Are these changes tested?

Yes.

- `go test ./parquet/...`
- `go test -race ./parquet/internal/encoding`
- `go vet -composites=false ./parquet/internal/encoding`
- Cross-compiled the encoding tests for linux/amd64 and linux/s390x.

No.

### DIFF
--- a/parquet/internal/encoding/boolean_encoder.go
+++ b/parquet/internal/encoding/boolean_encoder.go
@@ -133,12 +133,27 @@ func (enc *PlainBooleanEncoder) PutSpacedBitmap(bitmap []byte, bitmapOffset int6
 		if run.Length == 0 {
 			break
 		}
-		// Bitmap copies need word readers for unaligned ranges. For short runs,
-		// writing the bits directly avoids that setup and its allocations.
-		if run.Length < scalarBitmapRunLimit {
-			enc.putBitmapScalar(bitmap, bitmapOffset+run.Pos, run.Length)
-		} else {
-			enc.PutBitmap(bitmap, bitmapOffset+run.Pos, run.Length)
+		// Bitmap copies need word readers for unaligned ranges. Keep the
+		// scalar path for short runs unless the run is an exact byte multiple:
+		// when both ranges are byte-aligned, PutBitmap uses a cheaper byte copy.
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Streaming the set-bit runs directly into the encoder is the right approach, and the alignment-aware dispatch (scalar for short/unaligned runs, byte-copy only when both ends are aligned) is carefully done. The buffer-boundary tests at 8191/8192/8193 hit exactly the risky cases. Impressive allocation reduction.

---

PR #1176, opened 2026-08-11
perf(parquet): decode levels directly into int16 batches

fallintoplace:
### What changed

- decode RLE and bit-packed levels directly into `[]int16`
- count max definition levels during decoding
- remove the intermediate `uint64` level batch and its second conversion pass
- add benchmarks for all-defined, mostly-defined, and alternating levels
- check partial value counts when a literal run is truncated

### Benchmarks

Apple M1 Pro, `-benchtime=300ms -count=5`:

| RLE workload | Before | After | Change |
| --- | ---: | ---: | ---: |
| 1,024 all defined | 2,137 ns/op | 1,495 ns/op | -30% |
| 1,024 mostly defined | 5,740 ns/op | 5,329 ns/op | -7% |
| 65,536 all defined | 61,213 ns/op | 24,601 ns/op | -60% |
| 65,536 alternating | 120,994 ns/op | 89,543 ns/op | -26% |

Allocations were unchanged.

### DIFF
--- a/parquet/internal/encoding/levels.go
+++ b/parquet/internal/encoding/levels.go
@@ -257,10 +257,10 @@ func (l *LevelDecoder) SetDataV2(nbytes int32, maxLvl int16, nbuffered int, data
 // values exist to be read, along with any decoding error.
 func (l *LevelDecoder) Decode(levels []int16) (int, int64, error) {
 	var (
-		buf          [1024]uint64
 		totaldecoded int
 		decoded      int
 		valsToRead   int64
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM. The direct `int16` level-decoding path preserves RLE and bit-packed semantics across partial reads, maximum levels, split decoding, and malformed inputs. Package tests, race tests, the full Parquet suite, randomized differential probes, benchmarks, and CI all pass.

---

> will follow up.*
>

PR #1177, opened 2026-08-11
perf(parquet): encode RLE levels in batches

fallintoplace:
### What changed

- add a batch path for repetition and definition level encoding
- scan repeated runs once instead of calling the scalar encoder for every level
- keep the scalar path around literal boundaries
- preserve the existing encoded bytes across split streaming calls
- add level encoder and scalar-versus-batch benchmarks

### Benchmarks

Apple M1 Pro:

| RLE workload | Before | After | Change |
| --- | ---: | ---: | ---: |
| 65,536 all defined | 151,051 ns/op | 31,143 ns/op | -79% |
| 65,536 mostly defined | 324,228 ns/op | 252,108 ns/op | -22% |
| 65,536 alternating | 339,128 ns/op | 341,149 ns/op | +0.6% |

The alternating case is the fallback path. Allocations remain at zero.

### DIFF
--- a/parquet/internal/encoding/levels.go
+++ b/parquet/internal/encoding/levels.go
@@ -98,12 +98,7 @@ func (l *LevelEncoder) EncodeNoFlush(lvls []int16) (nencoded int, err error) {
 
 	switch l.encoding {
 	case format.Encoding_RLE:
-		for _, level := range lvls {
-			if err = l.rle.Put(uint64(level)); err != nil {
-				return
-			}
-			nencoded++
-		}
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
The batched RLE path preserves scalar encoding semantics across literal/repeated boundaries and split streaming calls, with appropriate maximum-run and error-path coverage. The Parquet suite, race tests, randomized byte-for-byte comparisons, and 386 compilation passed locally. Benchmarks confirm substantial gains for repeated-level workloads without allocations.

PR #1179, opened 2026-08-12
perf(parquet): batch DELTA_BYTE_ARRAY encoding

fallintoplace:
## Summary

- Batch DELTA_BYTE_ARRAY prefix lengths and suffixes in chunks of 256 values.
- Reuse fixed scratch space for DELTA_LENGTH_BYTE_ARRAY lengths.
- Preserve the previous value across batch boundaries and separate Put calls.
- Add a boundary round-trip test and representative benchmarks.

## Why

DELTA_BYTE_ARRAY was calling the prefix and suffix encoders once per value. Large pages therefore paid for many small calls and temporary length slices.

Apache Arrow C++ already uses fixed-size batching for this encoder. This change follows the same shape while keeping the existing Go encoding format and last-value behavior.

## Correctness

- The encoded layout is unchanged: prefix lengths first, followed by the suffix stream.
- The first value still has a zero prefix.
- Empty suffixes still work as before.
- The new test compares one large Put call with Put calls split around a batch boundary, then decodes and compares every value.

## Benchmark

64K values on an Apple M1 Pro, 1 second per sample, 3 samples:

| --- | ---: | ---: | ---: |
| prefix-heavy | 2.09 ms/op | 1.75 ms/op | ~16% faster |
| low-prefix | 1.68 ms/op | 1.47 ms/op | ~12% faster |

Allocation counts stayed the same in both cases.

## Checks

- `PARQUET_TEST_DATA=<parquet-testing-data> go test ./parquet/...`
- `go test -race ./parquet/internal/encoding -count=1`
- `git diff --check`

No public API changes.

### DIFF
--- a/parquet/internal/encoding/delta_byte_array.go
+++ b/parquet/internal/encoding/delta_byte_array.go
@@ -43,6 +43,16 @@ type DeltaByteArrayEncoder struct {
 
 const deltaByteArrayBatchSize = 256
 
+func commonPrefixLength(left, right parquet.ByteArray) int {
+	maximum := min(left.Len(), right.Len())
+	for i := 0; i < maximum; i++ {
+		if left[i] != right[i] {
+			return i
+		}
+	}
+	return maximum
+}
+
 func (enc *DeltaByteArrayEncoder) EstimatedDataEncodedSize() int64 {
 	prefixEstimatedSize := int64(0)
 	if enc.prefixEncoder != nil {
@@ -83,22 +93,13 @@ func (enc *DeltaByteArrayEncoder) Put(in []parquet.ByteArray) {
 	lastVal := enc.lastVal
 	var prefixLengths [deltaByteArrayBatchSize]int32
 	var suffixes [deltaByteArrayBatchSize]parquet.ByteArray
-	for i := 0; i < len(in); i += deltaByteArrayBatchSize {
-		batchSize := min(deltaByteArrayBatchSize, len(in)-i)
-		for j := 0; j < batchSize; j++ {
-			val := in[i+j]
-			commonPrefixLength := 0
-			maximumCommonPrefixLength := min(lastVal.Len(), val.Len())
-			for commonPrefixLength < maximumCommonPrefixLength {
-				if lastVal[commonPrefixLength] != val[commonPrefixLength] {
-					break
-				}
-				commonPrefixLength++
-			}
-
+	for offset := 0; offset < len(in); offset += deltaByteArrayBatchSize {
+		batchSize := min(deltaByteArrayBatchSize, len(in)-offset)
+		for i, val := range in[offset : offset+batchSize] {
+			prefixLength := commonPrefixLength(lastVal, val)
 			lastVal = val
-			prefixLengths[j] = int32(commonPrefixLength)
-			suffixes[j] = val[commonPrefixLength:]
+			prefixLengths[i] = int32(prefixLength)
+			suffixes[i] = val[prefixLength:]
 		}
 		enc.suffixEncoder.Put(suffixes[:batchSize])
 		enc.prefixEncoder.Put(prefixLengths[:batchSize])
--- a/parquet/internal/encoding/delta_byte_array.go
+++ b/parquet/internal/encoding/delta_byte_array.go
@@ -38,6 +38,9 @@ type DeltaByteArrayEncoder struct {
 	prefixEncoder *DeltaBitPackInt32Encoder
 	suffixEncoder *DeltaLengthByteArrayEncoder
 
+	prefixLengths [deltaByteArrayBatchSize]int32
+	suffixes      [deltaByteArrayBatchSize]parquet.ByteArray
+
 	lastVal parquet.ByteArray
 }
 
@@ -70,8 +73,8 @@ func (enc *DeltaByteArrayEncoder) initEncoders() {
 		encoder: newEncoderBase(enc.encoding, nil, enc.mem),
 	}
 	enc.suffixEncoder = &DeltaLengthByteArrayEncoder{
-		newEncoderBase(enc.encoding, nil, enc.mem),
-		&DeltaBitPackInt32Encoder{
+		encoder: newEncoderBase(enc.encoding, nil, enc.mem),
+		lengthEncoder: &DeltaBitPackInt32Encoder{
 			encoder: newEncoderBase(enc.encoding, nil, enc.mem),
 		},
 	}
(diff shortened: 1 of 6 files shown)


parquet/internal/encoding/delta_byte_array.go
```diff
@@ -74,39 +86,30 @@ func (enc *DeltaByteArrayEncoder) Put(in []parquet.ByteArray) {
...
-		l1 := enc.lastVal.Len()
-		l2 := val.Len()
-		j := 0
-		for j < l1 && j < l2 {
-			if enc.lastVal[j] != val[j] {
-				break
-			}
-			j++
+	lastVal := enc.lastVal
+	var prefixLengths [deltaByteArrayBatchSize]int32
```
zeroshade: **Blocking:** these fixed arrays cause roughly 7 KiB of stack memory to be zeroed on every `Put` call, regardless of how many values the call contains. The nested `DeltaLengthByteArrayEncoder.Put` adds another 1 KiB scratch array.

Comparing the actual merge base and this PR on an Apple M4 while encoding 64K prefix-heavy values:

- 1 value/`Put`: 2.72 ms → 9.57 ms, approximately **3.5× slower**
- 8 values/`Put`: 1.80 ms → 2.16 ms, approximately **20% slower**
- 256 values/`Put`: 1.57 ms → 1.26 ms, approximately **20% faster**

The crossover is around 32 values per call. Default writer batches benefit, but callers using small `WriteBatch` or encoder `Put` calls encounter a large performance cliff.

Please keep the scratch storage on the encoder and allocate it once rather than zeroing full stack arrays per call. The suffix scratch must be cleared after each used batch so it does not retain caller-owned buffers.

parquet/internal/encoding/delta_byte_array_benchmark_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+)
+
+func encodeDeltaByteArrayUnbatched(values []parquet.ByteArray) (Buffer, error) {
```
zeroshade: This hand-written “before” implementation is not the actual merge-base implementation: it links against this PR’s modified `DeltaLengthByteArrayEncoder.Put`. On the current head it reports approximately 3.45 ms → 1.28 ms, a roughly 63% improvement, which does not correspond to the PR description’s actual revision-to-revision result.

It also benchmarks only one large `Put`, so it cannot detect the small-call regression introduced by the fixed scratch arrays.

Please benchmark the production encoder with several `Put` chunk sizes—at least 1, 8, 32, 256, and all values—and use revision-to-revision `benchstat` results rather than retaining a duplicate historical implementation in the benchmark.

parquet/internal/encoding/delta_length_byte_array.go
```diff
@@ -42,14 +42,16 @@ type DeltaLengthByteArrayEncoder struct {
 
 // Put writes the provided slice of byte arrays to the encoder
 func (enc *DeltaLengthByteArrayEncoder) Put(in []parquet.ByteArray) {
-	lengths := make([]int32, len(in))
-	totalLen := int(0)
-	for idx, val := range in {
-		lengths[idx] = int32(val.Len())
-		totalLen += val.Len()
+	var lengths [deltaByteArrayBatchSize]int32
```
zeroshade: Please add a targeted standalone `DELTA_LENGTH_BYTE_ARRAY` test around 255/256/257 and 511/512/513 values, comparing one large `Put` with split `Put` calls and round-tripping the result.

The new DELTA_BYTE_ARRAY test does not exercise this loop’s multi-batch path because the outer encoder invokes the suffix encoder with at most 256 values at a time.

zeroshade (changes requested):
The encoded output remains compatible and I found no decoding or state bug introduced by the batching logic. The complete encoding tests, race tests, vet, and diff checks pass.

The fixed stack scratch arrays do, however, cause a substantial performance regression for small `Put` calls: approximately 3.5x slower for one value per call and 20% slower for eight values per call in direct merge-base versus PR measurements. The submitted benchmark exercises only one large call and cannot detect this regression. I left inline requests to reuse scratch storage, revise the benchmark, and add targeted standalone DELTA_LENGTH boundary coverage.

Separately, while tracing the encoder lifecycle I found a pre-existing multi-page bug: `DeltaByteArrayEncoder.FlushValues` does not reset `lastVal`, although the column writer reuses the encoder for subsequent pages. The next page can therefore start with a nonzero prefix and fail decoding with `parquet: first delta byte array prefix length must be zero`. This reproduces on the merge base, so it should not block this PR, but it warrants a separate issue/fix.

---

>

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -74,39 +89,29 @@ func (enc *DeltaByteArrayEncoder) Put(in []parquet.ByteArray) {
...
+		enc.prefixEncoder.Put(enc.prefixLengths[:batchSize])
+		clear(enc.suffixes[:batchSize])
 	}
 
 	// do the memcpy after the loops to keep a copy of the lastVal
 	// we do a copy here so that we only copy and keep a reference
 	// to the suffix, and aren't forcing the *entire* value to stay
 	// in memory while we have this reference to just the suffix.
-	enc.lastVal = append([]byte{}, enc.lastVal...)
+	enc.lastVal = append([]byte{}, lastVal...)
```
zeroshade: The original multi-fold performance cliff is fixed, but the one-value `Put` case remains approximately 12–15% slower than the merge base:

- merge base: approximately 2.72 ms
- current head: approximately 3.12 ms

The remaining dominant cost is this allocation and copy on every `Put`. Reusing the existing backing buffer is safe within a page:

```go
enc.lastVal = append(enc.lastVal[:0], lastVal...)
```

`FlushValues` can continue setting it to `nil` to reset page state and release the retained value. An independent probe found this change makes the one-value case faster than the merge base while preserving byte-identical output.

Since this is a performance PR and small `Put` calls were the original blocking concern, please eliminate the remaining regression and update the benchmark results.

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -38,9 +38,24 @@ type DeltaByteArrayEncoder struct {
 	prefixEncoder *DeltaBitPackInt32Encoder
 	suffixEncoder *DeltaLengthByteArrayEncoder
 
+	prefixLengths [deltaByteArrayBatchSize]int32
```
zeroshade, line 41: Moving these arrays onto the encoder makes `DeltaByteArrayEncoder` roughly 7 KiB, but its `Type` method still has a value receiver:

```go
func (DeltaByteArrayEncoder) Type() parquet.Type
```

Calling it through the encoder interface now copies the entire struct. `DeltaLengthByteArrayEncoder` has the same issue with its new roughly 1 KiB scratch field.

Both constructors return pointers, so please switch both `Type` methods to pointer receivers.

zeroshade (changes requested):
The updated implementation addresses most of the prior review:

- the 7 KiB per-`Put` stack clearing is gone;
- suffix scratch references are cleared;
- the synthetic baseline benchmark was replaced with a useful `Put`-size matrix;
- standalone DELTA_LENGTH boundary/split coverage was added;
- output remains byte-compatible;
- the additional `lastVal` page reset fixes the previously identified multi-page corruption.

The encoding and race suites, vet, and diff checks pass. I left two remaining performance comments inline: the one-value `Put` case is still approximately 12–15% slower than the merge base, and the newly enlarged encoder structs retain value-receiver `Type` methods that copy the complete scratch storage.

---

>

zeroshade (approved):
Assuming CI is all green, this LGTM

PR #1180, opened 2026-08-12
perf(arrow/array): bulk append fixed-width empty values

fallintoplace:
## What changed

`AppendEmptyValues` used to call `AppendEmptyValue` in a loop. For larger batches, each value repeated the reserve, bitmap, and storage work.

This change makes fixed-width empty appends bulk operations:

- reserve space once
- zero the appended value range in one operation
- set validity bits in bulk
- clear boolean value bits separately from the validity bitmap
- keep the existing single-value path for count 1

The numeric generated source and template are updated together. The covered builders are numeric, temporal, interval, float16, decimal, fixed-size binary, and boolean.

## Correctness

- Empty values stay valid, including when the builder reuses storage.
- Reused numeric and boolean storage is explicitly tested.
- `bufferBuilder.Advance` now zeroes skipped bytes as documented.
- `BooleanBuilder.Resize` now preserves the requested logical length when truncating.
- Zero and negative counts remain no-ops.

## Benchmark

`AppendEmptyValues(1024)` on an Apple M1 Pro. Three runs were used for each result.

| Builder | Before | After | Allocs |
| --- | ---: | ---: | ---: |
| Int32 | 5.15 us | 1.50 us | 11 -> 5 |
| Int64 | 6.44 us | 2.52 us | 11 -> 5 |
| Timestamp | 6.50 us | 3.22 us | 12 -> 6 |
| Boolean | 4.04 us | 0.30 us | 7 -> 5 |
| Fixed-size binary | 10.1 us | 5.69 us | 16 -> 7 |

Small counts keep the single-value path, so the biggest improvement is for larger batches.

## Tests

- `go test ./arrow/array`
- `go vet ./arrow/array`
- Added focused correctness tests and representative benchmarks

### DIFF
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -90,9 +90,16 @@ func (b *BooleanBuilder) AppendEmptyValue() {
 }
 
 func (b *BooleanBuilder) AppendEmptyValues(n int) {
-	for i := 0; i < n; i++ {
-		b.AppendEmptyValue()
+	if n <= 0 {
+		return
 	}
+	if n == 1 {
+		b.AppendEmptyValue()
+		return
+	}
+	b.Reserve(n)
+	bitutil.SetBitsTo(b.rawData, int64(b.length), int64(n), false)
+	b.unsafeSetValid(n)
 }
 
 func (b *BooleanBuilder) AppendValueFromString(s string) error {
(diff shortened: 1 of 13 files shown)


zeroshade (approved):
The bulk path ordering is correct (`Reserve` → zero range → `unsafeSetValid`), and both drive-by fixes — `Advance` zeroing per its documented contract and the `BooleanBuilder.Resize` truncation length — are genuine correctness improvements with test coverage. Benchmarks look convincing.

May need a rebase now that #1185 touched `builder.go`.

---

PR #1182, opened 2026-08-12
perf(parquet/pqarrow): decode booleans directly to bitmaps

fallintoplace:
## What does this PR do?

- decodes Parquet BOOLEAN values directly into a packed bitmap
- supports both PLAIN and RLE boolean pages
- expands nullable values inside the bitmap without creating a `[]bool`
- passes the record reader value bitmap directly to the Arrow Boolean array

## Why?

The read path currently stores decoded booleans as one byte per value. It then allocates an Arrow bitmap and packs the values again in `transferBool`.

This finishes wiring the direct bitmap decoder API from #707 into the pqarrow read path. The record reader now keeps boolean values packed from decoding through Arrow array construction.

The change also fixes decoder state after a partial PLAIN bitmap decode, so the next call resumes at the correct source bit.

## Benchmarks

Apple M1 Pro, 1M values, uncompressed Parquet, `GOMAXPROCS=1`:

| case | before | after | change | before B/op | after B/op | change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| dense | 1.48 ms | 0.15 ms | -90% | 1.61 MB | 0.56 MB | -65% |
| 50% null | 6.72 ms | 4.63 ms | -31% | 4.11 MB | 3.08 MB | -25% |

```text
GOMAXPROCS=1 go test ./parquet/pqarrow -run '^$' -bench '^BenchmarkBooleanBitmapRead$' -benchmem -benchtime=500ms -count=5
```

## Tests

- added direct bitmap tests for consecutive PLAIN decode calls
- added nullable bitmap tests for PLAIN and RLE with unaligned offsets
- ran `go test ./parquet/...`
- ran race tests for the touched encoding, file, and pqarrow packages
- ran `go vet` for the touched packages

### DIFF
--- a/parquet/file/record_reader.go
+++ b/parquet/file/record_reader.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
-	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
@@ -229,6 +228,9 @@ func (pr *primitiveRecordReader) ResetValues() {
 }
 
 func (pr *primitiveRecordReader) numBytesForValues(nitems int64) (num int64, err error) {
+	if pr.Descriptor().PhysicalType() == parquet.Types.Boolean {
+		return bitutil.BytesForBits(nitems), nil
+	}
 	typeSize := int64(pr.Descriptor().PhysicalType().ByteSize())
 	var ok bool
 	if num, ok = utils.Mul64(nitems, typeSize); !ok {
@@ -264,9 +266,8 @@ func (pr *primitiveRecordReader) ReserveValues(extra int64, hasNullable bool) er
 func (pr *primitiveRecordReader) ReadValuesDense(toRead int64) (err error) {
 	switch cr := pr.ColumnChunkReader.(type) {
 	case *BooleanColumnChunkReader:
-		data := pr.values.Bytes()[int(pr.valuesWritten):]
-		values := *(*[]bool)(unsafe.Pointer(&data))
-		_, err = cr.curDecoder.(encoding.BooleanDecoder).Decode(values[:toRead])
+		_, err = cr.curDecoder.(encoding.BooleanBitmapDecoder).DecodeToBitmap(
+			pr.values.Bytes(), pr.valuesWritten, int(toRead))
 	case *Int32ColumnChunkReader:
 		values := arrow.Int32Traits.CastFromBytes(pr.values.Bytes())[int(pr.valuesWritten):]
 		_, err = cr.curDecoder.(encoding.Int32Decoder).Decode(values[:toRead])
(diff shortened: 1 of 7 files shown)


parquet/file/record_reader.go
```diff
@@ -229,6 +228,9 @@ func (pr *primitiveRecordReader) ResetValues() {
 }
 
 func (pr *primitiveRecordReader) numBytesForValues(nitems int64) (num int64, err error) {
+	if pr.Descriptor().PhysicalType() == parquet.Types.Boolean {
```
zeroshade, line 296: **Blocking:** this silently changes the representation returned by the exported `file.RecordReader.Values()` and `ReleaseValues()` APIs for Boolean columns.

Previously, Boolean values used one byte per value, consistent with `Boolean.ByteSize() == 1`. After this change they are LSB-first packed bits. External code continues compiling but either reads garbage or indexes beyond the shorter buffer. `file.NewRecordReader` is exported, so this is observable outside pqarrow.

Please preserve the existing methods’ representation and expose the packed path through a Boolean-specific interface such as `ReleaseValueBitmap()`, which `pqarrow.transferBool` can detect and consume zero-copy. If this behavior break is intentionally accepted instead, it at least needs explicit API documentation and a release-note callout, but preserving compatibility would be preferable.

parquet/internal/encoding/boolean_decoder.go
```diff
@@ -205,6 +206,39 @@ func (dec *PlainBooleanDecoder) DecodeSpaced(out []bool, nullCount int, validBit
...
+		return valuesRead, errors.New("parquet: boolean decoder: number of values / definition levels read did not match")
+	}
+
+	// Expand the packed physical values backwards into their logical positions.
+	// Copying backwards keeps unread compact values from being overwritten.
+	physicalIndex := valuesToRead - 1
+	for logicalIndex := length - 1; logicalIndex >= 0 && physicalIndex >= 0; logicalIndex-- {
+		if bitutil.BitIsSet(validBits, int(validBitsOffset)+logicalIndex) {
+			value := bitutil.BitIsSet(out, int(outOffset)+physicalIndex)
+			bitutil.SetBitTo(out, int(outOffset)+logicalIndex, value)
```
zeroshade: This per-bit backwards expansion regresses sparsely-null Boolean columns despite the large dense/50%-null wins.

On an Apple M4 with `GOMAXPROCS=1`, the PR’s existing 1M-row, 10%-null benchmark was consistently approximately 4.86 ms versus 4.64 ms on the merge base—about 5% slower. An independent 1%-null run measured approximately 3.28 ms versus 2.62 ms, about 25% slower. Allocated bytes still improve, but sparse nulls are a common case and the PR description currently reports only dense and 50%-null results.

Please add 1% and 5% null-density cases and either avoid the regression or explicitly document and justify the latency-versus-memory trade-off.

parquet/pqarrow/boolean_bitmap_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/compress"
+	"github.com/apache/arrow-go/v18/parquet/file"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBooleanBitmapReadAcrossPages(t *testing.T) {
```
zeroshade, line 34: The cross-page test exercises the PLAIN bitmap decoder only. `RleBooleanDecoder.DecodeToBitmap` is also new, but the existing RLE file-reader test decodes through `[]bool` rather than through this pqarrow bitmap-transfer path.

Please parameterize this regression test over PLAIN and RLE encoding so both implementations are covered end to end across page boundaries.

zeroshade (changes requested):
The direct bitmap decoder is correct across source and destination alignments, consecutive partial calls, nullable expansion, PLAIN and RLE decoder state, page and row-group boundaries, seeking, and nested columns. Focused tests, race tests, and vet pass locally; CI is fully green.

I found two issues that need addressing:

1. The exported `file.RecordReader.Values()` and `ReleaseValues()` methods silently change Boolean representation from one byte per value to packed bits.
2. The per-bit nullable expansion causes a measurable 5–25% latency regression for sparse-null columns, a case not represented in the reported benchmark results.

I also left a test-coverage request for the new RLE bitmap path.

---

>

zeroshade (approved):

PR #1187, opened 2026-08-12
perf(arrow/array): bulk append list nulls and empty values

fallintoplace:
## What

- reserve List and ListView batches once
- update parent validity in bulk
- append repeated offsets and zero sizes through the unsafe builder path
- cover List, LargeList, ListView, and LargeListView

These bulk APIs currently call the scalar append path once per list slot. Empty and null variable-sized lists do not append child values, so only the parent validity and dimensions need updating.

This follows the batching shape used by the C++ list builders:

https://github.com/apache/arrow/blob/485499fd02ea2b0c323d67871fbe96aae4232504/cpp/src/arrow/array/builder_nested.h#L138-L160

For ListView, this keeps Go's current physical representation by repeating the current child offset with a zero size.

## Benchmark

Apple M1 Pro, Go 1.26.3. Representative medians from three to five runs with 65,536 slots:

| Builder | Operation | Before | After | Change |
|---|---|---:|---:|---:|
| List | nulls | 848 us | 201 us | -76% |
| List | empty | 848 us | 197 us | -77% |
| LargeList | nulls | 1.08 ms | 241 us | -78% |
| LargeList | empty | 1.05 ms | 245 us | -77% |
| ListView | nulls | 1.35 ms | 414 us | -69% |
| ListView | empty | 1.24 ms | 415 us | -67% |
| LargeListView | nulls | 1.47 ms | 509 us | -65% |
| LargeListView | empty | 1.50 ms | 505 us | -66% |

Regular-list allocations drop from 45 to 18 per batch. List-view allocations drop from 68 to 25. The `n=1` cases stay within benchmark noise.

## Tests

- `go test -race ./arrow/array -run 'TestListBuilderBulkAppendNullsAndEmptyValues|TestList|TestLargeList' -count=1`
- `go vet ./arrow/array`
- `go test -p 1 ./...`


arrow/array/list.go
```diff
@@ -436,19 +455,29 @@ func (b *baseListBuilder) AppendNull() {
 }
 
 func (b *baseListBuilder) AppendNulls(n int) {
```
zeroshade, line 457: `MapBuilder` wraps a `ListBuilder`, but its `AppendNulls` and `AppendEmptyValues` methods still call the scalar path in a loop.

After one `adjustStructBuilderLen()` call, they should be able to delegate directly to these new bulk methods:

```go
b.adjustStructBuilderLen()
b.listBuilder.AppendNulls(n)
```

and similarly for empty values. That would extend this optimization to map arrays without duplicating the dimension logic.

arrow/array/list_builder_bulk_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}},
+	{"list_view", true, func(mem memory.Allocator) array.VarLenListLikeBuilder {
+		return array.NewListViewBuilder(mem, arrow.PrimitiveTypes.Int32)
+	}},
+	{"large_list_view", true, func(mem memory.Allocator) array.VarLenListLikeBuilder {
+		return array.NewLargeListViewBuilder(mem, arrow.PrimitiveTypes.Int32)
+	}},
+}
+
+func TestListBuilderBulkAppendNullsAndEmptyValues(t *testing.T) {
```
zeroshade, line 52: Because the optimized path writes numeric builder storage and validity bits directly, a small bulk-versus-scalar parity matrix would provide stronger regression protection than one fixed sequence.

Please vary the starting parent length across bitmap boundaries, the batch size, and reuse after `NewArray()` to cover stale bits, non-byte-aligned writes, and dimension-builder length synchronization.

zeroshade (changes requested):
The bulk implementation preserves the scalar behavior for validity, offsets, sizes, child values, zero/negative counts, and builder reuse. Reserving the parent also reserves the offset/size builders before the direct writes, and final trailing-offset handling remains correct.

The array suite, targeted race tests, vet, and representative benchmarks pass locally. Independent differential probes across all four builder variants and thousands of mixed operation sequences found no divergence from element-wise appends.

I am requesting changes for the two follow-ups noted inline: extend the optimization to `MapBuilder` through delegation, and strengthen the tests with scalar-parity cases across bitmap boundaries. GitHub currently reports no CI checks for this head.

---

>

zeroshade (approved):

PR #1188, opened 2026-08-12
perf(arrow/array): optimize bulk appends for binary builders

fallintoplace:
## Summary

- reserve the full batch once in `BinaryBuilder.AppendNulls` and `AppendEmptyValues`
- fill the repeated current offsets directly for both 32-bit and 64-bit offset buffers
- update the validity bitmap and counters once per batch
- add coverage and benchmarks for Binary, LargeBinary, String, and LargeString

This follows the same batching shape used by [Arrow C++ binary builders](https://github.com/apache/arrow/blob/485499fd02ea2b0c323d67871fbe96aae4232504/cpp/src/arrow/array/builder_binary.h#L100-L140).

## Benchmark

65,536 appended values, representative medians on darwin/arm64:

| Builder | Operation | Before | After | Change |
| Binary | nulls | 403 us | 93 us | -77% |
| LargeBinary | nulls | 475 us | 158 us | -67% |
| String | nulls | 402 us | 85 us | -79% |
| String | empty | 458 us | 84 us | -82% |
| LargeString | nulls | 532 us | 159 us | -70% |
| LargeString | empty | 517 us | 154 us | -70% |

Allocations per build drop from 27 to 8 for 32-bit offsets and from 28 to 8 for 64-bit offsets. Single-value performance stays about the same.

## Tests

- `go test -race ./arrow/array -run TestBinaryBuilderBulkAppendNullsAndEmptyValues -count=1`
- `go vet ./arrow/array`

### DIFF
--- a/arrow/array/binarybuilder.go
+++ b/arrow/array/binarybuilder.go
@@ -25,6 +25,7 @@ import (
 	"unsafe"
 
 	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/bitutil"
 	"github.com/apache/arrow-go/v18/arrow/internal/debug"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/apache/arrow-go/v18/internal/json"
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified that the batched offset and validity updates preserve scalar append behavior across Binary, LargeBinary, String, and LargeString, including bitmap boundaries and builder reuse. The full array suite, targeted race tests, vet, formatting, diff checks, and benchmark all pass.

PR #1191, opened 2026-08-12
perf(arrow/ipc): slice already-zero-based offset buffers

fallintoplace:
### Rationale for this change

IPC currently allocates a new offsets buffer whenever a variable-length array has unused trailing offsets. For a prefix slice, the first logical offset is already zero, so the loop only copies every offset and subtracts zero.

Arrow C++ handles this case by slicing the existing offsets buffer.

- Read the first logical offset instead of treating every sliced array as needing rebasing.
- Rebase only when that offset is non-zero.
- Use `memory.SliceBuffer` when the required offsets are already zero-based.
- Add coverage for int32 and int64 offsets, middle slices, and slices with a non-zero array offset whose first logical value offset is zero.

The benchmark uses a 524,288-value slice from a 1,048,576-value backing array. Medians from 6 runs on an Apple M1 Pro were:

| prefix slice | main | this PR |
|---|---:|---:|
| int32 time/op | 238 us | 40.7 ns |
| int32 B/op | 2,106,000 | 80 |
| int64 time/op | 283 us | 40.7 ns |
| int64 B/op | 4,203,153 | 80 |
| allocs/op | 8 | 1 |

Unsliced buffers remain allocation-free. Middle slices that need rebasing are unchanged.

### Are these changes tested?

- `go test ./arrow/ipc -count=1`
- `go test ./arrow/... -count=1`
- `go test ./arrow/ipc -run "^$" -bench "^BenchmarkGetZeroBasedValueOffsets$" -benchmem -count=6`

No. The IPC output and public API are unchanged.

### DIFF
--- a/arrow/ipc/writer.go
+++ b/arrow/ipc/writer.go
@@ -922,29 +922,14 @@ func (w *recordEncoder) getZeroBasedValueOffsets(arr arrow.Array) *memory.Buffer
 
 	dataTypeWidth := arr.DataType().Layout().Buffers[1].ByteWidth
 
-	// if we have a non-zero offset, then the value offsets do not start at
-	// zero. we must a) create a new offsets array with shifted offsets and
-	// b) slice the values array accordingly
-	hasNonZeroOffset := data.Offset() != 0
-
-	// or if there are more value offsets than values (the array has been sliced)
-	// we need to trim off the trailing offsets
-	hasMoreOffsetsThanValues := offsetBytesNeeded < voffsets.Len()
-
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
The zero-based prefix-slice fast path preserves offset and buffer-lifetime semantics while avoiding the unnecessary rebase allocation. The added int32/int64 coverage exercises prefix, middle, and empty-prefix cases. IPC tests, race tests, vet, and benchmarks passed locally.

PR #1197, opened 2026-08-12
perf(compute): batch contiguous filter take indices

fallintoplace:
## Summary

* Add a range writer for uint16 and uint32 take indices.
* Use it when a filter block or bitmap run is fully selected.
* Keep the scalar path for mixed blocks.
* Add benchmarks for dense, clustered, alternating, short, and nullable filters.

## Benchmark

On an Apple M1 Pro, the 1M-row all-selected case improved from about 3.0 ms to 0.5 ms locally. Fragmented filters stayed close to the baseline.

## Tests

* `go test ./arrow/compute/internal/kernels ./arrow/compute`
* `go test -race ./arrow/compute/internal/kernels ./arrow/compute`
* Full package test run passed with the local Parquet test data checkout, excluding the CSV example that needs a separate Arrow CSV fixture.

### DIFF
--- a/arrow/compute/internal/kernels/vector_selection.go
+++ b/arrow/compute/internal/kernels/vector_selection.go
@@ -99,6 +99,19 @@ type builder[T any] interface {
 	UnsafeAppendBoolToBitmap(bool)
 }
 
+func unsafeAppendRange[T arrow.IntType | arrow.UintType](b *bufferBuilder[T], start T, n int) {
+	if n == 0 {
+		return
+	}
+
+	b.reserve(n)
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified that batched range writes preserve take-index ordering, sliced-filter offsets, and DropNulls/EmitNulls semantics across bitmap boundaries. Kernel and compute tests, race tests, vet, formatting, randomized probes, and diff checks pass. Benchmarks confirm substantial dense-filter gains without fragmented-filter regressions. GitHub currently reports no CI checks.

---

>

PR #1199, opened 2026-08-12
perf(arrow/compute): take chunked binary values without concatenation

fallintoplace:
## What

Chunked string and binary `take` currently concatenates all value chunks before selecting rows.

This adds a chunk-aware path for:

- STRING
- BINARY
- LARGE_STRING
- LARGE_BINARY

The new path resolves global indices to source chunks and copies only selected values. It keeps one output chunk per index chunk and preserves null and bounds behavior.

## Benchmark

64 chunks x 4096 rows, 32-byte values, random indices, Apple M1 Pro:

- 1% selected: 0.87 ms / 9.6 MB -> 0.08 ms / 0.11 MB
- 10% selected: 2.36 ms / 10.4 MB -> 1.43 ms / 0.96 MB

The benchmark covers 8 and 64 chunks, 1% / 10% / 50% / 100% selection, and all four types.

## Tests


arrow/compute/internal/kernels/vector_selection.go
```diff
@@ -1161,6 +1227,144 @@ func ChunkedPrimitiveTake(ctx *exec.KernelCtx, batch []*arrow.Chunked, out *exec
...
+		default:
+			validityBuilder.UnsafeAppendN(int64(block.Len), false)
+			for i := 0; i < int(block.Len); i++ {
+				appendNull()
+			}
+			pos += int64(block.Len)
+		}
+	}
+
+	offsetBuilder.unsafeAppend(offset)
```
zeroshade, line 1378: **Blocking:** a zero-length index chunk leaks its output offsets buffer.

This terminating offset correctly creates a real offsets buffer for an empty binary array. However, `vectorExecutor.WrapResults.toChunked` calls `ArrayDatum.Chunks()`, which retains the resulting array, and then silently drops zero-length chunks without releasing them.

With multi-chunk string values and an empty index array:

- merge base: 0 bytes remain allocated;
- this PR: 64 bytes leak per call.

A leading empty chunk in chunked indices leaks similarly.

Please fix the shared executor path to release zero-length arrays that it drops, mirroring the release behavior in the later accumulation branch. Add checked-allocator tests for empty, leading-empty, and trailing-empty index chunks.

arrow/compute/internal/kernels/vector_selection.go
```diff
@@ -1161,6 +1227,144 @@ func ChunkedPrimitiveTake(ctx *exec.KernelCtx, batch []*arrow.Chunked, out *exec
...
+
+func ChunkedVarBinaryTake[OffsetT int32 | int64](ctx *exec.KernelCtx, batch []*arrow.Chunked, out *exec.ExecResult) ([]*exec.ExecResult, error) {
+	values, indices := batch[0], batch[1]
+	if ctx.State.(TakeState).BoundsCheck {
+		if err := checkIndexBoundsChunked(indices, uint64(values.Len())); err != nil {
+			return nil, err
+		}
+	}
+
+	outData := make([]*exec.ExecResult, len(indices.Chunks()))
```
zeroshade, line 1406: **Blocking:** when `indices.Chunks()` is empty, this returns an empty result slice. `execChunked` then calls `output.MakeArray()` on the uninitialized output span, causing a panic such as:

```text
arrow/array: string offset buffer must have at least 4 values
```

The merge-base concatenate path returns a normal empty chunked result for the same input.

Please preserve that behavior by constructing a valid empty variable-binary result, or by fixing the shared empty-result path so it can create the required offsets buffer. Add a zero-chunk-index regression test for all four supported types.

arrow/compute/internal/kernels/vector_selection.go
```diff
@@ -1161,6 +1227,144 @@ func ChunkedPrimitiveTake(ctx *exec.KernelCtx, batch []*arrow.Chunked, out *exec
...
+			}
+			for newCap < needed {
+				newCap *= 2
+			}
+			dataBuilder.reserve(newCap - dataBuilder.len())
+			spaceAvail = dataBuilder.cap() - dataBuilder.len()
+		}
+		dataBuilder.unsafeAppendSlice(value)
+		spaceAvail -= len(value)
+		offset += OffsetT(len(value))
```
zeroshade: For the `int32` instantiation used by STRING/BINARY, this silently wraps once the selected data in an output index chunk exceeds `math.MaxInt32`.

The previous concatenate path rejected chunked Binary/String data that overflowed 32-bit offsets. Bypassing concatenation now allows large chunked sources, but the resulting individual output chunk must still satisfy the Binary offset limit.

Please check before incrementing and return an overflow/capacity error rather than emitting corrupt offsets. The capacity-growth arithmetic should likewise avoid integer overflow.

arrow/compute/chunked_take_bench_test.go
```diff
@@ -0,0 +1,193 @@
...
+				[]bool{false, true, true, true})
+			empty := newTestBinaryArray(mem, typ, nil, nil)
+			values := arrow.NewChunked(typ, []arrow.Array{empty, chunk0, empty, chunk1})
+			defer values.Release()
+			empty.Release()
+			chunk0.Release()
+			chunk0Full.Release()
+			chunk1.Release()
+
+			indices := newTestInt64Array(mem, []int64{4, 0, 2, 0, 5, 1}, []bool{true, true, true, false, true, true})
```
zeroshade, line 58: The correctness test uses only `Int64` indices, leaving the 8-, 16-, and 32-bit dispatch branches and unsigned index types untested.

Please table-test all supported signed and unsigned index widths, including:

- negative and upper-bound failures;
- out-of-range garbage under null index slots;
- sliced index arrays;
- zero-length and zero-chunk index inputs;
- explicit output chunk-count assertions for chunked indices.

The logical `ChunkedEqual` assertion currently cannot detect an incorrect output chunk layout.

zeroshade (changes requested):
The core chunk resolver, sliced-value access, null propagation, and signed/unsigned index handling are correct across randomized differential tests. The compute suite, focused race tests, targeted vet, and diff checks pass.

I am requesting changes for three correctness issues and the associated coverage gaps:

1. Empty or leading-empty index chunks leak their retained offsets buffer.
2. A zero-chunk index column now panics instead of returning an empty result.
3. STRING/BINARY output offsets can silently wrap past `math.MaxInt32`.

I left inline comments with reproductions and suggested regression coverage.

---

>

zeroshade (approved):
just needs to resolve the merge conflicts

PR #1206, opened 2026-08-16
feat(extensions): add VariantGet for path extraction from variant arrays

nssalian:
### Rationale for this change

Reading one field from a variant column today reassembles the whole value per row (`VariantArray.Value`) and then navigates to the field in Go. When the variant is shredded, that field is often already a typed column, so the full reassembly is wasted work.

- Adds `compute.VariantGet(ctx, *extensions.VariantArray, VariantGetOptions{Path, AsType, Strict})`, mirroring the arrow-rs [variant_get](https://github.com/apache/arrow-rs/blob/main/parquet-variant-compute/src/variant_get.rs). It follows the shredded `typed_value` columns as far as the path allows (stepping into struct fields directly, gathering list elements with the `take` kernel) and reassembles the residual `value` per-row only for what remains.
- `AsType == nil` returns a `VariantArray` pointing at the path; set it to get a typed array. Typed output is produced by grouping leaves by their natural Arrow type and casting each group with the existing `compute` cast kernels, so per-element values convert to the requested target independent of row order.
- `Strict` controls cast strictness: the default allows overflow/truncation via the cast kernels, `Strict` makes a lossy or impossible cast error (arrow-go has no per-value null-on-failure mode).
- Adds `variant.VariantPath` (an opaque `Field`/`Index`/`Join`/`Len`/`StepAt` builder) and `variant.Value.GetByPath`, plus a `VariantArray.VariantType()` accessor.

### Follow-ups (out of scope here):

- Nested `AsType` output: struct/list target types return `arrow.ErrNotImplemented`; materializing them would need a recursive per-field builder.

### Notes on cross-implementation behavior

Path navigation follows arrow-rs rather than Spark's `variant_get` in two places: a field access on a non-object errors (Spark returns null), and `Strict` treats in-range fractional truncation as an error (Spark truncates). Absent paths, out-of-range indices, and object/array-to-primitive casts match Spark (null, or error under `Strict`).

### Are these changes tested?

- `arrow/compute/variant_get_test.go` covers typed and variant output, nested field and list-index paths, residual/mixed-shredded rows, missing field, null rows, dictionary-encoded metadata, mixed-width/type leaves, order independence, the perfect-shredding fast path, `Strict` vs default cast outcomes, nested-type rejection, field-on-scalar errors, and `CheckedAllocator` leak checks.
- `parquet/variant/path_test.go` covers the `VariantPath` builder and `GetByPath` navigation, including the empty-string object key.

Yes - new exported API: `compute.VariantGet`, `compute.VariantGetOptions`, `variant.VariantPath`, `variant.Value.GetByPath`, and `extensions.VariantArray.VariantType()`. Additive only.


arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	"github.com/apache/arrow-go/v18/parquet/variant"
+)
+
+// VariantPathElement is a single step of a variant path: either an object field
+// name or an array index.
+type VariantPathElement struct {
+	name    string
+	index   int
+	isIndex bool
+}
```
zeroshade: Since none of the members here are exported, and there's nothing that accepts this, why are we exporting this struct?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	return VariantPathElement{name: name}
+}
+
+// VariantPathIndex returns a path element selecting the array element at index.
+func VariantPathIndex(index int) VariantPathElement {
+	return VariantPathElement{index: index, isIndex: true}
+}
+
+// VariantPath is an ordered list of path elements to extract from a variant value.
+type VariantPath []VariantPathElement
```
zeroshade: If the intent is for the `VariantPath` and `VariantPathElement` to be opaque, we can just export `VariantPath` without exporting `VariantPathElement` right? We'd just add a method for joining paths together etc.

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+
+// VariantPathField returns a path element selecting the named object field.
+func VariantPathField(name string) VariantPathElement {
+	return VariantPathElement{name: name}
+}
+
+// VariantPathIndex returns a path element selecting the array element at index.
+func VariantPathIndex(index int) VariantPathElement {
+	return VariantPathElement{index: index, isIndex: true}
+}
```
zeroshade: instead of having the `isIndex` can we just branch on `len(name) != 0`?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	// arrow-rs: a cast failure produces null.
+	Strict bool
+	// Mem is the allocator for output arrays; nil uses memory.DefaultAllocator.
+	Mem memory.Allocator
+}
+
+// VariantGet extracts opts.Path from each value of a VariantArray. It follows the
+// shredded typed_value columns as far as the path allows, then falls back to a
+// per-row walk of the residual value for the remainder.
+func VariantGet(input arrow.Array, opts GetOptions) (arrow.Array, error) {
```
zeroshade: why take an `arrow.Array` and then type assert it? The parameter should just be a `*VariantArray`

zeroshade: Can we push this to the `arrow/compute` package and implement it as a function named `variant_get`?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+// shreddingState is a (value?, typed_value?) column pair at one level of a shredded
+// variant, mirroring arrow-rs ShreddingState.
+type shreddingState struct {
+	value      arrow.TypedArray[[]byte]
+	typedValue arrow.Array
+	length     int
+}
+
+func stateFromVariant(va *VariantArray) shreddingState {
+	vt := va.ExtensionType().(*VariantType)
```
zeroshade: we should really add a function to `*VariantArray` that returns the `*VariantType` directly instead of requiring the `.ExtensionType().(*VariantType)` dance...

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+
+	var value arrow.TypedArray[[]byte]
+	if vt.valueFieldIdx != -1 {
+		value = st.Field(vt.valueFieldIdx).(arrow.TypedArray[[]byte])
+	}
+
+	var typed arrow.Array
+	if vt.typedValueFieldIdx != -1 {
+		typed = st.Field(vt.typedValueFieldIdx)
+	}
```
zeroshade: instead of doing this, use the existing accessors on `*VariantArray` for these: `va.Metadata()` and `va.Shredded()`

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	if !ok {
+		return pathStep{}, fmt.Errorf("%w: expected struct field %q while following path, got %s",
+			arrow.ErrInvalid, name, st.Field(idx).DataType())
+	}
+
+	return pathStep{kind: stepSuccess, state: stateFromFieldStruct(child)}, nil
+}
+
+func shreddedGetPath(va *VariantArray, opts GetOptions) (arrow.Array, error) {
+	state := stateFromVariant(va)
```
zeroshade: should we error if `opts.Path` is empty? or just call `va.Values()` ?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	return state.typedValue
+}
+
+// buildTargetVariant wraps the current shredding state as a VariantArray, carrying
+// the accumulated ancestor nulls onto the storage struct.
+func buildTargetVariant(va *VariantArray, state shreddingState, nulls *nullTracker, mem memory.Allocator) (*VariantArray, error) {
+	// Take the raw metadata array (not va.Metadata) so dictionary- or large-binary-
+	// encoded metadata is preserved and decoded by the target's own reader.
+	srcVT := va.ExtensionType().(*VariantType)
+	metadata := va.Storage().(*array.Struct).Field(srcVT.metadataFieldIdx)
```
zeroshade: Why not `va.Metadata()`?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+func allNullResult(va *VariantArray, opts GetOptions) (arrow.Array, error) {
+	if opts.AsType != nil {
+		return array.MakeArrayOfNull(opts.Mem, opts.AsType, va.Len()), nil
+	}
+
+	bldr := NewVariantBuilder(opts.Mem, NewDefaultVariantType())
+	defer bldr.Release()
+	for i := 0; i < va.Len(); i++ {
+		bldr.AppendNull()
+	}
```
zeroshade: can you add a comment for why we're not still just using MakeArrayOfNull for the variant type?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+		for i := range n.valid {
+			n.valid[i] = true
+		}
+	}
+	for i := 0; i < n.length; i++ {
+		if arr.IsNull(i) {
+			n.valid[i] = false
+		}
+	}
+}
```
zeroshade: why do this instead of keeping the bitmap and then using BitmapOr? That would likely be simpler and more performant

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+		field, err := obj.ValueByKey(elem.name)
+		if err != nil {
+			return variant.Value{}, false, nil
+		}
+		cur = field.Value
+	}
+
+	return cur, true, nil
+}
```
zeroshade: this feels like it should be a method on `variant.Value` for convenience rather than a free function here. Seems like it would be useful for variant values themselves to have a `GetByPath` method

zeroshade (changes requested):

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+
+		step, err := followFieldElement(state, elem.name)
+		if err != nil {
+			return nil, err
+		}
+
+		switch step.kind {
+		case stepSuccess:
+			nulls.apply(state.typedValue)
```
zeroshade: A valid shredded layout can have `typed_value == null` for one row while `value` contains that row's complete residual object. `VariantArray.Value()` would correctly reconstruct that row, but this is going to mark it null instead. For instance a two-row case of `[1, 2]` would return `[1, null]` if the `2` was in the residual `value` instead of in the `typedValue`. 

You should add root-level and intermediate nested mixed-row tests.

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if !ok || leaf.Type() == variant.Null {
+			bldr.AppendNull()
+
+			continue
+		}
+
+		if appendVariantToTypedBuilder(bldr, leaf) {
```
zeroshade: The `appendVariantToTypedBuilder` method is for the shredding writer where only lossless physical fits should succeed. As a result, ordinary conversions would silently produce null under the default mode such as:

* `int64` -> `float64`
* `int64` -> `decimal128`
* `decimal(scale 2)` -> `decimal(scale 10)`
* `string` -> `binary`

Once this moves to `arrow/compute` use the existing cast machinery there

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+
+		obj, ok := cur.Value().(variant.ObjectValue)
+		if !ok {
+			return variant.Value{}, false, nil
+		}
+		field, err := obj.ValueByKey(elem.name)
+		if err != nil {
+			return variant.Value{}, false, nil
+		}
```
zeroshade: Two distinct failures become a missing/null result:

* Field access on a scalar returns null, while the arrow-rs implementation specifies a cast/type error
* Every `ObjectValue.ValueByKey` error gets swallowed instead of propagated, only `arrow.ErrNotFound` is meaning absence, malformed metadata produces a different error that should propagate.

Probing for corruption currently returns a nil error and a null output.

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+// navigateValue walks path into a fully reassembled variant value.
+func navigateValue(v variant.Value, path VariantPath) (variant.Value, bool, error) {
+	cur := v
+	for _, elem := range path {
+		if elem.isIndex {
+			arr, ok := cur.Value().(variant.ArrayValue)
+			if !ok || elem.index < 0 || uint32(elem.index) >= arr.Len() {
+				return variant.Value{}, false, nil
+			}
+			el, err := arr.Value(uint32(elem.index))
```
zeroshade: a large index would wrap here. on a 64-bit platform, `VariantPathIndex(math.MaxUint32+1)` converts to zero and selects element 0. Should we check before narrowing?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+
+	return shreddedGetPath(va, opts)
+}
+
+// shreddingState is a (value?, typed_value?) column pair at one level of a shredded
+// variant, mirroring arrow-rs ShreddingState.
+type shreddingState struct {
+	value      arrow.TypedArray[[]byte]
+	typedValue arrow.Array
+	length     int
```
zeroshade: this is never used, remove it?

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+	defer target.Release()
+
+	if len(remaining) == 0 {
+		if opts.AsType == nil {
+			target.Retain()
+
+			return target, nil
+		}
+
+		if shredded := tryPerfectShredding(state, nulls, opts.AsType); shredded != nil {
```
zeroshade: try this *before* constructing the target, otherwise we're building and discarding a struct array and possible bitmap.

arrow/extensions/variant_get.go
```diff
@@ -0,0 +1,453 @@
...
+// VariantPathIndex returns a path element selecting the array element at index.
+func VariantPathIndex(index int) VariantPathElement {
+	return VariantPathElement{index: index, isIndex: true}
+}
+
+// VariantPath is an ordered list of path elements to extract from a variant value.
+type VariantPath []VariantPathElement
+
+// GetOptions controls VariantGet.
+type GetOptions struct {
```
zeroshade: use `VariantGetOptions` instead of `GetOptions` since this is exported at the package level.

zeroshade (changes requested):
Some edge case tests that are missing which should probably get added:

* Mixed shredded/residual rows at root and nested levels
* Supported `AsType` conversion matrix
* Corrupted object metadata versus genuinely missing keys
* Field-on-scalar versus index-on-scalar semantics
* Index greater than `math.MaxUint32`
* A missing path with a nested `AsType`
* Empty-string object key in the path
* Sliced and zero-length inputs

arrow/extensions/variant_get.go
nssalian: Done - it's now `arrow/compute/variant_get.go` as `compute.VariantGet`

nssalian: Signature is now `VariantGet(ctx, *extensions.VariantArray, VariantGetOptions)` - no `arrow.Array` assert.

arrow/extensions/variant_get.go
nssalian: Renamed

arrow/extensions/variant_get.go
nssalian: The element type is unexported now

arrow/extensions/variant_get.go
nssalian: `VariantPath` is opaque with `Field`/`Index`/`Join`/`StepAt`; `VariantPathElement` is gone

arrow/extensions/variant_get.go
nssalian: A step is now field vs index by `name != ""`, no `isIndex` bool

arrow/extensions/variant_get.go
nssalian: Navigation is now `variant.Value.GetByPath(path)`

arrow/extensions/variant_get.go
nssalian: Added `VariantArray.VariantType()`

arrow/extensions/variant_get.go
nssalian: Empty path with no cast returns the input array

arrow/extensions/variant_get.go
nssalian: the perfect-shredded fast path runs before building the target struct/bitmap

arrow/extensions/variant_get.go
nssalian: Uses `Metadata()`/`Shredded()`/`UntypedValues()`

arrow/extensions/variant_get.go
nssalian: Removed (the old `extensions/variant_get.go` is gone in the relocation)

arrow/extensions/variant_get.go
nssalian: Guarded - `uint64` compare against `Len()` before narrowing to `uint32`

arrow/extensions/variant_get.go
nssalian: `GetByPath` errors on a field step into a scalar and propagates corrupt-metadata errors; only `ErrNotFound` (and out-of-range / non-array index) map to absent

arrow/extensions/variant_get.go
nssalian: Typed output now goes through `compute.CastArray`. Note arrow-go `CastOptions` has no `safe=null` like arrow-rs, so `Strict` -> error-on-lossy and default -> `Allow*` (lenient); documented on the field. Flag if you want null-on-failure

arrow/extensions/variant_get.go
nssalian: Fixed - residual rows reassemble via `Value(i)`+`GetByPath`; `TestVariantGetMixedShreddedRows` pins the `[1,2]` case

arrow/extensions/variant_get.go
nssalian: Dropped the `[]bool` for a bitmap. Kept as `BitmapAndAlloc` on validity (1=valid), which is the validity-space equivalent of OR-ing null masks - commented on `nullTracker.merge`. Happy to spell it as `BitmapOr` on null-masks if you'd prefer

arrow/extensions/variant_get.go
nssalian: The residual target build reads the **raw** metadata column rather than `va.Metadata()`, because `va.Metadata()` asserts `TypedArray[[]byte]` and panics on dictionary-encoded metadata (spec-legal). The raw column preserves dict/large-binary encoding and is decoded by the target's own reader. Added `TestVariantGetDictMetadata`

arrow/extensions/variant_get.go
nssalian: Added - `MakeArrayOfNull` can't build the variant extension (non-nullable metadata/value), so the all-null case uses the builder

arrow/compute/variant_get.go
```diff
@@ -0,0 +1,658 @@
...
+			step, err = fieldStep(state, name)
+		} else {
+			step, err = indexStep(ctx, mem, state, index)
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		if step.kind == stepSuccess {
+			nulls.merge(state.typedValue)
```
zeroshade, line 250: **Blocking:** This still drops residual-backed rows whenever the path is non-empty. A successful schema-level step merges the parent `typed_value` validity into the output mask, so a row with null `typed_value` but a complete value in the residual column is forced null. The new mixed-row test uses an empty path and never executes this line. For `$.a`, `[{"a":1},{"a":2}]` with row 2 residual-backed returns `[1,null]`. Please implement per-row fallback/partitioning and cover root field, nested field, and list-index paths.

arrow/compute/variant_get.go
```diff
@@ -0,0 +1,658 @@
...
+}
+
+// buildNaturalArray materializes the leaves as an array of the first present leaf's
+// natural Arrow type so the cast kernels can convert it. Rows whose value does not
+// match that natural type become null. Returns nil when no leaf is present.
+func buildNaturalArray(mem memory.Allocator, leaves []variantLeaf) arrow.Array {
+	var natural arrow.DataType
+	for _, l := range leaves {
+		if l.present && l.value.Type() != variant.Null {
+			natural = naturalArrowType(l.value)
```
zeroshade: **Blocking:** Choosing the source type from the first present leaf silently discards later values of another natural width/type at lines 424–425. Variant integers commonly encode as different widths: values `1`, `1000`, and `5000000000` become `int8`, `int16`, and `int64`; requesting `AsType: int64` currently returns `[1,null,null]`. Results also become row-order-dependent and `Strict` behaves differently depending on the fast path. Please materialize against the requested target type or otherwise unify leaf types, with mixed-width and mixed-type regression coverage.

parquet/variant/path.go
```diff
@@ -0,0 +1,108 @@
...
+
+// VariantPath is an ordered list of steps to navigate into a variant value. The
+// zero value is the root path; extend it with Field and Index.
+type VariantPath struct {
+	elems []pathElem
+}
+
+// Field returns a copy of the path with an object-field step appended.
+func (p VariantPath) Field(name string) VariantPath {
+	return VariantPath{elems: append(p.grow(), pathElem{name: name})}
```
zeroshade: **Blocking:** `Field("")` produces the same zero-valued `pathElem` as `Index(0)`, and `StepAt`/`GetByPath` distinguish steps using `name != ""`. Empty-string object keys are valid, so extracting `{"":42}` with `Field("")` is interpreted as array index 0 and returns null. Please retain an explicit field/index discriminator through `StepAt` and path rebuilding, with an empty-key regression test.

zeroshade (changes requested):
Found three blocking data-correctness issues: residual-backed rows are lost for non-empty paths, heterogeneous numeric leaves are silently dropped based on row order, and empty-string object keys are encoded as array index zero.

---

>

arrow/compute/variant_get.go
```diff
@@ -0,0 +1,658 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if opts.AsType == nil {
+		return buildLeafVariantArray(mem, leaves), nil
+	}
+
+	src := buildNaturalArray(mem, leaves)
+	if src == nil {
+		return allNullResult(mem, len(leaves), opts.AsType), nil
```
zeroshade: Unsupported nested `AsType` extraction silently returns an all-null array. For example, extracting object `{"a":{"x":1}}` at `Field("a")` into a struct type makes `buildNaturalArray` return `nil`, after which this branch reports success with a null struct rather than returning `arrow.ErrNotImplemented`. Please restore the nested-target rejection or implement nested materialization, and restore regression coverage.

zeroshade (changes requested):
The three unresolved correctness findings remain on the current head: residual-backed rows are lost for non-empty paths, heterogeneous Variant widths are silently nulled based on row order, and empty object keys are indistinguishable from index zero.

I also found one additional blocking case inline: requesting an unsupported nested `AsType` silently produces an all-null array instead of returning an error.

CI is green, but these extraction paths can return incorrect successful results and need to be addressed before merging.

---

>

arrow/compute/variant_get.go
nssalian: Fixed. The descent loop now breaks as soon as a level carries any residual (if state.hasResidual() { break }), before merging that level's typed_value validity, and the remaining path is reassembled  per-row via buildTargetVariant + GetByPath. A row with null typed_value but a value in the residual column is reconstructed, not nulled — $.a over [{"a":1},{"a":2}] with row 2 residual-backed now returns [1,2]. Covered by TestVariantGetResidualBackedField (root field), TestVariantGetResidualBackedNestedField, TestVariantGetResidualBackedListIndex, and TestVariantGetResidualBackedDeepField (mid-level  residual).

arrow/compute/variant_get.go
nssalian: Fixed this as well. castLeaves no longer picks one source type from the first leaf. Each leaf is grouped by its own natural Arrow type, each group is cast to the requested AsType with the cast kernels, and results are scattered back to their original rows via Concatenate+Take. The output type is fixed by AsType before any row is read, so it's order-independent. 1/1000/5000000000 → AsType: int64 now returns all three.  Strict runs through the same CastArray on both the fast and slow paths. Coverage: TestVariantGetMixedWidthIntegers, TestVariantGetHeterogeneousLeaves, TestVariantGetTypeOrderIndependent,                 TestVariantGetInterleavedScatter (non-identity scatter), plus TestVariantGetMixedFloatWidths, MixedDecimalScales, MixedTimestampUnits.

parquet/variant/path.go
nssalian: This is now fixed too. Good catch. pathElem now carries an explicit isField bool; Field sets it, Index leaves it false, and StepAt/GetByPath/path rebuilding branch on isField instead of name != "". {"":42} extracted with Field("")  
resolves the empty key (returns 42), and Index(0) on it is absent. Covered by TestGetByPathEmptyKey and TestVariantGetEmptyKey.

nssalian:
Thanks for the detailed review. I fixed all the pending issues. One intentional divergence to flag: field-access-on-a-scalar errors (matching arrow-rs, per your earlier note) rather than Spark's variant_get, which returns null there; similarly Strict errors on in-range truncation.

arrow/compute/variant_get.go
nssalian: VariantGet now rejects a nested AsType up front with arrow.ErrNotImplemented (restoring the guard that was dropped when the code moved to arrow/compute), before any leaf work — struct and list  targets return the error instead of a silent all-null array. Covered by TestVariantGetNestedTypeNotImplemented (struct + list). Full nested materialization remains a follow-up.

arrow/compute/variant_get.go
```diff
@@ -0,0 +1,735 @@
...
+// the residual for any remaining path. With AsType nil it returns a VariantArray of
+// the extracted values; otherwise it casts them to AsType with the cast kernels.
+func VariantGet(ctx context.Context, input *extensions.VariantArray, opts VariantGetOptions) (arrow.Array, error) {
+	if input == nil {
+		return nil, fmt.Errorf("%w: VariantGet requires a non-nil VariantArray", arrow.ErrInvalid)
+	}
+
+	// Nested target types are not yet supported; reject up front rather than
+	// silently producing an all-null array from the leaf cast.
+	if _, ok := opts.AsType.(arrow.NestedType); ok {
```
zeroshade: This rejects UUID extraction because `*extensions.UUIDType` embeds `ExtensionBase` and therefore satisfies `arrow.NestedType`, even though its storage is fixed-size binary. That makes `VariantGet(..., AsType: extensions.NewUUIDType())` return `ErrNotImplemented` before reaching the UUID support already present in `naturalArrowType` and `appendNatural`. Please reject actual list/struct-like target storage rather than all `NestedType` implementations, and add a UUID extraction regression test.

arrow/compute/variant_get.go
```diff
@@ -0,0 +1,735 @@
...
+	perm := make([]uint64, len(leaves))
+	valid := make([]bool, len(leaves))
+	var casted []arrow.Array
+	defer func() { releaseAll(casted) }()
+
+	var pos uint64
+	for _, key := range order {
+		g := groups[key]
+		col := buildTypedColumn(mem, g.dt, leaves, g.rows)
+		cast, err := CastArray(ctx, col, NewCastOptions(asType, strict))
```
zeroshade, line 453: In non-strict mode this casts a whole natural-type group at once and discards the entire group if any row fails. For example, converting Variant strings `["1", "bad"]` to `int64` currently produces `[null, null]`, losing the valid first row. Please retry or partition failed group casts at row granularity so only failed conversions become null, and add a mixed-valid/invalid same-type test.

zeroshade (changes requested):
The previous review findings are fixed at the current head, including residual-backed rows, heterogeneous scatter, field-path errors, oversized indexes, and unsupported list/struct targets.

Two current-head issues remain: UUID targets are rejected before reaching the existing UUID conversion support, and non-strict conversion drops every row in a natural-type group when only one row fails. Details and fix directions are inline.

Focused tests, race tests, vet, and direct reproducers otherwise pass. No CI checks are currently reported for this head.

---

>

arrow/compute/variant_get.go
nssalian: Figured this out. The nested-target rejection was tripping on UUID because `*extensions.UUIDType` embeds `ExtensionBase`, which satisfies `arrow.NestedType` (it carries `Fields()`/`NumFields()`), so it was rejected before reaching the UUID support in `naturalArrowType`/`appendNatural`. `VariantGet` now unwraps an extension target to its storage type before the `NestedType` check - a fixed-size-binary-backed target (UUID) passes, actual struct/list storage still rejects. `VariantGet(..., AsType: extensions.NewUUIDType())` now round-trips. Add tests: `TestVariantGetUUIDTarget`; `TestVariantGetNestedTypeNotImplemented` still pins struct/list rejection.

arrow/compute/variant_get.go
nssalian: Non-strict mode was casting a whole natural-type group in one shot and dropping the entire group if any row failed. It now falls back to a per-row cast when the group cast errors, so only the inconvertible rows null - `["1","bad"]` -> `int64` now returns `[1, null]` instead of `[null, null]`. Strict is unchanged (first failure errors). Covered by `TestVariantGetPartialCastFailure`, which runs under a `CheckedAllocator` so the per-row fallback is leak-checked.

zeroshade (approved):
Looks good! Just waiting for CI to finish for confirmation before merging

PR #1214, opened 2026-08-26
perf(arrow/array): batch run-end encoded appends

fallintoplace:
## What

- Batch `RunEndEncodedBuilder.AppendNulls` into one physical null value and one run.
- Batch `RunEndEncodedBuilder.AppendEmptyValues` into one physical empty value and one run.
- Keep zero and negative counts as no-ops.
- Preserve the empty-value state used by JSON decoding.

The old implementation called the scalar append method once per logical value. This change keeps the same API while avoiding the repeated physical values and run ends.

## Benchmark

Apple M1 Pro, 65,536 rows, `int32` run ends and `int32` encoded values:

| --- | ---: | ---: | ---: |
| Nulls | ~1.0-1.6 ms, 1.16 MB, 55 allocs/op | ~1.0-1.5 us, 1.7 KB, 19 allocs/op | ~1,250x |
| Empty values | ~1.2-2.7 ms, 1.16 MB, 55 allocs/op | ~1.2-1.5 us, 1.7 KB, 19 allocs/op | ~1,750x |

The benchmark covers `int16`, `int32`, and `int64` run-end types, `int32` and string encoded values, and multiple batch sizes.

## Tests

- `go test ./... -run '^$' -count=1 -p 2`

### DIFF
--- a/arrow/array/encoded.go
+++ b/arrow/array/encoded.go
@@ -490,9 +490,13 @@ func (b *RunEndEncodedBuilder) AppendNull() {
 }
 
 func (b *RunEndEncodedBuilder) AppendNulls(n int) {
-	for i := 0; i < n; i++ {
-		b.AppendNull()
+	if n <= 0 {
+		return
 	}
+
+	b.finishRun()
+	b.values.AppendNull()
+	b.addLength(uint64(n))
 }
 
 func (b *RunEndEncodedBuilder) UnsafeAppendBoolToBitmap(v bool) {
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Reviewed at fa9d776. The bulk append implementation preserves logical behavior while reducing repeated null/empty appends to a single physical run. The full arrow/array suite, race tests, semantic equivalence probes, boundary coverage, formatting, vet, and diff hygiene all passed locally.

PR #1218, opened 2026-08-26
perf(parquet): bound Bloom filter hash batches

fallintoplace:
## Summary

- **Batch Bloom filter hashing into 1024-value chunks.**
- Insert each chunk directly from the column writers.
- Reuse fixed scratch storage for the built-in xxhash path.
- Keep the existing Hasher interface. Custom hashers still use Sum64s.
- Cover spaced values and boolean bitmap paths.
- Regenerate the typed column writer.

## Why

Large writes used to build byte-slice and hash arrays for the whole input before updating the filter. This keeps the temporary hashing memory bounded and follows the dictionary Bloom-filter work in #1164.

## Benchmark

On an Apple M1 Pro with 100k int32 values:

- **Materialized:** 838 KB/op, 3 allocs/op
- **Bounded:** 35 KB/op, 2 allocs/op

The bounded path was also about 10% faster in the local benchmark.

## Tests

- go test ./parquet/metadata
- go test ./parquet/file -run writer and Bloom-filter tests
- go test -race on the touched metadata and file tests
- go test ./... -run ^$

### DIFF
--- a/parquet/file/column_writer_types.gen.go
+++ b/parquet/file/column_writer_types.gen.go
@@ -211,7 +211,7 @@ func (w *Int32ColumnChunkWriter) writeValues(values []int32, numNulls int64) {
 		w.pageStatistics.(*metadata.Int32Statistics).Update(values, numNulls)
 	}
 	if w.bloomFilter != nil && w.currentEncoder.Encoding() != parquet.Encodings.PlainDict {
-		w.bloomFilter.InsertBulk(metadata.GetHashes(w.bloomFilter.Hasher(), values))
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
LGTM — Bloom-filter hashing is now bounded by batch size while preserving dense, spaced, bitmap, and custom-hasher behavior. The focused tests, randomized boundary probes, race checks, vet, and generated-file consistency check all passed locally. The GitHub workflows are awaiting maintainer approval rather than reporting failures.

---

>

PR #1220, opened 2026-08-27
perf(parquet): stop spaced expansion once aligned

fallintoplace:
## Summary

- Stop `spacedExpand` as soon as the remaining decoded prefix is already in its final positions.
- Avoid the final no-op `copy` for aligned valid prefixes.
- Add focused coverage for empty, all-null, leading, middle, trailing, clustered, alternating, and offset bitmaps.
- Add a 65,536-value benchmark for `int32` and `BYTE_ARRAY`.

## Benchmark

- **Machine:** Apple M1 Pro, arm64
- **Go:** 1.26.3
- **Command:** `go test ./parquet/internal/encoding -run '^$' -bench '^BenchmarkSpacedExpand$' -benchmem -benchtime=50ms -count=3`
- **Workloads:** trailing, late, clustered, random 10% nulls, and alternating validity layouts
- **Result:** `0 B/op` and `0 allocs/op` for every case. CPU impact is small and layout-dependent, with the clearest benefit when the reverse walk reaches an already-aligned prefix.

## Tests

- `go test ./parquet/internal/encoding`
- `go test ./parquet/internal/... ./parquet/file ./parquet/metadata ./parquet/schema ./parquet/variant ./parquet/compress ./parquet/cmd/...`

### DIFF
--- a/parquet/internal/encoding/decoder.go
+++ b/parquet/internal/encoding/decoder.go
@@ -228,6 +228,9 @@ func spacedExpand[T parquet.ColumnTypes](buffer []T, nullCount int, validBits []
 		// overwrite any existing data with the correctly spaced data. Any data that happens to be left in the null
 		// slots is fine since it shouldn't matter and saves us work.
 		idxDecode -= run.Length
+		if idxDecode == run.Pos {
+			return numValues
+		}
 		n := copy(buffer[run.Pos:], buffer[idxDecode:int64(idxDecode)+run.Length])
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
LGTM. Stopping spaced expansion once the remaining values are aligned preserves null placement and decoded values while avoiding the redundant self-copy. Package tests, race tests, vet, randomized bitmap/offset probes, benchmarks, and CI all pass.

---

> will follow up.*
>

PR #1221, opened 2026-08-27
perf(arrow/ipc): reuse record encoder scratch

fallintoplace:
## Summary

- Keep one `recordEncoder` on `Writer` and `FileWriter`.
- Reuse the `fields`, `meta`, and `variadicCounts` backing arrays.
- Restore the encoder recursion depth when resetting.
- Add a benchmark for repeated writes with 1, 16, and 64 columns.
- Add a regression test for resetting after a nested encode error.

## Benchmark

**Command**

```text
go test ./arrow/ipc -run '^$' -bench '^BenchmarkWriterRecordEncoderReuse$' -benchmem -benchtime=100ms -count=3
```

**Machine:** Apple M1 Pro, arm64  
**Go:** 1.26.3

The benchmark writes one warm-up batch before timing repeated writes. Each width was tested with 16, 256, and 4096 rows. The allocation results were the same across those row counts.

| Columns | Before | After |
| --- | ---: | ---: |
| 1 | 1,624 B/op, 21 allocs/op | 1,408 B/op, 18 allocs/op |
| 16 | 8,072 B/op, 33 allocs/op | 6,656 B/op, 26 allocs/op |
| 64 | 31,240 B/op, 41 allocs/op | 25,984 B/op, 32 allocs/op |

Timing was noisy on this machine, so the allocation reduction is the main result.

## Tests

- `go test ./arrow/ipc`
- `go test ./arrow/...`

### DIFF
--- a/arrow/ipc/file_writer.go
+++ b/arrow/ipc/file_writer.go
@@ -304,6 +304,8 @@ func (f *FileWriter) getRecordEncoder() *recordEncoder {
 }
 
 func (f *FileWriter) Close() error {
+	defer func() { f.encoder = nil }()
+
 	if f.closed {
 		return f.closeErr
 	}
--- a/arrow/ipc/file_writer.go
+++ b/arrow/ipc/file_writer.go
@@ -256,6 +256,7 @@ type FileWriter struct {
 	codec           flatbuf.CompressionType
 	compressNP      int
 	compressors     []compressor
+	encoder         *recordEncoder
 	minSpaceSavings float64
 
 	// map of the last written dictionaries by id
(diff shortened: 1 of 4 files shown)


arrow/ipc/writer.go
```diff
@@ -96,6 +96,7 @@ type Writer struct {
 	codec           flatbuf.CompressionType
 	compressNP      int
 	compressors     []compressor
+	encoder         *recordEncoder
```
zeroshade, line 99: This cached encoder retains its peak-capacity metadata scratch after `Close()` and close-after-failure. Please clear `w.encoder` on every terminal close path so a reachable closed stream writer does not retain potentially large scratch arrays.

arrow/ipc/file_writer.go
```diff
@@ -256,6 +256,7 @@ type FileWriter struct {
 	codec           flatbuf.CompressionType
 	compressNP      int
 	compressors     []compressor
+	encoder         *recordEncoder
```
zeroshade, line 259: Please likewise clear `f.encoder` when the file writer closes. A 4,096-column probe retained capacities of 4,096 fields and 8,192 metadata entries after close; a 100,000-column probe retained approximately 6.15 MiB until the writer became unreachable.

zeroshade (changes requested):
The encoder reuse preserves behavior and reproduces the allocation improvements. IPC, repeated encoder, race, vet, LSP, and CI checks pass. One lifecycle regression needs correction before merge.

zeroshade (approved):

PR #1223, opened 2026-08-27
perf(arrow/array): batch dictionary empty appends

fallintoplace:
## What does this change?

- `AppendEmptyValues` used to build an `n`-element values array and dictionary-encode every empty value.
- Resolve the empty value once, reserve the index builder once, and append the same dictionary index for the whole batch.
- Keep the existing null dictionary path unchanged.

## Benchmark

Command:

```text
go test ./arrow/array -run '^$' -bench '^BenchmarkDictionaryBuilderAppendEmptyValues$' -benchmem -benchtime=200ms -count=3
```

Apple M1 Pro, Go 1.26.3. The benchmark reuses the dictionary builder and resets the output length between iterations.

| Type | Count | Before ns/op | After ns/op | Before B/op | After B/op |
| int32 | 64 | 1,684 | 653 | 1,168 | 784 |
| int32 | 4,096 | 81,910 | 14,193 | 42,576 | 784 |
| int32 | 65,536 | 1,305,061 | 221,358 | 551,382 | 784 |
| string | 64 | 2,573 | 868 | 1,568 | 1,056 |
| string | 4,096 | 121,191 | 14,577 | 42,912 | 1,056 |
| string | 65,536 | 1,916,937 | 230,499 | 551,714 | 1,056 |

## Tests

### DIFF
--- a/arrow/array/dictionary.go
+++ b/arrow/array/dictionary.go
@@ -692,7 +692,7 @@ func (b *dictionaryBuilder) AppendEmptyValues(n int) {
 		panic(err)
 	}
 
-	b.idxBuilder.Builder.Reserve(n)
+	b.idxBuilder.Reserve(n)
 	for i := 0; i < n; i++ {
 		b.idxBuilder.UnsafeAppend(idx)
 	}
--- a/arrow/array/dictionary.go
+++ b/arrow/array/dictionary.go
@@ -683,13 +683,20 @@ func (b *dictionaryBuilder) AppendEmptyValues(n int) {
 
 	valueBuilder := NewBuilder(b.mem, b.dt.ValueType)
 	defer valueBuilder.Release()
-	valueBuilder.AppendEmptyValues(n)
+	valueBuilder.AppendEmptyValue()
 
 	values := valueBuilder.NewArray()
 	defer values.Release()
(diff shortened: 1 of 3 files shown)


arrow/array/dictionary.go
```diff
@@ -683,13 +683,20 @@ func (b *dictionaryBuilder) AppendEmptyValues(n int) {
...
 
 	values := valueBuilder.NewArray()
 	defer values.Release()
-	if err := b.AppendArray(values); err != nil {
+	idx, _, err := b.memoTable.GetOrInsert(getvalFn(values)(0))
+	if err != nil {
 		panic(err)
 	}
+
+	b.idxBuilder.Builder.Reserve(n)
```
zeroshade: The repository’s lint hook rewrites this to `b.idxBuilder.Reserve(n)`, causing the required Lint check to fail because the hook modifies the file. Please use the promoted method and rerun lint.

zeroshade (changes requested):
The batching change preserves behavior in the tested cases and reproduces the expected performance and allocation improvements. Array, repeated dictionary, race, vet, and LSP checks pass. One lint failure must be fixed before merge.

zeroshade (approved):

PR #1224, opened 2026-08-27
perf(arrow/array): concatenate bitmaps directly

fallintoplace:
## What does this change?

- `concat` used to build a temporary `[]bitmap` for each bitmap it concatenated.
- `concatBitmaps` now reads the input `ArrayData` directly by buffer index.
- Keep the existing nil bitmap behavior for all-valid inputs and preserve sliced bitmap offsets.

## Benchmark

Command:

```text
go test ./arrow/array -run '^$' -bench '^BenchmarkConcatenateBitmaps$' -benchmem -benchtime=200ms -count=3
```

Apple M1 Pro, Go 1.26.3. The benchmark concatenates 65,536 rows split across 64, 1,024, and 8,192 chunks.

| Input | Chunks | Before ns/op | After ns/op | Before B/op | After B/op | Allocs before/after |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| nullable Boolean | 64 | 6,965 | 5,689 | 25,824 | 20,448 | 10 / 8 |
| nullable Boolean | 1,024 | 55,375 | 41,568 | 119,648 | 37,728 | 10 / 8 |
| nullable Boolean | 8,192 | 412,768 | 276,285 | 805,730 | 150,368 | 10 / 8 |
| nullable int32 | 8,192 | 313,516 | 269,165 | 738,913 | 411,233 | 9 / 8 |

## Tests

### DIFF
--- a/arrow/array/concat.go
+++ b/arrow/array/concat.go
@@ -910,6 +910,12 @@ func concatBitmaps(data []arrow.ArrayData, idx int, mem memory.Allocator) (*memo
 	}
 
 	out := memory.NewResizableBuffer(mem)
+	success := false
+	defer func() {
+		if !success {
+			out.Release()
+		}
+	}()
 	out.Resize(int(bitutil.BytesForBits(int64(outlen))))
 	dst := out.Bytes()
 
@@ -922,6 +928,7 @@ func concatBitmaps(data []arrow.ArrayData, idx int, mem memory.Allocator) (*memo
 		}
 		offset += d.Len()
 	}
+	success = true
 	return out, nil
 }
 
--- a/arrow/array/concat.go
+++ b/arrow/array/concat.go
@@ -67,26 +67,6 @@ type rng struct {
 	offset, len int
 }
 
-// simple bitmap struct to reference a specific slice of a bitmap where the range
-// offset and length are in bits
-type bitmap struct {
-	data []byte
-	rng  rng
-}
-
-// gather up the bitmaps from the passed in data objects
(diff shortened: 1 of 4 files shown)


arrow/array/concat.go
```diff
@@ -823,13 +803,13 @@ func concatBitmaps(bitmaps []bitmap, mem memory.Allocator) (*memory.Buffer, erro
 	dst := out.Bytes()
 
 	offset := 0
-	for _, bm := range bitmaps {
-		if bm.data == nil { // if the bitmap is nil, that implies that the value is true for all elements
-			bitutil.SetBitsTo(out.Bytes(), int64(offset), int64(bm.rng.len), true)
+	for _, d := range data {
+		if buf := d.Buffers()[idx]; buf == nil { // if the bitmap is nil, that implies that the value is true for all elements
```
zeroshade, line 924: `out` has already been allocated here. If an `ArrayData` implementation lacks buffer slot `idx`, this access panics; the outer recovery returns an error, but the local buffer is unreachable and leaked. A checked-allocator probe retains 64 bytes on this head versus 0 on the base. Please release `out` on panic/error after allocation and add a checked-allocator regression test.

zeroshade (changes requested):
The performance and allocation improvements reproduce, and the normal-input correctness, race, vet, LSP, and CI checks pass. One error-path allocation regression needs correction before merge.

zeroshade (approved):

PR #1225, opened 2026-08-27
perf(arrow/array): copy same-dictionary indices directly

fallintoplace:
## What does this change?

- Same-dictionary concatenation used to create one temporary buffer wrapper per input chunk.
- Use `concatFixedWidthBuffers` directly for the dictionary indices.
- Keep the dictionary-unification path unchanged for different dictionaries.

## Benchmark

Command:

```text
go test ./arrow/array -run '^$' -bench '^BenchmarkConcatenateSameDictionary$' -benchmem -benchtime=200ms -count=3
```

Apple M1 Pro, Go 1.26.3. The benchmark concatenates 65,536 rows with the same four-value string dictionary, split across different chunk counts.

| Chunks | Before ns/op | After ns/op | Before B/op | After B/op | Allocs before/after |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 33,736 | 33,262 | 271,049 | 270,962 | 11 / 9 |
| 8 | 33,791 | 33,809 | 272,353 | 271,650 | 26 / 17 |
| 64 | 42,134 | 39,250 | 282,785 | 277,153 | 138 / 73 |
| 1,024 | 198,280 | 157,632 | 462,626 | 371,234 | 2,058 / 1,033 |
| 8,192 | 1,202,369 | 928,859 | 1,778,211 | 1,057,314 | 16,394 / 8,201 |

## Tests

### DIFF
--- a/arrow/array/concat.go
+++ b/arrow/array/concat.go
@@ -595,11 +595,10 @@ func concat(data []arrow.ArrayData, mem memory.Allocator) (arr arrow.ArrayData, 
 			dict.Release()
 		}
 
-		indexBuffers := gatherBuffersFixedWidthType(data, 1, idxType)
 		if dictsSame {
 			out.dictionary = dict0.Data().(*Data)
 			out.dictionary.Retain()
-			out.buffers[1] = concatBuffers(indexBuffers, mem)
+			out.buffers[1] = concatFixedWidthBuffers(data, 1, idxType.BitWidth()/8, out.length, mem)
 			break
 		}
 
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
This change safely reuses the fixed-width concatenation path for same-dictionary indices and substantially reduces allocations for large chunk counts. I verified parity across index widths, slices, offsets, nulls, empty inputs, equal-but-distinct dictionaries, and dictionary unification. Array, repeated, race, vet, and LSP checks pass; CI is green.

PR #1232, opened 2026-08-27
perf(arrow/array): bulk append BinaryView nulls

fallintoplace:
## What

- Replace the per-value validity updates in `BinaryViewBuilder.AppendNulls` with one bulk bitmap update.
- Update the length and null count once per batch.
- Treat zero and negative counts as no-ops.
- Add parity coverage for aligned and unaligned starting positions.

## Benchmark

Apple M1 Pro, Go 1.26.3. The builder is preallocated and reused, so this measures the bitmap append itself. Medians from 5 runs at offset 0:

Command: `go test -vet=off ./arrow/array -run "^$" -bench "^BenchmarkBinaryViewBuilderAppendNulls$" -benchmem -benchtime=500ms -count=5 -cpu=1`

| Values | Before | After |
| ---: | ---: | ---: |
| 1,024 | 764.8 ns/op | 19.66 ns/op |
| 65,536 | 40,266 ns/op | 152.2 ns/op |

Both versions use 0 B/op and 0 allocs/op in this preallocated benchmark.

## Tests

- `PARQUET_TEST_DATA=parquet-testing/data go test ./... -count=1`

### DIFF
--- a/arrow/array/binarybuilder.go
+++ b/arrow/array/binarybuilder.go
@@ -606,10 +606,17 @@ func (b *BinaryViewBuilder) AppendNull() {
 }
 
 func (b *BinaryViewBuilder) AppendNulls(n int) {
-	b.Reserve(n)
-	for i := 0; i < n; i++ {
-		b.UnsafeAppendBoolToBitmap(false)
+	if n <= 0 {
+		return
 	}
+	if n > math.MaxInt-b.length {
+		panic("arrow/array: builder length overflow")
+	}
+
+	b.Reserve(n)
+	bitutil.SetBitsTo(b.nullBitmap.Bytes(), int64(b.length), int64(n), false)
+	b.length += n
+	b.nulls += n
 }
 
 func (b *BinaryViewBuilder) AppendEmptyValue() {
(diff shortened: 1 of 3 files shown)


arrow/array/binarybuilder.go
```diff
@@ -606,10 +606,14 @@ func (b *BinaryViewBuilder) AppendNull() {
...
 
 func (b *BinaryViewBuilder) AppendNulls(n int) {
-	b.Reserve(n)
-	for i := 0; i < n; i++ {
-		b.UnsafeAppendBoolToBitmap(false)
+	if n <= 0 {
+		return
 	}
+
+	b.Reserve(n)
```
zeroshade, line 616: Please reject counts that would overflow `b.length + n` before calling `Reserve`.

With one existing value, `AppendNulls(math.MaxInt)` now returns with `Len() == math.MinInt` and `NullN() == math.MaxInt`. `NewBinaryViewArray()` then produces a negative-length array whose `ValidateFull()` still returns nil. A guard such as `n > math.MaxInt-b.length`, plus boundary regression coverage, would prevent this silent corruption.

zeroshade (changes requested):
The bitmap optimization and performance improvement are reproducible, and normal parity, race, and allocation checks pass. However, the new bulk path can silently corrupt builder state when the requested count overflows its length arithmetic.

zeroshade (approved):

PR #1233, opened 2026-08-27
perf(compute): fast-path single-span execution

fallintoplace:
### Rationale for this change

`iterateExecSpans` allocates three tracking slices before it knows whether the
batch can be processed as one contiguous span. Most non-chunked batches do not
need that state.

- Detect non-chunked batches that fit in one execution span.
- Return that span directly without allocating chunk iteration state.
- Keep the existing path for chunked batches and batches split by
  `maxChunkSize`.
- Preserve sliced array offsets, scalar promotion, and iterator completion
  behavior.
- Add focused tests and a benchmark.

### Benchmark

The benchmark covers one array, two arrays, scalar plus array, all scalars, and
a chunked control case at lengths 1, 16, 256, and 4096.

Representative results from one run on an Apple M1 Pro with Go 1.26.3:

| Case | Before | After | Allocations |
| --- | ---: | ---: | ---: |
| Array, length 1 | 342.0 ns/op, 528 B/op | 189.0 ns/op, 304 B/op | 8 -> 5 |
| Two arrays, length 1 | 451.4 ns/op, 776 B/op | 286.6 ns/op, 528 B/op | 8 -> 5 |
| Scalar plus array, length 1 | 433.8 ns/op, 776 B/op | 260.8 ns/op, 528 B/op | 8 -> 5 |
| All scalars | 468.1 ns/op, 784 B/op | 281.4 ns/op, 528 B/op | 10 -> 7 |
| Chunked control | 356.6 ns/op, 528 B/op | 339.2 ns/op, 528 B/op | 8 -> 8 |

Command:

```bash
go test ./arrow/compute -run '^$' -bench '^BenchmarkIterateExecSpans$' -benchmem -benchtime=100ms -count=1 -cpu=1
```

### Are these changes tested?

- `go test ./arrow/compute -count=1`
- `go test -race ./arrow/compute -count=1`

### DIFF
--- a/arrow/compute/exec_spans_benchmark_test.go
+++ b/arrow/compute/exec_spans_benchmark_test.go
@@ -1,0 +1,109 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified the single-span fast path preserves offsets, nested data, nulls, scalar promotion, iterator completion, and boundary behavior. Compute, race, LSP, and adversarial differential checks pass. The allocation reduction is reproducible, while the chunked fallback remains unchanged.

Nice improvement—thanks!

PR #1234, opened 2026-08-27
perf(compute): share logical row maps across aligned sorts

fallintoplace:
### Rationale for this change

Aligned multi-key sorts build the same `logicalRowMap` once per physical sort
column. The map only depends on the shared chunk boundaries, so the extra maps
use memory without adding any lookup information.

- Detect aligned chunk boundaries before creating sort comparators.
- Build one `logicalRowMap` for aligned multi-key columns.
- Pass that map to each physical comparator.
- Keep the existing per-column maps for misaligned and single-key sorts.

### Benchmark

This uses the existing `BenchmarkSortIndices_Int64_TwoKeys` benchmark with
65,536 rows and 64 aligned chunks.

Median of three runs on an Apple M1 Pro with Go 1.26.3:

| Case | Before | After |
| --- | ---: | ---: |
| Two aligned int64 keys | 12.65 ms/op, 3,165,232 B/op, 12 allocs/op | 12.39 ms/op, 2,116,656 B/op, 11 allocs/op |

Command:

```bash
go test ./arrow/compute/internal/kernels -run '^$' -bench '^BenchmarkSortIndices_Int64_TwoKeys$' -benchmem -benchtime=200ms -count=3 -cpu=1
```

### Are these changes tested?

- `go test ./arrow/compute/internal/kernels -count=1`
- `go test ./arrow/compute -count=1`
- `go test -race ./arrow/compute/internal/kernels -count=1`
- `go vet -composites=false ./arrow/compute/internal/kernels`

### DIFF
--- a/arrow/compute/internal/kernels/vector_sort.go
+++ b/arrow/compute/internal/kernels/vector_sort.go
@@ -140,7 +140,7 @@ func extensionStorageFixedSizeBinaryChunks(chunks []arrow.Array) ([]arrow.Array,
 	return out, nil
 }
 
-func newFixedSizeBinaryComparator(chunks []arrow.Array, numRows int, vn bool) (columnComparator, error) {
+func newFixedSizeBinaryComparator(chunks []arrow.Array, numRows int, vn bool, rowMap logicalRowMap) (columnComparator, error) {
 	f0, ok := chunks[0].(*array.FixedSizeBinary)
 	if !ok {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Verified the shared row map is limited to aligned multi-key columns and preserves behavior across aligned, misaligned, empty-chunk, null, NaN, mixed-type, extension, and non-radix cases. Kernel, compute, race, vet, LSP, and randomized oracle checks pass. The allocation and memory reduction is reproducible.

Nice improvement—thanks!

PR #1235, opened 2026-08-27
perf(parquet/pqarrow): use synchronous path for serial reads

fallintoplace:
## Summary

- `ArrowReadProperties.Parallel` is false by default.
- `GetFieldReaders` now uses a direct loop for serial reads.
- `ReadRowGroups` now reads columns directly for serial reads.
- The parallel path is unchanged.
- Added a benchmark for 1, 8, 32, and 128 columns with small row groups.

## Benchmark

Apple M1 Pro. Median of 5 runs. The benchmark uses `Parallel=false`.

| columns | upstream main | this PR | time change | allocs/op |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 156.9 us | 128.8 us | -18% | 1124 -> 1106 |
| 8 | 1.05 ms | 967 us | -8% | 7848 -> 7810 |
| 32 | 3.38 ms | 3.24 ms | -4% | 31413 -> 31296 |
| 128 | 11.92 ms | 11.18 ms | -6% | 124896 -> 124436 |

Command:

```text
go test ./parquet/pqarrow -run '^$' -bench '^BenchmarkReadTableSerial$' -benchmem -benchtime=1s -count=5
```

## Tests

- `go test ./parquet/pqarrow -count=1`
- `go test -race ./parquet/pqarrow -run '^(TestGetFieldReadersReleasesPartialReadersOnError|TestRecordReaderSerial|TestRecordReaderSeekToRow|TestRecordReaderParallel)$' -count=1`
- `go vet ./parquet/pqarrow`

### DIFF
--- a/parquet/pqarrow/file_reader.go
+++ b/parquet/pqarrow/file_reader.go
@@ -253,6 +253,24 @@ func (fr *FileReader) GetFieldReaders(ctx context.Context, colIndices, rowGroups
 	out := make([]*ColumnReader, len(fieldIndices))
 	outFields := make([]arrow.Field, len(fieldIndices))
 
+	if !fr.Props.Parallel {
+		for idx, fidx := range fieldIndices {
+			rdr, err := fr.GetFieldReader(ctx, fidx, includedLeaves, rowGroups)
+			if err != nil {
+				for _, rdr := range out {
+					if rdr != nil {
+						rdr.Release()
+					}
+				}
+				return nil, nil, err
+			}
+			outFields[idx] = *rdr.Field()
+			out[idx] = rdr
+		}
+
+		return out, arrow.NewSchema(outFields, fr.Manifest.SchemaMeta), nil
+	}
+
 	// Load batches in parallel
 	// When reading structs with large numbers of columns, the serial load is very slow.
 	// This is especially true when reading Cloud Storage. Loading concurrently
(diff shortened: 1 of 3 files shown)


parquet/pqarrow/file_reader.go
```diff
@@ -370,6 +385,39 @@ func (fr *FileReader) ReadRowGroups(ctx context.Context, indices, rowGroups []in
...
+				rdr.Release()
+			}
+		}()
+
+		for idx, rdr := range readers {
+			if err := ctx.Err(); err != nil {
+				return nil, err
+			}
+
+			data, err := fr.ReadColumn(rowGroups, rdr)
```
zeroshade: `ReadColumn` previously ran inside the worker’s deferred recovery boundary, which converted decoder panics into `panic while reading: …` errors. Calling it directly here lets those panics escape from `ReadTable`.

I reproduced this through the public `FileReader` flow with a malformed extension implementation: the base returns the recovered error, while this head panics. Please preserve the recovery boundary for synchronous reads—ideally through a shared guarded helper—and add regression coverage.

zeroshade (changes requested):
The serial-path performance and allocation improvements are reproducible, and the standard tests pass. However, the synchronous path removes an existing panic-recovery boundary, causing malformed decoder or extension-array input to panic callers rather than return an error.

zeroshade (approved):

PR #1236, opened 2026-08-27
perf(parquet): reuse fixed-length zero scratch

fallintoplace:
## Summary

- Reuse a zero-value buffer in `PlainFixedLenByteArrayEncoder`.
- Allocate it lazily only when `Put` sees a nil value.
- Clear the buffer reference when the encoder is released.
- Add coverage for lazy creation and reuse.
- Add benchmarks for widths 4, 16, and 32 with repeated small `Put` calls.

## Benchmark

Apple M1 Pro. Median of 3 runs. All-valid input.

| width | rows per put | puts | upstream main | this PR | change |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4 | 1 | 64 | 489 ns | 470 ns | -4% |
| 16 | 1 | 64 | 516 ns | 470 ns | -9% |
| 32 | 1 | 64 | 498 ns | 471 ns | -5% |

The benchmark also covers 1, 16, and 1024 rows with 1, 64, and 1024 repeated puts, plus sparse nil input.

```text
go test ./parquet/internal/encoding -run '^$' -bench 'BenchmarkPlainEncodingFixedLenByteArray' -benchmem -benchtime=100ms -count=3
```

## Tests

- `PARQUET_TEST_DATA=/path/to/parquet-testing/data go test ./parquet/file ./parquet/pqarrow -count=1`
- `PARQUET_TEST_DATA=/path/to/parquet-testing/data go test -race ./parquet/internal/encoding ./parquet/file ./parquet/pqarrow -count=1`
- `go vet ./parquet/internal/encoding ./parquet/file ./parquet/pqarrow`

### DIFF
--- a/parquet/internal/encoding/fixed_len_byte_array_encoder.go
+++ b/parquet/internal/encoding/fixed_len_byte_array_encoder.go
@@ -31,6 +31,7 @@ type PlainFixedLenByteArrayEncoder struct {
 	encoder
 
 	bitSetReader bitutils.SetBitRunReader
+	zeroValue    []byte
 }
 
 // Put writes the provided values to the encoder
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
Verified the fixed-length zero scratch is lazily allocated, safely copied, reused across encoder operations, and released correctly. Encoder, race, file, pqarrow, vet, and LSP checks pass. The claimed performance improvement is reproducible, and CI is green.

Nice improvement—thanks!

PR #1237, opened 2026-08-27
perf(arrow/array): batch-pack boolean values

fallintoplace:
## Summary

- Pack BooleanBuilder values into the bitmap one byte at a time.
- Handle unaligned prefixes and trailing values without changing neighboring bits.
- Reuse the 8-value pack helper for value and validity bitmaps.
- Add exhaustive bitmap coverage and a BooleanBuilder benchmark.

## Benchmark

Apple M1 Pro. 65,536 values. The builder is pre-reserved. Median of 5 runs.

| Benchmark | upstream main | this PR | change |
| --- | ---: | ---: | ---: |
| BooleanBuilder / all false | 145 us | 18.6 us | 7.8x |
| BooleanBuilder / alternating | 145 us | 18.6 us | 7.8x |
| Boolean / all valid | 160 us | 34.4 us | 4.6x |
| Boolean / 50% null | 161 us | 35.7 us | 4.5x |

The first two rows use nil validity. The last two use the existing explicit-validity benchmark.

```text
go test ./arrow/array -run '^$' -bench '^BenchmarkBooleanBuilderAppendValues$' -benchmem -benchtime=100ms -count=5
go test ./arrow/array -run '^$' -bench '^BenchmarkAppendValuesWithValidity/boolean/' -benchmem -benchtime=100ms -count=5
```

## Tests

- `go test ./arrow/bitutil ./arrow/compute/... -count=1`
- `go vet ./arrow/array ./arrow/bitutil`

### DIFF
--- a/arrow/array/booleanbuilder.go
+++ b/arrow/array/booleanbuilder.go
@@ -135,8 +135,12 @@ func (b *BooleanBuilder) AppendValues(v []bool, valid []bool) {
 	}
 
 	b.Reserve(len(v))
-	for i, vv := range v {
-		bitutil.SetBitTo(b.rawData, b.length+i, vv)
+	if len(v) < 8 {
+		for i, vv := range v {
+			bitutil.SetBitTo(b.rawData, b.length+i, vv)
+		}
+	} else {
+		packBoolsToBitmap(b.rawData, b.length, v)
 	}
 	b.unsafeAppendBoolsToBitmap(valid, len(v))
 }
--- a/arrow/array/booleanbuilder_benchmark_test.go
+++ b/arrow/array/booleanbuilder_benchmark_test.go
@@ -52,13 +52,15 @@ func BenchmarkBooleanBuilderAppendValuesSmall(b *testing.B) {
 	for _, length := range []int{1, 2, 3, 7, 8} {
 		b.Run(fmt.Sprintf("len=%d", length), func(b *testing.B) {
 			values := makeBooleanBenchmarkValues(length, func(i int) bool { return i%2 == 0 })
-			benchmarkAppendValues(b, func() (func(), func()) {
-				bldr := NewBooleanBuilder(memory.DefaultAllocator)
-				bldr.Reserve(length)
-				return func() {
-					bldr.AppendValues(values, nil)
-				}, bldr.Release
-			})
+			bldr := NewBooleanBuilder(memory.DefaultAllocator)
+			bldr.Reserve(len(values) * b.N)
+			b.ReportAllocs()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				bldr.AppendValues(values, nil)
+			}
+			b.StopTimer()
+			bldr.Release()
 		})
 	}
 }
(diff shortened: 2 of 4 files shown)


arrow/array/booleanbuilder.go
```diff
@@ -135,9 +135,7 @@ func (b *BooleanBuilder) AppendValues(v []bool, valid []bool) {
 	}
 
 	b.Reserve(len(v))
-	for i, vv := range v {
-		bitutil.SetBitTo(b.rawData, b.length+i, vv)
-	}
+	packBoolsToBitmap(b.rawData, b.length, v)
```
zeroshade: This regresses small `AppendValues` batches because `packBoolsToBitmap` adds setup overhead without packing a complete byte. Pre-reserved median results versus the exact base were: 1 value +35%, 2 values +23%, 3 values +15%, and 7 values +14%; the new implementation becomes faster at 8 values. Please retain the previous `SetBitTo` loop for small slices and add small-batch benchmark coverage.

zeroshade (changes requested):
The bitmap implementation is correct, and the large-batch performance improvement is reproducible. Array/race tests, vet, LSP diagnostics, and 5,000 randomized parity cases passed.

One performance regression remains: `AppendValues` batches smaller than eight elements now pay the setup cost of `packBoolsToBitmap` without reaching its byte-packing loop. Against the exact base, batches of 1, 2, 3, and 7 values regressed by approximately 35%, 23%, 15%, and 14%, respectively.

Please retain the previous `SetBitTo` loop for small slices—such as `len(v) < 8`—while using byte packing for larger batches, and add small-batch benchmarks to prevent recurrence.

arrow/array/booleanbuilder_benchmark_test.go
```diff
@@ -0,0 +1,72 @@
...
+			})
+		})
+	}
+}
+
+func BenchmarkBooleanBuilderAppendValuesSmall(b *testing.B) {
+	for _, length := range []int{1, 2, 3, 7, 8} {
+		b.Run(fmt.Sprintf("len=%d", length), func(b *testing.B) {
+			values := makeBooleanBenchmarkValues(length, func(i int) bool { return i%2 == 0 })
+			benchmarkAppendValues(b, func() (func(), func()) {
```
zeroshade: This benchmark delegates to `benchmarkAppendValues`, where builder creation and release occur outside the timed region. Because `b.N` is calibrated against only the roughly 65 ns append, the default benchmark performs millions of unmeasured allocations and releases. The first sub-benchmark did not finish within 180 seconds; even `-benchtime=1ms` took 5.9 seconds wall-clock. Please use a preallocated builder sized for the measured iterations, or include/reset lifecycle work without unbounded untimed setup.

zeroshade (changes requested):
The current head fixes the original small-batch performance regression by retaining the scalar path below eight values. Array, race, vet, and randomized append tests pass.

The new small-batch benchmark is not practically runnable with the default benchmark settings, however. Its untimed per-iteration builder setup scales with the very large `b.N` calibrated from the tiny timed append operation; the first default sub-benchmark exceeded 180 seconds. Details are inline.

---

>

zeroshade (approved):

PR #1238, opened 2026-08-27
perf(parquet/compress): reuse gzip writers

fallintoplace:
## Summary

- Pool gzip writers by compression level.
- Reset a writer before returning it to the pool.
- Avoid retaining the destination page buffer in the cached writer.
- Add a concurrent correctness test and page-size benchmarks.

## Benchmark

Apple M1 Pro. The output buffer is reused. Median of 3 runs.

| Input | upstream main | this PR | change |
| --- | ---: | ---: | ---: |
| Repeated, 64 KiB | 115 us, 1.08 MiB, 15 allocs | 25.7 us, 48 B, 1 alloc | 4.5x faster |
| Random, 64 KiB | 94.7 us, 1.08 MiB, 15 allocs | 15.6 us, 117 B, 1 alloc | 6.1x faster |

```text
go test ./parquet/compress -run '^$' -bench '^BenchmarkGzipEncodePages$' -benchmem -benchtime=200ms -count=3
```

## Tests

- `PARQUET_TEST_DATA=/path/to/parquet-testing/data go test ./parquet/... -count=1`
- `PARQUET_TEST_DATA=/path/to/parquet-testing/data go test -race ./parquet/compress ./parquet/file ./parquet/pqarrow -count=1`
- `go vet ./parquet/compress ./parquet/file`

### DIFF
--- a/parquet/compress/compress_test.go
+++ b/parquet/compress/compress_test.go
@@ -200,6 +200,43 @@ func TestGzipCompressBound(t *testing.T) {
 	}
 }
 
+func TestGzipEncodeLevelConcurrent(t *testing.T) {
+	codec, err := compress.GetCodec(compress.Codecs.Gzip)
+	assert.NoError(t, err)
+
+	src := makeCompressibleData(4 * 1024)
+	levels := []int{
+		gzip.DefaultCompression,
+		gzip.NoCompression,
+		gzip.BestSpeed,
+		gzip.BestCompression,
+		gzip.StatelessCompression,
+	}
+
+	const workers = 16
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
The level-keyed gzip writer pools preserve existing behavior across every accepted compression level. Exact compressed output matched the base under repeated reuse, and the reset sequence correctly avoids retaining destination buffers.

Full parquet tests, targeted race suites, vet, and LSP diagnostics passed. The performance improvement and reduction from 15 allocations to 1 are reproducible. The head is mergeable and all required CI checks pass.

PR #1239, opened 2026-08-27
perf(arrow/array): avoid temporary slices in Diff comparisons

fallintoplace:
### Rationale for this change

`array.Diff` compares each matching pair by calling `SliceEqual` on two one-element slices. This creates temporary Arrow arrays for every non-null comparison.

- Resolve a value comparator once when the Myers diff is created.
- Compare booleans, integers, temporal values, strings, and binary values directly.
- Keep floating-point, decimal, interval, and nested values on the existing `SliceEqual` fallback.
- Preserve the existing null handling.
- Add offset and null coverage for the direct comparison paths.
- Add benchmarks for equal arrays and a change at the end of the input.

Apple M1 Pro results with `GOMAXPROCS=1` and 65,536 values:

| Case | Before | After |
| --- | ---: | ---: |
| Int64 equal | 18.4 ms | 0.65 ms |
| Int64 changed last | 18.5 ms | 0.60 ms |
| String equal | 20.3 ms | 0.74 ms |
| String changed last | 20.5 ms | 0.73 ms |
| Binary equal | 21.8 ms | 0.82 ms |
| Binary changed last | 22.6 ms | 0.84 ms |

The equal cases go from 262,147 allocations to 4 allocations per operation.

### Are these changes tested?

- `go test ./...`

No.

### DIFF
--- a/arrow/array/diff.go
+++ b/arrow/array/diff.go
@@ -17,6 +17,7 @@
 package array
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
@@ -140,6 +141,7 @@ type editPoint struct {
 type quadraticSpaceMyersDiff struct {
 	base         arrow.Array
 	target       arrow.Array
+	valueEqual   func(baseIndex, targetIndex int) bool
 	finishIndex  int
 	editCount    int
 	endpointBase []int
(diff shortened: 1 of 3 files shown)


zeroshade (approved):
The direct comparators preserve existing `SliceEqual` semantics across nulls, sliced arrays, primitive and temporal types, binary variants, fixed-size binary, and extensions.

The claimed performance improvement is reproducible, reducing representative comparisons from roughly 10–12 ms to 0.35–0.49 ms and equal-case allocations from 262,147 to 4.

Full array, race, assert, repeated Diff tests, vet, and LSP diagnostics passed. The head is mergeable and all required CI checks pass.

Optional follow-up: extend the committed test matrix with temporal and fixed-size-binary cases so every fast-path family remains explicitly covered.

PR #1240, opened 2026-08-27
perf(arrow/array): use unsafe writes for binary bulk appends

fallintoplace:
## What

- Use unchecked offset writes after `Reserve` in `BinaryBuilder.AppendValues`.
- Use unchecked payload copies after `ReserveData`.
- Apply the same path to `AppendStringValues`.
- Keep scalar appends unchanged.
- Make `Reserve` cover the terminal offset when the builder has already grown its minimum capacity.
- Keep the current physical payload layout for null entries.
- Add mixed validity coverage and benchmarks for Binary, LargeBinary, String, and LargeString.

## Benchmark

Apple M1 Pro, 65,536 values x 4 bytes, representative run:

| Builder | Before | After | Change |
| --- | ---: | ---: | ---: |
| Binary | 545 us | 473 us | -13% |
| LargeBinary | 584 us | 535 us | -8% |

Allocations stayed at 9 per operation.

## Tests

- `PARQUET_TEST_DATA=parquet-testing/data go test ./...`
- `go test -tags=assert ./arrow/array -count=1`
- `go vet ./arrow/array`

### DIFF
--- a/arrow/array/binarybuilder.go
+++ b/arrow/array/binarybuilder.go
@@ -39,26 +39,29 @@ type BinaryBuilder struct {
 	offsets bufBuilder
 	values  *byteBufferBuilder
 
-	appendOffsetVal func(int)
-	getOffsetVal    func(int) int
-	maxCapacity     uint64
-	offsetByteWidth int
+	appendOffsetVal       func(int)
+	unsafeAppendOffsetVal func(int)
+	getOffsetVal          func(int) int
+	maxCapacity           uint64
+	offsetByteWidth       int
 }
 
 // NewBinaryBuilder can be used for any of the variable length binary types,
 // Binary, LargeBinary, String, LargeString by passing the appropriate data type
 func NewBinaryBuilder(mem memory.Allocator, dtype arrow.BinaryDataType) *BinaryBuilder {
 	var (
-		offsets         bufBuilder
-		offsetValFn     func(int)
-		maxCapacity     uint64
-		offsetByteWidth int
-		getOffsetVal    func(int) int
+		offsets           bufBuilder
+		offsetValFn       func(int)
+		unsafeOffsetValFn func(int)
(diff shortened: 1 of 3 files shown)


arrow/array/binarybuilder.go
```diff
@@ -263,6 +268,12 @@ func (b *BinaryBuilder) DataCap() int { return b.values.capacity }
 // by checking the capacity and calling Resize if necessary.
 func (b *BinaryBuilder) Reserve(n int) {
 	b.reserve(n, b.Resize)
+	if n > 0 {
```
zeroshade: This check now runs through `Reserve` for every scalar `Append`, `AppendNull`, and `AppendEmptyValue`. In an identical pre-reserved benchmark against the exact base, median `Append` time increased from 6.43 ns to 6.98 ns (+8.6%), and `AppendNull` increased from 3.96 ns to 4.68 ns (+18.4%). Please move the terminal-offset correction to the bulk/capacity-growth path so the scalar hot path remains unchanged, and add a scalar benchmark covering it.

zeroshade (changes requested):
The bulk optimization is correct and its performance improvement is reproducible. Full array tests, race/assert/checkptr builds, vet, and cross-compilation passed.

One performance regression remains: the additional terminal-offset capacity check in `BinaryBuilder.Reserve` also runs for every scalar append. With capacity pre-reserved, median `Append` time increased from 6.43 ns to 6.98 ns (+8.6%), while `AppendNull` increased from 3.96 ns to 4.68 ns (+18.4%). Both remained zero-allocation.

Please keep this offset-capacity correction off the scalar hot path—for example, by handling it in the bulk or capacity-growth path—and add scalar coverage to prevent recurrence.

The only incomplete CI check is a cancelled macOS CGO job.

zeroshade (approved):

PR #1241, opened 2026-08-28
fix(arrow/extensions): use canonical Variant extension name

### LINKED ISSUE #1203: [Parquet] Incorrect Variant extension type name
## Description

The canonical Variant extension type name is "arrow.parquet.variant", but arrow-go returns - 
https://github.com/apache/arrow-go/blob/986b027408e1d973cb78023fc74a2a1bc361c6a9/arrow/extensions/variant.go#L265

It seems historical, because Variant  was implemented in arrow-go prior to https://github.com/apache/arrow/pull/47456 got merged.

----

This causes:
- IPC output recordbatches with Variant ext type to be not readable by other arrow implementations
- And vice versa, other implementation's recordbatches to be not readable by arrow-go

### Component(s)

Parquet

dgvj-work:
### Rationale for this change

The canonical Variant extension type name is `arrow.parquet.variant`. arrow-go still writes `parquet.variant`, which predates https://github.com/apache/arrow/pull/47456. That makes Variant IPC unreadable by other Arrow implementations, and vice versa.

C++ already made this change in apache/arrow#49082.

Fixes #1203

- `VariantType.ExtensionName()` now returns `arrow.parquet.variant`
- Keep `parquet.variant` registered so older arrow-go IPC still deserializes as `VariantType`
- Parquet schema conversion recognizes both names

### Are these changes tested?

- `go test ./arrow/extensions`
- `go test ./parquet/pqarrow -run 'TestConvertSchemaParquetVariant|TestShreddedVariantSchema'`

Yes — newly written IPC/Flight metadata uses `arrow.parquet.variant`. Existing `parquet.variant` data still reads.

### DIFF
--- a/arrow/extensions/extensions.go
+++ b/arrow/extensions/extensions.go
@@ -37,8 +37,9 @@ func init() {
 	}
 
 	// arrow-go originally registered Variant as parquet.variant. Keep that
-	// name in the registry so older IPC still deserializes.
-	if err := arrow.RegisterExtensionType(&legacyVariantType{}); err != nil {
+	// name in the registry so older IPC still deserializes. Seed default
+	// storage so GetExtensionType("parquet.variant") is a complete type.
+	if err := arrow.RegisterExtensionType(&legacyVariantType{VariantType: *NewDefaultVariantType()}); err != nil {
 		panic(err)
 	}
 }
--- a/arrow/extensions/extensions.go
+++ b/arrow/extensions/extensions.go
@@ -17,6 +17,8 @@
 package extensions
 
 import (
+	"fmt"
+
 	"github.com/apache/arrow-go/v18/arrow"
 )
 
@@ -44,11 +46,23 @@ func init() {
 	}
 }
 
-// legacyVariantType exists only so GetExtensionType("parquet.variant") can
-// still reconstruct a VariantType. In-memory and newly written IPC use
-// VariantExtensionName.
+// legacyVariantType is a compatibility adapter for the historical
+// parquet.variant name. Deserialize always returns a canonical VariantType;
+// newly written IPC uses VariantExtensionName.
 type legacyVariantType struct {
 	VariantType
 }
 
 func (*legacyVariantType) ExtensionName() string { return LegacyVariantExtensionName }
+
+func (v *legacyVariantType) String() string {
+	return fmt.Sprintf("extension<%s>", v.ExtensionName())
+}
+
+func (v *legacyVariantType) ExtensionEquals(other arrow.ExtensionType) bool {
+	return variantExtensionEquals(v.StorageType(), other)
+}
+
+func (*legacyVariantType) Deserialize(storageType arrow.DataType, _ string) (arrow.ExtensionType, error) {
+	return NewVariantType(storageType)
+}
--- a/arrow/extensions/extensions.go
+++ b/arrow/extensions/extensions.go
@@ -35,4 +35,19 @@ func init() {
 			panic(err)
 		}
 	}
+
+	// arrow-go originally registered Variant as parquet.variant. Keep that
+	// name in the registry so older IPC still deserializes.
+	if err := arrow.RegisterExtensionType(&legacyVariantType{}); err != nil {
+		panic(err)
+	}
 }
+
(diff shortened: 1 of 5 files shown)


parquet/pqarrow/schema.go
```diff
@@ -353,7 +353,7 @@ func fieldToNode(name string, field arrow.Field, props *parquet.WriterProperties
 		return schema.MapOf(field.Name, keyNode, valueNode, repFromNullable(field.Nullable), fieldIDFromMeta(field.Metadata))
 	case arrow.EXTENSION:
 		extType := field.Type.(arrow.ExtensionType)
-		if extType.ExtensionName() == "parquet.variant" {
+		if extensions.IsVariantExtensionName(extType.ExtensionName()) {
 			return variantToNode(extType.(*extensions.VariantType), field, props, arrprops)
```
zeroshade: `arrow.GetExtensionType("parquet.variant")` now returns `*extensions.legacyVariantType`. The new name predicate accepts that type, but this unconditional assertion to `*extensions.VariantType` panics. `ToParquet` should return an error for unsupported wrappers or normalize compatible legacy storage through `NewVariantType`, rather than asserting based only on the extension name. Please add a regression test using the registry-returned legacy type.

zeroshade (changes requested):
The legacy Variant compatibility path works when reading old IPC data, but `ToParquet` can now panic for the newly registered legacy extension type.

The name-based predicate accepts `parquet.variant`, after which the code asserts that the value is the canonical `*extensions.VariantType`. A schema containing the registry-returned legacy type therefore panics instead of returning an error. Details are inline.

Targeted extension, IPC, pqarrow, race, and vet tests otherwise pass. This head currently has no CI results.

---

>

parquet/pqarrow/schema.go
dgvj-work: ToParquet now reconstructs a VariantType from the extension storage instead of asserting `*VariantType`. The registry-returned `parquet.variant` type converts the same way as the canonical type, with a regression test covering that path.

arrow/extensions/extensions.go
```diff
@@ -35,4 +35,20 @@ func init() {
...
+	if err := arrow.RegisterExtensionType(&legacyVariantType{VariantType: *NewDefaultVariantType()}); err != nil {
+		panic(err)
+	}
 }
+
+// legacyVariantType exists only so GetExtensionType("parquet.variant") can
+// still reconstruct a VariantType. In-memory and newly written IPC use
+// VariantExtensionName.
+type legacyVariantType struct {
+	VariantType
```
zeroshade, line 53: Embedding `VariantType` promotes canonical `String`/`ExtensionEquals` behavior, making legacy equality asymmetric; arrays built from the registry prototype can also panic because `VariantArray` asserts `*VariantType`. Please treat this as a compatibility adapter: override legacy identity/equality, explicitly deserialize old IPC to canonical `VariantType`, and normalize legacy/canonical types through one helper in `VariantArray` accessors. Add symmetry, legacy-string, IPC-canonicalization, and direct legacy-array tests.

zeroshade (changes requested):
The prior `ToParquet` panic is fixed, but the registry-returned legacy wrapper still has asymmetric equality and can panic in `VariantArray` accessors. Suggested compatibility-adapter approach is inline.

CI is also incomplete, with the macOS Go 1.25 test job currently failing.

---

>

arrow/extensions/extensions.go
dgvj-work: `legacyVariantType` now owns its own `String`/`ExtensionEquals`, and `Deserialize` always returns a canonical `VariantType`. `VariantArray` accessors go through `asVariantType` so a registry-built array no longer panics. Added tests for equality symmetry, the legacy string, IPC canonicalization, and direct legacy-array access.

dgvj-work:
Pushed a follow-up for the legacy compatibility adapter.

`legacyVariantType` now has its own identity/equality, old IPC deserializes to the canonical `VariantType`, and `VariantArray` accessors normalize both types through one helper. Registry-built arrays no longer panic.

@zeroshade this should be ready for another look.

zeroshade (approved):
LGTM just waiting for CI to finish

PR #1242, opened 2026-08-28
fix(parquet/pqarrow): require Variant value field when writing

### LINKED ISSUE #1204: [Parquet] Variant Value field is optional

Consensus was reached in https://github.com/apache/parquet-format/pull/591 that Variant `value` field is mandatory.

- Arrow-go should make the field mandatory for the writer.
- Reading invalid files should be allowed for backwards compatibility and ability to read out of spec - https://github.com/apache/parquet-format/pull/591#issuecomment-4907983005

### Component(s)

Parquet

dgvj-work:
### Rationale for this change

Consensus in apache/parquet-format#591 is that the Variant `value` field is mandatory for writers. arrow-go currently omits it when a shredded schema has only `metadata` + `typed_value`.

Reading files that already omit `value` should still work.

Fixes #1204

- `ToParquet` errors if a Variant extension type has no `value` field
- `FromParquet` still accepts Variant groups without `value`

### Are these changes tested?

- `go test ./parquet/pqarrow -run 'TestVariantValueRequiredOnWrite|TestReadVariantWithoutValueField|TestConvertSchemaParquetVariant|TestShreddedVariantSchema'`

Yes — writing a Variant schema without a `value` field now returns `arrow.ErrInvalid`. Reading such files is unchanged.

### DIFF
--- a/parquet/pqarrow/schema.go
+++ b/parquet/pqarrow/schema.go
@@ -265,13 +265,16 @@ func variantToNode(t *extensions.VariantType, field arrow.Field, props *parquet.
 		return nil, err
 	}
 
-	if value := t.Value(); value.Type != nil {
-		valueField, err := fieldToNode("value", value, props, arrProps)
-		if err != nil {
-			return nil, err
-		}
-		fields = append(fields, valueField)
+	value := t.Value()
+	if value.Type == nil {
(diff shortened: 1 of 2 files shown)


zeroshade (approved):
Thanks for this! I've been meaning to do this and kept getting side tracked

dgvj-work:
@zeroshade thanks for the review - CI is green on this one. Could you merge when you have a minute?

PR #1243, opened 2026-08-28
fix(arrow/extensions): reject Null Variant typed_value

### LINKED ISSUE #1205: [Parquet] Variant `Null` type `typed_value` field is supported

Tbd - https://github.com/apache/arrow/pull/50810

Per the discussion here - https://github.com/apache/arrow/issues/50622 - Variant shouldn't support `Null` type `typed_value` array.

Nulls meaning in the shredding spec:
https://github.com/apache/parquet-format/blob/2076361bb64e2de9ca6a8d06eda025a6fa4e9df6/VariantShredding.md?plain=1#L68-L73

### Component(s)

Parquet

dgvj-work:
### Rationale for this change

A Null `typed_value` column is not a valid shredded Variant type. Nulls are encoded in the `value` column as Variant null (`00`). A dedicated Null `typed_value` would also make “field present and null” indistinguishable from “field missing” for shredded objects.

See apache/arrow#50622 and apache/arrow#50810.

Fixes #1205

- `NewVariantType` rejects a Null `typed_value` field
- Nested shredded fields with a Null `typed_value` are also rejected

### Are these changes tested?

- `go test ./arrow/extensions`

Yes — constructing a Variant extension type with a Null `typed_value` now returns `arrow.ErrInvalid`.

### DIFF
--- a/arrow/extensions/variant.go
+++ b/arrow/extensions/variant.go
@@ -122,10 +122,10 @@ func createShreddedField(dt arrow.DataType) arrow.DataType {
 //	 ), Nullable: true})
 //
 // This is intended to be a convenient way to create a shredded variant type from a definition
-// of the fields to shred. If the provided data type is nil, it will create a default
-// variant type.
+// of the fields to shred. If the provided data type is nil or Null (including an
+// extension whose storage is Null), it will create a default variant type.
 func NewShreddedVariantType(dt arrow.DataType) *VariantType {
-	if dt == nil {
+	if dt == nil || isNullType(dt) {
 		return NewDefaultVariantType()
 	}
 
@@ -220,11 +220,7 @@ func NewVariantType(storage arrow.DataType) (*VariantType, error) {
 		return nil, fmt.Errorf("%w: typed_value field must be nullable, got %s", arrow.ErrInvalid, typedValueField.Type)
 	}
 
-	dt := typedValueField.Type
-	if dt.ID() == arrow.EXTENSION {
-		dt = dt.(arrow.ExtensionType).StorageType()
-	}
-
+	dt := storageType(typedValueField.Type)
 	if dt.ID() == arrow.NULL {
 		return nil, fmt.Errorf("%w: typed_value field must not be null type", arrow.ErrInvalid)
 	}
@@ -295,6 +291,17 @@ func isBinary(dt arrow.DataType) bool {
 		dt.ID() == arrow.BINARY_VIEW
 }
 
+func storageType(dt arrow.DataType) arrow.DataType {
+	if ext, ok := dt.(arrow.ExtensionType); ok {
+		return ext.StorageType()
+	}
+	return dt
+}
+
+func isNullType(dt arrow.DataType) bool {
+	return storageType(dt).ID() == arrow.NULL
+}
+
 func validStruct(s *arrow.StructType) bool {
 	switch s.NumFields() {
 	case 1:
@@ -302,7 +309,7 @@ func validStruct(s *arrow.StructType) bool {
 		if f.Name == "value" {
 			return isBinary(f.Type)
 		}
-		return f.Name == "typed_value" && f.Type.ID() != arrow.NULL
+		return f.Name == "typed_value" && !isNullType(f.Type)
 	case 2:
 		valField, ok := s.FieldByName("value")
 		if !ok || !valField.Nullable || !isBinary(valField.Type) {
@@ -318,7 +325,7 @@ func validStruct(s *arrow.StructType) bool {
 			return validNestedType(nt)
 		}
 
-		return typedField.Type.ID() != arrow.NULL
+		return !isNullType(typedField.Type)
 	default:
 		return false
 	}
--- a/arrow/extensions/variant.go
+++ b/arrow/extensions/variant.go
@@ -221,6 +221,9 @@ func NewVariantType(storage arrow.DataType) (*VariantType, error) {
 	}
 
 	dt := storageType(typedValueField.Type)
+	if dt == nil {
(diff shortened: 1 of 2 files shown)


arrow/extensions/variant.go
```diff
@@ -225,6 +225,10 @@ func NewVariantType(storage arrow.DataType) (*VariantType, error) {
 		dt = dt.(arrow.ExtensionType).StorageType()
 	}
 
+	if dt.ID() == arrow.NULL {
```
zeroshade, line 227: This new rejection makes `NewShreddedVariantType(arrow.Null)` silently return `nil`, because that constructor discards the error returned by `NewVariantType`. Normal use of the returned type then panics. Please handle `arrow.Null` with a valid non-nil representation or expose an error-returning construction path instead of discarding the validation failure.

arrow/extensions/variant.go
```diff
@@ -295,7 +299,10 @@ func validStruct(s *arrow.StructType) bool {
 	switch s.NumFields() {
 	case 1:
 		f := s.Field(0)
-		return (f.Name == "value" && isBinary(f.Type)) || f.Name == "typed_value"
+		if f.Name == "value" {
+			return isBinary(f.Type)
+		}
+		return f.Name == "typed_value" && f.Type.ID() != arrow.NULL
```
zeroshade: Checking only `f.Type.ID()` lets an extension whose storage type is `arrow.Null` bypass this validation because its ID is `EXTENSION`. The same bypass exists in the two-field path below. Please unwrap extension storage before checking for Null in both nested `typed_value` forms and add one- and two-field regression tests.

zeroshade (changes requested):
The direct Null `typed_value` rejection is incomplete in two ways.

First, `NewShreddedVariantType(arrow.Null)` now silently returns `nil` because its call to `NewVariantType` discards the new validation error. Second, nested `typed_value` fields can still wrap Null storage in an extension and bypass the new checks. Both cases were reproduced; details are inline.

Extension, race, vet, and focused pqarrow Variant tests otherwise pass. This head currently has no CI results.

---

>

arrow/extensions/variant.go
dgvj-work: Good catch. `NewShreddedVariantType(arrow.Null)` now returns the default non-shredded variant type, same as a nil argument, so callers always get a usable type.

arrow/extensions/variant.go
dgvj-work: Unwrapped extension storage before the Null checks in both the one-field and two-field nested `typed_value` forms, and added regression tests for those cases.

arrow/extensions/variant.go
```diff
@@ -291,11 +291,25 @@ func isBinary(dt arrow.DataType) bool {
 		dt.ID() == arrow.BINARY_VIEW
 }
 
+func storageType(dt arrow.DataType) arrow.DataType {
+	if ext, ok := dt.(arrow.ExtensionType); ok {
+		return ext.StorageType()
```
zeroshade: This unwraps only one extension layer. A double-wrapped Null such as `Opaque(Opaque(Null))` therefore bypasses both one- and two-field nested `typed_value` validation: `NewVariantType` succeeds, but builder creation then panics on the unsupported extension type; the shredded constructor can return nil after discarding the later construction error. Please recursively unwrap extension storage with cycle protection and add double-wrapped regressions for direct, one-field, two-field, and shredded construction.

arrow/extensions/variant.go
```diff
@@ -291,11 +291,25 @@ func isBinary(dt arrow.DataType) bool {
...
+	}
+	return dt
+}
+
+func isNullType(dt arrow.DataType) bool {
+	return storageType(dt).ID() == arrow.NULL
```
zeroshade: This dereferences the result of `storageType` without handling nil. Because `NewOpaqueType(nil, ...)` is public, using such a type as `typed_value` currently panics here instead of returning `arrow.ErrInvalid`, including through `NewShreddedVariantType`. Please make extension-storage unwrapping nil-safe, reject malformed storage in the error-returning constructor, and ensure the shredded convenience constructor remains non-panicking.

zeroshade (changes requested):
The two previously reported basic cases are fixed: bare Null now produces a usable default shredded type, and a single extension layer backed by Null is rejected in both nested forms.

Two related malformed/recursive extension cases remain. Recursive wrapping can still bypass validation and later panic during builder creation, while a public opaque extension with nil storage panics during validation itself. Details and fix directions are inline.

Extensions tests, race tests, and vet otherwise pass. CI is still running, and the two existing threads remain unresolved.

---

>

arrow/extensions/variant.go
dgvj-work: Updated `storageType` to unwrap extensions recursively, with a seen-set so a cycle just returns nil. Double-wrapped Null is now rejected for direct `NewVariantType`, both nested forms, and `NewShreddedVariantType`.

arrow/extensions/variant.go
dgvj-work: `storageType` is now nil-safe. `NewVariantType` returns `arrow.ErrInvalid` for an Opaque with nil storage, and `NewShreddedVariantType` falls back to the default type instead of panicking.

dgvj-work:
Pushed a follow-up for the remaining unwrap cases.

`storageType` now walks extension layers (with cycle protection) and is nil-safe. Double-wrapped Null is rejected in the direct, one-field, two-field, and shredded paths. `NewVariantType(NewOpaqueType(nil, ...))` returns `arrow.ErrInvalid`, and `NewShreddedVariantType` still returns the default type instead of panicking.

@zeroshade this should be ready for another look.

zeroshade (approved):
LGTM jsut waiting for CI to complete

PR #1246, opened 2026-08-28
perf(parquet): decode RLE booleans into bitmaps

fallintoplace:
## What does this PR do?

- **Writes repeated RLE boolean runs directly** into the output bitmap
- **Copies literal bit-packed runs directly** without a `[]uint64` staging buffer
- Keeps partial reads and truncated-page errors aligned with the existing decoder

## Why?

`RleBooleanDecoder.DecodeToBitmap` currently decodes through a `[1024]uint64` buffer and writes each value one at a time. This change uses bulk bitmap fills for repeated runs and packed bitmap copies for literal runs.

## Benchmarks

Apple M1 Pro, `GOMAXPROCS=1`, `-benchmem`, 1M values:

| pattern | before | after | change |
| --- | ---: | ---: | ---: |
| all true | 2.65 ms | 2.45 us | -99.9% |
| alternating | 3.45 ms | 0.20 ms | -94.2% |
| short runs | 3.50 ms | 0.20 ms | -94.3% |

The benchmark also covers 64K values, unaligned output offsets, and nullable spaced decoding.

## Tests

- `go test ./parquet/internal/encoding ./parquet/internal/utils ./parquet/pqarrow`
- `go test -race ./parquet/internal/encoding ./parquet/internal/utils ./parquet/pqarrow`
- `go test -tags noasm ./parquet/internal/encoding ./parquet/internal/utils`

### DIFF
--- a/parquet/internal/encoding/boolean_decoder.go
+++ b/parquet/internal/encoding/boolean_decoder.go
@@ -381,38 +381,18 @@ func (dec *RleBooleanDecoder) Decode(out []bool) (int, error) {
 
 func (dec *RleBooleanDecoder) DecodeToBitmap(out []byte, outOffset int64, length int) (int, error) {
 	max := shared_utils.Min(length, dec.nvals)
-	writer := bitutil.NewBitmapWriter(out, int(outOffset), max)
-
-	var (
-		buf [1024]uint64
-		n   = max
-	)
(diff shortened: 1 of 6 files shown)


zeroshade (approved):
Just needs the merge conflicts to get resolved

PR #1249, opened 2026-08-28
perf(arrow/array): skip full-range SliceEqual slices

fallintoplace:
## Summary

- **Skip temporary arrays when both ranges cover their full inputs.**
- Keep the existing `NewSlice` path for partial ranges.
- This also helps `ChunkedEqual` when chunk boundaries line up.
- Add full-range semantic coverage and benchmarks.

## Benchmark

Local Apple M1 Pro. `-benchtime=500ms -count=3`.

| Case | main | change |
| --- | ---: | ---: |
| `SliceEqual` int64, 64 values | ~700 ns, 352 B, 4 allocs | ~516 ns, 0 B, 0 allocs |
| `SliceEqual` string, 64 values | ~677 ns, 384 B, 4 allocs | ~474 ns, 0 B, 0 allocs |
| `ChunkedEqual`, 1,024 chunks x 64 values | ~756 us, 360 KB, 4,096 allocs | ~559 us, 0 B, 0 allocs |

## Tests

- `go test ./arrow/array`
- `go test ./... -run '^$'`
- `GOOS=linux GOARCH=386 go test -c -o /dev/null ./arrow/array`

### DIFF
--- a/arrow/array/compare.go
+++ b/arrow/array/compare.go
@@ -19,6 +19,7 @@ package array
 import (
 	"fmt"
 	"math"
+	"reflect"
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/bitutil"
@@ -356,7 +357,7 @@ func Equal(left, right arrow.Array) bool {
 func SliceEqual(left arrow.Array, lbeg, lend int64, right arrow.Array, rbeg, rend int64) bool {
 	if lbeg == 0 && lend == int64(left.Len()) &&
 		rbeg == 0 && rend == int64(right.Len()) &&
-		canEqualDirectly(left) && canEqualDirectly(right) {
+		canEqualDirectly(left, right) {
 		return Equal(left, right)
 	}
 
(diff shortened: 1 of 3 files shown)


arrow/array/compare.go
```diff
@@ -353,6 +353,10 @@ func Equal(left, right arrow.Array) bool {
 
 // SliceEqual reports whether slices left[lbeg:lend] and right[rbeg:rend] are equal.
 func SliceEqual(left arrow.Array, lbeg, lend int64, right arrow.Array, rbeg, rend int64) bool {
+	if lbeg == 0 && lend == int64(left.Len()) && rbeg == 0 && rend == int64(right.Len()) {
```
zeroshade: The full-range fast path calls `Equal` on the original interface values. `Equal` only handles Arrow-Go’s concrete array types or `ExtensionArray`, so another valid `arrow.Array` implementation now panics. Wrapping an `*array.Int64` in `struct{ arrow.Array }` and calling full-range `SliceEqual` previously returned `true`; this head panics with `arrow/array: unknown array type`. Please retain the slice-normalization path for unsupported implementations or make `Equal` safely normalize generic arrays, and add the wrapper regression test.

zeroshade (changes requested):
Skipping slice construction for full-range inputs changes compatibility with arbitrary `arrow.Array` implementations.

The previous path normalized inputs through `NewSlice`/`MakeFromData`. The new fast path calls `Equal` directly, which panics for valid wrapper implementations it does not recognize. I reproduced this with a wrapper around `*array.Int64`; the same issue propagates through one-chunk `ChunkedEqual`. Details are inline.

This head currently has no CI results.

---

>

zeroshade (approved):

zeroshade:
@fallintoplace just need to fix the failing tests

zeroshade (approved):

PR #1254, opened 2026-08-28
perf(parquet/compress): reuse Brotli writers across encodes

fallintoplace:
## Summary

- **Reuse Brotli writers** for one-shot `Encode` and `EncodeLevel` calls.
- Keep a separate `sync.Pool` for each supported compression level.
- Reset pooled writers with a nil destination before putting them back.
- Keep the streaming writer APIs unchanged.
- Add serial and parallel benchmarks plus sequential and concurrent correctness tests.

## Benchmark

Compared with `main` at `6b039a76`:

| Case | Main | This PR | Main allocations | This PR allocations |
| compressible / 256 KiB | ~0.78 ms/op | ~0.51 ms/op | ~22 allocs/op, ~11.9 MB/op | 1 alloc/op, ~10 KB/op |
| semi-random / 256 KiB | ~2.04 ms/op | ~1.77 ms/op | ~21 allocs/op, ~12.0 MB/op | 1 alloc/op, ~37 KB/op |

The parallel benchmark also stays at one allocation per operation in the pooled path.

## Checks

- `go test ./parquet/compress -count=1`
- `go test -race ./parquet/compress`
- `go vet ./parquet/compress`
- Parquet package tests with `PARQUET_TEST_DATA` set

### DIFF
--- a/parquet/compress/brotli.go
+++ b/parquet/compress/brotli.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"sync"
 
 	"github.com/andybalholm/brotli"
 	"github.com/apache/arrow-go/v18/parquet/internal/debug"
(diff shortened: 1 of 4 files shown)


parquet/compress/brotli.go
```diff
@@ -41,15 +44,44 @@ func (b brotliCodec) EncodeLevel(dst, src []byte, level int) []byte {
...
+	}
+	return &brotliWriterPools[level]
+}
+
+func releaseBrotliWriter(pool *sync.Pool, w *brotli.Writer) {
+	if pool == nil {
+		return
+	}
+	w.Reset(nil)
+	pool.Put(w)
```
zeroshade: Returning every Brotli writer to a process-global pool retains its large encoder workspace. In a fresh-process probe, one 256 KiB level-11 encode left roughly 44 MiB live after one GC (level 9 was about 43 MiB and the default level about 11 MiB), while the non-pooled writer retained none. Sustained or concurrent compression can retain multiple writers per level and processor. Please introduce a bounded-memory policy—for example, avoid pooling high-workspace levels or use a bounded cache—and add memory-retention coverage.

zeroshade (changes requested):
The reuse path preserves encoded output and passes focused, race, production-caller, and randomized differential tests.

However, pooling every Brotli writer introduces a substantial process-memory regression for high-workspace levels. A single small level-11 request can keep tens of MiB live through a GC, and concurrent compression can retain multiple such workspaces. Details and possible bounded-memory approaches are inline.

No CI checks are currently reported for this head.

---

>

zeroshade (approved):

PR #1256, opened 2026-08-28
fix(parquet): stop BYTE_STREAM_SPLIT FLBA spaced decode aliasing output

### LINKED ISSUE #1255: parquet: BYTE_STREAM_SPLIT FIXED_LEN_BYTE_ARRAY with nulls across multiple data pages decodes values shifted by one

A `FIXED_LEN_BYTE_ARRAY` column encoded with `BYTE_STREAM_SPLIT` decodes **values shifted by one position** when the column chunk contains **both nulls and more than one data page**. The data is silently wrong — no error is returned.

Found on `main` (`f85b0b25` / `477eb633`), linux/amd64, Go 1.24. This is not a recent regression; it reproduces on older commits as well. Surfaced while reviewing #1172, which is unrelated and does not cause it.

**Isolation, already narrowed:**

* Encoding-specific — `PLAIN` passes on byte-identical data and writer configuration; only `BYTE_STREAM_SPLIT` fails.
* Requires **both** conditions:
  * a single large data page (1 MiB) with nulls **passes**
  * multiple data pages without nulls **passes**
  * multiple data pages *with* nulls **fails**
* Reproduced at widths 2, 3, 4, 7, 8, 16, 17, and 32 — not width-specific.
* First mismatch appears partway through, not at the start (e.g. around index 259 for width 4).

Example failure at width 4:

```
mismatch at idx 259
expected: []byte{0x0c, 0x0d, 0x0e, 0x0f}
actual  : []byte{0x10, 0x11, 0x12, 0x13}
```

`0x10 0x11 0x12 0x13` is the value belonging to index 260, i.e. the stream is off by exactly one value from that point on.

**Not root-caused.** `ByteStreamSplitFixedLenByteArrayDecoder.SetData` recomputing `nvals = len(data)/typeLen` and setting `stride` looks correct on inspection, so the fault is likely elsewhere — either per-page decoder reset in the column reader or the encoder's per-page flush. I have not determined whether the file is written incorrectly or read incorrectly.

**Note on existing coverage:** the spaced-decoding unit tests use a single `SetData` followed by one `DecodeSpaced`, so they exercise only the single-page case and pass. A fix should come with a multi-page + nulls integration test.

### Reproducer

Drop into `parquet/pqarrow/` as `bss_flba_repro_test.go` and run
`go test ./parquet/pqarrow/ -run TestBSSFLBANullsMultiPage -v`.

```go
package pqarrow_test

import (
	"bytes"
	"context"
	"fmt"
	"testing"

	"github.com/apache/arrow-go/v18/parquet/file"
	"github.com/stretchr/testify/require"
)

func TestBSSFLBANullsMultiPage(t *testing.T) {
	encs := map[string]parquet.Encoding{
		"BSS":   parquet.Encodings.ByteStreamSplit,
		"PLAIN": parquet.Encodings.Plain,
	}
	for encName, enc := range encs {
		for _, pageSize := range []int64{512, 1 << 20} { // small pages vs one big page
			for _, width := range []int{4, 32} {
				for _, nulls := range []bool{false, true} {
					t.Run(fmt.Sprintf("%s/page%d/w%d/nulls=%v", encName, pageSize, width, nulls), func(t *testing.T) {
						mem := memory.DefaultAllocator
						const n = 5000

						dt := &arrow.FixedSizeBinaryType{ByteWidth: width}
						sc := arrow.NewSchema([]arrow.Field{{Name: "v", Type: dt, Nullable: nulls}}, nil)

						bldr := array.NewFixedSizeBinaryBuilder(mem, dt)
						defer bldr.Release()
						want := make([][]byte, 0, n)
						for i := 0; i < n; i++ {
							if nulls && i%7 == 3 {
								bldr.AppendNull()
								want = append(want, nil)
								continue
							}
							v := make([]byte, width)
							for j := range v {
								v[j] = byte(i*width + j)
							}
							bldr.Append(v)
							want = append(want, v)
						}
						arr := bldr.NewArray()
						defer arr.Release()
						rec := array.NewRecord(sc, []arrow.Array{arr}, int64(n))
						defer rec.Release()

						var buf bytes.Buffer
						props := parquet.NewWriterProperties(
							parquet.WithEncoding(enc),
							parquet.WithDictionaryDefault(false),
							parquet.WithDataPageSize(pageSize),
							parquet.WithBatchSize(128),
						)
						w, err := pqarrow.NewFileWriter(sc, &buf, props, pqarrow.DefaultWriterProps())
						require.NoError(t, err)
						require.NoError(t, w.Write(rec))
						require.NoError(t, w.Close())

						rdr, err := file.NewParquetReader(bytes.NewReader(buf.Bytes()))
						require.NoError(t, err)
						defer rdr.Close()

						fr, err := pqarrow.NewFileReader(rdr, pqarrow.ArrowReadProperties{BatchSize: 137}, mem)
						require.NoError(t, err)
						tbl, err := fr.ReadTable(context.Background())
						require.NoError(t, err)
						defer tbl.Release()

						require.EqualValues(t, n, tbl.NumRows())
						idx := 0
						for _, c := range tbl.Column(0).Data().Chunks() {
							fsb := c.(*array.FixedSizeBinary)
							for i := 0; i < fsb.Len(); i++ {
								if want[idx] == nil {
									require.True(t, fsb.IsNull(i), "idx %d should be null", idx)
								} else {
									require.False(t, fsb.IsNull(i), "idx %d should be valid", idx)
									require.Equal(t, want[idx], fsb.Value(i), "mismatch at idx %d", idx)
								}
								idx++
							}
						}
						require.Equal(t, n, idx)
					})
				}
			}
		}
	}
}
```

Observed result — only the BSS multi-page-with-nulls combinations fail:

```
--- PASS: TestBSSFLBANullsMultiPage/BSS/page512/w4/nulls=false
--- FAIL: TestBSSFLBANullsMultiPage/BSS/page512/w4/nulls=true
--- PASS: TestBSSFLBANullsMultiPage/BSS/page512/w32/nulls=false
--- FAIL: TestBSSFLBANullsMultiPage/BSS/page512/w32/nulls=true
--- PASS: TestBSSFLBANullsMultiPage/BSS/page1048576/w4/nulls=true
--- PASS: TestBSSFLBANullsMultiPage/BSS/page1048576/w32/nulls=true
--- PASS: TestBSSFLBANullsMultiPage/PLAIN/... (all)
```

### Component(s)

Parquet

zeroshade:
### Rationale for this change

Fixes #1255.

A `BYTE_STREAM_SPLIT` `FIXED_LEN_BYTE_ARRAY` column that contains nulls **and** spans more than one data page decodes values shifted by one position — silently, with no error returned.

`spacedExpand` moves values into their spaced positions with `copy` and deliberately does not clean up the null slots:

```go
// because we technically don't care what is in the null slots we don't actually have to clean
// up after ourselves ... Any data that happens to be left in the null slots is fine
```

That reasoning holds for the scalar column types, whose buffers hold values. But `ByteArray` / `FixedLenByteArray` buffers hold **slice headers**, so `copy` leaves *duplicate headers* behind: after expansion a null slot and a valid slot can reference the same backing array.

That is still harmless for a decoder that *replaces* the header, which is what `PlainFixedLenByteArrayDecoder` does:

```go
out[idx] = pflba.data[:pflba.typeLen]   // replaces
```

`ByteStreamSplitFixedLenByteArrayDecoder` instead writes **through** the caller's existing slice:

```go
out[idx] = out[idx][:dec.typeLen]       // reuses caller storage
...
out[element][stream] = data[encLoc]     // writes through it
```

So once `flbaRecordReader` reused its value buffer for the next page, two output slots shared one backing array and clobbered each other. This explains the full shape of the bug: BYTE_STREAM_SPLIT only, nulls required (to create the duplicates), and two or more pages required (the first page creates the aliases, the second decodes into them).

- Add `spacedExpandSwap`, which swaps instead of copying so the buffer remains a permutation of its original elements — no slot aliases another, and every slot keeps its reusable capacity.
- Use it from `ByteStreamSplitFixedLenByteArrayDecoder.DecodeSpaced`.
- Leave `spacedExpand` itself untouched, so every other column type and the non-spaced path are unaffected.
- Add a decoder-level regression test (two pages through one reused buffer, widths 2/3/4/7/8/16) and a randomized differential test asserting `spacedExpandSwap` places values in exactly the same slots as `spacedExpand` while never leaving duplicates.
- Add a `pqarrow` round-trip test over a multi-page, nullable BSS FLBA column — the integration-level case that was returning wrong data.

The decoder-level regression test fails at every width without the fix.

### Why swap rather than the simpler alternatives?

I measured two other approaches and rejected both:

**Making `spacedExpand` itself swap** is correct but replaces `memmove` with element-wise swaps for *every* column type, which is far too expensive on sparse-null runs:

```
SpacedExpandInt64/n65536/nullEvery0     4.457µ -> 81.265µ   +1723%
SpacedExpandFLBA/n65536/nullEvery0      3.969µ -> 173.959µ  +4283%
```

**`clear(out[:toRead])` in `DecodeSpaced`** is a one-liner, but discards all reusable headers. On 4096 slots / 3511 values / width 16 that is 3511 allocations per `DecodeSpaced` on `main` today (it would drop to 1 once #1172 lands, but the fix should not depend on that).

Swapping keeps the reuse. Cost on the affected path only (`BenchmarkBSSFLBADecodeSpaced`, 8192 slots):

```
w4/nullEvery100    41.64µ -> 60.55µ   +45%
w4/nullEvery7      49.45µ -> 71.95µ   +46%
w4/nullEvery2      95.83µ -> 89.59µ    -7%
w16/nullEvery100   213.9µ -> 385.2µ   +80%
w16/nullEvery7     244.7µ -> 251.3µ      ~
w16/nullEvery2     154.3µ -> 160.4µ      ~
geomean                             +25%
```

Steady-state allocations stay at 1–2 per call. A ~25% geomean cost on a path that is currently returning **incorrect data** seemed clearly worth it, and nothing outside BSS FLBA spaced decoding is touched.

### Are these changes tested?

Yes — new tests described above. `parquet/...` passes in full with `PARQUET_TEST_DATA` supplied, and `-race` is clean on `parquet/internal/encoding`.

Yes: BYTE_STREAM_SPLIT FIXED_LEN_BYTE_ARRAY columns with nulls spanning multiple data pages now decode correctly. Previously affected reads returned silently incorrect values.

**This PR contains a "Critical Fix".** Reading an affected file produced wrong values with no error, which could have been persisted or acted on downstream without any indication of a problem.

### DIFF
--- a/parquet/internal/encoding/byte_stream_split.go
+++ b/parquet/internal/encoding/byte_stream_split.go
@@ -96,15 +96,16 @@ func releaseBufferToPool(pooled *PooledBufferWriter) {
 }
 
 func validateByteStreamSplitPageData(typeLen, nvals int, data []byte) (int, error) {
-	if nvals*typeLen < len(data) {
+	encodedNvals, remainder := len(data)/typeLen, len(data)%typeLen
+	if encodedNvals > nvals || (encodedNvals == nvals && remainder != 0) {
 		return 0, fmt.Errorf("data size (%d) is too small for the number of values in in BYTE_STREAM_SPLIT (%d)", len(data), nvals)
 	}
 
-	if len(data)%typeLen != 0 {
+	if remainder != 0 {
 		return 0, fmt.Errorf("ByteStreamSplit data size %d not aligned with byte_width: %d", len(data), typeLen)
 	}
 
-	return len(data) / typeLen, nil
+	return encodedNvals, nil
 }
 
 type byteStreamSplitEncoder[T int32 | int64 | float32 | float64] struct {
--- a/parquet/internal/encoding/byte_stream_split_decode_test.go
+++ b/parquet/internal/encoding/byte_stream_split_decode_test.go
@@ -488,3 +488,120 @@ func TestSpacedExpandSwapMatchesSpacedExpand(t *testing.T) {
 			"iter %d n=%d nulls=%d: swap left duplicate entries", iter, n, nullCount)
 	}
 }
+
+// bssEncodeFLBA lays out values in BYTE_STREAM_SPLIT order: all byte 0s, then all
+// byte 1s, and so on.
+func bssEncodeFLBA(values []parquet.FixedLenByteArray, width int) []byte {
+	data := make([]byte, len(values)*width)
+	for vi, v := range values {
+		for bi, b := range v {
+			data[bi*len(values)+vi] = b
+		}
+	}
+	return data
+}
+
+// TestByteStreamSplitFLBADoesNotWriteThroughForeignBuffers guards the second half of the
+// aliasing hazard reported on GH-1255: this decoder reuses any output slice with enough
+// capacity and writes through it, so if the previous page left slices that point at
+// memory the decoder does not own, decoding corrupts that memory.
+//
+// Two producers leave such slices behind, and neither needs nulls to do it:
+//
+//   - RLE_DICTIONARY assigns dict[idx] into every slot for that index, so a repeated
+//     index leaves several slots aliasing one dictionary-backed slice. Writing through
+//     them both clobbers a decoded value and corrupts the dictionary itself.
+//   - PLAIN slices the page buffer directly, so every slot points into the previous
+//     page's data.
+func TestByteStreamSplitFLBADoesNotWriteThroughForeignBuffers(t *testing.T) {
+	for _, width := range []int{2, 3, 4, 7, 8, 16} {
+		t.Run(fmt.Sprintf("width=%d", width), func(t *testing.T) {
+			col := schema.NewColumn(schema.NewFixedLenByteArrayNode("v", parquet.Repetitions.Required, int32(width), -1), 0, 0)
+
+			// The BSS page: three distinct values, decoded into a reused buffer.
+			bssValues := make([]parquet.FixedLenByteArray, 3)
+			for i := range bssValues {
+				bssValues[i] = make(parquet.FixedLenByteArray, width)
+				for j := range bssValues[i] {
+					bssValues[i][j] = 100 + byte(i*width+j)
+				}
+			}
+			bssData := bssEncodeFLBA(bssValues, width)
+
+			t.Run("after RLE_DICTIONARY", func(t *testing.T) {
+				// Dictionary of two entries, referenced by indices [0, 0, 1] so that
+				// slots 0 and 1 alias the same dictionary-backed slice.
+				dictBuf := make([]byte, 2*width)
+				for i := range dictBuf {
+					dictBuf[i] = byte(i + 1)
+				}
+				// RLE_DICTIONARY index data: bit width byte, then a bit-packed run.
+				idxBuf := []byte{1, 0b00000011, 0b00000100}
+
+				plainDict := NewDecoder(parquet.Types.FixedLenByteArray, parquet.Encodings.Plain,
+					col, memory.DefaultAllocator)
+				require.NoError(t, plainDict.SetData(2, dictBuf))
+
+				dictDec := NewDictDecoder(parquet.Types.FixedLenByteArray, col, memory.DefaultAllocator).(*DictFixedLenByteArrayDecoder)
+				dictDec.SetDict(plainDict)
+				require.NoError(t, dictDec.SetData(3, idxBuf))
+
+				out := make([]parquet.FixedLenByteArray, 3)
+				n, err := dictDec.Decode(out)
+				require.NoError(t, err)
+				require.Equal(t, 3, n)
+
+				dictBefore := bytes.Clone(dictBuf)
+
+				bssDec := NewDecoder(parquet.Types.FixedLenByteArray, parquet.Encodings.ByteStreamSplit,
+					col, memory.DefaultAllocator).(FixedLenByteArrayDecoder)
+				require.NoError(t, bssDec.SetData(len(bssValues), bssData))
+				n, err = bssDec.Decode(out)
+				require.NoError(t, err)
+				require.Equal(t, len(bssValues), n)
+
+				for i, want := range bssValues {
+					require.Equalf(t, want, out[i], "slot %d decoded incorrectly", i)
+				}
+				require.Equal(t, dictBefore, dictBuf, "decoding wrote through into the dictionary page buffer")
+			})
+
+			t.Run("after PLAIN", func(t *testing.T) {
+				plainValues := make([]parquet.FixedLenByteArray, 3)
+				for i := range plainValues {
+					plainValues[i] = make(parquet.FixedLenByteArray, width)
+					for j := range plainValues[i] {
+						plainValues[i][j] = byte(i*width + j)
+					}
+				}
+				plainData := make([]byte, 0, len(plainValues)*width)
+				for _, v := range plainValues {
+					plainData = append(plainData, v...)
+				}
+
+				plainDec := NewDecoder(parquet.Types.FixedLenByteArray, parquet.Encodings.Plain,
+					col, memory.DefaultAllocator).(FixedLenByteArrayDecoder)
+				require.NoError(t, plainDec.SetData(len(plainValues), plainData))
+
+				out := make([]parquet.FixedLenByteArray, 3)
+				n, err := plainDec.Decode(out)
+				require.NoError(t, err)
+				require.Equal(t, len(plainValues), n)
+
+				plainBefore := bytes.Clone(plainData)
+
+				bssDec := NewDecoder(parquet.Types.FixedLenByteArray, parquet.Encodings.ByteStreamSplit,
+					col, memory.DefaultAllocator).(FixedLenByteArrayDecoder)
+				require.NoError(t, bssDec.SetData(len(bssValues), bssData))
+				n, err = bssDec.Decode(out)
+				require.NoError(t, err)
+				require.Equal(t, len(bssValues), n)
+
+				for i, want := range bssValues {
+					require.Equalf(t, want, out[i], "slot %d decoded incorrectly", i)
+				}
+				require.Equal(t, plainBefore, plainData, "decoding wrote through into the PLAIN page buffer")
+			})
+		})
+	}
+}
--- a/parquet/internal/encoding/byte_stream_split_decode_test.go
+++ b/parquet/internal/encoding/byte_stream_split_decode_test.go
@@ -19,11 +19,16 @@ package encoding
 import (
 	"bytes"
 	"fmt"
+	"math/rand"
 	"testing"
 	"unsafe"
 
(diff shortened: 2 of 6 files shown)


zeroshade:
Checked how this composes with #1172, since both touch `fixed_len_byte_array_decoder.go`.

They merge cleanly with no conflict (`prepareOutput` and `spacedExpandSwap` are independent), and the merged result passes `parquet/internal/encoding`, `parquet/file`, and `parquet/pqarrow` in full, including both PRs' own test sets.

Allocation behaviour for `DecodeSpaced` on 4096 slots / 3511 values / width 16:

| | allocs/op | correct? |
| --- | --- | --- |
| `main` today | 0 | **no** |
| `main` + this PR | 16 | yes |
| #1172 + this PR | **0** | yes |

So the two are complementary rather than competing: #1172's contiguous cold-output block absorbs the entries that need storage, and `spacedExpandSwap` preserves every header so the steady state needs no allocation at all. Either merge order works.

zeroshade:
Rebased onto current `main`, which now carries both #1172 and #1220.

#1220 added an early return to `spacedExpand` once the decoded prefix is already aligned. I mirrored it in `spacedExpandSwap` — at that point every remaining swap is a self-swap, so the same reasoning applies and the two functions stay consistent:

```go
idxDecode -= run.Length
// Once the decoded prefix is already aligned every remaining swap is a
// self-swap, so there is nothing left to do. Mirrors spacedExpand.
if idxDecode == run.Pos {
    return numValues
}
```

Re-verified on the new base: the randomized differential test now compares against the early-returning `spacedExpand` and still agrees at every valid position while never leaving duplicates; the decoder-level regression test still fails at all six widths without the fix, and the `pqarrow` round-trip still fails without it. `parquet/...` passes in full and `golangci-lint` reports 0 issues.

The earlier measurement stands: with #1172 now on `main`, `DecodeSpaced` is 0 allocs/op in steady state with this fix applied.

fallintoplace:
I would request changes here because I think there is still one case this doesn't handle.

I tried `PLAIN` nullable FLBA first, then `BYTE_STREAM_SPLIT` FLBA on the next page, reusing the same `valueBuf`.

After the PLAIN page, `spacedExpand` can leave two entries pointing to the same slice. Then BSS uses that same buffer and writes into the slices before `spacedExpandSwap` happens.

So something like `[P0, P1, P1, _, P2]` can be left from the first page. Then BSS decodes into the first 3 entries, and the second and third values can both write into `P1`.

So I think one value can already be overwritten before `spacedExpandSwap` gets a chance to fix the layout.

Maybe worth adding a `PLAIN -> BYTE_STREAM_SPLIT` multi-page test for this case.

zeroshade:
@fallintoplace I believe I've addressed your comment, it's a good point. Let me know what you think

fallintoplace:
I think there may still be a similar issue with `RLE_DICTIONARY -> BYTE_STREAM_SPLIT`.

Possible repro:

1. Decode an FLBA dictionary page with repeated values, e.g. indexes `[0, 0, 1]`. The first two entries in `valueBuf` can point to the same dictionary-backed slice.
2. Reuse the same `valueBuf` for a BYTE_STREAM_SPLIT page with distinct values `[A, B, C]`.
3. BSS writes through the existing slices before `spacedExpandSwap` runs, so writing `B` can overwrite `A` because the first two entries alias.

This one shouldn't need nulls. Could we add a `RLE_DICTIONARY -> BYTE_STREAM_SPLIT` regression test too?

zeroshade:
@fallintoplace confirmed, and thank you — you were right, and the underlying problem is broader than either of the two cases you named. Pushed 09060510.

I wrote your `RLE_DICTIONARY -> BYTE_STREAM_SPLIT` repro as a test first. It fails on the previous commit exactly as you predicted, at every width:

```
--- FAIL: .../width=2/after_RLE_DICTIONARY
    expected: parquet.FixedLenByteArray{0x64, 0x65}
    actual  : parquet.FixedLenByteArray{0x66, 0x67}
    Messages: slot 0 decoded incorrectly
```

`dictConverter.Copy` does `o[idx] = dc.dict[val]`, and `Fill` copies one header across an RLE run, so repeated indices leave several slots sharing one dictionary-backed slice. And as you said, no nulls needed.

While writing it I checked `PLAIN -> BYTE_STREAM_SPLIT` again and found my previous commit had only papered over it. `spacedExpandSwap` removes *duplicate* headers, but PLAIN's `Decode` does `out[idx] = pflba.data[:typeLen]`, so every slot points into the page buffer even with no duplicates at all. BSS then overwrites that buffer wholesale:

```
--- FAIL: .../width=2/after_PLAIN
    expected: []byte{0x0, 0x1, 0x2, 0x3, 0x4, 0x5}
    actual  : []byte{0x64, 0x65, 0x66, 0x67, 0x68, 0x69}
    Messages: decoding wrote through into the PLAIN page buffer
```

So the root cause isn't aliasing specifically — it's that a decoder which writes *through* caller headers was claiming them on capacity alone, while the value buffer is shared across pages and encodings. Capacity can't distinguish our storage from a dictionary entry or a page buffer.

Reuse is now gated on provenance: a header is only written through if it falls inside the block this decoder allocated; anything else is re-pointed at a fresh block.

### This narrows your GH-1172 contract, so please push back if you disagree

`TestByteStreamSplitFixedLenByteArrayDecoderReusesProvidedOutput` asserted that a caller-provided buffer with capacity gets filled in place. That is precisely the unsafe behavior, and I could not find a way to keep it — the record reader hands us the same `valueBuf` whether the headers are ours or were left by the previous page's decoder. I replaced it with two tests: one pinning that we do *not* write through caller buffers, one pinning that we still reuse our own storage.

Allocation is unchanged where it matters. Preparing only the decoded prefix regressed `DecodeSpaced` to 1 alloc/op, because the expansion permutes headers across the whole window and unowned ones migrate into the prefix. Preparing the whole window once, null slots included, restores it:

| | allocs/op | correct? |
| --- | --- | --- |
| before this commit | 0 | **no** |
| prefix-only preparation | 1 | yes |
| this commit | **0** | yes |

(4096 slots / 3009 values / width 16, steady state.) The dense `Decode` path is untouched — successive windows still allocate one block each, as before, and earlier windows keep their values.

`parquet/...` and `arrow/...` pass in full, `-race` is clean on `encoding`, `file` and `pqarrow`, and `golangci-lint` reports 0 issues. Both new tests fail without the fix.

One thing I'd like your read on: `owns` compares `unsafe.SliceData` pointers against the block bounds. It's the honest expression of "did I allocate this", and `go vet` is clean, but if you'd rather not have pointer-provenance in this path I'm open to alternatives — the only other one I found that is clearly correct costs an allocation per page.

fallintoplace:
Everything passed on my side.
No further changes from me. Thanks for following this through

PR #1262, opened 2026-08-29
perf(parquet/pqarrow): write binary values from Arrow offsets

fallintoplace:
## Summary

- **Writes Binary, String, LargeBinary, and LargeString values from Arrow buffers and offsets.**
- Removes the full `[]parquet.ByteArray` allocation from non-dictionary writes.
- Keeps nullable writes, page statistics, bloom filters, page limits, and dictionary fallback behavior.
- Adds coverage for 32-bit and 64-bit offsets, nulls, slices, small batches, and all byte-array encodings.

## Benchmark

64K String rows, uncompressed, dictionary disabled, Apple M1 Pro. Median of 3 runs with 3 seconds per run.

| stats | before | after | heap/op |
| --- | ---: | ---: | ---: |
| false | 1.798 ms | 1.260 ms | 10.62 MB -> 9.04 MB |
| true | 2.343 ms | 1.861 ms | 10.62 MB -> 9.06 MB |

Benchmark command: `go test ./parquet/pqarrow -run '^$' -bench '^BenchmarkWriteArrowBinaryOffsets$' -benchmem -benchtime=3s -count=3`

## Tests

- `go test ./parquet/...`
- `go vet ./parquet/internal/encoding ./parquet/metadata ./parquet/file ./parquet/pqarrow`

### DIFF
--- a/parquet/file/byte_array_column_writer_arrow.go
+++ b/parquet/file/byte_array_column_writer_arrow.go
@@ -111,8 +111,9 @@ func writeArrowValuesSpaced[T byteArrayArrowOffset](w *ByteArrayColumnChunkWrite
 			enc.PutArrow(values, offsets)
 		}
 		if w.pageStatistics != nil {
-			nulls := numValues - numRead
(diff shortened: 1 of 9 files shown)


zeroshade (approved):

zeroshade:
Just needs to fix the linting issue

PR #1264, opened 2026-08-29
perf(bitutil): add ARM64 NEON bitmap operations

fallintoplace:
This adds an ARM64 NEON fast path for aligned bitmap operations.

**What changed**

- Add NEON implementations for AND, OR, ANDNOT, XOR, and XNOR.
- Process 64 bytes per vector loop and keep scalar tails.
- Use the SIMD path only when ARM64 ASIMD is available.
- Keep the existing fallbacks for noasm and other architectures.
- Add aligned, unaligned, tail, and large-input correctness coverage.
- Add benchmarks for all five operations.

**Benchmark**

- **Machine:** Apple M1 Pro, arm64
- **Go:** 1.26.3
- **Method:** median of 5 runs, scalar `-tags noasm` compared with the default NEON path
- **Command:** `go test -run '^$' -bench '^BenchmarkBitmapAlignedOps/(32768|131072)/(and|or|and-not|xor|xnor)$' -benchmem -benchtime=300ms -count=5 ./arrow/bitutil`
- **Lower is better:** values are ns/op

| Size | Operation | Scalar | NEON | Speedup |
| 32 KiB | AND | 2,000 | 689 | 2.9x |
| 32 KiB | OR | 2,001 | 744 | 2.7x |
| 32 KiB | ANDNOT | 2,020 | 741 | 2.7x |
| 32 KiB | XOR | 2,012 | 741 | 2.7x |
| 32 KiB | XNOR | 2,009 | 793 | 2.5x |
| 128 KiB | AND | 8,051 | 3,652 | 2.2x |
| 128 KiB | OR | 8,093 | 3,602 | 2.2x |
| 128 KiB | ANDNOT | 8,041 | 3,670 | 2.2x |
| 128 KiB | XOR | 8,039 | 3,643 | 2.2x |
| 128 KiB | XNOR | 8,118 | 3,630 | 2.2x |

Both paths use zero allocations.

**Checks**

- `go test -count=1 ./arrow/bitutil`
- `go test -count=1 -tags noasm ./arrow/bitutil`
- `go test -race -count=1 ./arrow/bitutil`
- `go test -count=1 ./arrow/compute/...`
- `go vet -composites=false ./arrow/bitutil`
- Linux amd64 and arm64 test-package cross-builds

### DIFF
--- a/arrow/bitutil/bitmap_ops_neon_arm64.s
+++ b/arrow/bitutil/bitmap_ops_neon_arm64.s
@@ -119,3 +119,77 @@ and_not_tail_loop:
 	BNE and_not_tail_loop
 and_not_done:
 	RET
+
+TEXT ·_bitmap_aligned_xor_neon(SB), $0-32
+	MOVD left+0(FP), R0
+	MOVD right+8(FP), R1
+	MOVD out+16(FP), R2
+	MOVD length+24(FP), R3
+
+xor_loop:
+	CMP $64, R3
+	BLO xor_tail
+	VLD1.P 64(R0), [V0.B16, V1.B16, V2.B16, V3.B16]
+	VLD1.P 64(R1), [V4.B16, V5.B16, V6.B16, V7.B16]
+	VEOR V0.B16, V4.B16, V0.B16
+	VEOR V1.B16, V5.B16, V1.B16
+	VEOR V2.B16, V6.B16, V2.B16
+	VEOR V3.B16, V7.B16, V3.B16
+	VST1.P [V0.B16, V1.B16, V2.B16, V3.B16], 64(R2)
+	SUB $64, R3
+	B xor_loop
+
+xor_tail:
+	CBZ R3, xor_done
+xor_tail_loop:
+	MOVBU (R0), R4
+	MOVBU (R1), R5
+	EOR R5, R4, R4
+	MOVB R4, (R2)
+	ADD $1, R0
+	ADD $1, R1
+	ADD $1, R2
+	SUBS $1, R3
+	BNE xor_tail_loop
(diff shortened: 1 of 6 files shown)


arrow/bitutil/bitmap_ops_neon_arm64.s
```diff
@@ -0,0 +1,261 @@
...
+// limitations under the License.
+
+TEXT ·_bitmap_aligned_and_neon(SB), $0-32
+	MOVD left+0(FP), R0
+	MOVD right+8(FP), R1
+	MOVD out+16(FP), R2
+	MOVD length+24(FP), R3
+
+and_loop:
+	CMP $64, R3
```
zeroshade: These NEON loops do not preserve the existing behavior for partially overlapping input/output ranges. For example, using `backing[:256]` as input and `backing[1:257]` as output produces different results from `-tags noasm` for all five bitmap operations because each store can overwrite bytes a later block has not loaded. The amd64 SIMD wrappers already detect this case. Please add equivalent overlap detection to the ARM64 wrappers, fall back to the Go implementation, and add shifted-overlap regressions larger than 64 bytes.

zeroshade (changes requested):
The focused, race, noasm, App Engine, vet, and cross-platform suites pass, as do independent-buffer randomized differential tests.

Partially overlapping input/output slices still produce architecture-dependent results on ARM64: the new NEON loops overwrite bytes that later blocks have not loaded. The amd64 SIMD path already falls back for this case. Details and the fix direction are inline.

No CI checks are currently reported for this head.

---

>

zeroshade (approved):

PR #1265, opened 2026-08-29
perf(compute): add ARM64 NEON arithmetic kernels

fallintoplace:
## Summary

- **Added** ARM64 NEON kernels for primitive add, subtract, multiply, absolute value, and negate.
- **Covered** int32, uint32, int64, uint64, float32, and float64 values.
- **Kept** the generic path for unsupported widths and 64-bit integer multiply.
- **Added** edge cases for vector tails, wrapping arithmetic, minimum integers, signed zero, and the no-assembly fallback.

## Benchmark

Apple M1 Pro, 3 MiB input, no nulls, median of 3 runs from `BenchmarkScalarArithmetic`:

| Case | ARM64 NEON | `-tags noasm` | Speedup |
| --- | ---: | ---: | ---: |
| int64 add, array-array | 179,747 ns/op | 759,116 ns/op | 4.22x |
| int64 add, array-scalar | 160,435 ns/op | 697,497 ns/op | 4.35x |
| float64 add, array-array | 182,029 ns/op | 761,779 ns/op | 4.19x |
| float64 add, array-scalar | 174,644 ns/op | 714,281 ns/op | 4.09x |

## Tests

- `go test -tags noasm ./arrow/compute/... -count=1`
- `go test -race ./arrow/compute/internal/kernels ./arrow/compute -count=1`
- `go vet ./arrow/compute/internal/kernels`
- Linux ARM64 and AMD64 cross-builds for the touched packages

### DIFF
--- a/arrow/compute/internal/kernels/base_arithmetic_arm64.go
+++ b/arrow/compute/internal/kernels/base_arithmetic_arm64.go
@@ -1,0 +1,164 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
LGTM. The ARM64 NEON arithmetic kernels match scalar behavior across supported integer and floating-point types, aliases, vector tails, guard regions, and special values. Focused, race, noasm, cross-platform, benchmark, and CI checks pass.

---

> will follow up.*
>

PR #1268, opened 2026-08-30
perf(compute): add ARM64 NEON constant multiplication

fallintoplace:
This adds an ARM64 NEON fast path for constant multiplication in the compute kernels.

**What changed**

- adds NEON kernels for int32/int64 input and output combinations
- keeps the existing generic implementation when NEON is unavailable
- adds boundary and overflow correctness coverage for short and non-multiple-of-four inputs
- adds benchmarks for all four conversion combinations

**Benchmark**

Apple M1 Pro, darwin/arm64, 1 MiB input:

| operation | generic | NEON dispatch | speedup |
| --- | ---: | ---: | ---: |
| int32 -> int32 | 421,874 ns/op | 87,556 ns/op | 4.82x |
| int32 -> int64 | 423,170 ns/op | 296,760 ns/op | 1.43x |
| int64 -> int32 | 422,409 ns/op | 135,615 ns/op | 3.11x |
| int64 -> int64 | 438,196 ns/op | 302,779 ns/op | 1.45x |

**Tests**

- go test -count=1 ./arrow/compute/internal/kernels
- go test -run ^$ -bench ^BenchmarkMultiplyConstant$ -benchmem -count=1 ./arrow/compute/internal/kernels


zeroshade (changes requested):
The new ARM64 NEON wrappers pass Go `int` lengths to kernels whose assembly consumes only the low 32 bits. For arrays with length at least `1 << 31`, the length can become negative or wrap to a short prefix, causing `ShiftTime` conversions to silently skip or partially process values.

Please fall back to the scalar implementation above `math.MaxInt32`, or make the kernel length handling genuinely 64-bit, and add boundary coverage.

---

>

zeroshade (approved):
The 32-bit length truncation issue from my previous review is resolved: every NEON wrapper now falls back to the scalar implementation above `MaxInt32`, the NEON declarations are properly restricted to ARM64, and boundary coverage was added. Native ARM64 and `noasm` runs both pass locally.

---

PR #1273, opened 2026-09-01
fix(parquet): guard byte-stream-split encoder release against never-flushed buffer

tom-eon:
### Rationale for this change

`byteStreamSplitEncoder.Release` unconditionally passes `enc.flushBuffer` to `releaseBufferToPool`, which dereferences it. The flush buffer is allocated lazily on the first `FlushValues`, so releasing an encoder that never flushed — which `columnWriter.Close` does for a column chunk that received no values — dereferences nil and crashes the process with a SIGSEGV. `ByteStreamSplitFixedLenByteArrayEncoder.Release` has the same defect.

A nil guard at the top of `releaseBufferToPool`, covering both encoders, plus a regression test that releases a never-flushed BYTE_STREAM_SPLIT encoder for all five physical types (INT32, INT64, FLOAT, DOUBLE, FIXED_LEN_BYTE_ARRAY).

### Are these changes tested?

Yes. The new test panics on every subtest without the fix and passes with it; the existing `parquet/internal/encoding` and `parquet/file` suites pass.

No API changes. Writing a parquet file with a BYTE_STREAM_SPLIT column no longer panics when a column chunk ends up empty (e.g. an empty trailing row group).

### DIFF
--- a/parquet/internal/encoding/byte_stream_split.go
+++ b/parquet/internal/encoding/byte_stream_split.go
@@ -94,6 +94,12 @@ var (
 )
 
 func releaseBufferToPool(pooled *PooledBufferWriter) {
+	// the byte-stream-split encoders allocate their flush buffer lazily on the first
+	// FlushValues, so releasing an encoder that never flushed (e.g. a column chunk that
+	// received no values) passes nil here
+	if pooled == nil {
+		return
+	}
 	buf := pooled.buf
 	memory.Set(buf.Buf(), 0)
 	buf.ResizeNoShrink(0)
--- a/parquet/internal/encoding/byte_stream_split_release_test.go
+++ b/parquet/internal/encoding/byte_stream_split_release_test.go
@@ -1,0 +1,52 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package encoding
+
+import (
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/schema"
+	"github.com/stretchr/testify/assert"
+)
+
+// The byte-stream-split encoders allocate their flush buffer lazily on the first
+// FlushValues, so an encoder released without ever flushing (as columnWriter.Close
+// does for a column chunk that received no values) must not dereference the nil buffer.
+func TestByteStreamSplitReleaseWithoutFlush(t *testing.T) {
+	tests := []struct {
+		name    string
+		typ     parquet.Type
+		typeLen int32
+	}{
+		{"int32", parquet.Types.Int32, -1},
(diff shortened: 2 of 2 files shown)


zeroshade (approved):
Good catch — this is a real user-facing crash, not just defensive hardening. Approving.

I confirmed the unit test fails on all five physical types without the guard:

```
--- FAIL: TestByteStreamSplitReleaseWithoutFlush/int32
    Panic value: runtime error: invalid memory address or nil pointer dereference
```

More importantly, I verified your claim about the writer path end to end, because that is what determines how much this matters. `columnWriter.Close` calls `FlushBufferedDataPages()` before releasing the encoder, so it was not obvious from reading whether an empty chunk actually reaches `Release` with a nil buffer. It does. On `main`, writing a BYTE_STREAM_SPLIT column whose chunk receives no values takes the process down:

```
--- FAIL: TestTmpBSSEmptyColumnChunk/row_group_with_no_values
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0]
```

Both shapes crash: a row group that touches the column without writing, and an empty trailing row group after a populated one. With your patch both write clean files (112 and 244 bytes).

I also checked whether the guard is *complete*, since `PooledBufferWriter.Finish` sets `b.buf = nil` and I wanted to rule out a second nil case on the already-flushed path. It is fine: `Finish` calls `Reset(0)`, which calls `Reserve(0)`, which re-acquires from the pool when `buf == nil`. So a non-nil `pooled` always has a non-nil `pooled.buf`, and `pooled == nil` is the only case to handle.

On placement — `releaseBufferToPool` has exactly two callers, both `Release` methods with the same lazy-init pattern, so guarding once in the helper rather than twice at the call sites is the right call.

Verified: `go build ./...`, `go vet`, full `go test ./...`, and `-race` on `parquet/internal/encoding` and `parquet/file` all pass.

## One request: add the writer-level regression test

Your unit test is the right test for the encoder contract, but the bug users hit lives in `parquet/file`, and nothing there would catch a regression if the `Release` path changes shape again. Here is the test I used to confirm the crash — it fails with SIGSEGV on `main` and passes on your branch, so feel free to drop it in as-is:

```go
package file_test

// Writes a BYTE_STREAM_SPLIT column chunk that never receives a value. The
// encoder's flush buffer is allocated lazily on the first FlushValues, so
// releasing it on Close must not dereference nil. See GH-1273.
func TestByteStreamSplitEmptyColumnChunk(t *testing.T) {
	node, err := schema.NewPrimitiveNode("f", parquet.Repetitions.Required,
		parquet.Types.Float, -1, -1)
	require.NoError(t, err)
	sc, err := schema.NewGroupNode("schema", parquet.Repetitions.Required,
		schema.FieldList{node}, -1)
	require.NoError(t, err)

	props := parquet.NewWriterProperties(
		parquet.WithEncoding(parquet.Encodings.ByteStreamSplit),
		parquet.WithDictionaryDefault(false),
	)

	t.Run("row group with no values", func(t *testing.T) {
		var buf bytes.Buffer
		w := file.NewParquetWriter(&buf, sc, file.WithWriterProps(props))
		rgw := w.AppendRowGroup()
		cw, err := rgw.NextColumn()
		require.NoError(t, err)
		require.NoError(t, cw.Close())
		require.NoError(t, rgw.Close())
		require.NoError(t, w.Close())
	})

	t.Run("empty trailing row group", func(t *testing.T) {
		var buf bytes.Buffer
		w := file.NewParquetWriter(&buf, sc, file.WithWriterProps(props))

		rgw := w.AppendRowGroup()
		cw, err := rgw.NextColumn()
		require.NoError(t, err)
		_, err = cw.(*file.Float32ColumnChunkWriter).WriteBatch([]float32{1, 2, 3}, nil, nil)
		require.NoError(t, err)
		require.NoError(t, cw.Close())
		require.NoError(t, rgw.Close())

		rgw2 := w.AppendRowGroup()
		cw2, err := rgw2.NextColumn()
		require.NoError(t, err)
		require.NoError(t, cw2.Close())
		require.NoError(t, rgw2.Close())

		require.NoError(t, w.Close())
	})
}
```

Nit, entirely optional: `byte_stream_split_release_test.go` only uses exported identifiers, so it could live in `package encoding_test` and read as a black-box test. Not worth a round trip on its own.

Issue #44, opened 2023-09-20
[Go][Parquet] Writing a Parquet file from a slice of structs


tschaub:
### Describe the usage question you have. Please include as many useful details as  possible.

I'm hoping to get suggestions on the best way to use the library to write a Parquet file given a slice of structs (Golang structs instead of Arrow's [array.Struct](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/array#Struct)).

The [`parquet.NewSchemaFromStruct()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/parquet/schema#NewSchemaFromStruct) function looks like a useful starting point to generate a Parquet schema from a struct.

The [`pqarrow.NewFileWriter()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/parquet/pqarrow#NewFileWriter) function is helpful for creating a writer.  And I can see how to convert a Parquet schema to an Arrow schema with the [`pqarrow.FromParquet()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/parquet/pqarrow#FromParquet) function.

The [`writer.WriteBuffered()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/parquet/pqarrow#FileWriter.WriteBuffered) method looks like a convenient way to write an Arrow record.  So the gap is then to get from a slice of structs to the Arrow record.

I was looking for something like `array.RecordFromSlice()`.  The [`array.RecordFromStructArray()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/array#RecordFromStructArray) looks useful, but I think I would have to do a fair bit of reflection to work with the [struct builder](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/array#StructBuilder).  It looks like  [`array.RecordFromJSON()`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/array#RecordFromJSON) does the same sort of reflection that I would have to do to use the struct builder.

I know it is not efficient, but I see that I can encode my struct slice as JSON and then generate a record from that.  Here is a working test that uses the `pqarrow.FileWriter` to write a slice of structs as Parquet:
```go
package pqarrow_test

import (
	"bytes"
	"encoding/json"
	"strings"
	"testing"

	"github.com/apache/arrow/go/v14/arrow/array"
[...]

PR #1274, opened 2026-09-02
feat(arrgen): Add code gen go generate machinery for zero allocation and high performance

JayJamieson:
### Rationale for this change

Reduce allocations and performance in hot loops and a suggestion in issue #44.

New go module arrgen that is opt-in.

- arrgen/cmd used to generate code from tagged structs

### Are these changes tested?

Yes, equivalence tests and performance/allocation oriented

Not really, module is opt-in.


zeroshade:
It looks like the files here are duplicated in the tree at `arrgen` and `arrgen/arrgen`, can you fix that?

zeroshade (changes requested):
Three issues need to be addressed:

- Generated schemas diverge from `arreflect` for supported untagged `time.Time`, `decimal128.Num`, and `decimal256.Num` fields, despite the documented equivalence. Please either restore strict parity or clearly reject/document these cases and remove the equivalence-test skips.
- The repository CI does not build, race-test, vet, or verify clean generation for the new `arrgen` module because root module traversal excludes nested modules. Please add explicit CI coverage for it.
- The Quick Start's unversioned `go:generate` directive fails in a normal consumer module unless `arrgen` has already been added as a dependency. Please document the required setup or use a valid pinned-version workflow.

---

>

zeroshade:
Also can we use [testable examples](https://go.dev/blog/examples) rather than creating a separate "examples" folder?

JayJamieson:
> It looks like the files here are duplicated in the tree at `arrgen` and `arrgen/arrgen`, can you fix that?

Ah bugger, I'm not sure how I managed that sorry, will fix.

zeroshade (changes requested):
Thanks for this — it's well-built, and the equivalence framing holds up under scrutiny. I read `arrgen` alongside `arreflect` rather than taking the claims on faith, and the temporal math does match exactly: `timeOfDayNanos` and the generated inline block both do `.UTC()` → midnight-in-UTC → `Sub().Nanoseconds()`, both divide-then-cast, and `Date32/64FromTime` is called identically on either path with no `.UTC()` on either side. The multi-level-pointer nil-check codegen in `renderAppend` is correct too, including the `||` short-circuit ordering the deref depends on.

Test quality is above average for this repo: the `Row` fixture covers every supported shape, row counts include 0, `CheckedAllocator.AssertSize` catches leaks, there's a golden test *and* a drift test *and* a CI `go generate && git diff --exit-code`, and the allocation assertions correctly stand down under `-race`/`-asan`. The `go.work` CI pass is the right call — without it the equivalence tests would silently validate against released `v18.7.0`'s `arreflect` instead of the tree.

One bug to fix, plus a few nits.

## Bug: generic struct types silently emit uncompilable code

`findStruct` (`arrgen/generate.go`) accepts any `*types.Named` whose underlying type is a struct, without checking for type parameters. Given:

```go
type Gen[T any] struct {
    A int64  `arrow:"a"`
    B string `arrow:"b"`
}
```

`Generate` returns no error and writes a file containing `func (a *GenAppender) Append(v *Gen)`, which then fails to build:

```
gen_arrow.go:68:33: cannot use generic type Gen[T any] without instantiation
gen_arrow.go:74:40: cannot use generic type Gen[T any] without instantiation
gen_arrow.go:98:48: cannot use generic type Gen[T any] without instantiation
```

This is the exact failure mode `doc.go` promises not to have — "anything arrgen cannot map ... is a generate-time error naming the field." The fix is one condition in `findStruct`:

```go
if named.TypeParams().Len() > 0 {
    return nil, fmt.Errorf("generic type; arrgen generates for concrete struct types only")
}
```

`testdata/errors/errors.go` is otherwise impressively exhaustive, so a `Generic[T any]` case belongs alongside the others.

## Nits

- **`findStruct` doesn't unalias.** `type Alias = Real` errors with `not a named type`, which points the reader at the wrong thing. `resolveColumn` already calls `types.Unalias` on field types; doing the same on `obj.Type()` would either make aliases work or at least let the error say "type alias".

- **Dead branch in `tagOpts.validate`** (`arrgen/tag.go`). The `if o.REE` early return means `o.REE` inside the following `[]bool{o.Dict, o.View, o.REE}` loop is always false. Harmless, but the slice literal is also an allocation on a path that runs once per field.

- **`Append(v *T)` panics on a nil `v`**, which the generated doc comment doesn't mention.

zeroshade:
Aside from the issues above, what's the reasoning for making this an entirely separate module? I'm pretty sure all the dependencies are already part of the main arrow-go module so there's no real benefit I can see from making it a separate module that gets versioned separately. Is there a particular reason to do so?

JayJamieson:
> Aside from the issues above, what's the reasoning for making this an entirely separate module? I'm pretty sure all the dependencies are already part of the main arrow-go module so there's no real benefit I can see from making it a separate module that gets versioned separately. Is there a particular reason to do so?

@zeroshade

Thanks for the review, I'll make the fixes in follow-up commit.

I opted for a separate module as I wasn't sure how open the team would be to maintaining and how often arrow-go does releases. I figured a separate module would at least allow it to easily be disabled if release/build fails for some reason and not block anyone.

If it's not a problem I can rework it without being a separate module 🙏

PR #1281, opened 2026-09-02
perf(internal/utils): add ARM64 NEON min/max for 8- and 16-bit integers

fallintoplace:
## Summary

- Add ARM64 NEON min/max kernels for int8, uint8, int16, and uint16.
- Keep scalar tails and scalar fallback when ASIMD is unavailable or `noasm` is used.
- Add boundary, extreme-value, and benchmark coverage.

## Benchmark

Apple M1 Pro, n=65536, median of 3 runs:

| Type | noasm | NEON |
| --- | ---: | ---: |
| int8 | 65,642 ns/op | 1,433 ns/op |
| uint8 | 66,233 ns/op | 1,394 ns/op |
| int16 | 65,431 ns/op | 2,807 ns/op |
| uint16 | 66,756 ns/op | 2,726 ns/op |

## Checks

- `go test ./internal/utils -count=1`
- `go test -race ./internal/utils -count=1`
- `go test -tags noasm ./internal/utils -count=1`
- `ARM_ENABLE_EXT=bogus go test ./internal/utils -count=1`
- `go vet ./internal/utils`
- `GOARCH=amd64 go test -c -o /dev/null ./internal/utils`

### DIFF
--- a/internal/utils/min_max.go
+++ b/internal/utils/min_max.go
@@ -155,58 +155,50 @@ var minmaxFuncs = struct {
 	ui64 func([]uint64) (uint64, uint64)
 }{}
 
-// GetMinMaxInt8 returns the min and max for a int8 slice, using AVX2 or
-// SSE4 cpu extensions if available, falling back to a pure go implementation
-// if they are unavailable or built with the noasm tag.
+// GetMinMaxInt8 returns the min and max for an int8 slice, using a SIMD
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
Verified locally on Apple M1: native NEON tests, `-race`, `noasm` parity, and an amd64 compile check all pass. Dispatch correctly guards on ASIMD detection with scalar tails and fallback.

---

PR #1282, opened 2026-09-02
perf(compute): vectorize numeric to boolean casts

fallintoplace:
## What does this PR do?

- Adds an ARM64 NEON fast path for numeric to boolean casts.
- Reuses the existing NEON `array != scalar` comparison kernel to produce the output bitmap.
- Covers int32, uint32, int64, uint64, float32, and float64.
- Keeps the scalar path for narrow integers, noasm builds, unsupported CPUs, and appengine.
- Adds coverage for NaN, signed zero, infinities, sliced input, and bitmap tails.
- Adds a benchmark matrix for the supported types, sizes, and zero distributions.

## Benchmark

Apple M1 Pro, `GOMAXPROCS=1`, compared with `-tags noasm`.

- 32-bit types: about 9x faster at 65K values and about 14x faster at 1M values.
- 64-bit types: about 6x faster at 65K values and about 8x faster at 1M values.
- Small arrays keep the same allocation count and retain the scalar tail path.

## Tests

- `go test ./arrow/compute ./arrow/compute/internal/kernels`
- `go test -tags noasm ./arrow/compute ./arrow/compute/internal/kernels`
- amd64 compile check for the compute package noasm fallback

### DIFF
--- a/arrow/compute/boolean_cast_bench_test.go
+++ b/arrow/compute/boolean_cast_bench_test.go
@@ -1,0 +1,84 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


zeroshade (approved):
Verified locally on Apple M1: the NEON path is exercised natively and passes, including NaN, signed zero, infinities, sliced inputs, and bitmap tails; `noasm` parity passes. Reusing the existing comparison kernel keeps the semantics aligned with the scalar `isNonZero` path.

---

PR #1283, opened 2026-09-02
perf(arrow/array): SIMD-pack boolean slices into bitmaps

fallintoplace:
## Summary

- Add private AVX2 and ARM64 NEON kernels for packing `[]bool` into Arrow bitmap bytes.
- Keep the existing scalar handling for unaligned prefixes and tails.
- Reuse the SIMD path for BooleanBuilder values and validity bitmap packing.
- Keep noasm and unsupported architectures on the scalar implementation.
- Expand bitmap coverage across multiple SIMD blocks and offsets.

## Benchmark

Apple M1 Pro. 65,536 pre-reserved values. Representative runs:

| Benchmark | noasm | SIMD |
| --- | ---: | ---: |
| BooleanBuilder / all false | 23.4 us | 3.9 us |
| BooleanBuilder / all true | 22.6 us | 3.9 us |
| BooleanBuilder / alternating | 21.9 us | 4.6 us |
| BooleanBuilder / one in three | 22.6 us | 4.0 us |

The SIMD path is about 4 to 5x faster for this workload, with zero allocations in both cases.

## Tests

- `go test ./arrow/bitutil ./arrow/compute/... -count=1`
- `go test -tags noasm ./arrow/array -count=1`
- `go vet ./arrow/array ./arrow/bitutil`
- `GOARCH=amd64 go test -c -o /dev/null ./arrow/array`

### DIFF
--- a/arrow/array/builder.go
+++ b/arrow/array/builder.go
@@ -237,6 +237,14 @@ func (b *builder) unsafeAppendBoolsToBitmap(valid []bool, length int) {
 		byteOffset++
 	}
 
+	packed := packBoolsSIMD(nullBitmap[byteOffset:], valid)
+	for i := 0; i < packed/8; i++ {
+		bitSet := nullBitmap[byteOffset+i]
+		b.nulls += 8 - bits.OnesCount8(bitSet)
+	}
+	valid = valid[packed:]
+	byteOffset += packed / 8
+
 	for len(valid) >= 8 {
 		bitSet := packBoolsByte(valid)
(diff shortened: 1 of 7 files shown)


zeroshade (approved):
The NEON path is verified natively on Apple M1 including `-race`; the AVX2 assembly is correct by inspection (compare-with-zero + invert + `VPMOVMSKB`, LSB-first ordering matching Arrow bitmaps) and will be exercised by the AMD64 CI jobs I've just approved. Integration into the existing prefix/tail scalar handling is clean.

---

PR #1285, opened 2026-09-03
chore: fix failing build tags

zeroshade:
### Rationale for this change
Fixing the build on go1.27

Updates the build tag on arrow/compute/registry.go which would already fail on anything <= go1.22 so that the tag is correct.

### Are these changes tested?
Adds go1.27 runs to the CI for testing these types of issues in the future

Only fixes go1.27 builds that were previous failing, anything that builds successfully would still build successfully.

### DIFF
--- a/arrow/compute/registry.go
+++ b/arrow/compute/registry.go
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build go1.22
+//go:build go1.23
 
 package compute
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,22 +37,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        go: [1.25, 1.26, 1.27]
+        arch: [amd64, arm64v8]
         include:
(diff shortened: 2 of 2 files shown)


nssalian (approved):

arrow/compute/registry.go
```diff
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build go1.22
+//go:build go1.23
```
lidavidm, line 17: I guess what I don't understand is why we need this tag at all?

lidavidm (approved):

arrow/compute/registry.go
zeroshade: hmm, originally we had the tag to maintain build compatibility with older Go versions for the rest of the module by excluding the compute package from a `go build ./...` command if you weren't using an up to date Go version because we introduced the generics from go 1.18

At this point, I think we can actually get rid of the build tags since I doubt anyone is building this with Go versions earlier than 1.23, we can do that as a follow-up later on rather than as part of this release.

PR #1287, opened 2026-09-03
perf(compute): SIMD-compact fragmented filters into take indices

fallintoplace:
### Rationale for this change

`GetTakeIndices` currently grows a buffer once per selected run. Fragmented filters such as alternating and short random runs create many allocations and are much slower than dense filters.

- Add an ARM64 NEON path for large, byte-aligned, non-null `uint32` filters.
- Compact 4-bit mask lookups into exact-size take-index output.
- Keep `uint16`, nullable, unaligned, small, dense, long-run, and `EmitNulls` paths unchanged.
- Add coverage for all 4-bit masks, offsets, tails, padding bits, null behavior, and width boundaries.
- Add GetTakeIndices and large FilterRecordBatch benchmarks.

**Benchmark hardware:** Apple M1 Pro, macOS arm64

**Commands:**

```text
go test ./arrow/compute/internal/kernels -run '^$' -bench='^BenchmarkGetTakeIndices/(all-selected|all-clear|long-runs|alternating|short-runs|random)/' -benchmem -benchtime=3x -count=3
go test ./arrow/compute -run '^$' -bench='^BenchmarkFilterRecordBatchGetTakeIndices/(rows=65536|rows=1048576)$' -benchmem -benchtime=1x -count=3
```

Median of the three reported benchmark samples. Values are ns/op per call.

| Workload | Selection pattern | Existing | NEON path | Speedup |
| GetTakeIndices 64K | all-selected | 41.8 us | 41.2 us | 1.0x |
| GetTakeIndices 64K | all-clear | 2.54 us | 3.03 us | 0.84x |
| GetTakeIndices 64K | long-runs | 826 us | 923 us | 0.89x |
| GetTakeIndices 64K | alternating | 48.1 ms | 21.9 us | 2,198x |
| GetTakeIndices 64K | short-runs | 4.87 ms | 22.6 us | 215x |
| GetTakeIndices 64K | random | 73.5 ms | 90.0 us | 817x |
| GetTakeIndices 1M | all-selected | 1.80 ms | 0.80 ms | 2.3x* |
| GetTakeIndices 1M | all-clear | 37.8 us | 39.0 us | 0.97x |
| GetTakeIndices 1M | long-runs | 705 ms | 176 ms | 4.0x* |
| GetTakeIndices 1M | alternating | 4.31 s | 515 us | 8,369x |
| GetTakeIndices 1M | short-runs | 201 ms | 230 us | 873x |
| GetTakeIndices 1M | random | 3.56 s | 1.69 ms | 2,111x |

`*` These patterns use the existing fallback. The measured difference is run-to-run allocator and GC noise, not a new fast path.

Downstream serial `FilterRecordBatch` with one Int64 column and an alternating filter measured **17.7 ms to 152 us (116x)** at 64K rows and **2.06 s to 831 us (2,477x)** at 1M rows. This benchmark includes both take-index generation and the downstream Take call.

### Are these changes tested?

- `go test ./...` with the Arrow and Parquet test-data submodules initialized and `PARQUET_TEST_DATA` / `ARROW_TEST_DATA` set.
- `go test ./arrow/compute/...`
- `go test -tags noasm ./arrow/compute/internal/kernels -run 'TestGetTakeIndices(BatchedRanges|Uint32Coverage)$' -count=1`
- `GOOS=linux GOARCH=amd64 go test -c ./arrow/compute/internal/kernels -o /dev/null`

No API changes. The optimization is ARM64-only and only applies to the narrow `uint32`, non-null, `DropNulls` case.

### DIFF
--- a/arrow/compute/internal/kernels/get_take_indices_neon_arm64.go
+++ b/arrow/compute/internal/kernels/get_take_indices_neon_arm64.go
@@ -1,0 +1,107 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//go:build go1.18 && arm64 && !noasm && !appengine
+
+package kernels
+
+import (
+	"unsafe"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/bitutil"
+	"github.com/apache/arrow-go/v18/arrow/compute/exec"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"golang.org/x/sys/cpu"
+)
+
+var takeIndicesUint32NeonPositions = makeTakeIndicesUint32NeonPositions()
+
+var takeIndicesUint32NeonCounts = [16]uint8{
+	0, 1, 1, 2, 1, 2, 2, 3,
+	1, 2, 2, 3, 2, 3, 3, 4,
+}
+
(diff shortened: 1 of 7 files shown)


zeroshade (approved):
Verified locally on Apple M1: the full compute tree passes natively (NEON path exercised), plus `-race`, `noasm` parity, an amd64 compile check, and a linux/arm64 build.

The real win here is the preallocation commit — the old `unsafeAppendRange` reserved per selected run, which is effectively quadratic for fragmented filters, and the fix benefits every platform including the `uint16` and nullable paths. The NEON compactor itself is clean: exact-count nibble stores mean it never writes past the popcount-sized buffer, the running base is accumulated correctly (+8 per byte, +4 for the high nibble), and the tail byte is masked so padding bits can't leak into the output. Guards (ASIMD, null-free, byte-aligned offset, bounds, fragmentation sampling) all fall back to the generic path.

One non-blocking note: the NEON gate only triggers for `DropNulls`; a null-free filter under `EmitNulls` is semantically identical and could share the fast path in a follow-up.

---

PR #1288, opened 2026-09-03
perf(compute): add ARM64 NEON comparisons for 8- and 16-bit integers

fallintoplace:
## Summary

- Add ARM64 NEON comparisons for `int8`, `uint8`, `int16`, and `uint16`.
- Cover array-array, array-scalar, and scalar-array inputs.
- Keep the existing generic fallback for tails and bitmap offsets.
- Add correctness coverage for signed and unsigned boundaries and bit packing.
- Add native benchmarks for the narrow integer types.

## Benchmark

Apple M1 Pro, Go 1.26.3. `-benchtime=50ms -count=3`. Baseline is `-tags noasm`.

| Case | NEON | Generic | Speedup |
| --- | ---: | ---: | ---: |
| int8 array-array 64K alternating | 5.01 us | 75.7 us | 15.1x |
| uint8 array-scalar 1M random50 | 68.9 us | 1.00 ms | 14.5x |
| int16 scalar-array 64K clustered50 | 4.31 us | 59.1 us | 13.7x |
| uint16 array-array 1M all | 79.7 us | 1.02 ms | 12.8x |

The benchmark matrix also covers 1K, 64K, and 1M elements, all three operand shapes, alternating, random 25/50/75%, clustered 50%, all, and none patterns.

## Tests

- `go test ./arrow/compute/internal/kernels ./arrow/compute`
- `go test -race ./arrow/compute/internal/kernels ./arrow/compute`
- `go test -tags noasm ./arrow/compute/internal/kernels ./arrow/compute`
- Darwin/amd64 and Linux/amd64 compile checks

### DIFF
--- a/arrow/compute/internal/kernels/scalar_comparison_arm64.go
+++ b/arrow/compute/internal/kernels/scalar_comparison_arm64.go
@@ -34,9 +34,13 @@ const (
 //go:noescape
 func _comparison_neon(typ int, op int, shape int, left, right, out unsafe.Pointer, groups int64)
 
+//go:noescape
+func _comparison_narrow_neon(typ int, op int, shape int, left, right, out unsafe.Pointer, groups int64)
+
 func neonComparisonSupported(typ arrow.Type) bool {
 	switch typ {
-	case arrow.INT32, arrow.UINT32, arrow.INT64, arrow.UINT64,
+	case arrow.INT8, arrow.UINT8, arrow.INT16, arrow.UINT16,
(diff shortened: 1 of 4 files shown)


zeroshade (approved):
Verified locally on Apple M1: native NEON runs, `-race`, and `noasm` parity all pass, plus amd64 and linux/arm64 compile checks. The narrow kernels plug into the harness from #1266, so tails and bitmap offsets stay on the generic path, and the differential tests cover signed/unsigned boundaries, all operand shapes, and bit packing. Type and operator indices in the assembly match the arrow constants.

---

PR #1290, opened 2026-09-06
fix(parquet/variant): prevent metadata offset cursor overflow

### LINKED ISSUE #1291: [Parquet] VARIANT metadata reader panics when offset table crosses byte 255

`variant.NewMetadata` panics on valid VARIANT metadata containing 126 dictionary keys with two-byte offsets:

```text
panic: runtime error: slice bounds out of range [255:1]
github.com/apache/arrow-go/v18/parquet/variant.(*Metadata).loadDictionary
```

Observed with v18.7.0 and reproduced on main at `6542b630f84dda25715f6238027cc1eed9d114c6` (Go 1.26.4, darwin/arm64). This affects reading valid data, not only malformed metadata.

#### Standalone reproducer

```go
package main

import (
    "encoding/binary"
    "fmt"

    "github.com/apache/arrow-go/v18/parquet/variant"
)

func main() {
    const count = 126
    metadata := []byte{0x41} // version 1, two-byte offsets
    metadata = binary.LittleEndian.AppendUint16(metadata, count)
    keys := make([]string, count)
    var offset uint16
    for i := range keys {
        keys[i] = fmt.Sprintf("key_%04d", i)
        metadata = binary.LittleEndian.AppendUint16(metadata, offset)
        offset += uint16(len(keys[i]))
    }
    metadata = binary.LittleEndian.AppendUint16(metadata, offset)
    for _, key := range keys {
        metadata = append(metadata, key...)
    }

    decoded, err := variant.NewMetadata(metadata) // panics
    if err != nil {
        panic(err)
    }
    fmt.Println(decoded.DictionarySize()) // expected: 126
}
```

Expected: successful decoding and preservation of all dictionary keys. A 125-key dictionary with the same encoding works; 126 reaches the overflowing slice boundary. Three- and four-byte offset tables are affected as well.

#### Cause and related work

`offsetPos := hdrSizeBytes + offsetSz` infers `uint8` because `offsetSz` is `uint8`. Advancing the cursor and computing `offsetPos + offsetSz` can wrap at byte 256 even though the metadata buffer and offset table are valid. The byte position needs a wider indexing type than the wire-format offset-width value.

Searched open and closed issues for VARIANT, metadata offsets, dictionary overflow, `loadDictionary`, and slice-bounds panics before filing. The older #623 concerns a different array-decoding failure. The validation added in #1063 does not widen this cursor; the reproducer still panics on current main.

Proposed fix and boundary regression tests: #1290.

### Component(s)

Parquet

matanper:
### Rationale for this change

Reading valid VARIANT metadata with 126 dictionary keys and two-byte offsets panics with `slice bounds out of range [255:1]`. The metadata offset cursor inherits `uint8` from `offsetSz`; both advancing the cursor and computing the slice end can wrap at byte 256. Three- and four-byte offset tables hit the same problem with even fewer keys.

This also affects current main after #1063: validating the complete offset table does not prevent the cursor arithmetic from overflowing while reading that valid table.

Use an `int` offset width and cursor for indexing the metadata buffer. Keep the wire-format width and existing metadata validation unchanged.

Add synthetic valid-metadata tests immediately before, at, and after the failure boundary for two-, three-, and four-byte offsets, plus larger dictionaries. Check every decoded dictionary key. No external test data is required.

### Are these changes tested?

- Before the fix, `go test ./parquet/variant -run TestMetadataOffsetTableCrossesByteBoundary -count=1` reproduces the panic at `width_2/keys_126`.
- After the fix, `go test ./parquet/variant ./arrow/extensions` passes, including all 13 new boundary cases.

Valid VARIANT dictionaries spanning byte 255 decode successfully instead of panicking. There are no public API or format changes.

Fixes #1291.


zeroshade (approved):
Confirmed the bug is real and the fix is correct. I reproduced the panic on current `main` with this PR's test (`slice bounds out of range [255:1]` at `width_2/keys_126`), so valid VARIANT metadata does fail today. The `uint8` cursor wrapping at byte 256 is the root cause, and switching to an `int` cursor is safe because the existing `valuesStart` check already validates the full offset-table extent in `uint64` arithmetic before the loop, keeping every slice in bounds. All 14 boundary subtests plus `./parquet/variant` and `./arrow/extensions` pass locally on the fix. Thanks for the tight, well-tested patch.

---

PR #1293, opened 2026-09-06
fix(arrow/extensions): avoid child values for missing variant lists

### LINKED ISSUE #1292: [Parquet] Shredded VARIANT nested lists fail to write when the parent object is null or missing

Writing a shredded VARIANT column to Parquet fails when a nested list's parent object is null or missing between populated rows. The input JSON is valid:

```json
{"obj":{"items":[1]}}
{"obj":null}
{"obj":{"items":[2]}}
```

Observed with Apache Arrow Go v18.7.0 on darwin/arm64. The standalone program below reproduces the error without Iceberg, AWS, external fixtures, or application code:

```text
lists with non-zero length null components are not supported
```

Replacing the middle row with `{}` reproduces the same failure in an Iceberg append regression. Non-null parent objects with lists write successfully.

#### Standalone reproducer

```go
package main

import (
	"bytes"
	"fmt"
	"github.com/apache/arrow-go/v18/arrow/extensions"
)

func main() {
	typ := extensions.NewShreddedVariantType(arrow.StructOf(
		arrow.Field{Name: "obj", Nullable: true, Type: arrow.StructOf(
			arrow.Field{Name: "items", Nullable: true, Type: arrow.ListOf(arrow.PrimitiveTypes.Int64)},
		)},
	))
	b := extensions.NewVariantBuilder(memory.DefaultAllocator, typ)
	defer b.Release()
	err := b.UnmarshalJSON([]byte(`[{"obj":{"items":[1]}},{"obj":null},{"obj":{"items":[2]}}]`))
	if err != nil {
		panic(err)
	}
	values := b.NewArray()
	defer values.Release()
	schema := arrow.NewSchema([]arrow.Field{{Name: "event", Type: typ, Nullable: true}}, nil)
	rec := array.NewRecordBatch(schema, []arrow.Array{values}, 3)
	defer rec.Release()
	tbl := array.NewTableFromRecords(schema, []arrow.RecordBatch{rec})
	defer tbl.Release()
	var output bytes.Buffer
	err = pqarrow.WriteTable(tbl, &output, 1024, parquet.NewWriterProperties(), pqarrow.DefaultWriterProps())
	fmt.Println(err)
}
```

Expected: successful Parquet writing with the populated lists and null/missing parent preserved. Actual: `pqarrow.WriteTable` returns the error above.

#### Traced cause

`shreddedObjBuilder.AppendMissing` appends a null parent struct and recursively appends missing children. `shreddedArrayBuilder.AppendMissing` appends a valid list plus a placeholder child element. The nested list therefore has a nonempty child range beneath a null ancestor. The Parquet writer in `parquet/pqarrow/encode_arrow.go` rejects the resulting discontiguous visited leaf ranges.

A locally tested correction makes the missing-list builder append a null list without adding child elements. This removes the failing layout; regressions should cover absent and null parent objects between populated list rows and verify logical values after read-back. The same `AppendMissing` implementation is present on main at `6542b630f84dda25715f6238027cc1eed9d114c6`.

This is distinct from the dictionary-offset overflow reported in #1291. Searched open issues for VARIANT, `shreddedArrayBuilder`, non-zero null components, and null/list combinations before filing; no matching open report was found.

### Component(s)

Parquet

matanper:
### Rationale for this change

Writing shredded VARIANT values to Parquet fails with `lists with non-zero length null components are not supported` when a nested list has a null, missing, or scalar parent between populated rows. For example:

```json
{"obj":{"items":[1]}}
{"obj":null}
{"obj":{"items":[2]}}
```

`shreddedObjBuilder.AppendMissing` marks the parent struct null, but recursively calling `shreddedArrayBuilder.AppendMissing` adds a valid list and a placeholder child element. The Parquet writer rejects the resulting skipped child range.

Fixes #1292.

Make `shreddedArrayBuilder.AppendMissing` append a null list without adding child elements. The production change is two added lines (including a comment) and four removed lines.

Add a self-contained Parquet write/read regression for null, missing, and scalar parent values between populated lists. It verifies all decoded JSON values, row count, and checked-allocator cleanup. No external fixtures are required.

### Are these changes tested?

All three regression cases reproduce the original write error before the fix and pass afterward.

- `go test ./parquet/pqarrow -run TestShreddedVariantListsUnderMissingParent -count=1`
- `go test ./arrow/extensions ./parquet/variant`

These valid shredded VARIANT values can now be written to Parquet while preserving null, missing, and scalar parent values. No public API changes.


zeroshade (approved):
Confirmed and verified. I reproduced all three regression cases on current `main` (each fails with `lists with non-zero length null components are not supported`), and they pass on this branch along with the full `arrow/extensions`, `parquet/variant`, and `parquet/pqarrow` suites.

The fix is also structurally right: the parent `shreddedObjBuilder.AppendMissing` already nulls the field's `value` column before delegating, so the list typed builder only needs to null its own list — which matches how `tryTyped` represents the non-array residual case. The old code fabricated a placeholder child element beneath a null parent, which is exactly what the Parquet writer rejects.

---

Issue #1296, opened 2026-09-08
[Compute] The registry holds no aggregate functions (no sum, mean, min_max, count, hash_*), and the package documentation does not say so


singhpratech:
**Describe the enhancement requested**

`compute.GetFunctionRegistry()` in v18.7.0 registers 85 functions, all scalar (80) or vector (5).
`FuncScalarAgg` and `FuncHashAgg` exist as kinds, but no function of either kind is registered: `sum`,
`mean`, `min`, `max`, `min_max`, `count`, `count_distinct`, `any`, `all`, `variance`, `stddev`,
`hash_sum` and `hash_count` are all absent. The package documentation describes "an Acero-like arrow
compute engine" and marks the package experimental, but does not say that aggregation is not
implemented, so a reader coming from pyarrow.compute or the C++ function list discovers it by a
failed lookup.

Go 1.27.1, macOS 26.6 (arm64), arrow-go v18.7.0. On v18.8.0-rc1 the registry holds 89 functions (76
scalar, 8 vector, 5 meta), still none of aggregate kind.

```go
reg := compute.GetFunctionRegistry()
names := reg.GetFunctionNames()
kinds := map[string]int{}
for _, n := range names {
    fn, _ := reg.GetFunction(n)
    kinds[fn.Kind().String()]++
}
fmt.Println(len(names), kinds)
for _, n := range []string{"sum", "mean", "min_max", "count", "hash_sum"} {
    _, ok := reg.GetFunction(n)
    fmt.Println(n, ok)
}
```

Output:

```
85 map[Scalar:80 Vector:5]
sum false
mean false
min_max false
count false
hash_sum false
```

**What would help**

Either of two things, and ideally both: a sentence in the `compute` package documentation listing
what is and is not implemented (scalar and vector functions yes; scalar aggregates and hash
aggregates not yet), and the aggregates themselves, starting with `sum`, `min_max`, `count` and
`mean` over the primitive types. Context: a binding that uses arrow-go's kernels as the reference for
its own reductions and group-bys currently has to write those references as plain Go loops.

**Component(s)**

Go, Compute

zeroshade:
The aggregate functions have been on my todo list for a long while but I haven't gotten to them. I'd love to see these kernels get implemented in the compute package!

If you want to contribute the docs/comments updates that would also be welcome. Thanks!

singhpratech:
Thank you. The doc sentence is #1303. On the kernels: I would like to take sum, min_max, count,
[...]

PR #1303, opened 2026-09-08
docs(compute): state which function kinds the registry implements

singhpratech:
### Rationale for this change

The package comment describes an Acero-like engine but does not say that scalar aggregate and hash
aggregate functions are not implemented, so a reader coming from `pyarrow.compute` or the C++
function list finds out by a failed `GetFunction`. #1296.

A "What is implemented" section in the `compute` package comment listing the function kinds the
registry holds (scalar, vector, meta, with the families and the vector/meta names) and stating that
`FuncScalarAgg` and `FuncHashAgg` have no registered functions yet.

### Are these changes tested?

Documentation only; the names were taken from `GetFunctionRegistry()` on main.

Documentation only. The aggregate kernels themselves are a separate piece of work; see the issue.


arrow/compute/doc.go
```diff
@@ -24,6 +24,19 @@
 // native-go arrow compute engine.
 //
 // Everything in this package should be considered Experimental for now.
```
zeroshade: at this point i think we can remove this line :smile:

zeroshade (approved):
LGTM, just one comment

arrow/compute/doc.go
singhpratech: Done, removed.

PR #1306, opened 2026-09-08
fix(compute/exec): do not carry a cached null count across ArraySpan.SetSlice

### LINKED ISSUE #1305: [Compute] and_kleene / or_kleene return wrong values when ExecCtx.ChunkSize is smaller than the input: ArraySpan.SetSlice carries a stale cached null count into the next slice
**Describe the bug**

With `ExecCtx.ChunkSize` set below the input length, `and_kleene` and `or_kleene` return `false`
where the answer is null. The result depends on the chunk size, which it must not.

```go
// left:  [true, null, true, null, false, true, null, true]
// right: [null, true, true, false, null, null, true, true]
ectx := compute.DefaultExecCtx()
ectx.ChunkSize = 1
ctx := compute.SetExecCtx(context.Background(), ectx)
out, _ := compute.CallFunction(ctx, "and_kleene", nil, leftDatum, rightDatum)
```

| ChunkSize | and_kleene result |
|---|---|
| default | `[null null true false false null null true]` |
| 1 | `[null false true false false null false true]` |
| 2 | `[null null true false false false false true]` |
| 3 and above | correct |

arrow-go main (281776d) and v18.7.0, whose `SetSlice` is the same code; Go 1.27.1, macOS 26.6 arm64. Full test in the attached file; the
CheckedAllocator in it reports no leak, this is a values bug only.

**Cause**

`iterateExecSpans` reuses one `ArraySpan` per argument and advances it with `SetSlice`
(executor.go:902). `ArraySpan.SetSlice` (exec/span.go:211-230) reuses the cached `Nulls` for the
new slice when it is `0` ("no nulls") or equal to the old `Len` ("all null"), and only otherwise
resets it to `UnknownNullCount`. That is right only while `Nulls` describes the whole array. The
Kleene kernels call `UpdateNullCount()` on the span they are given (scalar_boolean.go:49, 54, 97,
121, 131, 174, 198), which stores the *current slice's* count into the shared span; on the next
`SetSlice` a slice that happened to be all valid makes every following slice "no nulls", and a
slice that was all null makes every following slice "all null". Any kernel that calls
`UpdateNullCount()` on its input span is exposed; the Kleene kernels are the ones on main that do.

**Expected behavior**

Results independent of `ChunkSize`. The C++ `ArraySpan::SetSlice` never carries a cached count
when a validity bitmap is present; the same rule here fixes it:

```go
if a.Type.ID() != arrow.NULL {
    if a.Nulls != 0 || len(a.Buffers[0].Buf) != 0 {
        a.Nulls = array.UnknownNullCount
    }
} else {
    a.Nulls = length
}
```

With that change the attached test passes for every chunk size and `go test ./arrow/compute/...`
still passes. I can open the PR with the fix and the test if that is welcome.

**Component(s)**

Go, Compute

<details>
<summary>and_kleene_chunksize_test.go (package compute_test; go test -run TestAndKleeneChunkSize ./arrow/compute/)</summary>

```go
package compute_test

import (
	"context"
	"testing"

)

// and_kleene over two boolean arrays with nulls, executed with different
// ExecCtx.ChunkSize values. The result must not depend on ChunkSize.
func TestAndKleeneChunkSize(t *testing.T) {
	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
	defer mem.AssertSize(t, 0)

	// left:  [true, null, true, null, false, true, null, true]
	// right: [null, true, true, false, null, null, true, true]
	lb := array.NewBooleanBuilder(mem)
	lb.AppendValues([]bool{true, false, true, false, false, true, false, true}, []bool{true, false, true, false, true, true, false, true})
	left := lb.NewArray()
	lb.Release()
	defer left.Release()
	rb := array.NewBooleanBuilder(mem)
	rb.AppendValues([]bool{false, true, true, false, false, false, true, true}, []bool{false, true, true, true, false, false, true, true})
	right := rb.NewArray()
	rb.Release()
	defer right.Release()

	run := func(chunk int64) string {
		ectx := compute.DefaultExecCtx()
		ectx.ChunkSize = chunk
		ctx := compute.SetExecCtx(compute.WithAllocator(context.Background(), mem), ectx)
		out, err := compute.CallFunction(ctx, "and_kleene", nil, &compute.ArrayDatum{Value: left.Data()}, &compute.ArrayDatum{Value: right.Data()})
		if err != nil {
			t.Fatal(err)
		}
		defer out.Release()
		arr := out.(*compute.ArrayDatum).MakeArray()
		defer arr.Release()
		return arr.String()
	}
	ref := run(compute.DefaultMaxChunkSize)
	t.Logf("ChunkSize default: %s", ref)
	for _, n := range []int64{1, 2, 3, 4, 8} {
		got := run(n)
		t.Logf("ChunkSize %d: %s", n, got)
		if got != ref {
			t.Errorf("ChunkSize %d: got %s, want %s", n, got, ref)
		}
	}
}
```

</details>

singhpratech:
### Rationale for this change

`and_kleene` and `or_kleene` return `false` where the answer is null when `ExecCtx.ChunkSize` is
below the input length (#1305). `ArraySpan.SetSlice` kept the cached null count for the new
slice when it was `0` or equal to the old length. That is only right while the count describes the
whole span. The executor reuses one span per argument across slices, and a kernel that calls
`UpdateNullCount()` on it stores the slice's count, so the next `SetSlice` treats the following
slice as all valid or all null.

`SetSlice` resets the count to `UnknownNullCount` whenever a validity bitmap is present, as the C++
`ArraySpan::SetSlice` does (`cpp/src/arrow/array/data.h`), and keeps it unknown when it was already
non-zero without a bitmap. The null type still gets the slice length; a span with no bitmap and a
zero count keeps zero. Two cases added to `TestArraySpan_SetSlice`, and a new test that runs both
Kleene kernels at chunk sizes 1, 2, 3, 4 and 8 against the default result.

Cost: a slice of an all-valid array with a bitmap now reports an unknown count, so a kernel that
asks pays one popcount over the slice instead of reading the cache. That is the C++ behaviour.

### Are these changes tested?

Yes. The new chunk-size test fails on main (four assertions, chunk sizes 1 and 2 for both kernels)
and passes with the change; `go test ./arrow/compute/...` passes.

Results of `and_kleene`/`or_kleene` no longer depend on `ExecCtx.ChunkSize`. `ArraySpan.Nulls` after
`SetSlice` is `UnknownNullCount` in cases where it was previously `0` or the slice length; code that
read it directly should call `UpdateNullCount()`, which is what the package's own kernels do.

Closes #1305.


arrow/compute/scalar_boolean_chunksize_test.go
```diff
@@ -0,0 +1,60 @@
+package compute_test
```
zeroshade, line 17: Please add the standard ASF license header to this new source file. Apache RAT 0.16.1, using this PR's `check_rat_report.py` and exclusion list, reports `NOT APPROVED: arrow/compute/scalar_boolean_chunksize_test.go`. The ordinary test file should be licensed rather than excluded from RAT.

zeroshade (changes requested):
The null-count invalidation fix and focused chunk-size tests pass. One required release-audit fix remains: the new test file needs the standard ASF header.

singhpratech:
Added the standard ASF header to arrow/compute/scalar_boolean_chunksize_test.go in the
latest commit. Thanks for the review.

PR #1310, opened 2026-09-11
perf(internal/hashing): avoid boxing byte slices in InsertOrGet

fallintoplace:
## What changed

- `InsertOrGet` already receives a `[]byte`.
- It still called `getHash`, which accepts `interface{}` and boxed the slice.
- This now calls `Hash(val, 0)` directly, matching the typed hashing path.
- There is no behavior change.

## Benchmark

The focused `BenchmarkBinaryMemoTableInsertOrGet` exercises this exact method with 65,536 lookups over 100 repeated values.

Apple M1 Pro, Go 1.26.3:

| | After |
| --- | ---: |
| ns/op | 1.28 ms |
| B/op | 0 |
| allocs/op | 0 |

The old Parquet byte-array benchmark was removed from this description because it uses a separate memo implementation.

## Testing

- `go test ./internal/hashing`
- `go test ./...` with `PARQUET_TEST_DATA` configured


zeroshade (approved):
The typed byte-slice hashing path preserves the existing hash/equality and ownership semantics. Verified with `go test -p 2 -count=1 ./internal/hashing`.

Issue #1298, opened 2026-09-08
[Compute] NewDatumWithoutOwning returns an ordinary Datum, so Release() on it frees the caller's array with no error; its doc comment is also truncated


singhpratech:
**Describe the bug**

`compute.NewDatumWithoutOwning` returns the same `Datum` implementation as `compute.NewDatum`, so
calling `Release()` on it compiles and runs like any other datum. But the datum did not retain the
value, so that `Release()` drops the reference the caller's array still depends on: the allocator
reports the buffers freed while the array is in scope and readable, and the array's own later
`Release()` does not panic. The doc comment says "should not have Release called on it", and ends
with a truncated sentence: "For the most part this is just a convenience function.+-" (datum.go
line 278).

With the Go allocator this is silent, because the memory stays alive under the garbage collector.
With a C allocator, or when the buffers were imported from C through the cdata package, the same
call frees memory under a live array.

Go 1.27.1, macOS 26.6 (arm64), arrow-go v18.7.0; identical on v18.8.0-rc1.

```go
mem := memory.NewCheckedAllocator(memory.NewGoAllocator())
b := array.NewInt64Builder(mem)
b.AppendValues([]int64{1, 2, 3}, nil)
arr := b.NewArray().(*array.Int64)
b.Release()
fmt.Println("after build:", mem.CurrentAlloc(), "bytes")          // 128 bytes

d := compute.NewDatumWithoutOwning(arr)
d.Release()                                                          // compiles; the doc says not to
fmt.Println("after datum.Release():", mem.CurrentAlloc(), "bytes, arr.Len()", arr.Len(), arr.Int64Values())
arr.Release()                                                        // no panic
```

Output:

```
after build: 128 bytes
after datum.Release(): 0 bytes, arr.Len() 3 [1 2 3]
```

For contrast, `compute.NewDatum(arr)` followed by `Release()` leaves 128 bytes allocated until
`arr.Release()`, as expected.

**Expected behavior**

One of: a non-owning datum whose `Release()` is a no-op (it owns nothing); or a distinct type that
does not expose `Release`; or, at least, a panic on the over-release. And the doc comment completed.

**Component(s)**

Go, Compute

leonkozlowski:
[...]

PR #1320, opened 2026-09-17
docs(compute): attach the NewDatumWithoutOwning doc comment and say what Release does to it

singhpratech:
### Rationale for this change

The comment above `NewDatumWithoutOwning` is separated from the declaration by a blank line, so
`go doc` and pkg.go.dev show the function with no documentation at all. The text itself has a
sentence with no main clause and ends in stray characters (`convenience function.+-`). The
function's one hazard, a `Release` call that releases the caller's reference (#1298), is the thing
a reader most needs to see and currently cannot.

The comment now sits on the declaration and states the contract: the returned `Datum` owns nothing,
the caller keeps the value alive for as long as the `Datum` is in use and must not call `Release` on
it, and since the `Datum` is an ordinary `ArrayDatum`/`ChunkedDatum`/`RecordDatum`/`TableDatum`/
`ScalarDatum`, a `Release` call compiles and releases the caller's reference, which with a C-backed
allocator or cdata-imported buffers frees memory under a live value. It ends by pointing at
`NewDatum` for the owning case.

### Are these changes tested?

Documentation only. `go doc ./arrow/compute NewDatumWithoutOwning` renders the two paragraphs;
gofmt and `go vet ./arrow/compute/` are clean.

Documentation only. Whether a non-owning `Datum` should refuse `Release` (a no-op, a distinct
type, or a panic) is a separate decision and stays with #1298; this change does not close it.


zeroshade (changes requested):
The proposed ownership rule is not correct for every accepted call pattern. `NewDatumWithoutOwning` does not add a reference, but that does not imply callers must never release the returned Datum.

`arrow/compute/exprs/exec.go` currently returns `compute.NewDatumWithoutOwning(scalar.NewMapScalar(kvArr))`. That releasable scalar is created inline, so the Datum is its only surviving handle; not releasing it leaks the scalar's retained Arrow data. `ScalarDatum.Release` delegates to the wrapped scalar's `Release`, and downstream expression evaluation currently releases these returned Datums.

Please describe the actual contract: no reference is added, and `Datum.Release` consumes the supplied reference. Borrowed inputs must remain owned and alive elsewhere; ownership-transferred temporaries must eventually be released through the Datum. Alternatively, forbid temporaries and update the existing caller.

singhpratech:
Yes, you're right, thanks. I'd looked at the datum types and their `Release` methods but not at who calls this, and the inline `NewMapScalar` in `exec.go` is exactly what the comment said you couldn't do. I've rewritten it along the lines you describe: no reference is added and `Release` consumes the one you passed in, so a borrowed value has to stay alive elsewhere and shouldn't be released through the Datum, while a temporary created for the Datum has to be. I also mention that `NewDatum` itself does the second thing.

I had a look at the other callers too. The sort and cumulative entry points borrow, as does the set-lookup value set; `NewDatum` and the map literal transfer. Both are in use, so describing the contract seemed better than forbidding temporaries.

zeroshade (approved):
Thanks for the rewrite — this addresses my earlier comment.

The contract is now stated correctly: no retain, and `Datum.Release` consumes the reference the caller supplied. Describing the two call patterns rather than forbidding temporaries is the right call, since both are in use. I checked the callers and it matches: the sort entry points, `safeCastScalar`, the set-lookup value set in `exprs/types.go`, and the tests all borrow; `exprs/exec.go`'s inline `NewMapScalar` and `NewDatum`'s `scalar.Releasable` branch transfer. The pointer back at `NewDatum` is accurate — it retains and hands the new reference to this function.

Also confirms clean: comment is attached to the declaration so `go doc` renders it, the stray `convenience function.+-` is gone, and gofmt/`go vet ./arrow/compute/` pass on the PR head.

One nit, not worth another round: `ScalarDatum.Release` type-asserts `releasable` and is a no-op otherwise, so "whose Release calls the wrapped value's Release" is a touch stronger than the code for a non-releasable scalar. The guidance it supports is still correct.

Agreed that whether a non-owning Datum should refuse `Release` stays with #1298.

PR #1321, opened 2026-09-18
perf(parquet): reuse DELTA_BYTE_ARRAY discard storage

fallintoplace:
**What**

- Reuse decoder-owned storage while `DELTA_BYTE_ARRAY` values are discarded.
- Add correctness tests and discard benchmarks.

**Why**

- `Discard` currently allocates a new byte slice for every discarded value with a non-empty suffix.
- Skipping 65,536 values creates 65,535 allocations only to throw the values away.

**Implementation**

- Copy the first discarded value into owned scratch because the decoded suffix can alias page data.
- Grow the scratch buffer amortized and reuse it across pages.
- Keep empty suffixes on the existing zero-copy path.
- Benchmark prefix-heavy and low-prefix data at 1,024 and 65,536 values.

The benchmark changes from 65,535 allocations to 0 allocations per operation after warm-up, with about 3.3x lower discard time for 65,536 values.

Tests:

- `go test ./parquet/file -run '^(TestWithEOFReader|TestInvalidHeaders|TestInvalidFooter|TestIncompleteMetadata|TestDeltaLengthByteArrayPackingWithNulls|TestDeltaBinaryPackedMultipleBatches|TestPageStreaming.*|TestPrimitiveReader|TestFullSeekRow|TestSkipEmptyRepeatedRows)$' -count=1`
- `go test -race ./parquet/internal/encoding -run 'TestDeltaByteArrayDecoder(DiscardsAllEmptyValues|DiscardCopiesFirstValue|ReusesDiscardScratch|RejectsInvalidPrefixes|KeepsPartialDecodeResults)$' -count=1`
- `go vet ./parquet/internal/encoding`


zeroshade (changes requested):
`discardScratch` can escape through `Decode` and then be overwritten by a later `Discard`.

A concrete sequence is values `["aa", "aa", "zz"]`: `Discard(1)` stores `"aa"` in the scratch; `Decode(1)` takes the empty-suffix path and returns a prefix backed by that scratch; the next `Discard(1)` reuses the scratch and changes the already-returned value to `"zz"`.

I reproduced this on the current head: the retained decoded value mutated from `"aa"` to `"zz"`. This breaks the decoder's existing result-lifetime invariant, including across `SetData` because the scratch is intentionally retained. Please ensure storage exposed by `Decode` cannot alias reusable discard scratch and add this Discard→Decode(empty suffix)→Discard transition as a regression case.

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -162,8 +163,9 @@ func (enc *DeltaByteArrayEncoder) FlushValues() (Buffer, error) {
 type DeltaByteArrayDecoder struct {
 	*DeltaLengthByteArrayDecoder
 
-	prefixLengths []int32
-	lastVal       parquet.ByteArray
+	prefixLengths  []int32
+	lastVal        parquet.ByteArray
+	discardScratch []byte
```
zeroshade, line 169: Worth documenting the invariant that makes the `slices.Clone` guard in `Decode` sufficient:

> Every `lastVal` produced by `Discard` begins at `&discardScratch[0]` — `setDiscardLastValue` assigns the whole buffer, and the empty-suffix branch narrows with `lastVal[:prefixLen:prefixLen]`, which preserves the base pointer.

That's what lets the guard get away with comparing only `&prefix[0] == &d.discardScratch[0]` instead of a range check. It holds today, and I couldn't construct a counterexample. But it's load-bearing and stated nowhere, so a future change to `Discard` that leaves `lastVal` at a non-zero offset into the scratch would defeat the guard *silently* — corrupted values, no panic, no failing assertion unless someone happens to hit the right interleaving.

A couple of sentences here (or on `setDiscardLastValue`) would make that much harder to break by accident.

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -363,6 +387,10 @@ func (d *DeltaByteArrayDecoder) Decode(out []parquet.ByteArray) (int, error) {
 
 		prefix := d.lastVal[:prefixLen:prefixLen]
 		if len(out[0]) == 0 {
+			// Decoded values must not escape through reusable discard storage.
+			if len(prefix) > 0 && len(d.discardScratch) > 0 && &prefix[0] == &d.discardScratch[0] {
+				prefix = slices.Clone(prefix)
+			}
```
zeroshade: This guard is doing more work than the comment lets on, and it's new — on `main` this path is safe for free because `Discard` never reuses a buffer.

I verified it's not decorative: deleting these four lines makes `TestDeltaByteArrayDecoderDiscardKeepsDecodedPrefixes` fail, and a randomized interleaved `Decode`/`Discard` stress test I wrote reports 44 corrupted values across 400 seeds. With the guard, both pass.

Suggest expanding the comment to say *why* — something like "after `Discard`, `lastVal` aliases the reusable scratch buffer; a later `Discard` would overwrite a value the caller still holds" — so the next person doesn't read it as a defensive nicety and delete it.

Optional alternative: hoisting the check to the top of `Decode` (clone `d.lastVal` once if it aliases scratch) would make this structural — "`lastVal` never aliases scratch once `Decode` starts" — rather than a per-value check inside the loop. Costs one clone in cases that don't need it, so the current lazy form is strictly cheaper. Your call; I'd take either.

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -246,23 +264,29 @@ func (d *DeltaByteArrayDecoder) Discard(n int) (int, error) {
...
+// decodeDiscardSuffix reads one suffix after Discard has bounded its count by
+// nvals. SetData has already validated the suffix lengths against the payload.
+func (d *DeltaByteArrayDecoder) decodeDiscardSuffix() parquet.ByteArray {
+	length := d.lengths[0]
+	suffix := d.data[:length:length]
+	d.data = d.data[length:]
+	d.nvals--
+	d.lengths = d.lengths[1:]
+	return suffix
+}
```
zeroshade: Consider putting this on `DeltaLengthByteArrayDecoder` instead. Every field it touches (`lengths`, `data`, `nvals`) is owned by that type, and the body duplicates its `Decode` loop for n=1 — so the two can drift independently. A `decodeOne()` there with the same "caller must ensure `nvals > 0`" precondition keeps them in sync by construction.

Not a correctness issue: `nvals` and `len(lengths)` move in lockstep (`SetData` establishes `nvals == len(lengths)`, `Discard` bounds itself by `min(n, d.nvals)`, and both decrement by one per iteration), so `d.lengths[0]` can't go out of range here. The unchecked index just reads scarier than it is, which is another argument for moving it next to the invariant it relies on.

zeroshade (commented):
### LGTM — nice result.

Reproduced the win locally (same benchmark file applied to both sides, `-count=3`, Ultra 7 155H):

| case | main | this PR |
|---|---|---|
| prefix-heavy/65536 | ~2.12 ms, 3,145,680 B, **65,535 allocs** | ~0.71 ms, 0 B, **0 allocs** |
| low-prefix/65536 | ~1.58 ms, 1,572,840 B, **65,535 allocs** | ~0.72 ms, 0 B, **0 allocs** |
| prefix-heavy/1024 | ~43 µs, 49,104 B, 1,023 allocs | ~12 µs, 0 B, 0 allocs |

~3x faster and allocation-free, as advertised.

`go test -race ./parquet/internal/encoding/...` passes. Full `./parquet/...` has two failures (`TestFileEncryptionDecryption`, `TestArrowReaderAdHocReadDecimals`) that reproduce identically on `main` — both are missing `PARQUET_TEST_DATA`, unrelated to this change. `go vet` and `gofmt` clean.

### On the aliasing hazard

The interesting part is that reusing the scratch buffer converts a previously harmless share into a write-after-share: `Decode`'s empty-suffix fast path hands `d.lastVal[:prefixLen:prefixLen]` straight to the caller, and after this change that can point into storage the next `Discard` overwrites. No panic — just silently mutated values the caller already holds.

I checked this rather than assuming. On `main`, `Discard` allocates a fresh buffer per reconstructed value that is written once and never touched again, so the same sharing is safe; the `discard_lifetime`/`discard_chunks` tests from this PR pass **unmodified against `main`'s source**. I also wrote a randomized stress test (400 seeds, interleaved `Decode`/`Discard`, re-verifying every previously returned value after each step): passes on `main`, passes on this PR, and produces 44 corruption failures against this PR with the `slices.Clone` guard deleted. So the hazard is introduced here, the guard is load-bearing, and `2956ddf2` is earning its keep.

That's not a complaint — it's the cost of the optimization and you've paid it. My only real ask is that the invariant the guard depends on gets written down, since breaking it fails silently rather than loudly. Details inline.

Splitting the new tests into `_discard_chunks_test.go` / `_discard_lifetime_test.go` matches the package's existing `_decode` / `_spaced` / `_validation` convention — no concern there.

Leaving this as a comment rather than an approval so the approve vote stays with a human reviewer; happy to switch it if you'd rather.

parquet/internal/encoding/delta_byte_array.go
```diff
@@ -173,6 +175,25 @@ func (DeltaByteArrayDecoder) Type() parquet.Type {
 
 func (d *DeltaByteArrayDecoder) Allocator() memory.Allocator { return d.mem }
 
+func (d *DeltaByteArrayDecoder) setDiscardLastValue(prefix, suffix parquet.ByteArray) {
+	valueLen := len(prefix) + len(suffix)
+	if valueLen == 0 {
+		if d.discardScratch == nil {
+			d.discardScratch = make([]byte, 0, 1)
```
zeroshade: Nit: the `, 1` isn't needed — `make([]byte, 0)` already returns a non-nil slice, which is all this branch needs to keep `lastVal != nil` working as the "first value of the page" sentinel. Harmless either way; if the intent was to signal "must be a real allocation, not nil", a short comment would carry that better than the capacity.

PR #1323, opened 2026-09-18
perf(arrow/array): concatenate list offsets directly

fallintoplace:
**What**

- Concatenate List and LargeList offsets directly into the output buffer.

**Why**

- Avoid temporary fixed-width buffer wrappers for every input chunk.
- Improve concatenation when arrays are split into many chunks.

**Implementation**

- Added `concatOffsetsData` using the existing direct 32-bit and 64-bit offset handlers.
- Updated only the List and LargeList paths. Map and Binary paths are unchanged.
- Added a benchmark covering both list types and 64, 1,024, and 8,192 chunks.

Benchmark on an Apple M1 Pro with 65,536 values:

| Case | ns/op before -> after | allocs/op before -> after |
| --- | ---: | ---: |
| List, 64 chunks | 86.5us -> 82.3us | 211 -> 145 |
| List, 1,024 chunks | 274.1us -> 229.5us | 3,309 -> 2,273 |
| List, 8,192 chunks | 1.512ms -> 1.220ms | 28,627 -> 19,656 |
| LargeList, 64 chunks | 93.7us -> 87.5us | 211 -> 146 |
| LargeList, 1,024 chunks | 262.6us -> 222.5us | 3,297 -> 2,254 |
| LargeList, 8,192 chunks | 1.467ms -> 1.192ms | 28,244 -> 19,217 |

Tests:

- `go vet ./arrow/array`
- `git diff --check`


zeroshade (commented):
Non-blocking diagnostic issue: both List branches now route through `handle32BitOffsetsData`/`handle64BitOffsetsData`, whose nil-buffer error says `binary array is missing an offset buffer`. A malformed List or LargeList therefore reports the wrong Arrow type. Please use a type-neutral offset-buffer message (or pass the logical type into the helper). The direct offset copying itself is equivalent to the previous sliced-buffer path for valid inputs.

zeroshade (commented):
### Re-review — prior feedback addressed, and the fix turned out to be worth more than I asked for.

`e93e04b` takes the type-neutral-message option, and the new `TestConcatMissingOffsetsDiagnostic` covers all four types that route through the shared helpers (List, LargeList, Binary, LargeBinary). That resolves my earlier comment.

**What I found while verifying it.** I ran the new test unmodified against current `main` to confirm it pins real behaviour. For Binary/LargeBinary it fails exactly as expected — wrong message, nothing else. For List/LargeList, `main` does something worse than report the wrong type:

```
expected: "array/concat: array is missing an offset buffer"
actual  : "arrow/concat: runtime error: index out of range [0] with length 0"
checked_allocator.go:191: LEAK of 64 bytes FROM
    github.com/apache/arrow-go/v18/arrow/array.concatOffsets
        arrow/array/concat.go:464
```

A recovered panic *plus* a leaked buffer. Cause: `gatherFixedBuffers` silently `continue`s past nil buffers, so `concatOffsets` returned a `valueRanges` shorter than `data`, and `gatherChildrenRanges` then indexed off the end — after `concatOffsets` had already allocated the output buffer it never released. Routing through `handle*OffsetsData` (which checks `d.Len() == 0` *before* the nil check, so legitimately-empty arrays still pass) and adding `out.Release()` on the error path fixes both. Worth mentioning in the PR description — it's a robustness fix, not just a diagnostic one.

**Performance.** Allocation counts are deterministic and consistently ~⅓ lower, which is the real signal here:

| case | main | this PR |
|---|---|---|
| list/chunks=64 | 211 allocs, 362,530 B | **145 allocs**, 356,855 B |
| list/chunks=1024 | 3,300 allocs, 638,800 B | **2,244 allocs**, 545,900 B |
| list/chunks=8192 | 28,530 allocs, 2,762,000 B | **19,250 allocs**, 1,988,900 B |
| large_list/chunks=8192 | 27,900 allocs, 2,994,000 B | **18,300 allocs**, 2,206,000 B |

That tracks: the per-input `memory.NewBufferBytes` wrapper from `gatherFixedBuffers` is gone.

On wall time I'll be honest that my box was too noisy to quote a number. My first run showed `large_list/chunks=8192` apparently 4-5x *slower* on this PR, which alarmed me — but the same run showed `list/chunks=8192` degrading monotonically 1.71→2.60 ms across its own five iterations, i.e. thermal throttling, and the large_list cases ran last. Re-running the two sides interleaved with cooldowns, this PR is faster in the clean rounds (1.61 vs 1.94 ms; 1.45 vs 1.83 ms) with one round where both sides blew up together. So: no regression, modest speedup, and don't trust my absolute timings.

`go test -race ./arrow/array/...` and the full `./arrow/...` suite pass. `go vet` and `gofmt` clean.

**CI:** the two Debian CGO jobs (Go 1.26 / 1.27) are red. Logs weren't available yet when I looked, and I can't reproduce locally (`ccalloc` needs Arrow C++ via pkg-config). The same two jobs failed on unrelated #1328 with a similar short runtime (~6m vs ~13-16m for green runs), which smells environmental rather than caused by this change — but that's a guess, so please confirm once the logs land.

One follow-up suggestion inline about `MapType`, which still has the exact bug this PR fixes. Leaving this as a comment rather than an approval so the approve vote stays with a human reviewer; happy to switch it.

arrow/array/concat.go
```diff
@@ -211,6 +211,27 @@ func concatBinaryBuffers(data []arrow.ArrayData, byteWidth, length int, out *Dat
...
+	switch byteWidth {
+	case arrow.Int64SizeBytes:
+		valueRanges, err = handle64BitOffsetsData(data, out, length)
+	default:
+		valueRanges, err = handle32BitOffsetsData(data, out, length)
+	}
+	if err != nil {
+		out.Release()
+		return nil, nil, err
+	}
```
zeroshade, line 231: This `out.Release()` is doing more than tidying up — it fixes a real leak, and I don't think the PR description gives it credit.

On `main`, the List/LargeList nil-offset path allocates the output buffer in `concatOffsets` and then dies via a recovered panic downstream, so the buffer is never released. Your own `TestConcatMissingOffsetsDiagnostic` catches it because of the `memory.NewCheckedAllocator` + `AssertSize(t, 0)` — running that test against `main` reports `LEAK of 64 bytes FROM ...concatOffsets`. Nice that the test covers the allocator state and not just the message; that's what made the leak visible.

Might be worth a line in the commit message so it's findable later.

arrow/array/concat.go
```diff
@@ -223,7 +244,7 @@ func handle32BitOffsetsData(data []arrow.ArrayData, out *memory.Buffer, outLen i
 
 		buf := d.Buffers()[1]
 		if buf == nil {
-			return nil, errors.New("array/concat: binary array is missing an offset buffer")
+			return nil, errors.New("array/concat: array is missing an offset buffer")
```
zeroshade, line 247: This resolves my earlier comment — thanks.

For anyone reading later: the type-neutral wording is the right call here because this helper is genuinely shared by four logical types (List, LargeList, Binary, LargeBinary), so there's no single correct type name to print. Threading the logical type through just to name it in an error wasn't worth the parameter.

Worth being aware the change is larger than the wording for lists: on `main` this branch wasn't reachable for List at all, because `gatherFixedBuffers` skipped nil buffers upstream and the failure surfaced later as `index out of range [0] with length 0`. So this is a new, earlier, typed failure rather than a reworded one.

arrow/array/concat.go
```diff
@@ -747,7 +768,7 @@ func concat(data []arrow.ArrayData, mem memory.Allocator) (arr arrow.ArrayData,
 		}
 	case *arrow.ListType:
 		offsetWidth := dt.Layout().Buffers[1].ByteWidth
-		offsetBuffer, valueRanges, err := concatOffsets(gatherFixedBuffers(data, 1, offsetWidth), offsetWidth, mem)
+		offsetBuffer, valueRanges, err := concatOffsetsData(data, offsetWidth, out.length, mem)
 		if err != nil {
 			return nil, err
 		}
```
zeroshade, line 774: Follow-up worth considering, either here or as a separate PR: `*arrow.MapType` further down still calls `concatOffsets(gatherFixedBuffers(data, 1, offsetWidth), offsetWidth, mem)` and therefore still has the exact bug you just fixed for List.

I probed it on this branch:

```
map err = arrow/concat: runtime error: index out of range [0] with length 0
checked_allocator.go:191: LEAK of 64 bytes
```

I tried the one-line swap to `concatOffsetsData(data, offsetWidth, out.length, mem)` and it works — the probe becomes `array/concat: array is missing an offset buffer` with no leak, and `./arrow/array/...` plus `./arrow/compute/...` pass clean.

That would also leave `concatOffsets`, `handle32BitOffsets`, and `handle64BitOffsets` with no callers, so they could be deleted. Bonus: that retires the `expand := src[:len(src)+1]` trick in `handle32BitOffsets`, which reads one element past a slice's length and relies on the gathered buffer having spare capacity — the direct `src[begin:end]` / `src[end]` form you've moved to is bounds-checked against the buffer's real length instead.

Entirely your call whether to widen this PR's scope; the List/LargeList change stands on its own.

PR #1324, opened 2026-09-18
perf(arrow/array): copy view buffers directly

fallintoplace:
**What**

- Copy BinaryView headers and ListView sizes directly from `ArrayData` during concatenation.

**Why**

- Avoid one temporary `memory.Buffer` wrapper per input chunk.
- Reduce allocations for fragmented view arrays.

**Implementation**

- Use the existing `concatFixedWidthBuffers` helper for BinaryView headers and ListView/LargeListView sizes.
- Added a benchmark for 64, 1,024, and 8,192 chunks.
- Existing concatenate tests cover sliced and nullable view arrays.

Benchmark on an Apple M1 Pro:

| Case | Runtime before -> after | Allocations before -> after |
| --- | ---: | ---: |
| BinaryView, 1,024 chunks | 1.02ms -> 0.93ms | 1,050 -> 25 |
| BinaryView, 8,192 chunks | roughly flat | 8,227 -> 34 |
| ListView, 8,192 chunks | 2.27ms -> 1.92ms | 26,477 -> 18,388 |
| LargeListView, 8,192 chunks | 2.28ms -> 1.95ms | 26,124 -> 17,304 |

Tests:

- `go vet ./arrow/array`
- `git diff --check`


zeroshade (changes requested):
`concatFixedWidthBuffers` introduces a recovered-panic allocation leak for malformed BinaryView/StringView data. It allocates and resizes the destination before slicing each source header buffer. If a non-empty source has a short non-nil header buffer, that slice panics; `concat` recovers and returns an error, but the local destination has not yet been assigned to `out.buffers[1]`, so `out.Release()` cannot free it.

I reproduced this on the current head with `CheckedAllocator`: one malformed one-element BinaryView leaked 64 bytes from `concatFixedWidthBuffers`. The old path sliced all sources in `gatherFixedBuffers` before allocating the destination. Please add failure cleanup around the new allocation (or publish it to the output before a bounds-sensitive operation) and cover the recovered short-header path.

zeroshade (approved):
Thanks — this addresses my earlier comment, and the fix is placed better than what I asked for.

Putting the `success`-flag defer inside `concatFixedWidthBuffers` rather than at the call site means it also covers the pre-existing callers (the `arrow.FixedWidthDataType` and dictionary-index paths), which carried the same latent exposure.

I verified the regression test three ways rather than just watching it pass:

- On the pre-fix commit `28c872ef` it reproduces the exact leak I reported — 64 bytes from `concatFixedWidthBuffers`, across all 8 subtests.
- On the head commit it passes clean.
- On `main` it also passes, since the old `gatherFixedBuffers` path sliced every source before allocating the destination. That confirms the leak was introduced by the first commit here and that the test is a real guard rather than an assertion that was always true.

**Correctness of the perf change.** `out.length` is `sum(d.Len())` at concat.go:653, so allocating `out.length * byteWidth` upfront is exactly equivalent to the old gather-then-sum. Nil-buffer handling is unchanged — `gatherFixedBuffers` had the identical `if buf == nil { continue }`. The BinaryView `bufferIndex` remapping still runs after the copy, in place on `out.buffers[1]`, and the variadic buffers are retained into `out.buffers` before the copy, so a panic there is already cleaned up by the existing recover handler.

I also chased a hypothesis that `concatListView` leaks `values` and `offsetBuffer` when the size-buffer copy panics, since they are allocated at lines 609/615 but not published until 623-625. Probing with a short sizes buffer under `CheckedAllocator` showed no leak on either `main` or this branch — the panic fires earlier, inside `minListViewOffset`, before anything is allocated. Noting it only so you know the path was checked.

**Perf.** Allocation counts are deterministic and match your table exactly:

| Case | main | PR |
| --- | ---: | ---: |
| BinaryView, 1,024 chunks | 1,050 | 25 |
| BinaryView, 8,192 chunks | 8,227 | 34 |
| ListView, 8,192 chunks | ~27,000 | ~18,000 |
| LargeListView, 8,192 chunks | ~27,800 | ~17,800 |

Runtime gains on x86 are larger than your M1 numbers — BinaryView/8192 went 30.9ms to 14.4ms here, where you reported roughly flat, and ListView/8192 went 9.8ms to 2.6ms. Your claims are conservative. No regressions.

`gofmt`, `go vet`, `git diff --check`, the full `arrow/array` suite, and `-race` are all clean.

One pre-existing issue, out of scope for this PR: at concat.go:617, if `concatListViewOffsets` returns its int32 offset-overflow error, `values` from line 609 is allocated but not yet assigned to `out.childData`, so it leaks. Triggering it needs more than 2^31 child elements, so it is not practically testable. It is present on `main` and untouched here — worth a separate issue rather than a change to this PR.

PR #1331, opened 2026-09-20
perf: zstd implement Pooling + DecodeAll with no new reader.

samueleresca:
### Rationale for this change

Performance improvement on the zstd path.

- zstd: decoders come from a `sync.Pool` instead of being created for every record batch, and each compressed buffer is decoded with one `DecodeAll` call straight into the destination buffer (decoder built with `WithDecoderConcurrency(1)` and `WithDecodeAllCapLimit(true))` instead of `Reset(bytes.NewReader)` + `io.ReadFull` on a streaming decoder.
- lz4 : no functional change. It keeps its original per-batch `lz4.Reader`; the library already pools its large block buffers internally. (Only a Decompress method wrapper was added.)

### Are these changes tested?

Added tests + benchmark for zstd (the changed path) - every case improves

```
SMALL buffers (4 KiB per column)
time      : -70.5% to -78.3% in all 8 cases
            1 col,  1 batch  (stream)  9.89 us  ->  2.92 us
            1 col, 64 batches (stream)  592.9 us -> 128.7 us
            16 cols, 64 batches (file)  3.67 ms  ->  894 us
B/op      : -60% (1 col, 1 batch), -64% (1 col, 64 batches), -22% (16 cols)
allocs/op : -45% (1 col, 1 batch) to -70% (16 cols, 64 batches)

LARGE buffers
per column   time                          B/op                        allocs/op
16 KiB       77.6 us  -> 24.0 us  (-69%)   306.9 KiB -> 157.1 KiB (-49%)   184 -> 78  (-58%)
256 KiB      796.7 us -> 191.7 us (-76%)   20.6 MiB  -> 2.05 MiB  (-90%)   185 -> 81  (-56%)
4 MiB        4.63 ms  -> 3.05 ms  (-34%)   51.1 MiB  -> 32.1 MiB  (-37%)   189 -> 79  (-58%)
```

Strictness change for zstd. A buffer whose decompressed size differs from the uncompressed-length prefix now returns an error. This matches other arrow libraries (e.g.: [arrow-cpp](https://github.com/apache/arrow/blob/f3b46fb7d4b8f5fb3fc459860038f0ed8b02bba5/cpp/src/arrow/ipc/reader.cc#L571-L575))


arrow/ipc/compression.go
```diff
@@ -81,29 +84,63 @@ func getCompressor(codec flatbuf.CompressionType) compressor {
 }
 
 type decompressor interface {
-	io.Reader
-	Reset(io.Reader)
+	Decompress(dst, src []byte) error
 	Close()
 }
```
zeroshade, line 94: `Close()` has quietly become an asymmetric contract: for zstd it now means "return to the pool", while for lz4 it still means "reset". A caller that calls `Close()` twice, or keeps using the value afterwards, hands one decoder to two owners.

Both current call sites are a single `defer codec.Close()` (`file_reader.go:535`, `:923`), so this is correct today — and even a double-`Put` wouldn't be a data race, since `DecodeAll` serializes on the `decoders` channel; it would just contend. But there's no guardrail, and the interface reads like an `io.Closer`.

Could you add a doc comment on the interface along the lines of: `Close releases the decompressor; it must not be used afterwards.`

arrow/ipc/compression.go
```diff
@@ -81,29 +84,63 @@ func getCompressor(codec flatbuf.CompressionType) compressor {
...
+	if len(out) != len(dst) {
+		return fmt.Errorf("arrow/ipc: zstd decompressed to %d bytes, expected %d", len(out), len(dst))
+	}
+	return nil
 }
 
 func (z *zstdDecompressor) Close() {
-	z.Decoder.Close()
+	zstdDecompressorPool.Put(z)
 }
```
zeroshade, line 141: Related to the interface comment above — worth a line here recording *why* `z.Decoder.Close()` is deliberately never called, i.e. that `zstd.NewReader(nil)` starts no goroutines (it returns before `d.Reset(r)`), so a pooled decoder holds only GC-reclaimable state. Without that note this looks like a missing `Close` and someone will eventually "fix" it back into a per-batch teardown.

arrow/ipc/compression.go
```diff
@@ -81,29 +84,63 @@ func getCompressor(codec flatbuf.CompressionType) compressor {
...
+	// dst's own capacity and fails rather than allocating a larger slice.
+	out, err := z.DecodeAll(src, dst[:0])
+	if err != nil {
+		return err
 	}
+	// Catch cases where the prefix says fewer bytes than the content, but the content fits in the dst's spare capacity
+	if len(out) != len(dst) {
+		return fmt.Errorf("arrow/ipc: zstd decompressed to %d bytes, expected %d", len(out), len(dst))
+	}
+	return nil
```
zeroshade, line 132: Good catch making this check unconditional — it's genuinely load-bearing, not belt-and-braces. `memory.Buffer.Bytes()` returns `b.buf[:b.length]` where `b.buf` is `roundUpToMultipleOf64`, so `cap(dst)` routinely exceeds `len(dst)` by up to 63 bytes; without this a frame declaring `len+k` (k ≤ 63) would decode "successfully" into the padding. `TestZstdDecompressorSpareCapacity` covers exactly that.

Question on the resulting strictness: a frame carrying *more* data than the uncompressed-length prefix claims previously truncated silently (`io.ReadFull` stops at `len(dst)`) and now errors. That looks like a fix rather than a regression — no Arrow producer should be emitting that — but it is an observable behavior change and isn't mentioned in the PR description. Was it deliberate? If so, could you note it under "Are there any user-facing changes?"

arrow/ipc/compression_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	var before, after runtime.MemStats
+	runtime.ReadMemStats(&before)
+	err := dec.Decompress(dst, frame)
+	runtime.ReadMemStats(&after)
+
+	require.Error(t, err)
+	assert.True(t, errors.Is(err, zstd.ErrDecoderSizeExceeded), "unexpected error: %v", err)
+	assert.Less(t, after.TotalAlloc-before.TotalAlloc, uint64(1<<20),
+		"decoding a frame declaring %d bytes allocated too much", declared)
```
zeroshade, line 174: `TotalAlloc` is process-wide and cumulative, so anything else allocating inside this window inflates the delta. The margin here is wide (1 MiB budget against a 64 MiB declared size) so it will almost always pass, but this is the shape of test that fails once a quarter on a loaded CI runner.

The intent — distinguishing "rejected up front" from "allocated, then failed" — is worth keeping; it's just the mechanism I'd like to be less ambient. A `runtime.GC()` before the first `ReadMemStats` would at least keep GC bookkeeping out of the window.

zeroshade (commented):
Nice work — the design is right, the risky part (bounding the decode to the destination buffer) is handled correctly, and the tests are unusually thorough for a perf PR. Two minor notes and one question inline; nothing blocking.

### What I verified locally

- `go build` + `go vet` clean, `go test -race ./arrow/ipc/` passes. The change is fully contained — `decompressor` has no other implementors or callers outside `arrow/ipc`.
- Perf claim reproduces. `codec=zstd/1cols/64batches/stream`, `-benchtime 20x`:

  | | main | PR |
  |---|---|---|
  | ns/op | 2,063,127 | 385,696 (**-81%**) |
  | B/op | 1,899,226 | 686,102 (**-64%**) |
  | allocs/op | 3,490 | 1,241 (**-64%**) |

- **Never calling `zstd.Decoder.Close()` does not leak goroutines.** `NewReader(nil)` returns at the `if r == nil` guard *before* `d.Reset(r)`, so no stream goroutines are ever started; a pooled decoder only retains a `blockDec`/`frameDec`, reclaimed by GC when the pool entry is dropped. Worth calling out because it's the obvious thing a future reader will try to "fix".
- **Bounding is sound even for frames with unknown content size.** I initially suspected `WithDecodeAllCapLimit` was a no-op for `fcsUnknown` frames, since the explicit `limitToCap` check in `DecodeAll` sits inside `if frame.FrameContentSize != fcsUnknown`. It isn't: `frameDec.runDecoder` sets `maxSyncLen = cap(dst)-len(dst)` unconditionally under `limitToCap`, and re-checks `len(history.b) > cap(dst)` after every block. Overshoot is bounded by one block.
- `wg.Go` needs Go 1.25+; `go.mod` is `go 1.26.0` and the CI matrix is `[1.26, 1.27]`, so that's fine. ASF headers present on both new files.

Leaving this as a comment rather than an approval only because of the question on the strictness change below — happy to approve once that's settled.

arrow/ipc/compression_test.go
samueleresca: Fixed in: https://github.com/apache/arrow-go/pull/1331/commits/0a1e5a5edf01557253078b7d339bbd9051ccaccf

arrow/ipc/compression.go
samueleresca: Fixed in: https://github.com/apache/arrow-go/pull/1331/commits/0a1e5a5edf01557253078b7d339bbd9051ccaccf

arrow/ipc/compression.go
samueleresca: Fixed in: https://github.com/apache/arrow-go/pull/1331/commits/0a1e5a5edf01557253078b7d339bbd9051ccaccf

arrow/ipc/compression.go
samueleresca: See updated PR description

